PR #199, opened 2012-11-03
Initial xdomainrequest support

enemaerke:
Based on issue 196 (https://github.com/cjohansen/Sinon.JS/issues/196) which I raised I have create a minimal implementation that worked for my needs. 

Implementation is very rudimentary and done in isolation which means no tests, no filtering, no integration into createServer, not added to any of your build scripts, very much a stripped down version of the FakeXMLHttpRequest but I thought I might as well share what got it working for me.

### DIFF
--- a/lib/sinon/util/fake_xdomain_request.js
+++ b/lib/sinon/util/fake_xdomain_request.js
@@ -1,0 +1,198 @@
+/**
+ * @depend ../../sinon.js
+ * @depend event.js
+ */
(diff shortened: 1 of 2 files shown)


nikhilmotiani:
Hi, is this integrated in latest version?

jsok:
Hi @nikhilmotiani, my initial PR #478 was merged, but a few tweaks to the XDR code have gone in since I believe.

PR #253, opened 2013-03-13
add support for twitter bower

agrberg:
the build script copies the built version of sinon to the root of the repository and includes this file


kazupooot:
:+1:

fatso83:
Be aware that those [Bower packages are from 2013](https://github.com/fabiosantoscode/sinon-built)

PR #278, opened 2013-04-27
Allow stubs to call through to other functions (original or new)

### LINKED ISSUE #169: Allow custom stub implementation after initial stubbing
You can do this:

``` javascript
this.sinon.stub(object, 'method', function() { return 'MyOwnComplexThing'; })
```

But you can't do this:

``` javascript
var stub = this.sinon.stub(object, 'method');
stub.runs(function() { return 'MyOwnComplexThing'; })
```

fizker:
Let's say that you want to stub a function, such as `fs.readFile`, for certain inputs, but most inputs are of no interest to you.

Currently, this requires you to replace the function with your own, and perform a number of comparisons in order to decide what to do.

But stubs already have a great syntax for determining calls (`withArgs`, so as a developer writing tests, I would love to utilize this.

Towards that end, I have added the following functions to the stub interface:
- `callFake(function)`: Pipes the last call through to the given function.
- `callFakeOn(function, context)`: Pipes the last call through to the given function, and force the given context.
- `callOriginal()`: Pipes the last call through to the function that was stubbed.
- `callsFake(function)`: Pipe future calls through to the given function.
- `callsFakeOn(function, context)`: Pipe future calls through to the given function, and force the given context.
- `callsOriginal()`: Pipe all future calls through to the stubbed function. This is the default behaviour of the sinon.spy() function.

In general, we would like to combine the `withArgs()` function of stubs with the ability to either call either the original or a specific fake function, thus utilizing the quite-good comparison behind that function.

Since all calls still go through the stubs, the various assert-methods also still function as expected.

Incidentally, this also renders the `sinon.spy()` function irrelevant, as it behaves the same as `sinon.stub().callsOriginal()`, except stubs have the powerful `withArgs` function.
## Scenarious where this could be useful

Replacing specific call to `fs.readFile`:

```
var fs = require('fs')
var sinon = require('sinon')
var readFileStub = sinon.stub(fs, 'readFile')

// We don't care about most files
readFileStub.callsOriginal()

// We do care about this specific file
readFileStub.withArgs('/path/to/interesting/file').yieldsAsync(ourFakeFile)
```

Overriding specific selector for `document.querySelector`:

```
var stub = sinon.stub(document, 'querySelector').callsOriginal()
stub.withArgs('.ourSpecificClass').returns(ourFakedElement)
```

Faking an ajax-response (this one might be a bit farfetched to see in practice):

```
var stub = sinon.stub(JSON, 'parse').callsOriginal()
stub.withArgs('fake-json').returns(ourFakeObject)

// This is responsible for parsing the JSON and parsing it on into the system
jsonFetched('fake-json')
```

(This potentially fixes #169, as I understood it).

### DIFF
--- a/lib/sinon.js
+++ b/lib/sinon.js
@@ -95,6 +95,7 @@ var sinon = (function (buster) {
                     object[property] = wrappedMethod;
                 }
             };
+            method.original = wrappedMethod;
 
             method.restore.sinon = true;
             mirrorProperties(method, wrappedMethod);
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -145,6 +145,24 @@
         }
     }
 
+    function callThrough(stub, thisValue, args) {
+        var func = stub._callback;
+        return func.apply(stub.callbackContext || thisValue, args);
+    }
+
+    function callOriginal(stub, thisValue, args) {
+        var func = stub.original;
+        return func.apply(thisValue, args);
+    }
+
+    function passLastCallThrough(stub, func, ctx) {
+        var calls = stub.callCount;
+        var args = stub.args[calls - 1];
+        var thisValue = stub.thisValues[calls - 1];
+
+        return (func || stub.original).apply(ctx || thisValue, args);
+    }
+
     var uuid = 0;
 
     sinon.extend(stub, (function () {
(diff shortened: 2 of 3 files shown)


oravecz:
These people won't let this die! :) OK, so developers have several use cases where `.callsOriginal()` will come in handy, and authors don't want to blur the line between Stubs and Spies.

Here is another use case where the proposed syntax is more than sugar. It comes into play when stubbing a prototype function.

```js
sinon.stub( EventWriter.prototype, "process", function(event) {
    if (event === 'some-event') {
        throw new Error('Some error I want to test for');
    } else {
        origProcess(event);  // <-- how do I invoke the method in the object that
                             // is constructed elsewhere?
    }
} )
```

While the new syntax would make simple work of it.

```js
sinon.stub( EventWriter.prototype, "process" )
    .withArgs( 'some-event' ).throws( new Error('Some error I want to test for') )
    .callsOriginal()
```

PR #442, opened 2014-03-25
Fixing a 'withArgs' and 'atLeast' issue with mock.js

stevesouth:
If a mock has a single ‘withArgs’ expectation that is expected to be called once, and another that is expected atLeast(1), the atLeast(1) expectations fails when called twice.

The "nots throw when exceeding at least expectation and withargs" test highlights the issue which is fixed with this commit.

### DIFF
--- a/lib/sinon/mock.js
+++ b/lib/sinon/mock.js
@@ -54,6 +54,16 @@
             }
         }
 
+        function arrayEquals(arr1, arr2, compareLength) {
+            if (compareLength && (arr1.length !== arr2.length)) return false;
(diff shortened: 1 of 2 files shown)


Druotic:
@cjohansen @stevesouth Are the only remaining steps here a rebase and some wording?  This issue occurs for more cases than just 'withArgs' and 'atLeast', and I believe this would fix the problem.  More generically, any time there are multiple expectations present, one of which is already satisfied (such as atMost/atLeast), the next call which is still valid as defined by the constraint of the already satisfied expectation will result in an unexpected call error.

PR #445, opened 2014-03-30
Matching multiple arguments at once

reimund:
These two commits makes it possible to match multiple arguments with a single matcher.

The first commit makes the following scenario work:

``` javascript
var obj = { foo: function() { }};
var ex = sinon.mock(obj).expects('foo');

obj.foo('bar', 'abc', 123);

sinon.assert.calledWith(ex, sinon.match(function(arg1, arg2, arg3) {
    return 'bar' == arg1
            && 'abc' == arg2
            && 123 == arg3;
}));
```

The second commit makes this, slighly less verbose, example work:

``` javascript
var obj = { foo: function() { }};
var mock = sinon.mock(obj);

mock.expects('foo').once()
    .withArgs(sinon.match.all(function(arg1, arg2, arg3) {
            return 'bar' == arg1 && 'abc' == arg2 && 123 == arg3;
    }));

obj.foo('bar', 'abc', 123);
```

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -45,9 +45,13 @@
         },
 
         calledWith: function calledWith() {
-            for (var i = 0, l = arguments.length; i < l; i += 1) {
-                if (!sinon.deepEqual(arguments[i], this.args[i])) {
-                    return false;
+            if (1 == arguments.length && sinon.match && sinon.match.isMatcher(arguments[0])) {
+                return arguments[0].test.apply(sinon, this.args);
+            } else {
+                for (var i = 0, l = arguments.length; i < l; i += 1) {
+                    if (!sinon.deepEqual(arguments[i], this.args[i])) {
+                        return false;
+                    }
                 }
             }
 
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -196,6 +196,13 @@
         }, "instanceOf(" + sinon.functionName(type) + ")");
     };
 
+    match.all = function (expectation, message) {
+        message = message || 'custom matcher (sinon.match.all)';
+        var m = match(expectation, message);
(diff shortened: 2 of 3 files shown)


mroderick:
This needs a rebase

fearphage:
Should we swap out the yoda conditions? They aren't present anywhere else.

Those should be [caught by ESLint](http://eslint.org/docs/rules/yoda) also.

> if we have this feature, wouldn't it make sense to make it a little more generic

One thing to think about, your suggested solution and the one in this PR don't allow for partial matches: `match(first3Args), match(last3Args)`

fatso83:
This has lingered far too long. Let's get this out for Christmas 🎁 

1. No one on the maintainer team has much interest in the mocks. I'd say let's get it merged if the test passes. On the way to Sinon 3 we will be splitting out the mocks into a separate project anyway.
2. I can do whatever rebasing needs to be done.
3. I'll rename `all` to `args`
4. We can do partial argument matching anyways using `arguments` (or [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) in ES2015). 
5. I'll fix the Yoda statements and add an ESLint rule.

Not sure what @fearphage means with the partial matching example, as I don't understand the code snippet without context.

Anyone opposed?

fearphage:
It doesn't seem like a way to not match all or the rest of the arguments. It's "greedy" in regular expression terms.

And to explain using code, there's no way to do this as I understand it: 

```js
calledWith(42, true, matchNext3Args, 'catpants', 'test');
```

The sample function call would match: `fn(42, true, 'foo', 'bar', 'baz', 'catpants', 'test')`

mroderick:
The original author has not participated in this PR for nearly three years. I am closing this

PR #453, opened 2014-04-06
be able to use only sinon.js file

javenwang:
I could not use only the sinon.js in Titanium environment, so I made some modifications.
As @omorandi said in https://github.com/cjohansen/Sinon.JS/issues/135#issuecomment-6047799, that is Titanium's problem which cannot handle relative paths correctly.


fatso83:
Closing this as the work on CommonJS support in 2.0 has made it redundant.

PR #466, opened 2014-04-19
Fix to allow fakeserver response body to be function

gdw2:
Something like this was supposed to be possible (pardon the coffeescript) using fakeserver.  The function was being used to perform the match, but not to generate the response.

```
server.respondWith("POST","/api/users",
  ()->
    params = JSON.parse(arguments[0].requestBody)
    return [200, {"Content-Type":"application/json"},JSON.stringify(params)]
)
```


mroderick:
This needs a rebase... and test coverage?

fatso83:
@gdw2: any chance of you completing this with a test?

mroderick:
This hasn't seen a response from @gdw2 in more than two years. I am closing it

Issue #458, opened 2014-04-15
Stubbing fs.readFileSync fails since 1.8.1


rprieto:
In Sinon `1.7.0`, the following works as expected:

``` js
sinon.stub(fs, 'readFileSync');
```

Later versions seem to throw:

```
TypeError: Cannot call method 'charCodeAt' of undefined
```

Unfortunately I can't seem to get a real stack trace for the error. Stubbing `fs.readFile` (not sync) works fine, and I'm running Node `v0.10.25`.

roc:
I also get this issue, in javascript... what should I add to usefully prove this?

fearphage:
Is the code open source? Have you tried upgrading to a later version of sinon?

roc:
@fearphage didn't have any free time to reproduce this until now. Here's an example with `proxyquire`, which I feel maybe muddies the waters of what's happening here, but I'd hoped it might be useful nonetheless:

https://github.com/roc/sinon-filesync

fearphage:
@roc Your test case is a no go. It's broken. No passing tests at all.

```shell
➜  sinon-filesync git:(master) mocha         


  loading a file and printing it
    1) should say hello m8

  replace file contents
    2) "before all" hook


  0 passing (13ms)
  2 failing

  1) loading a file and printing it should say hello m8:
     TypeError: Cannot read property 'charCodeAt' of undefined
      at Object.stripBOM (internal/module.js:48:14)
      at require (internal/module.js:20:19)
      at Context.it (test/index.js:11:17)

  2) replace file contents "before all" hook:
     TypeError: Cannot read property 'charCodeAt' of undefined
      at Object.stripBOM (internal/module.js:48:14)
      at Object.require.extensions.(anonymous function) (node_modules/proxyquire/lib/proxyquire.js:259:43)
      at Proxyquire._withoutCache (node_modules/proxyquire/lib/proxyquire.js:179:12)
      at Proxyquire.load (node_modules/proxyquire/lib/proxyquire.js:136:15)
      at Context.before (test/index.js:22:11)
```

fearphage:
```js
'use strict'

const file = require('fs').readFileSync(
  __dirname + '/file.txt', 'utf8'
);

module.exports = `hello ${file}`;
```

[...]

PR #563, opened 2014-09-15
refactored for running coffeescript tests, and added #458 test in coffee...

mjashanks:
issue #458 was closed as it could not be recreated. 

However, the proving test was created in JS  - the bug exists only for coffeescript.

I have added coffscript test support, and converted the test for coffeescript. The test currently fails.

I updated the "test" script in package.json to run coffeescript tests also.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     }
   ],
   "scripts": {
-    "test": "node test/node/run.js",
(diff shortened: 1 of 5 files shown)


mroderick:
I am closing this. Feel free to re-open with JavaScript reproducible errors.

PR #600, opened 2014-11-13
Sinon 2.0

cjohansen:
Making this a pull request that can be used to track progress and dependencies.

- [x] Build with browserify
- [x] Run browserified bundle in browsers
- [ ] Extract fake server
- [ ] Fix deferred tests
- [ ] Run CI tests against both source and built version(s)
- [x] Lint
- [ ] Figure out a solution for overridable sinon.log
- [ ] Figure out a solution for overridable sinon.format
- [ ] Figure out an alternative for sinon.config
- [ ] Figure which parts of the public API no longer needs to be public (extend, etc)
- [ ] Figure out which parts to extract as optional add-ons (perhaps assert, test)
- [ ] Use `this` as little as possible
- [ ] Get rid of the sinon.spy() / sinon.spy.create() duality. We don't need sinon.spy.create and the like
- [ ] Do we still want to support mocks? Are people using mocks?
- [ ] Clean up code organization: stream-line file names, flatten directory structure
- [x] Upgrade Buster to a more recent version
- [ ] Include https://github.com/mroderick/sinon-555 as an automated test somehow
- [ ] Stubbing an object should not execute getters (in supported browsers)
- [ ] Stubbing an object property should fail if it is a getter (in supported browsers)
- [ ] Make fake XHR/server load on Node - remove the checks


GCheung55:
I can think of four possible ways of loading Sinon.JS in the browser. 

1. Load each file in order via script tab. 
2. Load one built file via script tag
3. Using an AMD loader, such as require.js, load one built file and add a shim property to require.js config so it knows the built file exports a sinon object. 
4. Using an AMD loader to require Sinon.JS file as individual modules. 

Generating a single built file would be for items 2 and 3. 

Would you consider changing the files to be Common.Js syntax? This would still allow generating a built file from browserify but AMD style folders/files could also be generated. Or the other way around and changing the files to be AMD syntax and generate Common.Js files. Either way should cover all four ways of loading. 

This also means not having to split up the codebase into their own modules. One could just do

'''javascript
// common.js
var sinonSpy = require('Sinon.JS/spy')

// amd
require(['Sinon.JS/spy'], function(sinonSpy){...})
'''

cjohansen:
CommonJS/Browserify is the goal.

GCheung55:
@cjohansen Cool. Just realized there is a sinon-2.0-more-strict branch and it already has the CommonJS style.

vvo:
@cjohansen I just released a standalone XHR/XDR project. It has some more precise features on XDR (currently sinon does not implement XDR well, there should not be readystate for example).

I also have .response, .timeout working.

It may not fit all your needs but it still extracted and so if you want to forkit/useit feel free: https://github.com/algolia/faux-jax

mroderick:
> I just released a standalone XHR/XDR mocking project.

:+1:

vvo:
@mroderick @cjohansen finished "faux-jax", implement precisely from IE7 to 11.

IE does not requires to load additional files to overwrite XMLHttpRequest, may be useful for sinon, see this hack: https://github.com/algolia/faux-jax/blob/cc7243a35127cc6d944f72b1966fa4cac49f3f95/lib/ie.js

cjohansen:
Interesting. If it solves all the cases using it in Sinon 2.0 will be a no-brainer. It saves us the work of extracting our own (incomplete) implementation.

vvo:
@cjohansen I am updating it frequentlty, it's all TDD, follow closely the specs (XHR, XDR). Currently missing:
- https://xhr.spec.whatwg.org/#the-upload-attribute
- no check whether request should fail based on CORS rules (sinon.js does not does it either)
- no XMLHttpRequestUpload implem
- no "xhr with timers" implem

Given the need, it may be trivial to implement it, did not have the use case.

vvo:
Also my hack can be used to overwrite setTimeouts etc.. without the need to load separate files. Also providing easy commonjs/browserify (where you only have one entry point) use

mroderick:
> Given the need, it may be trivial to implement it, did not have the use case.

Perhaps some of the people asking for fully faked XHR2 api in Sinon.JS tickets could lend a hand if we campaign for it :)

cjohansen:
> Also my hack can be used to overwrite setTimeouts etc.. without the need to load separate files. Also providing easy commonjs/browserify (where you only have one entry point) use

This was actually fixed in Sinon a while back as well, a release is pending.

vvo:
> This was actually fixed in Sinon a while back as well, a release is pending.

Did you invent the trick or found it like me in dark places (google forums I guess)

cjohansen:
It was based on a very thorough article investigating the matter. I can't remember where I saw it, and unfortunately the commit does not honor the source :(

GCheung55:
If it's about overwriting setTimeout and such, I submitted a PR a few months ago... I know I included a link to the article.

GCheung55:
It was https://github.com/cjohansen/Sinon.JS/pull/628#issuecomment-67908525

vvo:
Made a module to allow overwriting writables: https://github.com/algolia/writable-window-method:

```js
var writable = require('writable-window-method');

writable([
  'XMLHttpRequest',
  'Date',
  'setTimeout'
]);

function XMLHttpRequest() {
  console.log('Owned');
}

var noNative = new XMLHttpRequest();
// 'Owned'

var xhr = new writable.original.XMLHttpRequest();
// a real XMLHttpRequest instance
```

Does not requires loading additionnal files, it may be handy to sinon.js npm users because when using commonjs, you cannot load additional files easily.

ponelat:
By supporting fakeServer on node.js would this allow for mocking nodej.js http.requests... using the same interface? Would mean that I can run http mock tests for nodejs *and* browser.... very cool.

vvo:
@ponelat This is exactly what you can do with https://github.com/algolia/faux-jax

ponelat:
@vvo thanks, I'll take a look into that.

mroderick:
Can we close this PR now?

cjohansen:
Unfortunately, yes...

vvo:
I did not get it, why unfortunately? Sinon.js 2.0 will not make it as planned? (no big deal, only asking)

cjohansen:
Because I put some work into this, and now it is outdated. Attempting to reconcile it with current master will be a bigger job than redoing the work.

MattKunze:
Does this mean sinon 2.0 is being abandoned, or is there another issue we can use to track the progress towards that? I'm primarily asking regarding usage with webpack (webpack/webpack#117) - we've been pointing at a the 2.0 branch for our dependencies, but would prefer to keep it closer to the mainstream release if possible.

cjohansen:
The existing branch is abondoned, but there will be a Sinon 2.0 eventually. CommonJS modules is highly desired, but someone needs to put in some work for it to happen.

frank-weindel:
I know Sinon 2.0 will happen eventually (and thank you for all the work you've put into this project!) but some people were relying on that branch for their webpack projects. For at least what I was using it for, it at least worked.

I recovered the branch on a form, whose commits still exist on github, which anyone can use if they'd like, until the real Sinon 2.0 is released:
https://github.com/onetwosee/sinon/tree/sinon-2.0

MattKunze:
FWIW `var sinon = require('imports?define=>false,require=>false!sinon/pkg/sinon')` is working for us with webpack and the current version of sinon (1.17.2), that's what we switched to when the 2.0 branch was deleted.

Pretty ugly, but I'd rather point to something closer to the mainstream development and then hopefully we can just clean up the require statement once CommonJS support exists

frank-weindel:
Thanks for that! That's a much better way to go. You can also configure your webpack.config.js in such a way so that you can just `require('sinon')` in your test files. For anyone reading this, make sure you npm install `imports-loader` as a dev-dependency.

```javascript
    module: {
      loaders: [
        {
            test: /sinon\.js$/,
            loader: 'imports?define=>false,require=>false'
        }
      ],
    },
    resolve: {
      alias: {
        sinon: 'sinon/pkg/sinon'
      }
    }
```

vieron:
Best solution for me (taken from https://github.com/webpack/webpack/issues/304):

```javascript
module: {
    noParse: [
        /node_modules\/sinon\//,
    ]
},
resolve: {
    alias: {
        'sinon': 'sinon/pkg/sinon'
    }
}
```

PR #628, opened 2014-12-16
Wrap IE specific files to properly set methods on window or global

GCheung55:
Unwrapped code can cause issues with `this` if the code is appended to other code that has 'use strict' defined outside of a function scope.

To prevent any issues in reference to `this`, wrap the IE code in a self-executing function and make any changes to a reference to `window` or `global`.

### DIFF
--- a/lib/sinon/util/timers_ie.js
+++ b/lib/sinon/util/timers_ie.js
@@ -1,4 +1,4 @@
-(function(global){
+(function (global) {
     "use strict";
 
     /**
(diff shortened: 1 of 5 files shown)


cjohansen:
That's awesome, thanks! https://github.com/cjohansen/Sinon.JS/commit/9ac683412f10aa3cf4cc85fae90165500f64cf15

GCheung55:
Closing because the changes were added.

PR #640, opened 2015-01-04
Add safety check for document.createElement and setInterval

kpdecker:
Mostly specific to https://github.com/walmartlabs/fruit-loops but adds one more layer of safety should other environments have a global named document that is not strictly a DOM document.

This showed test failures locally but these appeared to be unrelated to this change being in the clock logic and did not appear to fail consistently with or without this patch.

### DIFF
--- a/lib/sinon/util/core.js
+++ b/lib/sinon/util/core.js
@@ -12,7 +12,7 @@
 "use strict";
 
 (function (sinon) {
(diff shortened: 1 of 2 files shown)


cjohansen:
Travis fails on some linting issues, but apparently not in your code. The changes look innocent enough, I'll take my chances ;) Thanks!

PR #646, opened 2015-01-13
Include check for window.ActiveXObject using strict comparison to undefi...

WaffleSouffle:
...ned

https://github.com/cjohansen/Sinon.JS/issues/645


cjohansen:
I made this change slightly differently: https://github.com/cjohansen/Sinon.JS/commit/951a4dabd77a67019c5021c2927f26b821c19e89

PR #647, opened 2015-01-13
Aligned sandbox.useFakeXMLHttpRequest API to documentation - resolves #393

### LINKED ISSUE #393: sandbox.useFakeXMLHttpRequest is server instead of xhr
This seems rather confusing, was this on purpose or is this a bug?

[lib/sinon/sandbox.js#L120](https://github.com/cjohansen/Sinon.JS/blob/v1.8.1/lib/sinon/sandbox.js#L120):
```js
    sinon.sandbox.useFakeXMLHttpRequest = sinon.sandbox.useFakeServer;
```

As a result, code like the following (based on sinon-qunit documentation) doesn't actually work:
```js
QUnit.test('example', function (assert) {
    var xhr = this.sandbox.useFakeXMLHttpRequest(),
        requests = [];

    xhr.onCreate = function (request) {
        requests.push(request);
    };

    $.getJSON('/test');
    assert.equal(this.requests.length, 1);

    requests[0].respond(200, { 'Content-Type': 'application/json' }, '{}');
});
```

.. because that attaches a spurious `onCreate` method to the server object. The fake xhr manager is actually in `xhr.xhr` and the code is fixed when changed to `xhr.xhr.onCreate = function () { .. }`.

fearphage:
This makes the documentation correct and keeps the usage consistent with `useFakeServer` and `useFakeTimers`.

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -76,7 +76,7 @@
             },
 
             useFakeXMLHttpRequest: function useFakeXMLHttpRequest() {
-              return this.xhr = this.useFakeServer().xhr;
+                return this.xhr = this.useFakeServer().xhr;
             },
 
             inject: function (obj) {
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -75,6 +75,10 @@
                 return this.add(this.server);
             },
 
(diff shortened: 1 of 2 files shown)


fearphage:
Tests are (and have been) failing, but the failed tests are not relevant to this branch/pull request.

PR #653, opened 2015-01-19
Use FakeXDomainRequest when XHR does not support CORS. Fixes #584

### LINKED ISSUE #584: fakeServer not Working in IE9
I've got `fakeServer` working in all browsers except IE9. I'm not seeing any errors in IE9, I just get a response from the real server instead of `fakeServer`. It looks like [`XDomainRequest` support was added in May](https://github.com/cjohansen/Sinon.JS/blob/master/Changelog.txt#L18), so I'm not sure what is going wrong.

This is basically how I'm setting up the `fakeServer`. My app takes care of defining `url` and `response`;

```
server.autoRespond = true;

server.respondWith(url, function(r){
  r.respond(
    200,
    {"Content-Type":"application/json"},
    response
  );
});
```

And here's my function that creates the CORS request;
```
function _createCorsRequest(method, url, data) {

    var xhr;
    if ('withCredentials' in new XMLHttpRequest()) {
      xhr = new XMLHttpRequest();
      xhr.open(method, url, true);
    } else if (typeof XDomainRequest != 'undefined') {
      xhr = new XDomainRequest();
      //The next 6 lines must be defined here or IE9 will abort the request
      xhr.timeout = 3000;
      xhr.onload = function(){};
      xhr.onerror = function (){};
      xhr.ontimeout = function(){};
      xhr.onprogress = function(){};
      xhr.open('POST', url + '?' + data);
    } else {
      xhr = null;
    }
    return xhr;
}
```

Do I need to configure something differently in order for `fakeServer` to respond to an `XDomainRequest`?

eriwen:
Related to #590 - Tests were breaking because fake_server was not correctly depending on `./fake_xdomain_request.js`. 

Note that I branched off of `v1.12.2` because tests are (as of this moment) broken on master. 

Cheers,
Eric

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -1,4 +1,5 @@
 /**
+ * @depend fake_xdomain_request.js
  * @depend fake_xml_http_request.js
  * @depend ../format.js
  * @depend ../log_error.js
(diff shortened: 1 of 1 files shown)


eriwen:
@cjohansen Yowza, that was fast :)

PR #655, opened 2015-01-20
Enhancement: run builds on docker

till:
This should provide more resources and a little speed-up.


mroderick:
Do they start faster? Running time seems about same

till:
@mroderick for reference: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

PR #658, opened 2015-01-25
Remove request to add Github labels from CONTRIBUING.md

jmeas:
Only contributors to the Sinon project can modify an issue's labels.


jmeas:
ya I actually read it :eyes:

cjohansen:
Is there a way to make it possible for everyone to add labels?

jmeas:
I don't think so, @cjohansen :(

Issue #511, opened 2014-07-13
global not properly detected


dmlap:
On Chrome 35 (OS X) and PhantomJS 1.9.7, 85457384d040e46b05ca4819af4e650f4dcf6938 fails to properly pass through the global object to the body of fake_xml_http_request.js. That leads to `TypeError: 'undefined' is not an object (evaluating 'global.XMLHttpRequest')` errors. [This example](http://jsfiddle.net/bkS3y/1/) demonstrates the problem outside of sinon.

PR #659, opened 2015-01-25
Fix fakeXHR requests in Node.

### LINKED ISSUE #508: Failing tests when running in browsers
The error appearing when running tests from current `master` is: 

```
TypeError: 'undefined' is not a function (evaluating 'sinon.useFakeXMLHttpRequest()')
at Object.sinon.fakeServer.create (http://localhost:8123/lib/sinon/util/fake_server.js:82:30)
```

If my use of `git bisect` is correct, then c28f47f917984167c88e24b04f6aa400654cf749 is the first commit where the tests fail when run in a browser.

**NOTE:** You will need to run the tests from the test page served from a webserver, and not just loaded from the file system, or you will get other errors as well, making it more difficult to isolate this issue.

### LINKED ISSUE #657: JSDom + fakeXHR troubles
Sinon version: **v1.12.2**
Environment: **OSX 10.10.1**
Example URL: **N/A**

**Description**
I'm attempting to write unit tests for jQuery ajax requests in a JSDom environment, but Sinon's fake XHR is not intercepting the requests. The exact same suite **does** work in the browser.

I know the jQueryXHR + JSDom combo can cause some folks trouble, but I'm fairly certain that Ive got them configured correctly. I've set the `global.window` and `global.XMLHttpRequest` object from JSDom so that jQuery can make XHR requests, and I've verified that this works. The issue comes in when I try to fake the requests with Sinon: they aren't intercepted.

It comes down to [`sinonXhr.supportsXHR`](https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/util/fake_xml_http_request.js#L28) being calculated as false. This follows from [this line](https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/util/fake_xml_http_request.js#L25), where it looks for `XMLHttpRequest` on the local variable `global`. That local variable is passed in [here](https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/util/fake_xml_http_request.js#L614). In Node, `this` is an empty object, while `global` is where it **should** be looking at for the xml http request.

It seems unlikely to me that this would go unnoticed for so long, so maybe I'm configuring things oddly. Has anyone had success with jQuery+JSDom+Sinon's FakeXHRs?

jmeas:
Sinon was unable to detect XHR support when loaded in certain
Node environments. This was a regression introduced in #515
to resolve #508 and #511.

Resolves #657

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -611,4 +611,4 @@
         makeApi(sinon);
     }
 
-})(typeof self !== "undefined" ? self : this);
+})(global ? global : this);


jmeas:
Okay, so, it's been awhile since these issues have been looked at. Luckily, it's not too hard to follow along to see what's going on here.

Follow these steps to see this commit in action:

1. Revert the commit in #515 to see the bugs described by #508 and #511.
2. Now, introduce this commit. See that #508 and #511 are resolved.

To see that this resolves #657, you'll need to check the value of `global` within [fake_xml_http_request.js](https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/util/fake_xml_http_request.js). An easy way to do this is creating a new project with sinon as a dep. Create a script that does the following...

```js
// Set up Node to look as if it supports XHR requests
global.jsdom = require('jsdom').jsdom;
global.document = global.jsdom('<html><head><script></script></head><body></body></html>');
global.window = global.document.parentWindow;
global.XMLHttpRequest = global.window.XMLHttpRequest;

// load up sinon
var sinon = require('sinon');

// turn on the requests
sinon.useFakeXMLHttpRequest();
```

I admit that I didn't run that code myself, but it **should** display the issue. Before this commit, it would not find the XMLHttpRequest. After this commit, it does. Before #515 this script would also have worked.

jmeas:
:cry: c'mon travis.

jmeas:
Oh, well...it seems like master branch, and most recent PRs are failing, so it's hard to say whether my PR is breaking anything new.

What's up with the testing suite?

jmeas:
Gonna close. This still needs work to fix my browser tests. I'll reopen when/if it gets updated.

Issue #660, opened 2015-01-25
Build failing on master / recent PRs


jmeas:
:speak_no_evil:

jmeas:
It seems like maybe this was intentional given the commit msg [here](https://travis-ci.org/cjohansen/Sinon.JS/builds/43873418), but then maybe [this](https://travis-ci.org/cjohansen/Sinon.JS/builds/43874395) was supposed to stop the build from failing? Not too sure, but I figured I'd open an issue about it.

jmeas:
This was partially fixed by a recent commit, and then I went ahead and broke it in #661. But I've gone ahead and righted that wrong in #663

PR #661, opened 2015-01-25
Fix fakeXHR requests in Node.

### LINKED ISSUE #508: Failing tests when running in browsers

### LINKED ISSUE #657: JSDom + fakeXHR troubles

jmeas:
Sinon was unable to detect XHR support when loaded in certain
Node environments. This was a regression introduced in #515
to resolve #508 and #511.

Resolves #657 

===

Reimplements #659. This one actually seems to work. Refer to that PR for a description on how to test it.

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -611,4 +611,4 @@
         makeApi(sinon);
     }
 
-})(typeof self !== "undefined" ? self : this);
+})(typeof global !== 'undefined' ? global : this);


jmeas:
Test repo [here](https://github.com/jmeas/backbone.simple-auth).

Run `gulp` to run the node unit tests. They will pass.

Run `gulp test:browser` to start up a server for the client-side tests. Navigate to `http://localhost:7777/test/runner.html` to see that they, too, pass.

After you've done that, dig into `package.json` and install the normal version of sinon. The node tests will no longer pass.

cjohansen:
Thanks!

jmeas:
:dancers:

lib/sinon/util/fake_xml_http_request.js
```diff
@@ -611,4 +611,4 @@
         makeApi(sinon);
     }
 
-})(typeof self !== "undefined" ? self : this);
+})(typeof global !== 'undefined' ? global : this);
```
fearphage: @cjohansen The build is failing on master because there are single quotes here instead of the jscs required double quotes.

jmeas: lol oops : [  :speak_no_evil:

jmeas: I think it's failing already due to #660, but I'll go ahead and open a PR to correct this.

cjohansen: Thanks for discovering and fixing!

PR #666, opened 2015-01-29
sinon.assert with spyCall support 2.0

### LINKED ISSUE #580: Using sinon.assert with spyCall
Hi, it seems using `sinon.assert` methods with [spyCall](http://sinonjs.org/docs/#spycall) not possible now. 
when trying:
````js
sinon.assert.calledWithMatch(mystub.firstCall, 83, {asd: 1});
````
I'm getting err like this:
````
AssertError: update(83, [object Object]) is not a function
````
Is it correct behavior? 
and if so, any plans to implement spyCalls like spies for assertion?
thanks!

### PR DESCRIPTION
**Pull request by Asgaroth**
closes #580

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -30,13 +30,18 @@
                     assert.fail("fake is not a spy");
                 }
 
-                if (typeof method != "function") {
-                    assert.fail(method + " is not a function");
+                if (method.proxy) {
+                    verifyIsStub(method.proxy);
+                } else {
+                    if (typeof method != "function") {
+                        assert.fail(method + " is not a function");
+                    }
+
+                    if (typeof method.getCall != "function") {
+                        assert.fail(method + " is not stubbed");
+                    }
                 }
 
-                if (typeof method.getCall != "function") {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by cjohansen** [comment]
Congrats on requesting the pull of the beast!

Could you rebase with master and confirm that the tests run green?

**Comment by mroderick** [comment]
Since `assert.fail` throws errors, and we depend on this behaviour, I think we can make `verifyIsStub` a lot easier to read

Is this equivalent to the proposed new version?

```javascript
function verifyIsStub() {
    var method,
        i,
        l;

    for (i = 0, l = arguments.length; i < l; ++i){
        method = arguments[i];

        if (!method) {
            assert.fail("fake is not a spy");
        }

        if (method.proxy) {
            verifyIsStub(method.proxy);
        }

        if (typeof method != "function") {
            assert.fail(method + " is not a function");
        }

        if (typeof method.getCall != "function") {
            assert.fail(method + " is not stubbed");
        }
    }
}
```

Asgaroth:
@mroderick I tested your version but it fails, since it still tries to verify that the `spyCall` is a stub, which is why I have an `if..else`

PR #670, opened 2015-02-02
Update README.md with extra resources for Usage

mroderick:
Hoping to slow down the stream of usage questions posted as GitHub issues by pointing posters towards the mailing list and the irc channel.


mantoni:
:beers:

Issue #397, opened 2014-02-11
work cycles with setTimeout(func, 0) not testable


timotheeg:
So it's common in many libs to do a setTimeout(func, 0) to defer execution. func() itself might schedule another defer action, yet further, via one more setTimeout(another_func, 0). 

These 2 defers in the browser will not run within the same execution cycle (regardless of how much time the cycle actually takes...), but they will with sinon's fakeTimer.

For example, I expected the following test would be true within sinon's test suite:

```
    "understand browser work cycle": function() {
        this.clock = sinon.useFakeTimers();

        var cb2 = sinon.spy();
        var cb1 = sinon.spy(function() {
            setTimeout(cb2, 0);
        });

        setTimeout(cb1, 0);
        this.clock.tick(10);
        assert(   cb1.called);
        assert( ! cb2.called);

        this.clock.tick(10);
        assert(cb2.called);
    },
```

But that is not the case in sinon because the combined sinon functions tick() and firstTimerInRange() do not have a concept of work cycle.

Is it even desirable? Or is there another way to test work cycle defers with sinon?

mroderick:
I've verified the issue by running the tests provided.

fearphage:
I'm not sure what the solution is here, but as a workaround, couldn't you just capture the method to what to use before you mock them out?

```js
const originalSetTimeout = global.setTimeout;
this.clock = sinon.useFakeTimers();

originalSetTimeout(catpants, 0);
```

mroderick:
This hasn't seen a response from the original poster in more than 3 years. I am closing it

PR #674, opened 2015-02-05
Add test for issue #397

mroderick:
This test verifies issue #397.

It's commented out, because of previously mentioned issues with using `useFakeTimers` and `restore` more than once in the whole test run.


cjohansen:
:+1:

Issue #677, opened 2015-02-07
Losing stack trace on error


jj09:
Hi,

When I encounter an exception, Sinon is rethrowing it, but loosing stack:
![](http://snag.gy/hP1bi.jpg)
 

I found a workaround (rethrow ```exception.stack```): http://stackoverflow.com/questions/9006490/is-there-any-workaround-to-rethrow-an-exception-and-preserve-the-stacktrace-in-j

It works:
![](http://snag.gy/0w0vX.jpg)
 
Maybe we can change it in the library?

mroderick:
@jj09 thank you for reporting this.

Could you provide a code sample to help better understand the issue?

jj09:
@mroderick

Here is the sample: https://github.com/jj09/sinonjs-demo 
Clone, set revision to second commit and open index.html

Third commit fixes the issue.

I also created a pull request with the fix: https://github.com/cjohansen/Sinon.JS/pull/678

fatso83:
Re-reported as sinonjs/sinon-test#18

PR #678, opened 2015-02-08
Fix for: Losing stack trace on error #677

jj09:
Rethrowing exception.stack instead of exception.
This fix allow to not loose exception stack in Developer tools.

Rethrow exception:

Rethrow exception.stack:

### DIFF
--- a/lib/sinon/test.js
+++ b/lib/sinon/test.js
@@ -55,7 +55,7 @@
                 if (!doneIsWrapped) {
                     if (typeof exception !== "undefined") {
                         sandbox.restore();
-                        throw exception;
+                        throw exception.stack;
                     } else {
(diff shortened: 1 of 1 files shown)


mroderick:
:+1: 

Please fix the broken unit tests

mroderick:
@jj09 bump

jj09:
@mroderick Sorry for delay. I looked at the failing tests, and the fix is not straight forward. Requires me to dig in into sinon a bit. I will try to get back to that when I would have more time.

fatso83:
Had any chance to dig into this?

fatso83:
The `test.js` file was extracted into its own project: [sinon-test](https://github.com/sinonjs/sinon-test). No longer relevant here.

PR #679, opened 2015-02-10
(AMD) use explicit define in built version

mroderick:
This should fix issues in test runners (Jasmine 2 with weird grunt
require plugins, karma, etc.) where users get the much dreaded
[mismatched anonymous define modules](http://requirejs.org/docs/errors.html#mismatch)
error when loading Sinon.JS via RequireJS.

I am not yet sure how to test this, feedback welcome.
One idea would be to write a few tests using RequireJS, that can be run via specific npm statements. 

This idea came from https://github.com/cjohansen/Sinon.JS/issues/591#issuecomment-67765972

See also: https://github.com/cjohansen/Sinon.JS/issues/639

### DIFF
--- a/build
+++ b/build
@@ -48,7 +48,7 @@ formatio not found, skipping. To build with formatio support:
   File.open(file, "w") do |f|
     f.puts("(function (root, factory) {")
     f.puts("  if (typeof define === 'function' && define.amd) {")
-    f.puts("    define([], function () {")
+    f.puts("    define('sinon', [], function () {")
     f.puts("      return (root.sinon = factory());")
     f.puts("    });")
     f.puts("  } else if (typeof exports === 'object') {")


mroderick:
BTW: The solution works for me in a separate project, where test dependencies are loaded via some weird grunt-requirejs plugin.

mroderick:
If we can get this working, then I think we should re-consider the stance of having a built version in the repository. The combination of these two would solve a lot of problems for people, and wouldn't be that difficult to implement, since we already have `release.sh`

Munter:
+1 for checking in the build version. This will also immediately solve installation via bower

cjohansen:
Thanks! This is great. I'm still on the fence with regards to bundling the build, specifically because I don't want to encourage further use of Bower. What other cases would that solve?

mroderick:
> Thanks! This is great. I'm still on the fence with regards to bundling the build, specifically because I don't want to encourage further use of Bower. 

I am aware of that.

Bower is just one single tool. I am sure the future will bring more (and hopefully better) tools for dependency management.

> What other cases would that solve?

Since we're targeting three different ways of loading the script (`<script>`, AMD and CommonJS), it'd be nice if Sinon.JS just worked straight out of the box with no configuration for all of them (when loaded from GitHub or a git clone).

As developers, we have tooling that works well with npm / package.json, but for browsers we can't use Sinon from `node_modules` loaded via `<script>` or AMD.

Someone suggested targeted builds, I am not sure I think we need to go that far. But, a bit more convenience when managing Sinon.JS as a dependency would be nice.

Munter:
I don't understand the extreme aversion against bower. It's not the best in the world, but it's a pragmatic temporary alternative to git submodules, synmlinking into node_modules or similar hacks,

It's a convenience for those of us who would like to use Sinon in browser testing. I am one of those people. And currently I have to check in a hard copy of Sinon into my own repository in order to run my tests, thus risking losing track of versions, having stupid interns accidentally change the file etc.

cjohansen:
> I don't understand the extreme aversion against bower. It's not the best in the world, but it's a pragmatic temporary alternative to git submodules, symlinking into node_modules or similar hacks,

I don't understand why loading files from `bower_components` is fine while loading them from `node_modules` is a hack.

I dislike Bower because I find it to be completely superfluous, and it makes source code repositories installable packages with no option to opt out and no way to influence the installation process. Basically I have to dump generated crap in there for it to work, and take on supporting people running master, expecting it to always be stable.

I guess we could include a built version in the tagged commits. That way released versions would be available from git.

Munter:
Bower lets you configure where it puts the modules, enabling me to put it inside my web root so URLs work. I cannot do that with node_modules.

Always running against master is a falsum. Bower has semver capability running against semver tags in the repository, and by default bower will only get the HEAD if there are no semver tags at all on the repos.

Having the check in the compiled source code, yes, that is true. It is however possible (allthough tedious) to unregister a bower package registration and move it to a different repository, for example one which only contents are released versions over time

mroderick:
> and take on supporting people running master, expecting it to always be stable

Yeah, that's a retarded practice.

Don't get me started on using loose dependency versions in `package.json` and then inventing `npm shrinkwrap` as a fix to npm's biggest design flaw.

> I guess we could include a built version in the tagged commits. That way released versions would be available from git.

I guess then we would have to have a releases branch, so the built stuff never makes it into master.
That works for me.

cjohansen:
> I guess then we would have to have a releases branch, so the built stuff never makes it into master.
That works for me.

Either that, or tag the commit and pop it, e.g.:

```sh
git commit -am "Prepare release"
./build.sh
git add pkg
git commit -am "Sinon x.y.z"
git tag vx.y.z
git reset --hard HEAD^
git push && git push --tags
```

ryanflowers:
I see that the fix for #679 has been merged to master as of 8 days ago. Any idea when this will be pushed out into the npm package?

PR #680, opened 2015-02-11
Fix #650, always throw errors

### LINKED ISSUE #650: sinon.test swallows exceptions which occur in an asyc test case
I noticed if I'm wrapping an async test spec with sinon.test and if the test spec runs to some exception, the test will hang, without either printing the stack trace nor the timeout error.

Here is an example to illustrate the problem
````
sinon = require("sinon");

describe("test", function() {
  return it("should raise exception", sinon.test(function(done) {
    throw Error("an error");
    return done();
  }));
});
````
When you run this spec with mocha, the mocha will just sit there without any progression.

According to https://github.com/cjohansen/Sinon.JS/issues/610, I added a line sinon.defaultConfig.useFakeTimers = false in the example code. With this change,the test will timeout eventually, but still does not give any indication of the original exception occurred.

mroderick:
This fixed the issue for the test case reported as #650, and breaks no tests.

However, I can't quite figure out how to write a failing test that will get resolved with this pull request. 

I am open to ideas

### DIFF
--- a/lib/sinon/test.js
+++ b/lib/sinon/test.js
@@ -52,13 +52,13 @@
                     exception = e;
                 }
 
+                if (typeof exception !== "undefined") {
+                    sandbox.restore();
+                    throw exception;
+                }
+
                 if (!doneIsWrapped) {
-                    if (typeof exception !== "undefined") {
-                        sandbox.restore();
-                        throw exception;
-                    } else {
-                        sandbox.verifyAndRestore();
-                    }
+                    sandbox.verifyAndRestore();
                 }
 
                 return result;


cjohansen:
Hmm, I'm not sure either. Capture the case from #650 as a test?

mroderick:
The code in `lib/sinon/test.js` is very hard to follow. Perhaps it would be clearer, if we tidied it up a bit first

cjohansen:
Be my guest :)

peterjwest:
Wouldn't a test case look something like this?

  describe('sinon.test', function() {
    it('should forward exceptions from the test method when used asynchronously', function(done) {
      try {
        sinon.test(function(done) {
          expect(false).to.be.true;
          done();
        })();
      }
      catch(e) {
        expect(e).to.equal(
          'AssertionError: expected false to be true'
        );
      }
    });
  });

fatso83:
`sinon.test` has been moved into a package of its own, and has since been removed in #968. This PR should be filed against `sinonjs/sinon-test`. peterjwest's test could then possibly be included there.

peterjwest:
Issue raised on sinonjs/sinon-test

PR #683, opened 2015-02-14
Fixed assert when spyCall failed

Asgaroth:
Assertions on spycalls worked "well" when it didn't fail, but broken for assertions that fail, this fixes it.


mroderick:
Thanks :+1:

PR #684, opened 2015-02-15
Use vanilla Buster.JS

mroderick:
This PR is an attempt at switching to using versioned Buster.JS, instead of the patched version we're currently using.

There are a number of benefits to using versioned Buster.JS

* Easy upgrade!
* Focus rocket (I so miss focus rocket)
* Easily run tests in browsers

The only thing I couldn't make work, was the use of `buster.nextTick` in `test-test.js`, but only in node... it works fine in browsers.

I've left a `FIXME` comment to allow someone else to take a deeper looking into this.


mroderick:
@cjohansen @mantoni in this PR all tests pass on vanilla BusterJS when running in local Chrome, local PhantomJS and Travis PhantomJS. Currently, testing only in PhantomJS on Travis due to configuration challenges.

I would like to move to a situation where we run CI tests in this order:
* linting, in latest node
* unit tests in node 0.6, 0.8, 0.10, 0.12
* unit tests in latest PhantomJS
* unit tests in real browsers using SauceLabs or similar

If you're OK with it, to get this PR merged (since rebasing it is not that much fun). Once merged, I'll open a new PR and get to work on the test setup described above.

Oh, and I should probably squash down all the commits into a single one, once you're happy with the PR

mroderick:
The only thing I couldn't make work, was the use of `buster.nextTick` in `test-test.js`, but only in node... it works fine in browsers.

I've left a `FIXME` comment to allow someone else to take a deeper looking into this.

mroderick:
This is now ready for review, ping @cjohansen @mantoni 

It runs tests in node 0.6, 0.8, 0.10, 0.12 and the globally installed PhantomJS on Travis.

.travis.yml
```diff
@@ -14,11 +14,11 @@ before_install:
   - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.8" ]; then npm install -g npm@1.4; fi'
 
 before_script:
-  - 'if [ "x$TRAVIS_NODE_VERSION" != "x0.6" ]; then $(npm bin)/jscs .; fi'
+  # we only need to run jscs once per build, so let's conserve a few resources
+  - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.12" ]; then $(npm bin)/jscs .; fi'
```
mantoni: I've seen this somewhere else before. What is the "x" for?

Gruntfile.js
```diff
@@ -0,0 +1,22 @@
...
+                }
+            }
+        }
+    });
+
+    grunt.loadNpmTasks("grunt-buster");
+
+    grunt.registerTask("test", ["buster"]);
+    grunt.registerTask("default", "test");
+};
```
mantoni: I don't get this whole grunt thing. I'd personally prefer [using plain npm scripts as a build tool](http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/), but maybe you can convince me?

mantoni:
Hm, I can't run the tests because it can't find PhantomJS, although it's in my `$PATH`:

```
$ npm t
# ...
No code style errors found.
Running "buster:all" (buster) task
>> PhantomJS not found. Run `npm install phantomjs` to install.
Warning: Task "buster:all" failed. Use --force to continue.

Aborted due to warnings.
npm ERR! Test failed.  See above for more details.
$ which phantomjs
/usr/local/bin/phantomjs
```

Any ideas?

mantoni:
Update: `npm run ci-test` works well, everything green. Had one failure on the first run, but all attempts to reproduce failed. Some magic timing thing, I guess.

From looking around in the test sources it all looks great. Decent work @mroderick :+1:

mroderick:
> Hm, I can't run the tests because it can't find PhantomJS, although it's in my $PATH:
> 
> $ npm t
> # ...
> No code style errors found.
> Running "buster:all" (buster) task
> >> PhantomJS not found. Run `npm install phantomjs` to install.
> Warning: Task "buster:all" failed. Use --force to continue.
> 
> Aborted due to warnings.
> npm ERR! Test failed.  See above for more details.
> $ which phantomjs
> /usr/local/bin/phantomjs
> Any ideas?

Ah, yes ... I need to fix that, somehow I managed to forget about it ... it's been a busy week :)

Gruntfile.js
mroderick: I'll remove it again, I don't need it any longer (since I am now starting buster-server myself from ci-test.sh)

.travis.yml
mroderick: I am not sure either. I didn't write this version of this part of the `.yml` file

cjohansen:
I checked it out and ran `npm install`, then `npm test`. Seeing a bunch of lint errors from `.git/gitweb/*. Anyone else seeing this?

mantoni:
No. The linter passed without error for me.

mroderick:
Rebased

`rm -rf node_modules && npm install && npm test` passes on my machine

mroderick:
Once we're all happy with this PR, I'll squash it down to a single commit, so history won't be as crazy.

cjohansen:
I'm still getting warnings from `.git/gitweb` files... Removing them helps, but we should probably fix that some time. Although, I suspect this isn't even related to this PR. And I don't know where that `gitweb` directory came from :)

This looks solid to me, very nice to have this properly in order. Well done, @mroderick!

mroderick:
Squashed, ready for merging :rocket:

cjohansen:
:tophat:

Krinkle:
Commit ad81c89ed18f587f0b804a2c465926223034a5bd adds the following:

```diff
diff --git a/lib/sinon/util/timers_ie.js b/lib/sinon/util/timers_ie.js
--- a/lib/sinon/util/timers_ie.js
+++ b/lib/sinon/util/timers_ie.js
-function setTimeout() {}
-function clearTimeout() {}
-function setImmediate() {}
+if (typeof window !== "undefined") {
+    function setTimeout() {}
+    function clearTimeout() {}
+    function setImmediate() {}

-setTimeout = sinon.timers.setTimeout;
+    setTimeout = sinon.timers.setTimeout;

```

Afaik you cannot capture a function declaration inside a conditional block. JavaScript doesn't have that kind of block scope for functions. The assignment remains inside the if-block however, but the function would be hoisted. So unless this is different in IE, the function would be unconditionally (re)defined in the global scope as an empty function. And only if `window` exists will it then replace the empty function with Sinon's replacement.

mroderick:
I agree with Krinkle's analysis

Krinkle:
@mroderick I don't recall any issues since I'm using Sinon in a browser only, so those functions are expected to be created.

However I assume there is a reason that conditional exists and I'm fairly certain that either that behaviour is now broken or (if there is no problem) the conditional should be removed since the functions are created either way due to hoisting.

mroderick:
> the conditional should be removed since the functions are created either way due to hoisting.

That's a very good point.

.travis.yml
Krinkle: This is an ancient shell script coding pattern to defend against the case where the environment variable has an empty string (or doesn't exist).

If executed as shell command, it would crash in that case, because the `[` command (alias for the `test` command) would not be receiving the expected arguments. E.g. it would see `test`, `=`, `"0.12"` and then missing the third parameter (first equal signs and 0.12 as middle, instead of actual value as first, equal sings in the middle, expected as third).

See also:
* <https://stackoverflow.com/q/6852612/319266>
* <https://stackoverflow.com/a/2237103/319266>
* <https://stackoverflow.com/a/2188223/319266>

Long story short, if you know the shell in question is a Bash shell, and are using `[[` instead of `[` then you do not need this hack. There are also a number of other ways to solve it without this hack.

PR #686, opened 2015-02-17
Cleaning up test.js

fearphage:
Simplified test.js *a little*  based on cjohansen/Sinon.JS#680

### DIFF
--- a/lib/sinon/test.js
+++ b/lib/sinon/test.js
@@ -16,6 +16,8 @@
 
 (function (sinon) {
     function makeApi(sinon) {
+        var slice = Array.prototype.slice;
+
         function test(callback) {
             var type = typeof callback;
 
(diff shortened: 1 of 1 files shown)


fearphage:
Note: [line 37](686/files#diff-9b1597fcd86fcb2f9919de55d87e6af5R37) overrides the captured result from [line 49](686/files#diff-9b1597fcd86fcb2f9919de55d87e6af5R49). Is that intentional? This was the same way in old code. I think one of them needs to be killed. I couldn't see which one.

cjohansen:
Thanks! I'm not sure either of those has to be killed? This code still has improvement potential :) Thanks for contributing to making it better!

fearphage:
Well there's no reason to store the `result` if it's never used or referenced:

```` javascript
result = callback.apply(this, args.concat(sandbox.args));
````

That's wasted code. I say remove the assignment since no one references it or is able to reference it even.

cjohansen:
That value is later returned: https://github.com/cjohansen/Sinon.JS/blob/32c54084d381293e7452eb982167be9b805933d7/lib/sinon/test.js#L63

fearphage:
Ahhh, I missed that. Thank you.

Issue #690, opened 2015-02-24
restore doesn't restore properties properly


simonzack:
`object[property] =` occurs in numerous places in the code base to restore properties, but this does not restore properties properly in modern javascript. 

`Object.getOwnPropertyDescriptor` & `Object.defineProperty` should be used instead.

mantoni:
Can you provide an example where the current implementation fails?

simonzack:
var obj = {};
    Object.defineProperty(obj, 'nonWritable', {value: function(){
      return 1;
    }});
    var stub = sinon.stub(obj, 'nonWritable').returns(2);
    console.log(obj.nonWritable());
    stub.restore();

1 is printed instead of 2.

simonzack:
Sure, in the following code:

    var obj = {};
    Object.defineProperty(obj, 'nonWritable', {value: function(){
      return 1;
    }});
    var stub = sinon.stub(obj, 'nonWritable', function(){return 2;});
    console.log(obj.nonWritable());
    stub.restore();

1 is printed instead of 2.

mantoni:
This is an example of `sinon.stub` not working with `Object.defineProperty`, but I see your point ;-)

Issue #88, opened 2012-03-03
Wrap ES5 getters and setters


rvagg:
``` js
var o = {
  get foobar() { return 'foobar' },
  set foobar(s) { throw 'bzzt!' },
  foo: function() { return 'foo' },
  bar: 'bar'
}; // -> { foobar: [Getter/Setter], foo: [Function], bar: 'bar' }

sinon.mock(o).expects('foobar').returns('woohoo'); // doesn't because sinon.wrapMethod() does an object[property] which accesses the getter which returns a string
```

To support getters and setters you need to do a `Object.getOwnPropertyDescriptor(obj, prop)` for the object and then a recursive `Object.getPrototypeOf(obj)` if it doesn't exist, the descriptor has a `'get'` and `'set'` property where they are defined.

But, you may want to mock out just the getter or just the setter or both, so how do you specify that? Some options:

``` js
mock(o).expects('foobar'); // ambiguous, the setter? getter? or both?
mock(o).expects('get foobar'); // ambiguous, there may be a legitimate property named "get foobar"
mock(o).expectsGet('foobar'); // clear but more code in Sinon and more to document ?
mock(o).expectsSet('foobar'); // ditto
```

And also, do you have a convention in place already for dealing with features that may not be present in the current environment? If someone tries to do this in IE6 what should happen?

nunofgs:
:+1: I also just felt the need for this feature.

simonzack:
:+1: This needs to be re-opened. Especially with ES6 catching on, this is more important than ever.

mkay581:
+1 I wholeheartedly agree with @simonzack

machineghost:
+1

mroderick:
> I'd also like to see this reopened. It'd be useful specifically for stubbing fields of a wrapped location object.

Does https://github.com/mroderick/wrapple plug that gap for you?

qubyte:
Nowadays I use wrapple quite extensively, and I'd forgotten about commenting on this issue. That probably means that wrapple has plugged the gap as you say.

derwaldgeist:
+1 just stumbled upon sinon not supporting getter mocking

lucasfcosta:
Hi @derwaldgeist, are you using Sinon's latest version?

The latest one supports stubbing getters with both `sandboxes` and `sinon.stub`.

All you gotta do is use the `get` function, for example:

```js
var myObj = {
    prop: "foo"
};

createStub(myObj, "prop").get(function getterFn() {
[...]

PR #692, opened 2015-02-26
Feature: getters & setters

simonzack:
This PR addresses #88. I didn't touch mock as I've read about the plans to remove it in 2.0.

I've tried to make as little modification to the existing codebase as possible, and integrated getters & setters by providing an additional `type` parameter, which can take the value `get` or `set`. Using `defineProperty` and friends also solves #690.

Example usage:

    var obj = {};
    Object.defineProperty(obj, 'property', {
      get: function(){
        return 1;
      },
      configurable: true
    });
    var stub = sinon.stub(obj, 'get', 'property', function(){return 2;});
    console.log(obj.property);
    stub.restore();

@cjohansen I understand that you're not really a fan of getters & setters, but a lot of people do have a need for this (see above issue). Since sinon is the best (and only) mocking framework around, I think that getter & setter support will really benefit the community.

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -16,12 +16,20 @@
 "use strict";
 
 (function (sinon) {
+    function getPropertyDescriptor(object, property) {
+        var proto = object, descriptor;
+        while (proto && !(descriptor = Object.getOwnPropertyDescriptor(proto, property))) {
+            proto = Object.getPrototypeOf(proto);
+        }
+        return descriptor;
+    }
+
     function makeApi(sinon) {
         var push = Array.prototype.push;
         var slice = Array.prototype.slice;
         var callId = 0;
 
-        function spy(object, property) {
+        function spy(object, property, types) {
             if (!property && typeof object == "function") {
                 return spy.create(object);
             }
@@ -30,8 +38,16 @@
                 return spy.create(function () { });
             }
 
-            var method = object[property];
-            return sinon.wrapMethod(object, property, spy.create(method));
+            if (types) {
+                var methodDesc = getPropertyDescriptor(object, property);
+                for (var i = 0; i < types.length; i++) {
+                    methodDesc[types[i]] = spy.create(methodDesc[types[i]]);
+                }
+                return sinon.wrapMethod(object, property, methodDesc);
+            } else {
+                var method = object[property];
+                return sinon.wrapMethod(object, property, spy.create(method));
+            }
         }
 
         function matchingFake(fakes, args, strict) {
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -17,14 +17,24 @@
 (function (sinon) {
     function makeApi(sinon) {
         function stub(object, property, func) {
-            if (!!func && typeof func != "function") {
-                throw new TypeError("Custom stub should be function");
+            if (!!func && typeof func != "function" && typeof func != "object") {
+                throw new TypeError("Custom stub should be a function or a property descriptor");
             }
 
             var wrapper;
 
             if (func) {
-                wrapper = sinon.spy && sinon.spy.create ? sinon.spy.create(func) : func;
+                if (typeof func == "function") {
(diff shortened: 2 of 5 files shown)


cjohansen:
My issue with this is that if you need to stub out properties, you probably shouldn't be using getters, but functions. Getters are nice for accessing private data or making quick computations over other data, but in my mind, anything complicated enough warranting stubbing out in tests seems just like bad design for me. It has always been a design goal for Sinon not to enable or make it too easy to do stupid things.

Can you come up with some convincing use cases for this? If we do go for this, and I think @mantoni and @mroderick should pitch in as well, I'm not ecstatic about the suggested API. I think we should either have a dedicated function, like `sinon.stubProperty` or maybe something like:

```js
sinon.stub(object, 'property', propertyDescriptor);
```

That last one will be a little more verbose, which is a good thing. Maybe it'll make people think twice.

As a last note, I personally feel that a better solution is to have a section on stubbing properties in the docs, laying down the case that you shouldn't do it, would be preferable. But it's not up to me alone.

mantoni:
Like @cjohansen, I wouldn't want to add an API for this, for the said resons. However, I understand the need to stub a property that was defined with a getter. I'd use the existing sandbox functionality, but this does not work at the moment:

```js
var sinon = require('sinon');

var o = {};
Object.defineProperty(o, 'p', { get : function () { return 1; }, configurable : true });

var s = sinon.sandbox.create();
s.stub(o, 'p', 2); // TypeError: Cannot set property p of #<Object> which has only a getter
```

simonzack:
My own use case of this is to stub a configuration class, which abstracts over a key value store. Getters and setters are used to interact with the store. I then provide the getter with a value, test whether the setter is called with the appropriate value, called only once, etc. I could use functions instead, but I personally find the getter/setter approach to be simpler.

I think this is more of a design preference, so if it is not convincing enough, I'm sure many others would also like the chime in here.

I agree with @mantoni that adding new APIs is not ideal, but I cannot see how both getters and setters can be used with the existing api. I'll update my PR with @cjohansen last api as it feels cleaner than mine.

simonzack:
@cjohansen I've switched to your version of the interface.

cjohansen:
Build is failing

simonzack:
@cjohansen Could you run the build again? Only 0.10 is failing and it says "this potentially indicates a stalled build", could be a travis issue.

mroderick:
Re-started the build, it passed

cjohansen:
Thanks for your work on this. Let's hope people won't use it too much ;)

hswolff:
Just a heads up, but this change broke sinon.js for IE<9 as it uses ES5 methods. Trying to fix it locally, it's a current WIP.

sungwoncho:
Is this documented?

vitalets:
+1 for documentation

vitalets:
Finally working example to stub getter:
````js
var o = {
    get foo() { return 'foo' }
};

sinon.stub(o, 'foo', { get: function () { return 'foo stubbed' }});
console.log(o.foo); // foo stubbed
````

MaerF0x0:
How would someone use this to stub an existing setter/getter ?  
ex:  [Mongodb Collections](http://mongodb.github.io/node-mongodb-native/2.0/api/Collection.html) uses a getter for [collectionName](https://github.com/mongodb/node-mongodb-native/blob/2.0/lib/collection.js#L123)

novemberborn:
In the example above, `sinon.stub(o, 'foo', { get: function () { return 'foo stubbed' }})` returns a *spy*. As far as I can tell it's not possible to change the behavior of the getter without replacing it entirely. Is this intentional?

Similarly when using `sinon.stub({ get foo () {} })` it won't stub the getter, it only stubs functions on the object.

fatso83:
@novemberborn : I do not know. Maybe some of the others might.

fatso83:
@simonzack This seems to have introduced a regression affecting Chrome. See #1046. Could you have a look?

PR #694, opened 2015-03-01
Implement yieldsRight

### LINKED ISSUE #689: Feature proposal: yieldsLast
The current implementation of `yields` calls the first callback it identifies with the arguments specified. This is most of the time sufficient.

There are however situations where more than one function is passed into a function so yielding is limited here.

```js
// in any module
module.exports = {
    foo: function (anyFunction, anyData, callback) {
        var result = anyData;

        if (!result) {
            return callback(new Error('any error'));
        }

        return callback(null, result);
    }
};

// in the test
var sinon = require('sinon'),
    anyModule = require('./anyModule'),
    anyData = { any: 'data' },
    anyResult = { any: 'result' },
    anyFunction = function () {
        console.log('first function called:', arguments);
    }
    callback = function (error, result) {
        console.log('callback function called:', arguments);
    },
    foo;

foo = sinon.stub(anyModule, 'foo');
foo.yields(null, anyResult);

foo(anyFunction, anyData, callback);
```

As expected, the result is that `anyFunction` is called with the parameters specified in the `yields` statement. The callback function which is the last argument for `foo` is not called at all.

This could be solved by adding a `yieldsLast` feature that yields not for the first but for the last function argument found (`callback` in the example above). 

I'm aware of the `callsArg` function which can be used to call a callback function at a specified arguments index. For this to work you need to know the exact index of the callback which is not very convenient.

If there is interest in this feature I would look into it if I'm able to send a pull request for it.

lxanders:
I implemented this together with @fhinkel (thanks for the nice coding sessions!).

Fixes #689.

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -16,6 +16,8 @@
 (function (sinon) {
     var slice = Array.prototype.slice;
     var join = Array.prototype.join;
+    var useLeftMostCallback = -1;
+    var useRightMostCallback = -2;
 
     var nextTick = (function () {
         if (typeof process === "object" && typeof process.nextTick === "function") {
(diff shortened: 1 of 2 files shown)


cjohansen:
:+1:

fhinkel:
Thanks!

lee-houghton:
Thanks for this, it's very useful!

I couldn't find this in the documentation however. I only found this when I was looking at the source code trying to figure out if there was a simple way of stubbing functions that take callbacks, and discovered that it was already implemented. Is it an officially supported API that I can use without fear of it disappearing?

fatso83:
Honestly, I think this has been left out by mistake. We can try to do it right once #991 ships, I hope.

PR #696, opened 2015-03-01
Fix documentation for linting

### LINKED ISSUE #695: Small error in CONTRIBUTING.md
There is a line about checking the format:

```
$ $(npm bin)/jscs .
```

This is not the easiest way anymore as you added a script for this. So running `npm run lint` is the best way now which is not reflected in this documentation.

lxanders:
One note: There is a usage of the direct call to the `jscs` command in the travis configuration file. As this issue is only about the documentation I didn't change that.

Fixes #695.


mroderick:
Thanks!

PR #697, opened 2015-03-01
Fix spy.calledWith(undefined) to return false if it was called without args

mantoni:
Either this sneaked in somehow or it was never really working. I think this is a bug and here is the fix. Anything else to consider @cjohansen @mroderick?

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -36,7 +36,11 @@
             },
 
             calledWith: function calledWith() {
-                for (var i = 0, l = arguments.length; i < l; i += 1) {
+                var l = arguments.length;
+                if (l > this.args.length) {
+                    return false;
+                }
+                for (var i = 0; i < l; i += 1) {
                     if (!sinon.deepEqual(arguments[i], this.args[i])) {
                         return false;
                     }
@@ -46,7 +50,11 @@
             },
 
             calledWithMatch: function calledWithMatch() {
-                for (var i = 0, l = arguments.length; i < l; i += 1) {
+                var l = arguments.length;
+                if (l > this.args.length) {
+                    return false;
+                }
+                for (var i = 0; i < l; i += 1) {
                     var actual = this.args[i];
                     var expectation = arguments[i];
                     if (!sinon.match || !sinon.match(expectation).test(actual)) {
--- a/test/sinon/spy_test.js
+++ b/test/sinon/spy_test.js
@@ -44,6 +44,12 @@ if (typeof require === "function" && typeof module === "object") {
                 assert.isFalse(this.spy[method](1, 2, 3));
             },
 
+            "returns false if not called with undefined": function () {
+                this.spy();
+
+                assert.isFalse(this.spy[method](undefined));
+            },
+
             "returns true for partial match": function () {
                 this.spy(1, 3, 3);
                 this.spy(2);


cjohansen:
Don't think it ever worked, thanks for fixing!

cjohansen:
The pull request was green, but master failed on node 0.12: "Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL file://test/phantom/run-tests-in-phantom.js. Domains, protocols and ports must match."

Travis hiccup?

mantoni:
Travis hickup, I guess. When I added the Node 0.12 config to master, it was green as well.

bedney:
This patch broke Sinon's `calledWith` call for me. The problem is that, if this is invoked without the same number of arguments as the original call, this patch returns false and moves on.
So:
```
foo = function(bar, baz) {
}
spy = sinon.spy(foo);
spy(1,2);
spy.calledWith(1); <- used to return true, now returns false
```
This could be due to a misunderstanding on my part of the description of `calledWith`, but my reading of it certainly implies that this would work (i.e. partial matching): "Can be used for partial matching, Sinon only checks the provided arguments against actual arguments, so a call that received the provided arguments (in the same spots) and possibly others as well will return true."

I believe the original intent here was to check for undefined if it is the only argument (and the test case certainly implies that). But, IMHO, the example above should also handle undefined if it is given as part of the signature:

```
spy.calledWith(1, undefined); <- same as above, should return true
spy.calledWith(undefined, 2); <- should also return true
```

Thoughts?

mantoni:
I'm suprised this does not work for you since the test case right after the new one verifies that partial matches still work. Have to look into that when I come close to a computer again.

Regarding your suggestions I have to disagree. First because that would make it impossible to check whether a function is actually called with `undefined`, secondly you can use `sinon.match.any` to express that you don't care about a particular argument.

bedney:
Thanks for the quick reply - I appreciate it.

The problem is that, in a language like JS with variable arity, you can't predict how a particular function got invoked (the `baz` parameter in `foo()` above might be optional).

In the extreme case, where a Function has no formal signature (its parameter list is empty), I can still do this:

```
foo = function() {
for (var i = 0; i < arguments.length; i++) {
}
}

So, I guess I could write things like:
```
spy.calledWith(sinon.match.any,sinon.match.any,sinon.match.any,sinon.match.any,sinon.match.any,sinon.match.any,sinon.match.any,sinon.match.any);
```
and take a guess at how many arguments got passed to `foo()`, but wow is that ugly.

**Comment by mantoni** [comment]
Well, sinon does not perform check based on arity, it checks what was actually passed as arguments to the function. Maybe I don't really understand your point.

**Comment by mantoni** [comment]
I just tried your initial snippet with the latest version and it works as it should.

```js
spy.calledWith(1); // returns true
```

bedney:
Thanks again for looking at this. I'm digging deeper here to understand what's really happening and I guess it comes down to what is meant by 'partial matching'. I did a longer test case here, and was able to fully reproduce what I'm seeing.

Assume:

```
aFunc = function(required, optional) {
console.log('required parameter: ' + required);
console.log('optional parameter: ' + optional);
}

aFuncSpy = sinon.spy(aFunc);
aFuncSpy(1);
```
And then:
```
aFuncSpy.calledWith(1) <- returns true, as expected
aFuncSpy.calledWith(1, 2) <- returns false, as expected
```
So here's comes the subtlety:
```
aFuncSpy.calledWith(1, sinon.match(
			function(val) {
				//	If there was no argument for this position in the current call record, return true
				if (!val) return true;

				//	val needs to be less than 10
				return val < 10;
			}))
```

That used to return `true`, but now returns `false`. Here's the rub: As you can see, if there was no second argument in that particular call record, I can choose to return `true` (which I do). But, with your patch in place, this matcher will never be called because `calledWith` is early exited because the number of arguments to `calledWith` here is two and we've only ever called the spy with one argument.

This happens to me because I have a function that sometimes gets called with one argument, sometimes two, sometimes three... and sometimes I want to check those values and sometimes I don't and I just want to return `true` from my matcher function.

I might not be using the right technique for doing this - if you can think of a better way, I'm all ears :-).

Thank you.

mantoni:
Aha, now I understand.

The question is: do you really need a matcher with a conditional? You should maybe have an assert that does one check in one specific way. If you expect your spy to be called with only `(1)`, I would assert it with `calledWithExactly(1)`. If you expect your test case to invoke it with `val < 10`, I would use your matcher, but without the if statement in there.

Given this is a very special case and because there are ways to do the assertions in a different way, I don't see much value in having a special logic to make sure any given matchers are always invoked. But I'm all ears to arguments. /cc @cjohansen @mroderick

bedney:
Yea, as I stated earlier, this all comes down to what is meant by 'partial match'.

The argument I would make is that, given that JavaScript is a very flexible language whereby calls can be made with greater or fewer arguments than the given arity, this kind of flexibility is needed in Sinon to allow me to determine, on the fly via a matcher, whether a particular argument is valid or not (and therefore whether the call was valid or not). I may choose to use multiple matchers in the `calledWith` that leverage closures to share data about whether a particular argument is valid or not (i.e. my matcher might say "You have to be less than 10 if argument #1 is 'foo', otherwise you have to be null" or "you have to be less than 10 if this other condition that I'm testing outside of the matcher than I reference in the matcher is not equal to the phase of the moon" ;-)).

My point is that there's no possible way for Sinon to determine these kinds of conditions on its side of the matchers loop call... the matchers have to always be invoked and let the tester determine what is best for them. I would also argue that while, this might seem like a 'very special case', that it's not really, again given the incredibly flexible nature of JavaScript call semantics.

Anyway, my 2 cents. Thanks again for being responsive - I appreciate it.

bedney:
@cjohansen, @mroderick - Any thoughts here? This is currently breaking for me.

An alternative approach that I considered this past week would be to introduce a new method, something like `calledWithVarArgs` or some such.

cjohansen:
This is a long thread, a summary would be most welcome ;)

bedney:
Sorry for the delay in replying.

So, basically, I was relying on the fact that SinonJS matchers get called whether or not the number of arguments to the `calledWith()` call when I invoke it are the same as the number of arguments I originally use to call the spy with. I spy on a call that can take a variable number of arguments (not an uncommon pattern in JavaScript) and I use smart matchers to determine (based on a set of criteria which might include the matchers coordinating their result via a closure) whether a particular call record for that spy is correct.

This patch breaks my pattern because it alters the logic to say that if the number of arguments supplied to the `calledWith()` call exceeds the number of arguments for that call record, that it should just return false.

My contention here is that, in a sense, that's a 'first order filter' that Sinon should not be making on it's own - that it cannot make that assumption. IMO, Sinon should always invoke the matchers because it cannot know what arbitrarily complex rules the tester has coded into the matchers that are appropriate for that spied call.

bedney:
@cjohansen - Any thoughts?

Should we open a separate issue for this? I would be happy to provide a pull request with `calledWithVarArgs` if that's what's decided here.

mantoni:
I'm still not convinced that there is a need to change anything. If you still think there is an issue, please provide an example that demonstrates the issue so that we're all talking about the same thing.

PR #698, opened 2015-03-02
mock: fixes #613

### LINKED ISSUE #613: Sinon mock exploding in node 0.11
The error ([ref](https://travis-ci.org/bitcoinjs/bitcoinjs-lib/jobs/42356908#L1478))
``` javascript
TypeError: Cannot assign to read only property 'length' of function Object() { [native code] }
at Object.expects (/home/travis/build/bitcoinjs/bitcoinjs-lib/node_modules/sinon/lib/sinon/mock.js:76:22)
```

The impl. code ([ref](https://github.com/bitcoinjs/bitcoinjs-lib/blob/2.0.0/test/ecpair.js#L145-L146))
``` javascript
      it('passes the options param', sinon.test(function() {
        var options = { compressed: true }

        this.mock(ECPair).expects('constructor')
          .once().calledWith(options)

        ECPair.makeRandom(options)
      }))
```

Relevant lines of sinon.js code: https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/mock.js#L76

Not sure why this is blowing up, but I suspect its probably something related to `[].push` being used the way it is.  Perhaps `Array.prototype.push` would avoid this?

dcousens:
As outlined, this fixes #613.  I haven't had time to isolate a test case for this as of yet.

### DIFF
--- a/lib/sinon/mock.js
+++ b/lib/sinon/mock.js
@@ -17,7 +17,7 @@
 
 (function (sinon) {
     function makeApi(sinon) {
-        var push = [].push;
+        var push = Array.prototype.push;
         var match = sinon.match;
 
         function mock(object) {


dcousens:
It is weird since in most cases I've always seen `[].push === Array.prototype.push`, so this may in fact be V8 doing something *smart* and breaking.

mroderick:
> It is weird since, at least in any test case I have, [].push === Array.prototype.push, so this may in fact be V8 doing something "smart" and breaking.

Wow, that's just .. wow. 

Would it be possible for you to do a reduced test case and show that this is a bug in node/v8?

I am not saying we shouldn't fix this in the short term for Sinon.JS, but I'd like it to be short term and not become a permanent workaround in all our codebases, like a lot of stuff that we do because of old IE

mroderick:
I am sorry, you did say that you haven't had time to do an isolated test case yet. (I am lacking sleep)

dcousens:
Disregard, this is not root cause.  The reason my non-isolated tests passed was due to how a dependency was being pulled in locally.

PR #699, opened 2015-03-02
fix mismatches between @depends and require calls

neonstalwart:
i'm back! stuff is broken again...

i obviously don't update sinon often and it's because whenever i do, i find these kinds of problems and lose time having to fix them.

1. introducing lolex broke AMD because lolex doesn't load via AMD - it needs an extra build step or something... i don't know, i don't use it so i haven't figured it out but i can't load sinon from source via AMD because of it.  so, i've changed it to only load it (via `require("lolex")`) in node.  if someone wants to use it with AMD, they can load it with a script tag first or figure out how to make it work.
1. since i don't need lolex, i thought "no problem.  i can work around this by loading stub and spy directly and lolex won't even come into the picture".  unfortunately, when format, extend, etc were moved to their own modules, the calls to `require` were not added.

hopefully the dependencies can be kept working with keeping these things in mind:

1. nothing within sinon depends on sinon/lib/sinon.js - anything that needs a reference to the `sinon` handle gets it from sinon/lib/util/core.js or else there will be circular dependencies
1. the @depends annotations and the calls to `require` in `loadDependencies` need to be kept in sync.  ideally, you would do away with the annotations and just let the code dictate what the dependencies are so that you don't have to duplicate this but that's up to you.

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -1,7 +1,7 @@
 /**
  * @depend times_in_words.js
  * @depend util/core.js
- * @depend stub.js
+ * @depend match.js
  * @depend format.js
  */
 /**
@@ -197,6 +197,7 @@
     function loadDependencies(require, exports, module) {
         var sinon = require("./util/core");
         require("./match");
+        require("./format");
         module.exports = makeApi(sinon);
     }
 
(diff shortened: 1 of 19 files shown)


cjohansen:
> ideally, you would do away with the annotations and just let the code dictate what the dependencies are so that you don't have to duplicate this but that's up to you.

Yup, that's the goal. I did most of the required work a while back to port Sinon over to Browserify, but I have limited time to follow up, and now that branch is far behind... We'll get there eventually.

cjohansen:
Thanks for fixing up this, and sorry for screwing it up for you again.

Issue #498, opened 2014-06-24
Feature request: Optionally have fakeServer remove each response from the stack as used.


muffinresearch:
I'd like to be able to use the fakeServer but instead of it respond with the most recently added match, I'd like it to operate on a stack of responses such that with each response the response is removed.

This is nice for testing because it makes responses more explicit and makes sure things aren't just happening because you are receiving the same response more than once if called a second time.

Here's an example: If I want to test some code to ensure that if a poll response changes it's handled correctly it would be much easier if I could make multiple `respondWith` calls knowing that they will be used in the same order they were added.

To make this work there's really only a couple of minor changes required. I'd suggest this could be an optional feature. E.g. you'd pass some kind of consumeStack=true into create or something like that. This way it won't break backwards compatibility.

I see a similar ticket was raised #368 but the solution given assumes a more manual approach to set up. I don't want to have to set up responses around the code I'm testing as I'm testing at a higher flow level.

I'm currently overriding `processRequest` but that requires adding a lot of the other util functions which aren't otherwise exposed.

Another possibility might be to change the API so that `respondWith` can optionally remove the response on first use. Or have something like `respondWithOnce()`

I'd be happy to work on a PR for this but before doing so I wanted to sanity check the idea and check I'm not missing some obvious other way to do the same thing.

blimmer:
I would really like something like this, as well. `stub` has a [really nice API](http://sinonjs.org/docs/#stub-onCall) to do this with `stub.onCall`, and it would be great to be able to do something like this.

PR #721, opened 2015-04-01
Make fake server more readily customizable.

muffinresearch:
This is a step towards #498 by making more of the API to fakeServer more easily customizable.  From here I think it will be easier to go further to making request processing more pluggable.

It doesn't add any exising functionality so should be covered by existing tests.

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -23,70 +23,30 @@ if (typeof sinon == "undefined") {
 
 (function () {
     var push = [].push;
-    function F() {}
+
+    function RequestProcessor() {
+        this.requests = [];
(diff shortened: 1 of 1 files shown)


cjohansen:
So if I read this correctly it basically opens up for newing up a server with a completely different server prototype? I'm not sure I see the value of that - that would require you to override all the functionality, why not just use that other implementation directly? Am I missing something?

muffinresearch:
> So if I read this correctly it basically opens up for newing up a server with a completely different server prototype? I'm not sure I see the value of that - that would require you to override all the functionality, why not just use that other implementation directly? Am I missing something?

If anything it's really less than that. It's the same server code but the matching funcs and other utils are exposed.

PR #722, opened 2015-04-08
Reorder events dispatched in `readyStateChange` in `FakeXMLHttpRequest`

gavboulton:
Event listeners for `progress`, `load` and `readyStateChange` function in  `FakeXMLHttpRequest` are dispatched in a different order in comparison to a browser. Reorder the events dispatched to reflect general browser behaviour.

Add a test to verify the dispatch order.

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -367,19 +367,19 @@
                     }
                 }
 
-                this.dispatchEvent(new sinon.Event("readystatechange"));
-
                 switch (this.readyState) {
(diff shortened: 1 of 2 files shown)


gavboulton:
I added and then removed a file and then pushed to force a new travis test. All tests passed. When removing the two commits to return this to the original PR the tests have gone back to their failed state. I think the first test just timed out so if you can re-run this the tests will pass.

mroderick:
I triggered a new build in Travis, and all is green.

Thanks! :+1:

gavboulton:
Awesome. Glad I could help!

PR #728, opened 2015-04-16
Added onLastCall function

arvindrajnaidu:
When tests change the behavior of a stub, they have to mention the index of the call to the stub.

Most of the time a test is only concerned with the latest behavior, which is onCall(this.behaviors.length)

I have added this convenience function along with the existing onFirstCall, onSecondCall, onThirdCall functions.

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -133,7 +133,7 @@
             onThirdCall: function onThirdCall() {
                 return this.onCall(2);
             },
-            
+
             onLastCall: function onLastCall() {
                 return this.onCall(this.behaviors.length);
             }
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -115,7 +115,9 @@
             },
 
             onCall: function onCall(index) {
-                if (!this.behaviors[index]) {
+                if (typeof index === "undefined") {
+                    return this.onCall(this.behaviors.length);
+                } else if (!this.behaviors[index]) {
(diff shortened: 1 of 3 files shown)


mantoni:
Hm. This could be miss-understood as "the last time I call the stub" which is impossible to implement.
With your implementation "last" actually means "add one more behavior". Maybe some other wording works better: `onNextCall()` or just `onCall()` without arguments?

arvindrajnaidu:
I agree, it could be mis-understood. I was thinking of onMostRecentCall, but went with onLastCall because of symmetry with the onFirstCall function. 

I like the onCall without args idea, I'll update the PR.

mantoni:
After thinking some more about it, I don't like overloading `onCall` either. I think it doesn't work well when used on the first call:

```js
sinon.stub().onCall().returns(1);
```

Thoughts?

lib/sinon/stub.js
```diff
@@ -115,7 +115,9 @@
             },
 
             onCall: function onCall(index) {
-                if (!this.behaviors[index]) {
+                if (typeof index === "undefined") {
+                    return this.onCall(this.behaviors.length);
```
fearphage: This could be simplified a bit.

```js
if (typeof index === "undefined") {
  index = this.behaviors.length;
}
```

fatso83:
As this has been dead in the water for more than a year, and I think this doesn't really read very well, I am closing this.

PR #729, opened 2015-04-17
Use `files` in package.json

SimenB:
Simpler than using `.npmignore`

Before: http://hastebin.com/arocakulox.sh
After: http://hastebin.com/arocakulox.sh

The before is with IntelliJ files as well, but it kinda illustrates the advantage of whitelist vs. blacklist :smile:


package.json
```diff
@@ -19,6 +19,7 @@
     }
   ],
   "scripts": {
+    "pretest": "npm run lint",
```
SimenB: Also lint the code when running tests

.npmignore
```diff
@@ -1,12 +0,0 @@
...
-/AUTHORS
-/build
-/test
-/Changelog.txt
-/jsl.conf
-/release.sh
-
-/pkg
-!/pkg/sinon.js
-!/pkg/sinon-ie.js
```
SimenB: I don't have ruby installed, and I'd prefer not to. Should `pkg` be included as well?

cjohansen:
This commit includes a bunch of IDEA files. Also, the pretest hook is kinda off-topic, and is in conflict with another PR to be merged.

SimenB:
@cjohansen Yeah I noticed after pushing. I already fixed it.

Want me to strip pretest as well?

cjohansen:
Yup.

SimenB:
Done. What about the `pkg` directory?

cjohansen:
Please keep the PR focused on one thing.

SimenB:
@cjohansen Sure. Is it good now?

cjohansen:
Sure, thanks!

PR #736, opened 2015-04-23
Remove unused Gruntfile

SimenB:
From https://github.com/cjohansen/Sinon.JS/pull/684#discussion_r28623586, it was supposed to be removed


mantoni:
Thanks!

PR #740, opened 2015-04-24
Run unit tests against packaged version in CI environment

mroderick:
This PR allows us to run tests against the packaged version in CI environments (or anywhere we care to run the `ci-test.sh` script).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ before_install:
   - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.8" ]; then npm install -g npm@1.4; fi'
 
 before_script:
+  - rvm install 2.2.0
+  - rvm use 2.2.0 --default
+  - gem install juicer
   # we only need to run jscs once per build, so let's conserve a few resources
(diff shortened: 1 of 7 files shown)


mantoni:
I love this commit history. It's very entertaining. Although you're most likely banging your head against the wall right now.

mroderick:
> I love this commit history. It's very entertaining. Although you're most likely banging your head against the wall right now.

I am ... but I might be close

mroderick:
@mantoni good to go

mantoni:
Yay! :+1: Looks good to me.

I'm wondering if we should change the dependencies to use at least the `~` prefix instead of being super strict. Especially with `formatio` requiring `samsam` and then `samsam` now being required by Sinon directly as well .... If the versions are not identical, we end up with two versions in the tree.

mroderick:
> I'm wondering if we should change the dependencies to use at least the ~ prefix instead of being super strict. Especially with formatio requiring samsam and then samsam now being required by Sinon directly as well .... If the versions are not identical, we end up with two versions in the tree.

Let's burn that bridge when we get to it

cjohansen:
:+1:

PR #744, opened 2015-05-06
Improve RELEASE.md to reduce effort when cutting a new release

mroderick:
* Generate Changelog.txt via `git-extras`
* Generate AUTHORS file via `git-extras`
* Create new version via `npm version` (updates package.json and creates a tag)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,11 +1,38 @@
 # Rolling Sinon releases
 
-* Compile interesting highlights from `git log vX.Y.Z..master` into Changelog.md
-* Put new authors in AUTHORS
-* Update version in package.json
-* Commit and tag version
-* `npm publish`
-* Copy files into the static site:
+You'll need a working installation of [git-extras](https://github.com/tj/git-extras) for this.
+
+## Update Changelog.txt
+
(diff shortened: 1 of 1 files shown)


cjohansen:
For versions - it seems the bower part is still in there? Also, I suspect we can just use `npm version`?

```sh
       Run this in a package directory to bump the version and write the new data back to the package.json file.

       The  newversion  argument  should be a valid semver string, or a valid second argument to semver.inc (one of "patch", "minor", "major", "prepatch", "premi-
       nor", "premajor", "prerelease"). In the second case, the existing version will be incremented by 1 in the specified field.

       If run in a git repo, it will also create a version commit and tag, and fail if the repo is not clean.
```

As for the changelog: +1

mroderick:
> For versions - it seems the bower part is still in there?

I am not sure how I managed that ... I'll tidy that up ;)

> Also, I suspect we can just use npm version?

I'll look into that :)

mroderick:
I am happy with this PR now, thank you for feedback and ideas!

cjohansen:
Looks very nice :)

PR #745, opened 2015-05-06
Deprecate mock

mroderick:
Add a deprecation warning for whenever `mock` is called, saying that it will be removed in v2.0

### DIFF
--- a/lib/sinon/mock.js
+++ b/lib/sinon/mock.js
@@ -24,6 +24,10 @@
         var match = sinon.match;
 
         function mock(object) {
+            if (typeof console !== undefined && console.warn) {
+                console.warn("mock will be removed from Sinon.JS v2.0");
+            }
+
             if (!object) {
                 return sinon.expectation.create("Anonymous mock");
             }


mantoni:
:ship:

Krinkle:
I noticed this in the release notes for 1.15.0 but doesn't mention why or what should be migrated toward.

What are the plans for this in v2.0? Should people use stubs instead? Or a different library? Including a short mention of the intended migration path in the warning message would be useful.

And perhaps on http://sinonjs.org/docs/#mocks-api as well.

davethehat:
Seconded. Why is mock being deprecated? There are (valid) styles of unit testing that rely on mocks, this is a significant breaking change that warrants more information than an output warning when running our tests.

cjohansen:
I agree we should say something in the docs about this. Why are we removing mocks? Main reasons:

1) Mocks couple fake behavior with expectations and are inherently more complex than using stubs and assertions. Using mocks bears the risk of over-specifying your tests. For this reason, we wouldn't recommend using mocks in your tests.

2) The mock API in Sinon is not very good

3) None of the contributors use mocks or have any interest in the continued maintenance of this part of the code-base

Of course it sucks to have mocks removed if you depend on them, which is why we're issuing an early deprecation, so you have time to move over to spies and stubs.

We definitely need to cover this and how to migrate in the docs before actually removing them.

davethehat:
Thanks for the update. I agree that mocks are often misused and can lead to real tangles in tests: when used with discipline (and of course we use them with discipline :-) ) they're indispensable, and I can't help feeling that without them people will kludge stubs, spies and assertions to achieve the same ends. I'm happy to contribute as a maintainer of the mock API if that'll help ... !

mantoni:
@davethehat The solution that I would prefer is to have a separate module that uses Sinon stubs to create mocks. They don't need to be a "core" Sinon.JS thing.

nzakas:
My company uses mocks extensively, and its unlikely we will be able to switch over every project (that's a ton of work). Just to play devil's advocate here: is there any scenario where you would consider leaving mocks as part of the library? 

I'd also +1 a separately maintained plugin.

cjohansen:
I guess a plugin such as @mantoni suggests is the way to go. That would take very little effort to migrate to, and we could move forward with less baggage.

cjohansen:
@mantoni @mroderick I'm starting to think we should ease into this even more carefully. I suggest we remove the warning for now. We will then extract mocks into a drop-in replacement extension/separate module, add the warning along with docs/migration path in 2.0, and remove it in 3.0. Right now we're just alerting people to it's being removed with no reason and no course of action, it's a bit rough.

mantoni:
@cjohansen @mroderick I agree that it would be nicer if there was an alternative when printing a deprecation warning. I understand that @davethehat already volunteered to maintain the mocks. So if the existing code could be moved into a separate module, that co-exists with the current built-in solution, we could keep the warning and point people to the alternative. In v2 we can then remove the mocks entirely, as the message already says.

davethehat:
I think @mantoni's suggestion is excellent. I'm happy to help, but away for a few days. I'll check this thread early next week when I'm back - and thanks for the response and the enlightened discussion ... :-)

nzakas:
If someone will maintain mocks, can it stay in the core? :

TheSavior:
We should also use node's util.deprecate so that warning only shows up once. https://nodejs.org/api/util.html#util_util_deprecate_function_string

sungwoncho:
+1 for more information about what we should use instead.

sebv:
+1 to not having deprecation message showing up hundred times! (Specially since there is no quick fix.)

blimmer:
We use mocks, not extensively, but enough to make this change a non-negligible amount of work. I very much support keeping mocks in the core framework (if there's support from @davethehat and others) or creating a plugin.

cjohansen:
I just put out 1.15.4 which silences the deprecation warning until we have a proper plan and supporting documentation.

mroderick:
:+1:

mantoni:
Now that master is the going to be a 2.0 release, are we removing the mocks?

mroderick:
> Now that master is the going to be a 2.0 release, are we removing the mocks?

I would say no. Let's get CommonJS kinks ironed out. And then we can extract it to a separate module, if we still feel like it.

fatso83:
@mroderick Just to make sure I understand you correctly: _after_ the CommonJS kinks have been ironed out, we can still extract it to a separate module before releasing what will be the final 2.0?

mroderick:
Sorry, I can see how that was a bit unclear.

Let me try I again, I think we should

1. fix outstanding major issues in `master` (and 1.7 branch)
2. publish a `v2.0` release
3. Then consider what's next. 

One next move I can see, would be to extract mocks into their own (perhaps optional) module.

theghostbel:
Do you have a simple example how to spy+stub same method?
For example we have
`var obj = { api: function() { return 1 } }`

I know how to stub it
```
var apiStub = sinon.stub(obj , 'api')
apiStub.returns(0)
obj.api() // => 0
```

I know how to spy on it
```
var apiSpy = sinon.spy()
obj.api = apiSpy
... // somewhere we do obj.api()
apiSpy.should.have.be.calledOnce
```

But how to do same at same time? I'm thinking about this:
```
var apiSpy = sinon.spy()
obj.api = apiSpy

var apiStub = sinon.stub(obj , 'api')
apiStub.returns(0)

obj.api() // => 0, not 1 as original implementation
apiSpy.should.have.be.calledOnce

**Comment by fatso83** [comment]
Read the docs. A stub is a spy.
28. apr. 2016 12.59 skrev "theghostbel" <notifications@github.com>:

**Comment by theghostbel** [comment]
Thanks, @fatso83! Somehow I've missed that part from docs. For others: this is how to stub _and_ spy same method:
```
var obj = { api: function() { return 1 } }
var apiStub = sinon.stub(obj , 'api')
apiStub.returns(0)

obj.api() // => 0, not 1 as original implementation
apiStub.should.have.been.calledOnce
```
You just apply spy's methods on stub object, as you normally will do with spies created by `sinon.spy()`

Issue #749, opened 2015-05-13
An aborted XHR does not have responseHeaders


delitescere:
After aborting an XHR, there should be an empty `responseHeaders` object available. This is because after a send, the abort can occur. An `onReadyState` is fired and a callback that checks if `readyState == 4` that proceeds to check `responseHeaders`.

In a browser, there is an empty `responseHeaders` after an aborted XHR. In Sinon, it is `undefined`.

A PR is coming that fixes it (with tests).

mroderick:
The PR has been merged, thanks!

delitescere:
Cheers :beers:

PR #750, opened 2015-05-13
Provide an empty `responseHeaders` after an aborted XHR

delitescere:
Fixes issue #749

Also fixes a previously existing lint failure in `test/stub-test.js`.


mantoni:
Thanks! LGTM, but I don't know enough about the XHR stuff. @mroderick @cjohansen

cjohansen:
Thanks!

PR #757, opened 2015-05-29
Fixed bug where assertions don't handle functions that have a propert…

marco-ramirez:
I was bitten when trying to spy on jQuery and then using sinon.assert.  This fixes the issue by signing the proxy created by the spy with a property named isSinonProxy.  Then the assertions check for the presence of the isSinonProxy property before assuming a property named proxy is a sinon proxy.


cjohansen:
:+1:

PR #759, opened 2015-06-05
Add support for responseType=json

anatomic:
When working with FakeXMLHttpRequest I found that there were several instances where my tests were failing because xhr.response wasn't correctly parsed as JSON.

I have added code to support setting responseType = "json" and return a response inline with the spec here: https://xhr.spec.whatwg.org/#dom-xmlhttprequest-responsetype

There are tests around this new functionality and nothing else is reporting as a fail.

Thanks for the great package!

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -53,6 +53,10 @@
     };
     /*jsl:end*/
 
+    // Note that for FakeXMLHttpRequest to work pre ES5
+    // we lose some of the alignment with the spec.
+    // To ensure as close a match as possible,
+    // set responseType before calling open, send or respond;
     function FakeXMLHttpRequest() {
         this.readyState = FakeXMLHttpRequest.UNSENT;
         this.requestHeaders = {};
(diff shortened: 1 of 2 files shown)


cjohansen:
Thanks. The use of `defineProperty` will kill support for IE < 9. Not sure if we're there yet. @mantoni @mroderick ?

mantoni:
Personally, I think we should help killing anything < IE 9. However, we should still make it a major release then.

mroderick:
I feel the same way. A 2.0 release where we no linger actively support IE<9 would be good.

People can still use 1.x, which is awesome

anatomic:
Hey, if you like I could re-work this to include support < ie 9?

Thanks :)

mroderick:
> Hey, if you like I could re-work this to include support < ie 9?

Sure, that'd be great :)

anatomic:
Ok, I'll try and take a look at this a little later today.  Would you want me to kill off Object.defineProperty completely or would you prefer to check for the capability and provide a fallback if not present?

anatomic:
I've refactored to remove the Object.defineProperty usage - it has meant that the behaviour doesn't exactly match the specification but I think it's as close as will be needed. Obviously if support for < IE9 is dropped then I'd advocate going back to the first version as it offered a more robust copy of the intended behaviour.

Cheers! :)

mroderick:
:+1:

Issue #747, opened 2015-05-12
Capture stack trace for each spy invocation?


papandreou:
While brainstorming possible future additions for unexpected-sinon I found out that I would like to display (parts of) the stack trace as part of the context for each invocation. From a cursory glance it seems like `sinon.spy` doesn't currently capture that information.

How would you feel about receiving a PR that captures the stack by creating a new `Error` object on each invocation and stores it on the spy instance alongside `spy.args`, `spy.exceptions`, `spy.returnValues` etc.? It seems to be the only missing piece of the context.

mroderick:
That sounds interesting.

Could you give a (pseudo) code example of how it would look in use?

papandreou:
I had something like this in mind when eg. asserting that a spy was never called:

```
expected spy was not called
  expected invocations[
    spy(42, { foo: 'bar' }) at someFile.js:100:33,
    spy('baz') at someOtherFile.js:134:33
  ] to be empty

  invocations[
    spy(42, { foo: 'bar' }) at someFile.js:100:33, // should be removed
    spy('baz') at someOtherFile.js:134:33 // should be removed
  ]
```

But basically whenever an invocation is mentioned, it would carry the call site along with it.

lucasfcosta:
By seeing the merged PR I think this one can be closed. Right?

papandreou:
Oh yes, thanks for the reminder!

PR #763, opened 2015-06-12
Spy call: Capture the stack

papandreou:
I tried implementing my suggestion from #747 and added the top stack frame to the `spyCall.toString()` output.

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -149,6 +149,11 @@
                     "' since no callback was passed.", args);
             },
 
+            getStackFrames: function () {
+                // Omit the error message and the two top stack frames in sinon itself:
+                return this.stack && this.stack.split("\n").slice(3);
+            },
+
             toString: function () {
                 var callStr = this.proxy.toString() + "(";
                 var args = [];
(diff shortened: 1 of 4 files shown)


cjohansen:
This is very cool! Did you test it in multiple browsers? Browsers differ in how they format the stack, would be nice if it worked the same in all of 'em.

papandreou:
You are correct, browsers put different info into the stack traces, and the `stack` property isn't standardized. There are libraries for attempting to normalize and parse the string, and V8 even has a really nice callsite API for avoiding the stringification.

However, introducing something like that would add overhead and could lead to a "lowest common denominator" scenario where actually relevant information would end up being stripped.

In Sinon's case I think the only use case for capturing the stack is to display the non-Sinon stack frames as extra info in `spyCall.toString()`, and I think this PR accomplishes that well enough. But maybe you have other features in mind?

cjohansen:
I guess I was mostly thinking that we should make sure it works the same enough for the tests to pass in all browsers we support. Did you run the tests in multiple browsers?

papandreou:
Tested in:
* Chrome 37.0.2062.120 on Linux
* Opera 12.16 on Linux
* Firefox 26.0 on Linux

Tried hooking additional browsers up to the buster-server setup via browserstack.com, but couldn't get that to work, and not on master either. Got an "All slaves timed out" message.

papandreou:
Thanks :)

PR #768, opened 2015-06-21
Add callsFake

edeustace:
This adds `callsFake` function similar to the [jasmine's callFake](http://jasmine.github.io/2.0/introduction.html#section-Spies:_<code>and.callFake</code>)

The benefits of such a function is that you can add a mock function inline with the parameters of the function declared. I find this easier to read than `callsArgWith` as this requires you to think about the function you want to mock in a different way. I find it more natural and readable sometimes to drop in a replacement function.

example: 
   
```javascript

    var foo = {
       bar: function(a, b, c){
           // do stuff..
           return 'baz';
        },
        barWithCallback: function(a, callback){
          callback(null, 'ok');
        }
    }

    foo.bar = sinon.stub.create().callsFake(function(a, b, c){
      return 'baz';  
    });

    //more readable than callsArgWith(1, null, 'ok');?
    foo.barWithCallback = sinon.stub.create().callsFake(function(a, callback){
      callback(null, 'ok');
    });

   
```

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -137,7 +137,12 @@
             },
 
             invoke: function invoke(context, args) {
-                callCallback(this, args);
+
+                if (this.callFakeFn) {
+                    this.callFakeFn.apply(null, args);
+                } else {
+                    callCallback(this, args);
+                }
 
                 if (this.exception) {
                     throw this.exception;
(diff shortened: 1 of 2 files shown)


mantoni:
Can you explain what the difference is to these two use cases:


```js
foo.barWithCallback = sinon.stub().yields(null, 'ok');
```

or

```js
foo.barWithCallback = sinon.spy(function (a, callback) {
  callback(null, 'ok');
});
```

Thanks.

edeustace:
Ah - my apologies - I didn't see the `sinon.spy(function(){})` api when going through the docs.

PR #769, opened 2015-06-27
Add pre-commit hook to validate staged files with JSCS

mroderick:
@cjohansen @mantoni you should test this locally before merging.

1. `npm install`
2. Modify a file, making an invalid change that JSCS looks for
3. Modify another file, making an invalid change that JSCS looks for
3. Stage the changes from one of the files
4. `git commit`

If all works as intended, then JSCS will complain about just the staged changes, ignoring the other file

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,6 +79,8 @@ Node installed, and install Sinon's dependencies:
 
     $ npm install
 
+This will also install a pre-commit hook, that runs style validation on staged files.
+
 #### PhantomJS
 
 In order to run the tests, you'll need a [PhantomJS](http://phantomjs.org) global.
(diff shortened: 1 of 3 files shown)


cjohansen:
I tried it, but I'm unable to come up with something JSCS finds unacceptable :) I first removed a couple of semicolons - nothing. I then ran `npm run lint` just to make sure it caught them, but that also reported OK. Then I added this line to two files:

```js
fuck = yeah;
```

`npm run lint` (and the pre-commit hook) still consider this OK. Uhm?

mroderick:
I've pushed a better version of the pre-commit script, that is much, much simpler.

The pre-commit hook only does things for staged files, so if you didn't stage your changes, you won't see any errors.

Also, JSCS mostly cares about *style*, whitespace, indentation, etc. This is one of the reasons I'd like to also introduce ESLint to our toolchain.

Anyway, update, modify (try some whitespace in method signatures), stage, commit, fail.

cjohansen:
Working great, nice work!

PR #774, opened 2015-06-30
Quote yield as it's a reserved keyword in strict mode.

ngryman:
Hey,

Android 4.0 fails when using `yield` in strict mode.
This resolves the issue.

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -118,7 +118,7 @@
                 this.args[pos].apply(thisValue, args);
             },
 
-            yield: function () {
+            "yield": function () {
                 this.yieldOn.apply(this, [null].concat(slice.call(arguments, 0)));
             },
 


mroderick:

> Android 4.0 fails when using yield in strict mode.

WOW! That's a really odd behaviour ... would you mind adding a comment above the change to explain why this particular property is quoted, preferable with a url for further reading?

ngryman:
@mroderick Sure :)

`es5` strict mode reserves `es6` keywords: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Paving_the_way_for_future_ECMAScript_versions.

I'm currently developing https://github.com/ngryman/simulator and testing it with [saucelabs]. Everything was fine on every browser of my grid, except for Android 4.0 precisely (>= 4.0 work fine).

Here is my failing test: https://saucelabs.com/tests/8b03a25aad8a4e249ede19e734ed79c4.
When diving on Browserstack and showing the JavaScript console, there was the following message: `yield: Unexpected Strict mode reserved word`.
My patch did fix the error, and now everything works fine.

Other seem to have the same issue, it seems that it is a strange behavior of 4.0 stock browser:
- https://groups.google.com/forum/#!topic/cujojs/K8YHOX7Up3Y
- https://github.com/angular/angular.js/issues/7969

[saucelabs]: https://saucelabs.com/

mroderick:
> es5 strict mode reserves es6 keywords: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Paving_the_way_for_future_ECMAScript_versions.

I meant add it to your commit, so future readers won't "tidy up" the file :)

mantoni:
Hm, I remember putting this change in quite some time ago. Must have been lost on the way ...

ngryman:
@mroderick Done :)

mroderick:
Thank you!

PR #775, opened 2015-06-30
Add ESLint

mroderick:
This pull request adds validatation of all `.js` files with ESLint 0.24, allowing for `browser`, `node` and `amd` environments.

There is also a `eslint-pre-commit` script to validate **staged** files before committing.

Hopefully this PR will save everyone some time:
* reviewers won't have to review invalid code, or explain linting errors to contributors, it won't get committed and won't show up in pull requests
* contributors will get immediate feedback via the `eslint-pre-commit` script and won't have to revisit pull requests just to learn about linting errors after Travis builds them

##### Before
```bash
$ eslint .
✖ 3359 problems (3359 errors, 0 warnings)
```
##### After
```bash
$ eslint .
✖ 45 problems (0 errors, 45 warnings)
```

As @gyandeeps pointed out, upcoming `eslint@1.0.0` will have no default configuration, which means we will have to create one, or be stuck on `0.24` for awhile. I'll conduct some experiments with these configs, to see how well suited they are the Sinon.JS codebase

I've created an `.eslintrc` file with some defaults. We can always tweak it as we go along. Thankfully, ESLint has support for warnings.

@cjohansen @mantoni @gyandeeps I think this is ready for review now

--- 

Once we're happy to merge this PR, I'll squash it down to a single commit


cjohansen:
:+1:

gyandeeps:
My suggestion:
- Not use deprecated rules
- Since with release 1.0.0, all the rules will be turned off by default. It would be a good idea to build a config module and then use that. example: https://www.npmjs.com/package/eslint-config-ideal
- Use inline disable if we want a particular line to not throw errors. Docs: http://eslint.org/docs/user-guide/configuring#configuring-rules
example:
```js
alert('foo'); // eslint-disable-line no-alert
```

Do let me know if I can be of any help.

test/assert-test.js
```diff
@@ -577,7 +577,9 @@
             },
 
             "calls pass callback": function () {
-                var a = new this.stub();
+                /*eslint-disable no-new, new-cap*/
+                new this.stub();
```
gyandeeps: Perfect place to do inline disable. its cleaner.

test/call-test.js
```diff
@@ -795,7 +799,10 @@
 
                 try {
                     object.doIt();
-                } catch (e) {}
+                }
+                /*eslint-disable no-empty*/
+                catch (e) {}
```
gyandeeps: Wed can also do this rather than disabling the rule.
```js
catch (ex) {
    // Do nothing
}
```

mroderick:
> Use inline disable if we want a particular line to not throw errors. Docs: http://eslint.org/docs/user-guide/configuring#configuring-rules example:
> `alert('foo'); // eslint-disable-line no-alert`

I was not aware of that, thank you for the tip! I've pushed new commits that tidies this up :+1:

mroderick:
> Since with release 1.0.0, all the rules will be turned off by default. It would be a good idea to build a config module and then use that.

I was also unaware of that. I took a look around, and https://github.com/walmartlabs/eslint-config-defaults has some appealing features for me.

Let us see how the rest of this PR turns out, before I start adding more complexity.

.eslintrc
```diff
@@ -0,0 +1,186 @@
+{
+    "ecmaFeatures": {},
+    "parser": "espree",
```
gyandeeps: You can remove this as you always want to use the eslint default parser. And as of now its espree. But I think we dont have to mention this.

gyandeeps:
LGTM.
Suggestion: I would activate more stylistic rules as it helps the project to have consist styled code. I find it very helpful.

mroderick:
> Question: Is this project going to be linted by 2 tools? Why I ask is because I see jscs also.

If ESLint can do everything that JSCS can do, then I would rather just have one tool.

gyandeeps:
Sounds good.
This PR looks good. :+1:

mroderick:
@cjohansen @mantoni I think this is ready

cjohansen:
I merged and pushed this to master, but it seems GitHub is not following along. Can you verify that master is like it should?

mroderick:
Not entirely sure how you managed that, but the changes look to be in there.

mroderick:
Thanks to @gyandeeps for in-depth review of this!

gyandeeps:
np :+1:

PR #785, opened 2015-07-20
fakeServer.create accepts configuration settings

### LINKED ISSUE #784: Pass options to fakeServer.create
One annoyance I have with the otherwise awesome fakeServer is that it does not take initialization arguments. What I want to be able to do is:

```js
server = sinon.fakeServer.create({
  autoRespond: true,
  autoRespondAfter: 1
});
```

Instead of; 
```js
server.autoRespond = true;
server.autoRespondAfter = 1;
```

This seems pretty trivial to implement or am I missing something?

maxcal:
This allows fakeServer to be created with a whitelisted set of settings and adds a configure method which allows the server to be configured from an object literal. Fixes #784

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -105,7 +105,7 @@ if (typeof sinon == "undefined") {
                 setting;
                 config = config || {};
                 for (setting in config) {
-                    if (whitelist[setting]) {
+                    if (whitelist[setting] && config.hasOwnProperty(setting)) {
                         this[setting] = config[setting];
                     }
                 }
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -79,8 +79,9 @@ if (typeof sinon == "undefined") {
 
     function makeApi(sinon) {
         sinon.fakeServer = {
-            create: function () {
+            create: function (config) {
                 var server = create(this);
+                server.configure(config);
                 if (!sinon.xhr.supportsCORS) {
                     this.xhr = sinon.useFakeXDomainRequest();
                 } else {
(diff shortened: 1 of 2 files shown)


mroderick:

lib/sinon/util/fake_server.js
```diff
@@ -94,7 +95,21 @@ if (typeof sinon == "undefined") {
...
+            configure: function (config) {
+                var whitelist = {
+                    "autoRespond": true,
+                    "autoRespondAfter": true,
+                    "respondImmediately": true,
+                    "fakeHTTPMethods": true
+                },
+                setting;
+                config = config || {};
+                for (setting in config) {
```
mroderick: Sinon.JS needs to be very defensive and shouldn't fail even when end users do questionable things like extending `Object.prototype`.

With that in mind, you'll need to protect `for..in` with `hasOwnProperty`

The [PR to add ESLint for validation](https://github.com/cjohansen/Sinon.JS/pull/775) would have caught this, but is not merged yet.

test/fake-server-test.js
```diff
@@ -0,0 +1,31 @@
...
+    "use strict";
+
+    var buster = root.buster || require("buster"),
+        sinon = root.sinon || require("../lib/sinon"),
+        assert = buster.assert,
+        refute = buster.refute;
+
+    buster.testCase("sinon.fakeServer", {
+        ".create": {
+            "allows valid init settings" : function () {
```
mroderick: Try to avoid using words like "valid" and "correct" in test names. Be specific about what your expectations are.

```javascript
"allows 'autoRespond' setting": function() {
```

You should create a test for each of the white listed settings

PR #788, opened 2015-07-20
Pass options to fakeserver create 784

maxcal:
This PR fixes the issues from my previous PR #785.
This allows fakeServer to be created with a whitelisted set of settings and adds a configure method which allows the server to be configured from an object literal.

So sorry for the repeated PR's. Realized that in order to safeguard completely against `Object.prototype` extensions I need to check `hasOwnProperty` on both the whitelist and settings.


maxcal:
issue #784

mroderick:
Very nice. Could you squash it down to just one commit before we merge it?

PR #789, opened 2015-07-20
fakeServer.create accepts configuration settings

### LINKED ISSUE #784: Pass options to fakeServer.create

maxcal:
This allows fakeServer to be created with a whitelisted set of settings and adds a configure method which allows the server to be configured from an object literal. Fixes #784

All squished into one nice commit.

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -79,8 +79,9 @@ if (typeof sinon == "undefined") {
 
     function makeApi(sinon) {
         sinon.fakeServer = {
-            create: function () {
+            create: function (config) {
                 var server = create(this);
+                server.configure(config);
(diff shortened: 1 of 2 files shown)


mroderick:
For future pull requests, it would be easier for both you and reviewers, if you rebase and push to the same branch.

Thank you for your contribution :+1:, let's see if we can get @cjohansen to bake a new release soon :fireworks:

PR #790, opened 2015-07-20
move tests for fakeServer into existing test file

maxcal:
Also removes the duplication of sinon.create from fakeServer.


mroderick:
Thanks!

PR #795, opened 2015-07-28
Fix #777: Support non-enumerable props when stubbing objects

### LINKED ISSUE #777: Proposal: Stub non-enumerable properties when only object is given
Using sinon v1.15.3, the following code snippet does not work

```javascript
class Foo {
  doSomething() {
    var thing = 0;
    // do stuff with thing...
    return thing;
  }
}

const fooStub = sinon.createStubInstance(Foo);
fooStub.doSomething.returns(1);
```

The last line throws a `TypeError`, saying that `doSomething` does not have a `returns` method. This is due to the fact that function defined in class expressions are attached as [non-enumerable properties](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-classdefinitionevaluation) (see step 21.a), but [within sinon.stub](https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/stub.js#L51-L55) a `for..in` loop is used, meaning any non-enumerable property is excluded. 

I think it would help with the future-forwardness of sinon to use something like `Object.getOwnPropertyNames` instead of `for..in` when using stub with only an object, because it would allow `createStubInstance` to interop seamlessly with class definitions, and I would be happy to submit a PR for this. I was just wondering if there are any arguments against stubbing non-enumerable properties and if so what would be a good solution to allow stubbing them in certain cases?

traviskaufman:
This commit modifies sinon.stub() such that when only an object is
given, all non-enumerable properties as well as enumerable properties
that are functions will be stubbed.

The iteration over the object is accomplished by using sinon.walk, a new
utility function that supports iterating over all properties of an
object up the prototype chain, falling back to for..in when
getOwnPropertyNames is not available.

This functionality allows sinon.stub() - and especially methods like
sinon.createStubInstance() - to be future-forward by facilitating the
fact that methods defined in class declarations are non-enumerable by
default. It also allows stubbing of native objects such as Arrays and
Dates without having to explictly pass property names.

### DIFF
--- a/test/walk-test.js
+++ b/test/walk-test.js
@@ -97,9 +97,9 @@
                 placeholder; // eslint-disable-line no-unused-vars
 
             Target.prototype.foo = 15;
-            delete Object.getOwnPropertyNames;
+            Object.getOwnPropertyNames = null;
 
-            // Different environments are inconsistent in how they handle for..in, therefore we
+            // Different environments may be inconsistent in how they handle for..in, therefore we
             // use it to track the number of expected calls, rather than setting it to a hard
             // number.
             /* eslint-disable guard-for-in */
--- a/test/walk-test.js
+++ b/test/walk-test.js
@@ -1,0 +1,115 @@
+(function (root) {
+    "use strict";
+
+    var buster = root.buster || require("buster"),
+        sinon = root.sinon || require("../lib/sinon"),
+        assert = buster.assert;
+
+    buster.testCase("sinon.walk", {
+        "should call iterator with value, key, and obj, with context as the receiver": function () {
+            var target = Object.create(null),
+                rcvr = {},
+                iterator = sinon.spy();
+
+            target.hello = "world";
+            target.foo = 15;
+
+            sinon.walk(target, iterator, rcvr);
+
+            assert(iterator.calledTwice);
+            assert(iterator.alwaysCalledOn(rcvr));
+            assert(iterator.calledWithExactly("world", "hello", target));
+            assert(iterator.calledWithExactly(15, "foo", target));
+        },
+
+        "should work with non-enumerable properties": function () {
+            var target = Object.create(null),
+                iterator = sinon.spy();
+
+            target.hello = "world";
+            Object.defineProperty(target, "foo", {
+                value: 15
+            });
+
+            sinon.walk(target, iterator);
+
+            assert(iterator.calledTwice);
+            assert(iterator.calledWith("world", "hello"));
+            assert(iterator.calledWith(15, "foo"));
+        },
+
+        "should walk the prototype chain of an object": function () {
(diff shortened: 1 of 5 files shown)


traviskaufman:
:( This is failing because of how PhantomJS 1.9.8 deals with object enumeration. I will look into it and see if I can make the tests more robust.

traviskaufman:
@mantoni @cjohansen possible to get some eyes on these changes?

test/walk-test.js
```diff
@@ -0,0 +1,126 @@
...
+                Target = function Target() {
+                    this.hello = "world";
+                },
+                target = new Target(),
+                rcvr = {},
+                iterator = sinon.spy(),
+                err = null,
+                numCalls = 0,
+                placeholder; // eslint-disable-line no-unused-vars
+
```
cjohansen: Nitpicking, but I prefer indvidual vars

traviskaufman: No problem. I noticed in some of the other tests the vars are grouped this way, which is the reason I opted for it. Would you prefer individual var decls throughout all the tests?

cjohansen: Sloppy reviewing on my end then probably :) I would prefer individual declarations.

mroderick: I'll update the `.eslintrc` to only allow individual `var` statements

mroderick: I've created #809 to update `.eslintrc` so ESLint will give better feedback about this.

nii236:
This will allow ES2015 classes to be stubbed out correctly, yes? Until the pull request goes through will there be an alternative workaround?

mantoni:
@nii236 This PR can't be merged. Please rebase.

traviskaufman:
@mantoni @nil236 yeah sorry for the delay here, starting a new job and whatnot I've been really busy. I'll re base and re-submit when I get home tonight.

nii236:
@traviskaufman: Thanks! I look forward to it!

traviskaufman:
@mantoni @mroderick made all the changes but the current build failed

```
Error: PhantomJS 1.9.8, Linux | sinon.FakeXMLHttpRequest | .setResponseHeaders | does not call readyStateChange if sync
  Error: INVALID_STATE_ERR - 4
    at verifyRequestOpened (./pkg/sinon.js:4928)
      at setResponseHeaders (./pkg/sinon.js:5144)
      at ./test/util/fake-xml-http-request-test.js:519
Running 1490 tests in 1 runtime ... 88% done
```

It passed for me on my local machine, so this looks like a false negative. Would one of you mind rebuilding?

mantoni:
Nice work! Thank you :+1:

nii236:
Thanks! :+1:

traviskaufman:
np! Thanks for merging!

Issue #796, opened 2015-07-28
Multiple expectations and already passing expectation results in unexpected call error


Druotic:
Sinon version: 1.14.1
Environment OSX node v0.10.38
Gist: https://gist.github.com/Druotic/ca90d0f80c63d442f9ad

If there are two or more expectations present on a mock, one of which uses atMost/atLeast (some constraint which passes early and can continue to pass), when another call is made which continues to satisfy the already met expectation, an unexpected call error is thrown. See gist for example.

I am almost certain this is a bug, but I just wanted to check before I submit a fix to ensure that I am not simply misusing atMost/atLeast somehow.  

Edit: #442 seems to address this problem, but is collecting dust.

Druotic:
Fix has been merged. Closing.

PR #797, opened 2015-07-28
Fix #442/#796: Rebase/cleanup of existing fix: "Fixing a 'withArgs' and 'atLeast' issue with mock.js"

Druotic:
I have simply rebased @stevesouth's commit and fixed some linting, file naming, etc to fit current expectations.  Note that I did not write this logic, I assumed it has already been reviewed due to the discussion in #442.  This should fix #442 and #796


Druotic:
@cjohansen I was hoping that this might get merged pretty soon since it has already been reviewed/discussed in #442, without the fix atMost/atLeast only works in the most basic of cases (when there is only one expectation).  I'm surprised this bug hasn't been more visible before now.

PR #801, opened 2015-08-03
Reapply: Returns supersedes previous throws

dlau:
Allows overriding a `throws()` call with `returns()`. Currently `stub.throws().returns()` will throw instead of returning.

original: https://github.com/cjohansen/Sinon.JS/pull/644
ref: https://github.com/cjohansen/Sinon.JS/issues/643


hurrymaplelad:
:+1: Thanks @dlau

mroderick:
Thanks @dlau @hurrymaplelad

PR #809, opened 2015-08-17
Set ESLint `one-var` to error (2)

### LINKED ISSUE #807: Update `.eslintrc` to only allow single var statements
The ESLint confit should not be ambiguous, when there's a clearly stated preference

mroderick:
This PR sets the ESLint `one-var` setting to be an error (2).

```javascript
"one-var": [2, {
    // Exactly one declaration for uninitialized variables per 
    // function (var) or block (let or const)
    "uninitialized": "always",
    // Exactly one declarator per initialized variable declaration
    // per function (var) or block (let or const)
    "initialized": "never"
}],
```

This means that correct formatting is as follows:

```javascript
var initialized1 = 1;
var initialized2 = 2;
var uninitialized1,
    uninitialized2;
```

I believe this matches @cjohansen's opinions on `var` declarations.

Fixes #807

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -149,7 +149,12 @@
         "newline-after-var": 0,
         "object-curly-spacing": [0, "never"],
         "object-shorthand": 0,
-        "one-var": 0,
+        "one-var": [2, {
+             // Exactly one declaration for uninitialized variables per function (var) or block (let or const)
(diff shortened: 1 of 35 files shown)


cjohansen:
:+1: It sure does!

Issue #569, opened 2014-10-01
Feature Request: make XHR2 testing easier


bitinn:
A notable issue being that fakeServer and fakeXHR do not populate `xhr.response`, this is quite critical to  [XHR2](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) as users can define `xhr.responseType` and expect `xhr.response` to contain formatted data:

eg. `xhr.responseType = 'json'` will result in a Javascript object of parsed json response at `xhr.response`, without needing to do extra `JSON.parse`; the gain is more significant for `Blob` type binary data, where dev no longer need to [hack mime type](http://www.html5rocks.com/en/tutorials/file/xhr2/) then read data as plain/text.

XHR2 support is [fairly wide nowadays](http://caniuse.com/#feat=xhr2), it would seem like a good time to add such support.

mroderick:
@nicolai86 all help is greatly appreciated :+1:

bitinn:
I think people mostly use Fetch instead of XHR2 (much easier to mock, I hope), closing it for good.

PR #813, opened 2015-08-24
Add support for arraybuffer, blob responseTypes

henryptung:
Based on spec: http://www.w3.org/TR/XMLHttpRequest/

Binary response types (`arraybuffer`, `blob`) expect binary strings (strings with `charCode`s in the range `0x00`-`0xFF`) in place of normal strings.

Align slightly more closely to spec (e.g. `responseText` should be `""`, not `null` when opening an XHR of `text` or unspecified `responseType`; `responseXML` should only be populated when `responseType` is `""` or `"document"`, etc.).

Does not support browsers that only implement the `BlobBuilder` API (like PhantomJS 1.9.x).

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -34,6 +34,8 @@
     var supportsProgress = typeof ProgressEvent !== "undefined";
     var supportsCustomEvent = typeof CustomEvent !== "undefined";
     var supportsFormData = typeof FormData !== "undefined";
+    var supportsArrayBuffer = typeof ArrayBuffer !== "undefined";
+    var supportsBlob = typeof Blob === "function";
     var sinonXhr = { XMLHttpRequest: global.XMLHttpRequest };
     sinonXhr.GlobalXMLHttpRequest = global.XMLHttpRequest;
     sinonXhr.GlobalActiveXObject = global.ActiveXObject;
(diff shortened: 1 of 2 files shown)


henryptung:
Targets #569.

henryptung:
As a side note, might be worth setting up a Travis CI build with PhantomJS 2 so the `ArrayBuffer`/`Blob` support can actually be tested. I tested locally with PhantomJS 2.0.0 and `npm run test`, FWIW.

Could also try to backfill for browsers that only support `BlobBuilder` (like PhantomJS 1.9.x) but that would require me not to be lazy.

cjohansen:
Wow, nice work! Looks solid to me, @mantoni @mroderick anything to add?

mantoni:
Nothing to add from my side. Good stuff :+1:

mroderick:
I've created #815 to add PhantomJS 2.0 to our test stack.

mroderick:
Nice work @henryptung :rocket:

henryptung:
Whoa, fast turnaround! Haven't had a chance to keep tabs, but thanks a ton guys :)

pghalliday:
This is awesome - will it be released soon? ;)
I made a dodgy hacked fork so i could install from git repo :o

henryptung:
Warning: there's been a regression since https://github.com/sinonjs/sinon/issues/875 changed interpretation in arraybuffer/blob mode back to text (rather than binary string). Fix forthcoming, but I'm currently a bit busy; will come back to this in a week or two.

Issue #702, opened 2015-03-07
sinon.test is broken


mohayonao:
on v1.13.0

```js
describe("sinon.test", function() {
  it("works?", sinon.test(function(done) {
    done(); // undefined is not a function
  });
});
```

https://github.com/cjohansen/Sinon.JS/commit/5b47ab5b500de1fbeb229bc1bc93f4c272e46740#diff-9b1597fcd86fcb2f9919de55d87e6af5L68

I think the argument `callback` of `sinonAsyncSandboxedTest` is necessary for mocha async test.

SamVerschueren:
Ran into this issue aswell today. Reverted back to v1.12.2 for now.

PR #819, opened 2015-08-26
Fixed async tests that had async bits commented out

fatso83:
These tests were expecting buster to have been extended with the utils from buster-core, but this failed on Node, but since the failing require() call exception was swallowed by the buster test runner it was hard to debug why.

Explicitly requiring buster-core fixes the problem.

Related to #702


mroderick:
Very nice! Thanks!

Issue #818, opened 2015-08-26
sinon.test() does not pass on errors to callbacks


fatso83:
`sinon.test(fn)` is supposedly supposed to do the same as `fn` (except for sandboxing), but this is not true when it comes to callbacks, as the following tests are meant to illustrate.

```javascript
it('works', function () {

    var normalFn= (function (callback) {
        callback(new Error('myerror'));
    });

    function errorChecker (err) {
        expect(err).to.not.be(undefined);
        errorChecker.called = true
    }

    normalFn(errorChecker)
    expect(errorChecker.called).to.be(true);
});

it('shows sinon.test bug - as this fails', function () {

    var sandboxedFn = sinon.test(function (callback) {
        callback(new Error('myerror'));
    });

    // is never called
    function errorChecker (err) {
        expect(err).to.not.be(undefined);
        errorChecker.called = true
    }

    sandboxedFn(errorChecker)
    expect(errorChecker.called).to.be(true);
});
```

I think the bug is caused by [some code in `lib/sinon/test.js`](https://github.com/cjohansen/Sinon.JS/blob/bff6358be696d3c662557cef96345468007762eb/lib/sinon/test.js#L36), but as I am not quite sure how it works, I haven't fixed it yet.
```javascript
if (typeof oldDone === "function") {
    args[args.length - 1] = function sinonDone(res) {
        if (res) {
            sandbox.restore();
            throw exception;
        } else {
            sandbox.verifyAndRestore();
        }
        oldDone(res);
    };
}```

mroderick:
https://github.com/cjohansen/Sinon.JS/blob/master/CONTRIBUTING.md#reporting-an-issue

mroderick:
Whoops, wrong button

fatso83:
Yeah, sorry about submitting before fleshing out the details; was caused by a "Tab,Enter" user bug ... Detailing it now.

fatso83:
This behavior is verified in Chrome 44 and PhantomJS 1.9.2, btw (through the Buster tests). Anything else that prevents the PR from being merged (except for time)?

mroderick:
Fixed via #820

PR #820, opened 2015-08-26
Fix for issue 818

fatso83:
Issue #818 makes SinonJS break async mocha tests. After examining the sinon code, I am assuming duplicated code led to some invalid logic where the exception was thrown, rather than being passed on to the callback. This PR fixes issue #818 while is seems not to break anything else.


mroderick:
Thank you for your contribution :+1:

Issue #172, opened 2012-08-20
sinon.logError should throw synchronously instead of after a timeout


airhorns:
In Sinon right now the following function gets executed when say a fake server encounters an error while responding to requests.

```javascript
        logError: function (label, err) {
            var msg = label + " threw exception: "
            sinon.log(msg + "[" + err.name + "] " + err.message);
            if (err.stack) { sinon.log(err.stack); }

            setTimeout(function () {
                err.message = msg + err.message;
                throw err;
            }, 0);
```

The error is thrown after a timeout instead of synchronously. This means that if for example I want to test that my code errors if the server returns a `500`, I can't wrap my test in a try catch because the error is caught and then thrown after the stack unwinds. I am sure there is a good reason why the error is through asynchronously right now but it would make it way easier to test code that throws on purpose in response to servers responding.

jfirebaugh:
@mroderick Thanks for adding the flag. Can you elaborate on why it's important to preserve the default behavior of using a timeout? From my perspective, it seems like the `useImmediateExceptions = true` behavior is more useful in the vast majority of cases, if not 100% of the time.

mroderick:
> Can you elaborate on why it's important to preserve the default behavior of using a timeout?

We're trying to be very careful about not breaking backwards compatibility, so that upgrading Sinon.JS in a project doesn't have to be a scary task. In the context of this ticket, it might seem most natural to change the default. However, there's likely to be many tests out there, that would fail in new and interesting ways if we change this behaviour.

I agree that we should probably change the default to be `useImmediateExceptions = true`, but I'd like to give the users some fair warning via some deprecation mechanism.

There has been some discussion on how best to get deprecation warnings to users. If you have good ideas on how to do this for a library like Sinon, then a new issue would be very welcome.

I've created https://github.com/cjohansen/Sinon.JS/issues/841 to discuss any changes to defaults in the next major release.

jfirebaugh:
Thanks, that makes sense! :+1:

tombh:
Just wanted to say I got caught out with this and sadly lost a lot of frustrated hours tracking where my errors were being swallowed :(

But `sinon.logError.useImmediateExceptions = true` solved everything.

PR #825, opened 2015-08-31
sinon.logError.useImmediateExceptions

D1SoveR:
Implementation of idea proposed in #172, a flag on `sinon.logError` that, when set to true, makes the method throw the errors in the same execution frame, thus allowing for easier debugging and/or better handling in test cases that deal with XHR requests.


mroderick:
@D1SoveR thank you for your pull request!

mroderick:
I've added https://github.com/sinonjs/sinon-docs/issues/22 for documentating the `logError` feature

Issue #811, opened 2015-08-17
Remove sinon-timers from build


mroderick:
`lolex` was extracted to it's own module about a year ago, it's time to stop packaging `sinon-timers` as an independent module and encourage people to use it's successor `lolex` instead.
- [x] Stop generating `sinon-timers.js` into `pkg/`
- [x] Update [clock documentation](http://sinonjs.org/docs/#clock) and point to using `lolex` for standalone fake timers without Sinon.JS

evdb:
On http://sinonjs.org/download/ the links under "Stand-alone fake timers" 404 - which I think is related to this issue as the code is no longer being generated?

pimterry:
How are built lolex packages for the browser supposed to be made available in future?

Since the Sinon.js download page now has a 404, as you're no longer building sinon-timers within Sinon, I went to look at the lolex package itself and the release notes there say "if you want to use Lolex in a browser, you have a few options. Releases are hosted on the sinonjs.org website". Feels like that just sends you back round in circles. It works if you're using NPM of course, but not everybody will be.

Are you going to have a download site for lolex, or some other release mechanism over there? At the moment these files are just completely unavailable (although you can just guess URLs to get an old version of course).

mantoni:
@pimterry Yes, we need to think about how we package standalone versions. Sorry for the confusion. In the meantime, you can use the [lolex.js](https://github.com/sinonjs/lolex/blob/master/lolex.js) file.

mroderick:
This has been implemented with https://github.com/grassator/Sinon.JS/commit/5ee643df517f6ed7296b4a1ce0a81317949526d6

mroderick:
@pimterry the browserified version of `lolex` is available via npm. Do you still think people will want to download this manually?

pimterry:
Hard to say whether others will want this, but that sounds fine to me personally.

mroderick:
@pimterry I've created an issue for the direct download of lolex.js

PR #832, opened 2015-09-09
Removed sinon-timers from the build. refs #811

grassator:
I'm not sure if this is all I have to do, but it's a start.

### DIFF
--- a/build
+++ b/build
@@ -109,11 +109,6 @@ Dir.chdir(File.dirname(__FILE__)) do
 
   add_license("pkg/sinon-ie#{version_string}.js", version)
 
-  FileUtils.cp("lib/sinon/util/fake_timers.js", "pkg/sinon-timers#{version_string}.js")
-  add_license("pkg/sinon-timers#{version_string}.js", version)
-  FileUtils.cp("lib/sinon/util/timers_ie.js", "pkg/sinon-timers-ie#{version_string}.js")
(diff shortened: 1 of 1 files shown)


mroderick:
Nice! I'll have to find some time to check it out and run it locally.

build
```diff
@@ -109,20 +109,11 @@ Dir.chdir(File.dirname(__FILE__)) do
...
-  add_license("pkg/sinon-timers#{version_string}.js", version)
-  FileUtils.cp("lib/sinon/util/timers_ie.js", "pkg/sinon-timers-ie#{version_string}.js")
-  add_license("pkg/sinon-timers-ie#{version_string}.js", version)
-
   merger = Juicer::Merger::JavaScriptMerger.new
   merger << "lib/sinon/util/fake_server_with_clock.js"
   merger.save("pkg/sinon-server#{version_string}.js")
   add_license("pkg/sinon-server#{version_string}.js", version)
   FileUtils.cp(output, 'pkg/sinon.js')
-  FileUtils.cp("pkg/sinon-ie#{version_string}.js", 'pkg/sinon-ie.js')
```
mroderick: We probably want to keep making `sinon-ie.js` and `sinon-server.js` for a little while longer.

grassator:
@mroderick sorry about that, now should be ok.

mroderick:
:+1: could you squash it down to just one commit before we merge it?

grassator:
@mroderick squashed

PR #837, opened 2015-09-14
stub.reset also resets behavior

### LINKED ISSUE #806: reset() does not reset a stub
`reset` resets only a spy but not a stub

Consider this code:
```js
var sinon = require('sinon');
var stub = sinon.stub();
stub.withArgs(1).returns('Bye');
stub.reset();
stub.returns('Hello');
console.log('stub(1)', stub(1));
```
it prints
```
stub(1) Bye
```
if I change `stub.reset();` to `stub.resetBehavior();` it works as expected, i.e. prints `stub(1) Hello`.
But `resetBehavior` is not documented (#267) and it would be more consistent and natural if `reset` works for both spies and stubs.

Another option is to change `stub.reset();` to `stub = sinon.stub();`.
But It is not always feasible to create a new stub in each test. Consider this example:
```js
var stub = spy.stub();
var obj = createExpensiveObject(stub);

it('test1', function() {
  // setup stub in some way
  obj.test1();
});

it('test2', function() {
  // setup stub in a different way
  obj.test2();
});
```

dotchev:
Now reset() resets the complete stub - both call history and behavior.
This is incompatible but more consistent.

Fixes #806

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -125,6 +125,11 @@
                 }
             },
 
+            reset: function () {
+                sinon.spy.reset.call(this);
+                this.resetBehavior();
+            },
+
             onCall: function onCall(index) {
                 if (!this.behaviors[index]) {
                     this.behaviors[index] = sinon.behavior.create(this);
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -1490,8 +1490,8 @@
 
                     stub.reset();
 
-                    assert.same(stub(5), 1);
-                    assert.same(stub(5), 2);
+                    assert.same(stub(5), undefined);
+                    assert.same(stub(5), undefined);
                     refute.defined(stub(5));
                 },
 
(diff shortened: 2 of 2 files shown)


mroderick:

This has the potential of messing up a lot of tests out there. If we merge this, I think we will have to bump the major version. (Which is probably a good thing to get going on the next version) ... but then we might as well get in some other changes at the same time.

@cjohansen @mantoni what do you think?

cjohansen:
I think we should just merge this and immediately push 2.0. Waiting for 2.0 to be something big isn't doing us any favors.

cjohansen:
(If there are other breaking changes in the pipeline that could be landed soon, that'd be fine as well)

mroderick:
We might change some defaults, while we are at it: https://github.com/sinonjs/sinon/issues/841

cjohansen:
:+1:

Issue #835, opened 2015-09-11
Uncaught TypeError: Cannot assign to read only property 'message' of error!


MichaelMcKechnie:
I was seeing this error in my log, and found that it happened because of my bone-headed decision to throw a constant string :) like this:

var handler = function (xhr) {
                xhr.onerror();
                throw "error!";
            };

The sinon code causes a type error when the following code runs (because you can't set the message property on the constant string "error!":

        function logError(label, err) {
            var msg = label + " threw exception: ";

            sinon.log(msg + "[" + err.name + "] " + err.message);

            if (err.stack) {
                sinon.log(err.stack);
            }

            logError.setTimeout(function () {
                err.message = msg + err.message;
                throw err;
            }, 0);
        };

I've fixed my own code, but I'm happy to alter sinon as well so that it handles the exception if you like.

mroderick:
> I've fixed my own code, but I'm happy to alter sinon as well so that it handles the exception if you like.

That would be most welcome :+1:

PR #844, opened 2015-09-20
issue #835 (uncaught TypeError)

mmckech:
LogError expects second parameter to be an instance of Error() but it can also come from end user handler code. So, it can be anything. I transferred error parameter in LogError to a new object


mroderick:
@mmckech thank you for your pull request. Would you mind rebasing it, so we can merge it?

fatso83:
Redone as #979 against the latest master

Issue #834, opened 2015-09-10
Use CommonJS


mroderick:
We would like to get rid of the UMD like wrapper in source files and adopt CommonJS in the entire codebase (for near future at least).

For this to happen there are are number of things that we need to do
- [ ] convert all the files to use CommonJS
- [ ] create UMD wrapped pre-built version(s) using Browserify / webbpack or similar
- [ ] run all tests for source version in node
- [ ] run all tests for built version(s) in PhantomJS (as a minimum)

Making the switch to CommonJS is not a small task, there are many dependencies to manage and the current toolchain is not very forgiving.

I am probably forgetting half the complexities, so expect this issue to be expanded, when I do remember.

If you have any ideas on how we can get towards CommonJS incrementally, please speak up!

cjohansen:
I think that we need to drop `sinon.config` in order to solve this properly. I can't imagine how we'll get the whole thing common js'd without doing that. I am also guessing very few people actually use that feature, so dropping it should be uncontroversial?

mantoni:
The takeaway from my attempt to CommonJSify Sinon was that the core `equals` function and the matchers depend on each other. Even though it's tempting to try and separate those concerns, I wouldn't take them apart.

I agree with @cjohansen that the configs should probably be removed.

Re testing: @cjohansen Are Browserify and Buster playing nicely together? I remember it didn't work for Lolex.

grassator:
I would like to take a stub at this, but even though I worked with sinon in many projects, I never saw the `sinon.config`, so any pointers on dealing with it would be appreciated.

My thoughts on the build system were that it might be possible to use special webpack loaders to replace the ruby build system without modifying the code, as that will at least allow people to work on the build process in a familiar language / environment.

Hopefully I will have some time this week to check this out and dive deeper into sinon internals.

cjohansen:
Nice to hear that you want to look at it. FYI: We've previously discussed using Browserify, and that is still what I am the most keen on, simply because I know it well, and I have zero experience with webpack. I'm not even sure if this affects the coding work, but is there any reason to prefer one over the other?

The Ruby "build system" is definitely going away. The sooner the better.

There's a couple things that you can currently just "stick" on the `sinon` object to have magic things happen. These include `sinon.config` and `sinon.log`. I'm not sure how that'll work in a CommonJS world - probably it won't work at all. `sinon.config` will not be missed. It's used to set defaults for the fake timers and fake XHR.

grassator:
[...]

Issue #848, opened 2015-09-23
sinon-ie not working with IE11 (win 7)


epikhighs:
So I did some debugging and for some reason if I emulate IE 10 and below, then the code in sinon-ie.js works as intended.  However, if I use IE 11, then the following code returns an 'Object doesn't support property' error, where getTime is not defined on Date

`(new Date()).getTime()`

The entire prototype chain of Date is just a plain object without any of the Date specific methods.  All the global methods and classes that sinon-ie tries to override has this problem in IE11.  For example, the following will not work as well in IE11.

`setTimeout(function () {console.log(2)}, 500)`

Not sure why emulating IE 10 and below works though.  Has anyone else encountered this?

Thanks.

mroderick:
Let's see if I've understood your description correctly.

When `sinon-ie.js` is loaded in IE11, then `(new Date()).getTime()` throws an Error. But, when emulating IE10 in IE11, it doesn't?

Does it work when `sinon-ie.js` is not loaded?

Perhaps we should recommend people to only use `sinon-ie.js` for IE10 and below, or use conditional comments in `sinon-ie.js`.

Suggestions welcome

timwienk:
I can confirm this issue on both IE11 and Edge (12), I ran into this problem today (or actually a few days ago already, but we upgraded more than just sinon, it just took a while to figure out what the hell was going on). It appears indeed to be the case that all global methods and classes set in `sinon-ie.js` cause problems. The `setTimeout`-like functions don't do anything (no errors either, just nothing) and the constructor functions have no prototype chain. Example errors below (collected with Karma):

Calling a Date method:
```
IE 11.0.0 (Windows 10 0.0.0) ERROR
  Object doesn't support property or method 'getTime'
```
Calling an XMLHttpRequest method:
```
IE 11.0.0 (Windows 10 0.0.0) ERROR
  Object doesn't support property or method 'setRequestHeader'
```

As a last resort kind of thing, and while eliminating as many things as possible, I compared the previous version of `sinon-ie` we used to the version I was trying to use. It turned out that the only actual changes are the `if (typeof window !== "undefined")` conditionals around the different pieces of code. In the spirit of _"whatever remains, no matter how improbable, must be the truth"_ I proceeded to remove these conditionals.

Yes, indeed, removing these conditionals solved the problem. I have no idea what the actual reason is this happens (I suspect it must have to do with some weird kind of caching/optimising IE tries to do), but apparently these conditionals _do_ have practical effect (#840).

PR #850, opened 2015-09-23
First effort towards moving to CommonJS

grassator:
This PR contains minimal set of changes required to proceed with issue #834. I'm not sure it should be merged straight away as there some things to discuss but should be a good start.

Since I tried to minimize code changes, this PR is using CommonJS quite pervertedly — a `core` module is used as a kind of global namespace instead of previous `sinon` global.

The packaging is done via browserify but that ruby build file is still there for now for orchestration and can be easily removed in a separate PR. Reformatting the code and getting rid of extra wrappers can also be done in a separate PR to keep things manageable. And final part is to actually export the correct things in each of the modules and assemble them.

The biggest problem with testing is that it looks like PhantomJS 1.X incorrectly handles `"use strict"` statements and somehow inherits "strictness" from the calling function. I found where I need to insert additional "use strict" in the code generated by browserify, but this would be really hacky and unnecessary according to the standard. Switch to PhantomJS 2.X solved the problem.

Additional problem / annoyance is that previously non-packages browser build was used for code coverage, but since it's no more, I had to switch to node one, which doesn't run all of the tests, so coverage dropped. Using package version with source maps also won't work because `lcov` reporter in instanbul doesn't support source maps. Not quite sure what to do here.

More stuff about testing:
* CI build is working after switch to PhantomJS 2: https://travis-ci.org/grassator/Sinon.JS
* :warning: There is no BusterJS setup for sinon-server and I have no idea how to test it.
* It might be a good idea to try to run packaged tests in different IE versions (how far back do we want to support?)

As a bonus this PR should fix issue with requiring sinon from Webpack, but needs to be confirmed.

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -1,10 +1,4 @@
 /**
- * @depend times_in_words.js
- * @depend util/core.js
- * @depend match.js
- * @depend format.js
- */
-/**
  * Assertions matching the test spy retrieval interface.
  *
  * @author Christian Johansen (christian@cjohansen.no)
@@ -12,8 +6,12 @@
  *
  * Copyright (c) 2010-2013 Christian Johansen
  */
-(function (sinonGlobal, global) {
-    "use strict";
+"use strict";
+(function () {
+
+    require("./times_in_words");
+    require("./match");
+    require("./format");
 
     var slice = Array.prototype.slice;
 
@@ -197,30 +195,6 @@
         return assert;
     }
 
-    var isNode = typeof module !== "undefined" && module.exports && typeof require === "function";
-    var isAMD = typeof define === "function" && typeof define.amd === "object" && define.amd;
+    makeApi(require("./util/core"));
 
-    function loadDependencies(require, exports, module) {
-        var sinon = require("./util/core");
-        require("./match");
-        require("./format");
-        module.exports = makeApi(sinon);
-    }
-
-    if (isAMD) {
-        define(loadDependencies);
-        return;
-    }
-
-    if (isNode) {
-        loadDependencies(require, module.exports, module);
-        return;
-    }
-
-    if (sinonGlobal) {
-        makeApi(sinonGlobal);
-    }
-}(
-    typeof sinon === "object" && sinon, // eslint-disable-line no-undef
-    typeof global !== "undefined" ? global : self
-));
+}());
--- a/test/buster.js
+++ b/test/buster.js
@@ -1,39 +1,5 @@
 var config = module.exports;
 
-config.browser = {
-    environment: "browser",
-    rootPath: "../",
-    libs: [
-        "node_modules/lolex/lolex.js",
-        "node_modules/samsam/lib/samsam.js",
-        "node_modules/formatio/lib/formatio.js"
-    ],
-    sources: [
-        "lib/sinon.js",
-        "lib/sinon/util/core.js",
-        "lib/sinon/typeOf.js",
-        "lib/**/*.js"
-    ],
-    testHelpers: [
-        "test/test-helper.js"
-    ],
-    tests: [
-        "test/**/*-test.js"
-    ]
-};
-
-config.coverage = {
-    extends: "browser",
-    "buster-istanbul": {
-        outputDirectory: "coverage",
-        format: "lcov",
-        excludes: ["**/*.json"]
-    },
-    extensions: [
-        require("buster-istanbul")
-    ]
-};
-
 config.node = {
     environment: "node",
     rootPath: "../",
@@ -43,5 +9,17 @@ config.node = {
     ],
     tests: [
         "test/**/*.js"
+    ]
+};
+
+config.coverage = {
+    extends: "node",
+    "buster-istanbul": {
+        outputDirectory: "coverage",
+        format: "lcov",
+        excludes: ["**/*.json"]
+    },
+    extensions: [
+        require("buster-istanbul")
     ]
 };
--- a/.eslintrc
+++ b/.eslintrc
@@ -132,7 +132,7 @@
         "func-style": [0, "declaration"],
         "generator-star": 0,
         "generator-star-spacing": 0,
-        "global-strict": [2, "never"],
+        "global-strict": [2, "always"],
         "guard-for-in": 2,
         "handle-callback-err": 0,
         "indent": [2, 4, {"indentSwitchCase": true}],
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,6 @@ before_install:
   - npm install buster@0.7.18
 
 before_script:
-  - rvm install 2.2.0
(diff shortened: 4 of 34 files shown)


mantoni:
This is a very clever approach. I like how it's basically just removing code. Well done.

I'm personally not concerned with the coverage drop for now. Also, moving on with PhantomJS 2.0 is fine as long as 1.9 works as an environment that still runs Sinon in the end.

Surgo:
:+1:

grassator:
> Also, moving on with PhantomJS 2.0 is fine as long as 1.9 works as an environment that still runs Sinon in the end.
@mantoni After carefully consider the implications I think I will have to rollback this change and go with extra `"use strict"` hack, because if people rely on checking that their function has been called without an explicit scope as opposed `window` all kinds of failures can happen in their tests and unfortunately lots of people still use PhantomJS 1.x.

I will do that change, write a giant comment explaining why I'm doing hacky stuff in the code and push updated branch a little later today.

grassator:
Ok, now the package tests pass on PhantomJS 1.x so it's more safe to merge this PR.

I've also went through all of the Microsoft browsers and results are the same on the sinon master and this branch, which are IE8-9 didn't get to the end, IE9-11 some failed tests, Edge passes.

The only real worry I have now is `sinon-server`, but as I mentioned I have no idea how to test it.

fatso83:
> IE8-9 didn't get to the end, IE9-11 some failed tests,
If this is the case for the master branch, then we have a serious regression, as the homepage of Sinon says it works with IE6+, and perhaps we should open a separate issue for that once it is confirmed with 1.17

lib/sinon/assert.js
```diff
@@ -12,8 +6,12 @@
  *
  * Copyright (c) 2010-2013 Christian Johansen
  */
-(function (sinonGlobal, global) {
-    "use strict";
+"use strict";
+(function () {
```
mroderick: Is it still necessary to wrap files in an IIFE when using CommonJS?

test/buster.js
```diff
@@ -1,29 +1,19 @@
...
-        "lib/sinon/util/core.js",
-        "lib/sinon/typeOf.js",
         "lib/**/*.js"
     ],
-    testHelpers: [
-        "test/test-helper.js"
-    ],
     tests: [
-        "test/**/*-test.js"
+        "test/**/*.js"
```
mroderick: Is there any benefit to also loading the `buster.js` file?

lib/sinon/assert.js
grassator: It's totally unnecessary but I wanted this PR to clearly show that the actual code didn't change. A next pass can get rid of them.

test/buster.js
grassator: If you look at the node config in master, it has this and I didn't change it. This "change" is just an artifact of how git compares.

lib/sinon/assert.js
mroderick: :+1:

test/buster.js
mroderick: :

mroderick:
> If this is the case for the master branch, then we have a serious regression, as the homepage of Sinon says it works with IE6+, and perhaps we should open a separate issue for that once it is confirmed with 1.17

This could be related: https://github.com/sinonjs/sinon/issues/848

grassator:
@mroderick, @fatso83 The wrong results for testing is totally on me – I forgot about `sinon-ie`. In my defence sinon should probably have a separate BusterJS profile for IE. I will try to rerun the tests with sinon-ie included asap.

As far as IE11 goes, I think it should work without `sinon-ie` and if I remember correctly only a couple of tests failed there.

In general as a user of the library I would prefer that library takes care of compatibility problems by itself instead shifting it to the user especially since for sinon the package size doesn't really matter.

Also it's probably worth investing in setting up sauce labs integration with something like https://www.npmjs.com/package/buster-reporter-sauce so that regressions if they happen don't go unnoticed.

fatso83:
Sauce Labs or Browserstack integration is one option to consider, as I
believe they sponsor open source projects, but a problem is that tests
sometimes fail due to connection problems, etc. So it comes with a downside
in terms of false alarms making noise. I might look into it (as a separate
issue) once I have some spare time.

mroderick:
@grassator https://github.com/sinonjs/sinon/pull/858 might be worth importing into (a branch of) your branch, to see if that makes some of the IE issues go away.

grassator:
@mroderick using #858 sinon-ie or the master one doesn't seems to make any difference for me. The same tests seem to fail with or without including sinon-ie into the the packaged buster config, at least for IE8-11. Here's the setup I was using:

```js
config.ie = {
    extends: "packaged",
    sources: [
        "pkg/sinon-ie.js",
        "pkg/sinon.js"
    ]
};
```

And then:

```bash
./node_modules/buster/bin/buster-test --config test/buster-packaged.js --config-group=ie
```

Sinewyk:
what's missing from this PR ?

grassator:
@Sinewyk the only possible issue here that I can think of could be with `sinon-server` because there are no tests for this separated file.

Sinewyk:
was it tested before ? Looked through tests, and I can't see what this buster stuff doing.

dmwyatt:
Just wanted to say I've been using this for a week or so on a project with extensive testing that I'm moving over to webpack, and it's been working great.

mroderick:
> Just wanted to say I've been using this for a week or so on a project with extensive testing that I'm moving over to webpack, and it's been working great.

@dmwyatt have you been using this branch in your project?

dmwyatt:
@mroderick, sorry I wasn't clear enough on that.  I cloned this repo locally, fetched and merged this pull request, built the project, and then `npm link`ed it into my own sinon-using project.

So yes, I'm using sinon with this pull request.

Sinewyk:
same here. Forked it for guaranteed stable sha1 and then
```json
"sinon": "lrqdo/Sinon.JS#db24a7a7cb643d076537458a4147adc29641420f",
```
and problem solved.

mantoni:
Also ran some local tests. I can confirm that it works in one of my projects

- Using Mocha on Node 4.0
- Using Mochify (browserified) in PhantomJS 1.9, latest Chrome, latest Firefox, IE 9, IE 10 and IE 11

:+1:

mroderick:
Great work @grassator. Thank you all for reviews and testing!

fearphage:
@mroderick @mantoni @cjohansen Do you plan to move away from modules modify "globals" (read: sinon core) as well?

mantoni:
Yes. In my view the work that was done so far is the door opener to get into the CommonJS world without a massive rewrite (see our previous attempts). The plan is to pull things like the fake server into separate modules, like the timers that went into Lolex. This would allow using individual parts of Sinon. The interesting question is, where the actual module borders are.

fearphage:
Good to hear. :+1:

Issue #852, opened 2015-09-24
Sinon 1.17.0: sinon.createStubInstance cannot be used twice on the same constructor


mantoni:
This is a new issue in Sinon 1.17.0 and it's working fine with 1.16.1:

```js
var a = sinon.createStubInstance((MyThing);
var b = sinon.createStubInstance((MyThing);
```

Fails on the second line with:

```
TypeError: Attempted to wrap constructor which is already wrapped
  at checkWrappedMethod (node_modules/sinon/lib/sinon/util/core.js:91)
  at wrapMethod (node_modules/sinon/lib/sinon/util/core.js:121)
  at stub (node_modules/sinon/lib/sinon/stub.js:66)
  at node_modules/sinon/lib/sinon/stub.js:59
  at node_modules/sinon/lib/sinon/walk.js:34
  at forEach (null:null)
  at walk (node_modules/sinon/lib/sinon/walk.js:35)
  at walk (node_modules/sinon/lib/sinon/walk.js:39)
  at walk (node_modules/sinon/lib/sinon/walk.js:39)
  at stub (node_modules/sinon/lib/sinon/stub.js:61)
  at node_modules/sinon/lib/sinon/util/core.js:361
  ...
```

Environment: Node 4.0.0

fatso83:
I am unable to recreate this? Tried running it with everything from 1.16.1 to the latest `master`
```
describe('issue #852', function() {
    function MyThing() {}
    it('should not fail', function() {
        var a = sinon.createStubInstance(MyThing);
        var b = sinon.createStubInstance(MyThing);
    });
});
```

fatso83:
Tried this in Node 0.12 and 4.1, as well as PhantomJS. Never a failing test.

mantoni:
Sorry, left this here without further investigating. Had a hell of a day. I have a large test case that has this issue. Will see if I can extract the gist of it tomorrow.

flexfrank:
Creating a stub of a subclass reproduces the same error.

~~~JavaScript
describe('', function() {
    var A = function() {};
    var B = function() {};
    B.prototype = Object.create(A.prototype);
    B.prototype.constructor = A;

    it('create stub instance of a subclass', function() {
        var b = sinon.createStubInstance(B);
    });
});
~~~

~~~
  ✗ create stub instance of a subclass
	TypeError: Attempted to wrap constructor which is already wrapped
	    at checkWrappedMethod (node_modules/sinon/pkg/sinon.js:1239)
[...]

Issue #847, opened 2015-09-23
property getter broken in 1.17.0


MaerF0x0:
The following code works in 1.16.1 but fails in 1.17.0
(with mocha and mongodb)

```coffeescript 
sinon = require 'sinon'
assert = require 'assert'

describe "mock collection", ->
   beforeEach ->
     @collection = sinon.createStubInstance(Collection)
     sinon.stub @collection, 'collectionName', get: -> 'users'

   it 'should return users', ->
     assert.equal @collection.collectionName , 'users'
```
Will throw this err: 

```
1) permissions: scoped_query "before each" hook:
     TypeError: Cannot read property 'name' of undefined
    at [object Object].Object.defineProperty.get [as collectionName] (/Users/user/code/api/node_modules/mongodb/lib/collection.js:130:52)
    at /Users/user/code/api/node_modules/sinon/lib/sinon/walk.js:34:43
    at Array.forEach (native)
    at walk (/Users/user/code/api/node_modules/sinon/lib/sinon/walk.js:33:45)
    at Object.walk (/Users/user/code/api/node_modules/sinon/lib/sinon/walk.js:39:17)
    at Object.stub (/Users/user/code/api/node_modules/sinon/lib/sinon/stub.js:52:23)
    at Object.sinon.createStubInstance (/Users/user/code/api/node_modules/sinon/lib/sinon/util/core.js:361:26)
    at Context.<anonymous> (/Users/user/code/api/test/permissions.coffee:277:7)
    at Hook.Runnable.run (/Users/user/code/api/node_modules/mocha/lib/runnable.js:200:32)
    at next (/Users/user/code/api/node_modules/mocha/lib/runner.js:201:10)
    at /Users/user/code/api/node_modules/mocha/lib/runner.js:212:5
    at process._tickDomainCallback (node.js:492:13)
```

For reference that line looks like:
```javascript
Object.defineProperty(Collection.prototype, 'collectionName', {
  enumerable: true, get: function() { return this.s.name; }
});
```

mrjackdavis:
I'm having a similar problem, where createStubInstance is throwing an error.

    var sequelize = sinon.createStubInstance(Sequelize);

results in:

    TypeError: Cannot read property 'getConnectorManager' of undefined
      at Object.defineProperty.get (node_modules/sequelize/lib/sequelize.js:279:35)
      at node_modules/sinon/lib/sinon/walk.js:34:43
      at Array.forEach (native)
[...]

PR #854, opened 2015-09-25
Add failing test for #847

mroderick:
This is a failing test for #847, #852, which can be used in the pull request to fix that error. We don't need to merge this on it's own.

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -101,6 +101,20 @@
                 clock = sinon.useFakeTimers(new Date("2015-1-5").getTime());
                 assert.equals(clock.now, Date.now());
             }
+        },
+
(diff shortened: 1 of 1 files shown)


mantoni:
Thanks @mroderick. Was just about to look into this. :star: :star: :star: :star: :star: 

@fatso83 It could be merged, but the build is failing - intentionally in this case. There is no solution for the problem in the diff yet.

mroderick:
> Looks good. What is preventing a merge? Should another maintainer always look through the changes, or let it rest for a day or something?

I was hoping that this could be merged along with a fix for the issue(s), so that we could detect any regressions

mroderick:
Perhaps it can just be the basis of someone else's pull request

PR #855, opened 2015-09-25
Run tests in node 4.1.x also

mroderick:
Let's see if I've understood the Travis CI documentation correctly.

>  nvm handles version resolution, so any version or alias of Node or io.js that nvm can install is available.


mantoni:
Looking good :+1:

PR #858, opened 2015-09-26
lib/sinon/util: Remove window conditionals from IE files.

### LINKED ISSUE #840: Remove condtional in timers-ie
Remove the conditional in https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/util/timers_ie.js#L15.

It doesn't have any practical effect, since the functions are hoisted to the top of the scope anyway.

See https://github.com/cjohansen/Sinon.JS/pull/684#issuecomment-137636562

### LINKED ISSUE #848: sinon-ie not working with IE11 (win 7)

timwienk:
The removed conditionals should not have any functional benefit, so in
that sense it's safe to remove these. (As described in #840.)

However, it appears that these conditionals have an unintended effect
in IE 11 and Edge 12, where the functions defined within the
conditionals actually stop working as expected. (See #848.)

While it is true these IE utils are not necessary for the newer IEs,
they are often still loaded as part of a generic test package.

Closes #840.
Fixes #848.


timwienk:
Tests in node failed with a `sinon is not defined` error. This is because the `lib/sinon/util/*_ie.js` files were loaded fine in node, but didn't actually do anything because of the `if (typeof window !== 'undefined')` conditional.

With the conditional gone, the behaviour is the same with these files just not included.

build
```diff
@@ -101,10 +101,10 @@ Dir.chdir(File.dirname(__FILE__)) do
   add_license(add_deps(output), version)
 
   File.open("pkg/sinon-ie#{version_string}.js", "w") do |f|
-    f.puts(File.read("lib/sinon/util/timers_ie.js"))
+    f.puts(File.read("lib/sinon/util_ie/timers.js"))
```
mroderick: In Sinon.JS we use `kebab-case` for filenames. Would you mind renaming the folder to follow that pattern?

mroderick:
@timwienk thank you for your detailed analysis and pull request!

Did you run the tests in a broad selection of IE?

build
timwienk: Yea, sure. I figured I'd change as little as possible, so I basically moved `"_ie"` from the file name to the directory name.

timwienk:
I did run the tests in multiple IEs (both before the changes and after). The difficulty with this is that before the changes, there were already failing tests in IE (and IE11 and Edge didn't even run, because of #848).

In terms of test results:
- I did not see any difference in IE<11 (same tests still failing),
- in IE11 it now runs, and only one error: `"TypeError: Object doesn't support property or method 'readAsBinaryString'"` (http://caniuse.com/#search=file),
- in Edge, Firefox and Chrome the tests pass successfully.

mroderick:
Could you also squash the commits down to just one, so we can keep git history neat?

Thanks!

timwienk:
Done.

Please note, though, that by doing that git may not track the change from `util/*_ie.js` to `util-ie/*.js` as renames if the file contents have changed significantly (which is the case because of indentation changes in these files), so history might get disjointed there.

timwienk:
If preferred (so all renames are tracked correctly) I can just re-squash this, so only the last two commits become one, and the first one stays separate.

mroderick:
This is lovely! But, now it needs a rebase (I should have merged your branch first, I apologise).

timwienk:
Rebasing done. The order of merging is quite alright, I guess. Either one of the branches would need to resolve the conflicts, might as well be the smaller one. :-)

I rebased the [original work][] without the squash as well (in fact, I rebased that, then re-squashed, for easier conflict resolution), so the option is still there for the two-commit version for the sake of rename tracking. It's no difference to me.

[original work]: https://github.com/sinonjs/sinon/compare/a973746...322d101

timwienk:
And rebased again, noticed you did a new push to your master branch. No conflicts this time.

timwienk:
Rebased again, fixed some conflicts due to ESLint comment changes.

Are there plans to merge this any time soon?

mroderick:
Thank you :+1:

dominykas:
Is there a plan to merge this also into the 1.x branch?

mroderick:
> Is there a plan to merge this also into the 1.x branch?

I am not aware of any such plans. We are hoping to release 2.0 soon

PR #859, opened 2015-09-27
Use SlimerJS for unit tests

mroderick:
**TL;DR**
This PR replaces the use of PhantomJS for SlimerJS, hoping for more stability and fewer false positives (PhantomJS is often the cause for failed builds on Travis CI)

* Install and start SlimerJS on Travis
* Kill any leftover xulrunner processes after tests are done, when we're not using Travis
* Stop running tests in node@0.8, SlimerJS dependency `ncp` uses `setImmediate`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 coverage/
 pkg/
+scripts/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "0.8"
   - "0.10"
   - "0.12"
   - "4.2"
@@ -11,12 +10,13 @@ cache:
   directories:
   - node_modules
 
-before_install:
-  - npm config set strict-ssl false
-  # use a more modern npm than ships with 0.8
-  - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.8" ]; then npm install -g npm@1.4; fi'
+env:
+  - SLIMERJSLAUNCHER=$(which firefox) DISPLAY=:99.0 PATH=$TRAVIS_BUILD_DIR/slimerjs:$PATH
+addons:
+  firefox: "25.0"
 
 before_script:
+  - ./scripts/travis-install-slimerjs
   # we only need to run eslint once per build, so let's conserve a few resources
   - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.12" ]; then $(npm bin)/eslint .; fi'
 
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "buster-istanbul": "0.1.13",
     "eslint": "^1.7.1",
     "eslint-config-defaults": "^2.1.0",
-    "pre-commit": "1.0.10"
+    "pre-commit": "1.0.10",
+    "slimerjs": "^0.9.6"
   },
   "files": [
     "lib",
--- a/scripts/ci-test.sh
+++ b/scripts/ci-test.sh
@@ -5,11 +5,12 @@ function finish {
     if [ -n "${TRAVIS+1}" ]; then
       echo "TRAVIS detected, skip killing child processes"
     else
-      kill $(jobs -pr)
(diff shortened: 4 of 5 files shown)


fatso83:
How does this run on Travis when it's not headless?

mroderick:
> How does this run on Travis when it's not headless?

You can run a headless version by invoking some secret incantations. I figured that I'd worry about Travis, once I have it working locally.

mroderick:
@fatso83 @cjohansen @mantoni this is now ready for review

fatso83:
No further comments on my part :-) I was about to say something about the line where we run a script called `phantom.js` with Slimer, but that was internal to the `buster` module, so it has nothing to do with this. :+1:

mroderick:
I fixed a merge mistake in `.travis.yml` and removed an accidental branch `add-slimers` from this repo ... and now one of the builds fail, but the other doesn't.

The tests have run, it should still be safe to merge.

fatso83:
I have seen through it while it was still passing in Travis so it should be ok. I'll merge it once I have just downloaded it locally and made sure it passes against the latest HEAD commit.

fatso83:
Hey, this might not be related to this commit, but the `prepublish` script is running `browserify` before the package has been installed. That is not so hot.

```
npm install
> ...
>  sinon@1.17.0 prepublish /Users/carl-erik.kopseng/dev-priv/Sinon.JS
> ./build

./build:53:in ``': No such file or directory - ./node_modules/.bin/browserify (Errno::ENOENT)
	from ./build:53:in `block in <main>'
	from ./build:46:in `chdir'
	from ./build:46:in `<main>'
```

On the `prepublish` script (from NPM docs)
> prepublish: Run BEFORE the package is published. (Also run on local npm install without any arguments.)

mroderick:
> Hey, this might not be related to this commit, but the prepublish script is running browserify before the package has been installed. That is not so hot.

You're right. We should tidy this up. 

Will you create an issue for it, or shall I?

fatso83:
@mroderick : Added #893

fatso83:
I cannot for the life of me understand why Buster complains that these tests are failing? I ran `npm test`. Is the coverage below a certain threshold?

screendump-failing tests

mroderick:
Perhaps it's an issue with exit codes? Buster seems happy enough

mroderick:
There's nothing configured to fail the test because of coverage

mroderick:
Hmm, it leaks a xulrunner process on my machine ... I thought I'd fixed that

mroderick:
@tenzer do you have an idea why it's leaking a xulrunner process on osx?

fatso83:
@mroderick You are right about the exit codes. It exited with a non-zero exit code as the `finish` function never managed to kill anything. The reason it did not work in the first place is that that the parent id of the `xulrunner` is never 1 (only if you think recursively). Finding the ppid using pstree and replacing `1` with that worked, but involved parsing and messy stuff.

I sent you a PR that simply kills the entire process group. Much simpler :-)

mroderick:
Hmm, this fails on ArrayBuffer related code introduced in #813 

According to MDN docs, then [TextDecoder.prototype.decode has been available in Firefox since 19](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/decode). It expects something that implements `ArrayBufferView`.

According the the [SlimerJS FAQ](https://slimerjs.org/faq.html#firefox-compatibility), then SlimerJS uses Gecko 38, which is compatible with Firefox 38.

Running the tests in Firefox 41 doesn't throw any Errors.

@henryptung did test this in Firefox? Did you see any Errors?

Perhaps this is just the universe telling us to stop running test in pseudo browsers like PhantomJS and SlimerJS, and start running them in real browsers (via BrowserStack perhaps). @fatso83 @mantoni @cjohansen opinions?

henryptung:
Looks like TextDecoder was introduced by https://github.com/sinonjs/sinon/commit/e71564e534bfdfdf6cfb6b384c0babf4ca478ef4. I didn't notice any issues when I was testing, though if it wants a view the fix there should be trivial.

Issue #271, opened 2013-04-19
$.ajax or fakeServer issue with jQuery 2.0.0 


philfreo:
http://jsfiddle.net/philfreo/7gUBk/

Passes in jQuery 1.9.1
Fails in jQuery 2.0.0

Not really sure if it's a jQuery or SinonJS bug, but it's a really simple test case that should pass.

PR #861, opened 2015-09-30
Fix fake xhr events

fearphage:
From [v1.12.2 to v1.13](https://github.com/sinonjs/sinon/compare/v1.12.2...v1.13.0) a bug was introduced. On errors and aborted calls, a `progress` event started being fired with `e.lengthComputable` set to true (This was the thing that broke my tests). The `progress` event was erraneously being fired with `{loaded: 100, total: 100}` also. In tracking down this issue, I ended up fixing a few other things along the way. I had to look at the XHR spec to get an idea of what was to be expected. So I updated `readyStateChange` to align more closesly to the [w3c spec](https://xhr.spec.whatwg.org/#request-error-steps). 

Here are the changes I've made based on the spec:
    
* aligned order of events (xhr && xhr.upload)
* added 'loadend' event to xhr.upload
* all `done` events are now progress events (`e.loaded`, `e.total`, `e.lengthComputable`)
* *bug fix*: 'load' was fired before 'abort'
* *bug fix*: progress events were showing complete (100 of 100) on error/abort (instead of 0 of 0)
* *bug fix*: `e.lengthComputable` was true in error/abort events
* *bug fix*: `xhr.onerror` was being called for aborted requests

### DIFF
--- a/lib/sinon/util/event.js
+++ b/lib/sinon/util/event.js
@@ -36,8 +36,8 @@ sinon.Event.prototype = {
 
 sinon.ProgressEvent = function ProgressEvent(type, progressEventRaw, target) {
     this.initEvent(type, false, false, target);
-    this.loaded = progressEventRaw.loaded || null;
-    this.total = progressEventRaw.total || null;
+    this.loaded = typeof progressEventRaw.loaded === "number" ? progressEventRaw.loaded : null;
+    this.total = typeof progressEventRaw.total === "number" ? progressEventRaw.total : null;
     this.lengthComputable = !!progressEventRaw.total;
 };
 
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -70,10 +70,11 @@ var unsafeHeaders = {
 // and uploadError.
 function UploadProgress() {
     this.eventListeners = {
-        progress: [],
-        load: [],
         abort: [],
-        error: []
+        error: [],
+        load: [],
+        loadend: [],
+        progress: []
     };
 }
 
@@ -439,6 +440,7 @@ sinon.extend(FakeXMLHttpRequest.prototype, sinon.EventTarget, {
         this.readyState = state;
 
         var readyStateChangeEvent = new sinon.Event("readystatechange", false, false, this);
+        var event, progress;
 
         if (typeof this.onreadystatechange === "function") {
             try {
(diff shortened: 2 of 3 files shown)


mantoni:
Thanks a lot. One question: the `onerror` handler being called on `abort` seemed to be there purposely because it had a test. It might not be "by the spec" but could reflect browser specific behavior. Are we sure this can be deleted?

fearphage:
Well it was doing the wrong thing. `.abort()` is a user-intiated action. `onabort` should be triggered. `onerror` is for... errors. Nothing has gone wrong. The API is doing exactly as requested. Logically it seems like calling `xhr.onerror` in the `xhr.abort` method was a bug. I can put it back though.

mantoni:
@fearphage Yes, I agree that it was doing the wrong thing. Just double checking. Maybe @cjohansen or @mroderick remember why it was put there in the first place.

fearphage:
This is where it came to be. # #271

fearphage:
What's the verdict, folks? @cjohansen @mantoni @mroderick 

http://www.w3.org/TR/XMLHttpRequest/#the-abort()-method

cjohansen:
Seems like this was put in place in error. This looks solid to me, well done!

cjohansen:
Btw, mind rebasing with master?

fearphage:
Does rebasing generally require a `push -f`? It did this time. I'm unfamiliar with rebasing. You made a lot of good changes I see.

cjohansen:
Yes it does, so you should be very careful doing it on shared remotes :) Thanks!

Tuxdude:
This change has not been merged into the latest v1.17.3 release, although it is part of v2.0.0-pre tag. Any chance this is planned to be merged into a stable release that will be pushed to npm ?

Without this change, error events are not propagated to FakeXMLHttpRequest, since the response status codes are not checked otherwise and always fires a load event instead.

Thanks!

fatso83:
I think it unlikely to happen by itself, but if you are willing to try and make a pull request with these changes against the 1.17 branch we can merge it.

fearphage:
@fatso83 You rang? :wink:

fatso83:
@fearphage Ah, sorry about that. Did not mean to imply that _you_ needed to do anything, but I see you already created PR #1017, so thanks a lot for that :-D I'll try to explicitly ping the right individual next time.

fearphage:
@fatso83 I didn't do it, because you asked me to or wanted me to. I did it, because I thought it needed to be done.

PR #865, opened 2015-10-03
Implemented stub#resetHistory method - fixes #863

### LINKED ISSUE #863: .resetHistory
With https://github.com/dotchev/Sinon.JS/commit/7078dd8c60273ddda97d868a57f02d73f0edf54e it is no longer possible to only reset the call history of a stub.

As discussed in https://github.com/sinonjs/sinon/issues/806#issuecomment-144327382, we could expand the api a tiny bit.

```javascript
// resets both call history and behaviour
stub.reset();

// NEW: reset call history 
stub.resetHistory();

// reset behaviour
stub.resetBehavior()
```

fearphage:
I refactored `stub#reset` to call `.resetHistory()` and `.resetBehvaior()`.

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -126,8 +126,10 @@
                 }
             },
 
+            resetHistory: sinon.spy.reset,
+
             reset: function () {
-                sinon.spy.reset.call(this);
+                this.resetHistory();
                 this.resetBehavior();
             },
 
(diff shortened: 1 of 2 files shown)


fatso83:
Seems solid. Thanks!

mroderick:
Thanks @fearphage!

PR #866, opened 2015-10-05
Allow yieldsOn, callsArgOn to use any defined context (not just an object)

shaine:
Currently, sinon throws an error if a stub is told to call an argument with a context that is not of type object, "TypeError: argument context is not an object", via the yieldsOn, callsArgOn, etc methods.

In certain rare circumstances, this is not desirable. For example, https://github.com/substack/node-seq calls its chained functions with a `this` context of its async callback. While this is a very strange and generally non-standard behavior, it is permitted by JS, so sinon should permit it.

This PR updates the error check to verify that `undefined` isn't being passed as the context, otherwise permits all JS types as contexts. I could update the PR to permit only functions or objects, but feel this is the better approach as `(function() {}).call(5)` seems to be valid.

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -191,9 +191,6 @@ var proto = {
         if (typeof pos !== "number") {
             throw new TypeError("argument index is not number");
         }
-        if (typeof context !== "object") {
-            throw new TypeError("argument context is not an object");
-        }
 
         this.callArgAt = pos;
         this.callbackArguments = [];
@@ -221,9 +218,6 @@ var proto = {
     callsArgOnWith: function callsArgWith(pos, context) {
         if (typeof pos !== "number") {
             throw new TypeError("argument index is not number");
-        }
-        if (typeof context !== "object") {
-            throw new TypeError("argument context is not an object");
         }
 
         this.callArgAt = pos;
@@ -256,10 +250,6 @@ var proto = {
     },
 
     yieldsOn: function (context) {
-        if (typeof context !== "object") {
-            throw new TypeError("argument context is not an object");
-        }
-
         this.callArgAt = useLeftMostCallback;
         this.callbackArguments = slice.call(arguments, 1);
         this.callbackContext = context;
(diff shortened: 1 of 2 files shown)


mroderick:

I think your arguments are solid. 

Since we're doing input validation, perhaps it would be worth throwing an Error for all invalid inputs? 

Passing `null` as context wouldn't produce results different from `undefined`, would it?

shaine:
Hrm, good point. I only checked for undefined because I assumed it would help preserve the author's intention, but my actual preference would be to permit undefined as well. Are there any cases where one may want to specify that undefined is the context, maybe just for specificity's sake?

fatso83:
I am wondering what would be The Right Thing:tm: to do. Usually, passing null or undefined as the context would be an error in a test, but it might not be. One could of course always as a user achieve the same thing (undefined context) by simply using `yieldsTo`, but for some given case that might make things more complex than need be.

I am leaning towards having the same lax behaviour as `Function.prototype.call` (and friends `apply`, etc), where it is fine to do `fn.apply(null, args)`.

shaine:
I agree. While essentially the same thing could be accomplished with `yieldsTo`, the user might mean to explicitly set it so as to communicate to future readers that they really did want it called with undefined vs an unintended side effect.

shaine:
(Oops, clumsy thumbs). 

I'll update the PR to remove the checks for now.

mroderick:
@shaine master has moved forwards, would you mind rebasing your branch so we can merge it?

fearphage:
You removed tests but didn't add any. I believe you should confirm that you can now call the methods on numbers, booleans, arrays, etc.

mroderick:
Yep, let's make sure we cover changes with tests

shaine:
Good point, was letting the unit tests test what was actually being done versus testing what change the PR allowed for. Added unit tests verifying that `undefined` and `5` are permissible context. Rebased with master.

Added updates to callsArgOnWith, yieldsToOn to also not type check the context so as to be consistent with the updates on yieldsOn, callsArgOn.

fatso83:
Great stuff - mind squashing the commits in the branch into one commit before we merge this?

shaine:
Anything else I can do on this?

fatso83:
Was just lost in the sea of notifications :-) Thanks!

PR #868, opened 2015-10-06
Fix #867: Walk properties only once

### LINKED ISSUE #867: Prototype chain and stub breaking behaviours between 1.16.1 and 1.17.1
I've spot another BC behavior between sinon 0.16.1 and sinon 0.17.1. I've see https://github.com/sinonjs/sinon/issues/860 but I'm not sure this is the same problem.

This code do not throw any error with sinon 1.16.1:

```javascript
'use strict';

var sinon = require('sinon');

var A = function A() {
    console.log('> Hello A');
};

A.prototype.foo = function foo() {
    return 42;
};


var B = function B() {
    console.log('> Hello B');
};

B.prototype = new A();

B.prototype.foo = function foo() {
    return 51;
};

var b = sinon.createStubInstance(B); // works in 1.6.1, error in 1.7.1 (see below)
```

In 1.7.1, the last line trigger this error: 

```
TypeError: Attempted to wrap foo which is already wrapped
    at checkWrappedMethod (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/util/core.js:81:29)
    at Object.wrapMethod (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/util/core.js:121:21)
    at stub (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/stub.js:67:26)
    at /Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/stub.js:60:25
    at /Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:26:26
    at Array.forEach (native)
    at walkInternal (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:23:45)
    at walkInternal (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:31:17)
    at walkInternal (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:31:17)
    at Object.walk (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:46:20)
--------------
Error: Stack Trace for original
    at Object.wrapMethod (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/util/core.js:140:34)
    at stub (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/stub.js:67:26)
    at /Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/stub.js:60:25
    at /Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:26:26
    at Array.forEach (native)
    at walkInternal (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:23:45)
    at walkInternal (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:31:17)
    at Object.walk (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/walk.js:46:20)
    at Object.stub (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/stub.js:52:23)
    at Object.sinon.createStubInstance (/Users/spy-seth/workspace/myproject/node_modules/sinon/lib/sinon/util/core.js:361:26)
```

mantoni:
This fixes `createStubInstance` for prototype chains with overrides as reported in #867. One could argue that the fix belongs into the iterator function in `sinon.stub`. I tried adding a check for whether a property is already stubbed and that also does the trick, but it seemed a bit messy.

Thoughts?

### DIFF
--- a/lib/sinon/walk.js
+++ b/lib/sinon/walk.js
@@ -2,7 +2,7 @@
 
 var sinon = require("./util/core");
 
-function walkInternal(obj, iterator, context, originalObj) {
+function walkInternal(obj, iterator, context, originalObj, seen) {
     var proto, prop;
 
     if (typeof Object.getOwnPropertyNames !== "function") {
@@ -18,14 +18,17 @@ function walkInternal(obj, iterator, context, originalObj) {
     }
 
     Object.getOwnPropertyNames(obj).forEach(function (k) {
-        var target = typeof Object.getOwnPropertyDescriptor(obj, k).get === "function" ?
-            originalObj : obj;
-        iterator.call(context, target[k], k, target);
+        if (!seen[k]) {
+            seen[k] = true;
+            var target = typeof Object.getOwnPropertyDescriptor(obj, k).get === "function" ?
+                originalObj : obj;
+            iterator.call(context, target[k], k, target);
+        }
     });
 
     proto = Object.getPrototypeOf(obj);
     if (proto) {
-        walkInternal(proto, iterator, context, originalObj);
+        walkInternal(proto, iterator, context, originalObj, seen);
     }
 }
 
@@ -40,7 +43,7 @@ function walkInternal(obj, iterator, context, originalObj) {
  * context - (Optional) When given, the iterator will be called with this object as the receiver.
  */
 function walk(obj, iterator, context) {
-    return walkInternal(obj, iterator, context, obj);
+    return walkInternal(obj, iterator, context, obj, {});
 }
 
 sinon.walk = walk;
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -702,6 +702,18 @@
                 refute.isFunction(obj.toString.restore);
                 refute.isFunction(obj.toLocaleString.restore);
                 refute.isFunction(obj.propertyIsEnumerable.restore);
+            },
+
+            "does not fail on overrides": function () {
+                var parent = {
+                    func: function () {}
+                };
+                var child = sinon.create(parent);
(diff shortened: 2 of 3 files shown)


fatso83:
Good work, I think this looks pretty clean. Just to see if I understand you correctly; the alternative would be to create an analog of your `seen` object  in [this part](https://github.com/sinonjs/sinon/blob/7e3e0b1881198ef73ceaa271a406d5afa2785a82/lib/sinon/stub.js#L56-L67) of sinon.stub, right? In that case, your take indeed seems cleaner

fatso83:
I am not sure if it would be right to include a test case for another issue, but if this fixes #867 then maybe it would be a nice regression test to have that test case put in as well?

mantoni:
Alternatively, there could be an if-statement around `sinon.stub(object, prop)` to check whether it's already stubbed.

mroderick:
Needs a rebase

mantoni:
@mroderick was already on it ;)

mroderick:
Will you also make a PR into https://github.com/sinonjs/sinon/tree/v1.17 branch, so we can publish a new 1.17.x version?

mroderick:
(yes, we really need to get back to having master as the branch we're shipping from)

mantoni:
Ah, good point. Will do.

PR #873, opened 2015-10-08
CommonJS-ified *some* of the things

fearphage:
This is not my cleanest pull request. I moved `core`, `extend`, `typeOf`, and `match` to follow the CommonJS style. This means there are no more referencing "globals" (`sinon.*`).

```shell
➜  Sinon.JS git:(commonjs-ify) grep -r sinon lib/sinon/util/core
lib/sinon/util/core/wrap-method.js:        } else if (wrappedMethod.restore && wrappedMethod.restore.sinon) {
lib/sinon/util/core/wrap-method.js:        } else if (wrappedMethodDesc.restore && wrappedMethodDesc.restore.sinon) {
lib/sinon/util/core/wrap-method.js:    method.restore.sinon = true;
lib/sinon/util/core/restore.js:    return typeof obj === "function" && typeof obj.restore === "function" && obj.restore.sinon;
lib/sinon/util/core/object-keys.js:        throw new TypeError("sinon.objectKeys called on a non-object");
lib/sinon/util/core/function-to-string.js:    return this.displayName || "sinon fake";
```

## Notable changes

* rewrote times-in-words for clarity and brevity
* I worked around a cyclic dependency between `core#deepEqual` && `match` by injecting the dependency
* `lib/sinon/util/core.js` became [lib/sinon/util/core/index.js](https://github.com/sinonjs/sinon/pull/873/files#diff-6af3eef342dccc1bf5f74a907c5f2ea1)
* I pulled each core function into its own file

## Notes:

The notion of `core` seems a little convoluted. `createStubInstance` seems to have gotten lost in core. I moved it to `lib/sinon/stub`. Tests are passing. Everything else that was in core, I left it there. I was going to make core not act as the namespace for sinon, but that changed spidered out into a big mess.


lib/sinon/util/core/times-in-words.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+var array = [null, "once", "twice", "thrice"];
+
+module.exports = function timesInWords(count) {
+    return array[count] || (count || 0) + " times";
```
fearphage: Rewrite.

lib/sinon/util/core/deep-equal.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function isElement(obj) {
+    return div && obj && obj.nodeType === 1 && isDOMNode(obj);
+}
+
+module.exports = function deepEqual(a, b) {
+    var match = deepEqual.match;
+
+    if (match && match.isMatcher(a)) {
```
fearphage: Injected dependency.

fearphage:
Did you know there are 2 different implementations of `times-in-words`? 

[One in core](https://github.com/sinonjs/sinon/pull/873/files#diff-fcd4362a232a46da8739cb9b2de652fcL324) and https://github.com/sinonjs/sinon/blob/a97374641599ed2f451e9a22f784060cf7a38b3e/lib/sinon/times_in_words.js

I didn't touch the extra copy in this PR.

lib/sinon/extend.js
```diff
@@ -54,13 +52,12 @@ var hasDontEnumBug = (function () {
...
+module.exports = function extend(target /*, sources */) {
     var sources = Array.prototype.slice.call(arguments, 1);
-    var source, i, prop;
-
-    for (i = 0; i < sources.length; i++) {
-        source = sources[i];
+    var i = 0;
+    var source, prop;
 
+    while ((source = sources[i++])) {
```
mroderick: This change doesn't actually change the behaviour of the code, but is is a departure from existing style. Please try to keep to the existing style.

lib/sinon.js
```diff
@@ -7,7 +7,21 @@
...
+exports.extend = require("./sinon/extend");
+exports.match = match;
+exports.typeOf = require("./sinon/typeOf");
+
+/*
+ * allow deepEqual to check equality of matchers through
+ * dependency injectection. Otherwise we get a circular
+ * dependency
+ */
+exports.deepEqual.match = match;
```
mroderick: Dependency injection is usually done via the arguments of a function. In this pattern, `deepEqual` will have different behaviour if it's extended by this file or not.

```javascript
function deepEqual(match, a, b) {
```

I would much prefer using an extra argument, if that mean solving circular dependency problems

lib/sinon/extend.js
fatso83: Agreed, I find the change to be for the worse regarding readability. If it were to be changed for fewer variables I'd much more prefer to simply use a `sources.forEach()` construct, but that would leave ES3 browsers out in the cold. So just keep the existing variant.

lib/sinon.js
fatso83: I think modifying an other module's exported function is surprising behaviour, and I'd much rather have explicit code than implicit code to avoid surprises :-)

fearphage: @mroderick Who would make that call? That's breaking backwords compatibility. `deepEqual(matcher, foo)` will change behavior. That's what I was trying to avoid.

fearphage: I had some other incantations that I tried.

```js
module.exports = exports = function(match) {
  exports.deepEqual = function(a, b) {
    if (match.isMatcher(a)) {
      return a.test(b);
    }

    return deepEqual(a, b);
  };
};
```

That felt gross.

fearphage: What about this?

```js
module.exports = exports = function(match) {
  exports.deepEqual.match = match;
};
```

or I could go with private data:

```js
var match;

module.exports = exports = function(_match) {
  match = _match;
};
// ...
function deepEqual(a, b) {
  if (match && match.isMatcher(a)) {
    return a.test(b);
  }
};
```

lib/sinon/extend.js
fearphage: Understood.

lib/sinon.js
fearphage: Or I could require a function call to get deepEqual returned:

```js
module.exports = function(match) {
  return function deepEqual(a, b) {
    if (match && match.isMatcher(a, b)) {
  };
};
```

Imports would look like

```js
let deepEqualWithoutMatcher = require('.../core/deep-equal')();

let deepEqualWithMatcher = require('.../core/deep-equal')(require('sinon/match'));
```

This also breaks backcompat.

mroderick:
Yeah, this is pretty convoluted ... one of the reasons my past attempts at CommonJS never got completed

fearphage:
The entire PR or something in particular?

lib/sinon.js
fatso83: Such a nasty business, but in order not to break backwards compatibility I would go in the direction of your upgrade function mixed with the last version. What about ending up with something like this:

```
let matcher = require('sinon/match');
let deepEqualWithoutMatcher = require('.../core/deep-equal');
let deepEqualWithMatcher = require('.../core/deep-equal').withMatcher(matcher)
```

You could keep your old and get a new version as well.

fatso83: Regarding the code in my previous comment, what about something like this?

``` javascript
function createDeepEquals(match){
    return function(a,b){
        if (match.isMatcher(a)) {
            return a.test(b);
        }
        return deepEquals(a,b);
    }
}

function deepEquals(){
    // existing code
}
deepEquals.using = createDeepEquals;

module.exports.deepEquals = normalDeepEquals;
```

fearphage: @fatso83 Impressive. I deleted that exact code. I even used the name `withMatcher` which I thought was suboptimal.

That works for me. Do we need more people to weigh in?

mroderick:
It's not my day for clear communication. I meant the the current implementation is convoluted

lib/sinon.js
fatso83: @fearphage Discussion by text is hard - not sure I am reading you correctly :-) So to clear things up; did you somehow create that same snippet of code in some earlier commit that was then purged in some rebase, or was something like this present in the existing code. In both cases, I clearly missed it somehow, and in any case, it shows that we are living in the same code culture. 

I also found `withMatcher` to be suboptimal. Not sure if my rename to `.using` is any better, but I think it reads pretty well (?). 

If it seems fine by you, might as well hear if the API fits the mind of @mroderick

fearphage: On the road to the solution you see here, I had `require('.../core/deep-equal').withMatcher(match)` in my code at some point. I removed it in favor of this solution which seemed cleaner at the time.

I also tried:

* `deepEqual.upgrade(m)`
* `deepEqual.allowMatcher(m)`
* `deepEqual.checkMatcher(m)`

Just throwing out ideas. `.use(match)` for some reason sounds better to me than `.using`. I'm flexible though.

fatso83: +1 on `.use(match)`. Btw, you are 9-10 hours after us, so we are already way into the weekend, so I wouldn't wait up for feedback :-) Checking out for now.

fatso83:
To me, if `lib/sinon.js` was modified to have `module.exports.deepEqual = deepEqual.use(match)`, and added the corresponding `use` (or similar) attribute/upgrade function to `deepEqual` in [`core/deep-equal`](https://github.com/sinonjs/sinon/pull/873/files?diff=split#diff-7a3ea2d1691331a2e906b8f6a01a531cR32), this pull looks to be in pretty good shape as it is not changing existing behaviour.

@cjohansen As a side-note, since this is mostly about reorganizing, would it make sense (in some later PR) to pull the entire hierarchy under lib/sinon one level up, replacing `lib/sinon.js` with `lib/index.js`? That extra level of indirection really isn't adding much today.

lib/sinon/util/core/deep-equal.js
```diff
@@ -0,0 +1,103 @@
...
+    for (prop in a) {
+        if (a.hasOwnProperty(prop)) {
+            aLength += 1;
+
+            if (!(prop in b)) {
+                return false;
+            }
+
+            // allow alternative function for recursion
+            if (!(arguments[2] || deepEqual)(a[prop], b[prop])) {
```
fearphage: I did it this way to maintain the airity of `deepEqual`.

lib/sinon.js
```diff
@@ -7,7 +7,21 @@
...
+exports.extend = require("./sinon/extend");
+exports.match = match;
+exports.typeOf = require("./sinon/typeOf");
+
+/*
+ * allow deepEqual to check equality of matchers through
+ * dependency injectection. Otherwise we get a circular
+ * dependency
+ */
+exports.deepEqual = exports.deepEqual.use(match);
```
fearphage: This is still changing the `core` "global". Once this is include, the only way to access `deepEqual` *without* matchers is to include it directly. This is a temporary problem. This will be fixed in the future when the commonjs-ification is completed and/or when `core` goes away.

fearphage:
My local build is not failing on `│v0.12.7`. I don't know.

fatso83:
I restarted the build. Passing with flying colors.  Good work! Also like that you spend some time commenting on your design choices.  Makes it easier to grasp your thinking.

fearphage:
> Also like that you spend some time commenting on your design choices. Makes it easier to grasp your thinking.

Just saw this. Thanks. Glad I could make it better/easier to digest. :+1:

PR #879, opened 2015-10-14
Removed duplicate implementation of sinon.timesInWords

fearphage:
`sinon.timesInWords` is already a [part of core](https://github.com/sinonjs/sinon/blob/22e39e5f07a306604f40c1b52a6a586d70626a36/lib/sinon/util/core.js#L324-L329).


mroderick:

It's nice to get rid of duplication. However, I would rather keep the standalone version and be rid of the one from `core.js`

fearphage:
After a few merges, the [one in core](https://github.com/sinonjs/sinon/blob/acfb4b284048c4f1c37756eff68a5761a7e42506/lib/sinon/util/core/times-in-words.js) is now *more* standalone (not changing another module's export) than this one.

mroderick:
> After a few merges, the one in core is now more standalone (not changing another module's export) than this one.

Cheers!

PR #881, opened 2015-10-15
Make travis cache node_modules to speed up builds

mroderick:
#### Before

2015-10-15 at 16 55

#### After
2015-10-15 at 16 56

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ node_js:
 
 sudo: false
 
+cache:
+  directories:
+  - node_modules
+
 before_install:
   - npm config set strict-ssl false
   # use a more modern npm than ships with 0.8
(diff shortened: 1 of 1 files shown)


mantoni:
How cool is that! Didn't know this is possible :+1:

.travis.yml
```diff
@@ -7,17 +7,14 @@ node_js:
...
   - npm config set strict-ssl false
   # use a more modern npm than ships with 0.8
   - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.8" ]; then npm install -g npm@1.4; fi'
-  # Workaround for intermittent build failures
-  # install buster separately from `npm install`, as it uses the C compiler, which might cause the
-  #   npm ERR! cb() never called!
-  # errror to appear, causing the builds to fail
-  #
-  # See this discussion: https://github.com/npm/npm/issues/7014
-  - npm install buster@0.7.18
```
fearphage: Will this break when you `rebuild without cache` (or whatever it's called in Travis)? I can't see the button myself.

mroderick: Let's find out!

I've nuked the cache and restarted one of the builds

mroderick: Still works, guess which one I restarted ;-)

2015-10-15 at 21 22

PR #882, opened 2015-10-15
Removed unneeded path resolution

fearphage:
Every `package.json` script is run with `$(npm bin)` in the path. So we don't need to manually add it.


mantoni:
I don't remember why, but this was put there on purpose.

mroderick:
I remember this as being important, but can't remember why. I'll take it for a spin

mroderick:
It still works on my machine and on Travis CI, I think we should merge it

fearphage:
The only way this is able to fail would be if something runs it outside of `npm`.

PR #886, opened 2015-10-20
Fix #875 Proper support UTF8 payloads

### LINKED ISSUE #875: Proper support UTF8 payloads
Sinon version : 1.17.1 Environment : Windows 8.1  all Browsers (tested with IE9-11, Firefox latest, Chrome latest)

Bug description

With the following commit https://github.com/sinonjs/sinon/commit/f8091656c6da4411cf25aeaeef6327405bd10356
ArrayBuffer and Blob support are introduced. So every response body now gets encoded according the request.responseType setting. 

One problem with the current implementation of that feature is proper UTF8 support. The converstion for now can only handle ASCII characters https://github.com/sinonjs/sinon/commit/f8091656c6da4411cf25aeaeef6327405bd10356#diff-4b4d9095007a078c5beda46091f1111aR293. 

It would be great to proper encode the body with the help of TextEncoder. An implementation example can be found in the specs https://encoding.spec.whatwg.org/#utf-8-encode. For cross browser support <IE10 there exists even a polyfill https://github.com/inexorabletash/text-encoding

If you accept that proposal I could write a PR for that. 
Thx for your effort so far.

johann-sonntagbauer:
* introduced new dependency "text-encoding"
* delegate encoding operations to TextEncoder/TextDecoder
* added unit test to verify proper utf-8 encoding

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -11,6 +11,7 @@
 require("../extend");
 require("./event");
 require("../log_error");
+var TextEncoder = require("text-encoding").TextEncoder;
 
 var sinon = require("./core");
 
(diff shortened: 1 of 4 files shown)


johann-sonntagbauer:
Strange the testcase works on my machine. Is there some problem with the test environment?

fatso83:
It sometimes happens that the builds fail. I'll restart it.

johann-sonntagbauer:
:D thx

mroderick:
This looks good to merge to me.

Issue #887, opened 2015-10-20
deepEqual fails on objects without a prototype or that override hasOwnProperty


jollytoad:
I'll raise a PR to illustrate the problem with a test and supply a fix.

jollytoad:
https://github.com/sinonjs/sinon/pull/888

PR #889, opened 2015-10-20
Tests and fixes for Issue #887 (single commit)

jollytoad:
Use Object.prototype.hasOwnProperty in deepEqual to cope with cases where hasOwnProperty doesn't exist, ie. Object.create(null), or has been overridden on an object.
With tests.

Fixes: #887

### DIFF
--- a/lib/sinon/util/core/deep-equal.js
+++ b/lib/sinon/util/core/deep-equal.js
@@ -69,7 +69,7 @@ var deepEqual = module.exports = function deepEqual(a, b) {
     }
 
     for (prop in a) {
-        if (a.hasOwnProperty(prop)) {
+        if (Object.prototype.hasOwnProperty.call(a, prop)) {
             aLength += 1;
 
             if (!(prop in b)) {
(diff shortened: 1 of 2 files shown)


fatso83:
Thanks! I think the commit looks good enough to merge, but I could like some input on semantics before merging. Is an object without a prototype truly equal to one without? I think it is - in the way this would mostly be used, but perhaps someone might cook up some situation I haven't thought of.

jollytoad:
Well the current behaviour of deepEqual is to not check whether the prototypes differ anyway, so whether a prototype is non-existent should make no difference either.

Test to confirm this behaviour:
```js
            "passes equal objects that have different prototypes": function () {
                var obj1 = Object.create({ one: 1, two: 2 });
                obj1.a = 1;
                obj1.b = 2;
                obj1.c = "hey";

                var obj2 = Object.create({ one: "a", two: "b" });
                obj2.a = 1;
                obj2.b = 2;
                obj2.c = "hey";

                assert(sinon.deepEqual(obj1, obj2));
            }
```

mroderick:
@jollytoad nice work!

I think we should merge this

PR #891, opened 2015-10-21
Use immediate exceptions

mroderick:
Since we're getting ready to do a new major release, I think now is a good time to change the default to use immediate exceptions instead of throwing Error in a new call stack.

People are still getting confused, see #172


mantoni:
Makes sense :+1:

mroderick:
I am merging this!

PR #894, opened 2015-10-23
Allow requestBody in FakeXDomainRequest, FakeXMLHttpRequest for GET requests

JoshuaKGoldberg:
It looks like you're explicitly stopping both GET and HEAD requests from allowing the request body to have a body. 

Why not GET? It's in the spec that GETs should be allowed a body, which makes sense for APIs that take in large amounts of data for GET requests.

Reference quoted from http://bugs.jquery.com/ticket/8961:

> This is all explained in the  HTTP 1.1 Spec, RFC 2616...
> 
> Essentially it breaks down to the fact that the message body should be allowed unless specified otherwise in the request type specifications, as in HEAD. The GET request does not mention the message body at all, so by default **should be able to use a message body**. This is useful if you want to query the server with structured (e.g. JSON) data.

### DIFF
--- a/lib/sinon/util/fake_xdomain_request.js
+++ b/lib/sinon/util/fake_xdomain_request.js
@@ -107,7 +107,7 @@ sinon.extend(FakeXDomainRequest.prototype, sinon.EventTarget, {
     send: function send(data) {
         verifyState(this);
 
-        if (!/^(get|head)$/i.test(this.method)) {
+        if (!/^(head)$/i.test(this.method)) {
             this.requestBody = data;
         }
         this.requestHeaders["Content-Type"] = "text/plain;charset=utf-8";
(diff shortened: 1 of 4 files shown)


JoshuaKGoldberg:
*(I'll update the unit tests later today)*

fatso83:
Seems fine to me, although an unconventional use case :-) Thanks for the doc links.

fearphage:
You should add a positive test.

JoshuaKGoldberg:
Will do.

JoshuaKGoldberg:
Added the tests and realized the problem also exists for FakeXMLHttpRequest, so I added the change (and tests) there too.

mroderick:

Would you mind squashing it down to just one commit, so we can merge it?

JoshuaKGoldberg:
Done!

JoshuaKGoldberg:
:) Thanks for accepting the PR!

PR #895, opened 2015-10-24
Removed unit test for GET body being null

JoshuaKGoldberg:
Now that GET bodies are allowed (as per the spec), this is no longer
true.


JoshuaKGoldberg:
Ugh clearly I don't know how to use the GitHub pull requests.

PR #898, opened 2015-10-26
Removed eslint switch statement warnings

fearphage:
* added an eslint ignore for a valid switch statement
* replaced readyStateChange switch with if statement

Now there are no warnings or errors (when everything is done well).

### DIFF
--- a/lib/sinon/util/fake_xdomain_request.js
+++ b/lib/sinon/util/fake_xdomain_request.js
@@ -70,26 +70,20 @@ sinon.extend(FakeXDomainRequest.prototype, sinon.EventTarget, {
     readyStateChange: function readyStateChange(state) {
         this.readyState = state;
         var eventName = "";
-        switch (this.readyState) {
-            case FakeXDomainRequest.UNSENT:
-                break;
-            case FakeXDomainRequest.OPENED:
-                break;
-            case FakeXDomainRequest.LOADING:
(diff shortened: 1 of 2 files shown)


mroderick:
The `default-case` warning is there for a reason, it's to make authors consider what should happen in the default case, when none of the other cases have matched.

Rewriting the code to use `if-else` is not really addressing the un-handled case, nor does disabling the warning.

I would rather see the a pull request that addresses the situation by handling un-handled cases in a `default` case.

fearphage:
> Rewriting the code to use if-else is not really addressing the un-handled case

I think it shouldn't have been a `switch` statement in the first place is the real answer. I don't think there's a problem that there is no work to be done in some of the XHR states.

> I would rather see the a pull request that addresses the situation by handling un-handled cases in a default case.

Should it throw an error or what?

fearphage:
@mroderick Bump. See questions above.

fatso83:
@fearphage A shame to leave you waiting for months, but I guess that is how open source maintenance works at times. I would like to get this PR closed, but I thought I might as well try to fix the issues @mroderick referred to, and I did that in #978. I found that the empty, but explicit switch cases were more telling than leaving them out, and so I just added `default` cases of throwing errors. 
All the cases seem to be handled, but at least now we will know if we have left anything out ...

PR #899, opened 2015-10-26
Replaced switch statement in favor of object lookup

fearphage:
So this is a little easier to follow and it's more performant.

### DIFF
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -55,68 +55,68 @@ function matchObject(expectation, actual) {
     return true;
 }
 
+var TYPE_MAP = {
+    "function": function (m, expectation, message) {
+        m.test = expectation;
+        m.message = message || "match(" + functionName(expectation) + ")";
+    },
+    number: function (m, expectation) {
+        m.test = function (actual) {
+            // we need type coercion here
+            return expectation == actual; // eslint-disable-line eqeqeq
+        };
+    },
+    object: function (m, expectation) {
+        var array = [];
+        var key;
+
+        if (typeof expectation.test === "function") {
+            m.test = function (actual) {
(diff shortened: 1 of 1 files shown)


mroderick:
> and it's more performant.

Have you measured it using http://benchmarkjs.com?

fearphage:
Switch statements, like if statements, get slower the lower you go in.

```js
switch(foo): {
  case 1: //fastest
  case 2: // slower
  case 1000: // slowest
}
```

The same can be said for `if-else` statements with a ton of `else-if`'s.

`foo in bar` is constant time `O(1)`.
`switch` and `if` statements are `O(n)`

I can delete that from the pull request message. It was merely a bonus and not the focal point of this PR. The main point is the code is cleaner and easier to read and maintain in my opinion.

fatso83:
Performance wise, this is probably quite a bit slower than a small lookup table that can be held in L1 cache, but we are talking nano-seconds here. On the big-Oh notation you are correct, but there is quite a bit of work involved in the lookup which is not done in a small switch statement, so there is a big C in those O(1)+C you are missing. Anyway, this is not a library focusing on performant code, and this code has not been identified as slow/problematic. Profile before refactoring for speed.

The important thing to me is readability, and I think you might have made an improvement in that regard. I would not mind merging it, but for later reference I think efforts are best spent on the bigger problems we are facing rather than micro-optimizations like this :-)

fearphage:
I was **not** attempting to make a performance improvement. I regret mentioning it at all now. I was attempting to simplify the code.

fatso83:
Haha, never mind. This seems legit, and I think it improves the readability, so I cannot see any reason not to merge it.

Thanks!

PR #910, opened 2015-11-01
Use ESLint --cache for CI speedup

mroderick:
With `eslint@1.4.0` came the new [`--cache` flag](http://eslint.org/blog/2015/09/eslint-v1.4.0-released/#cached-results).

#### Before

type | Local | Travis
--------|---------|-----
real | 0m3.444s | 0m4.984s
user | 0m3.323s | 0m5.428s
sys | 0m0.132s | 0m0.196s

#### After

type | Local | Travis
--------|---------|-----
real | 0m0.949s | 0m4.916s
user | 0m0.854s | 0m5.211s
sys | 0m0.095s | 0m0.136s

I've moved the `.eslintcache` file into `node_modules`, which gets cached in the Travis build cache. And yet, it's still just as slow on Travis, yet speed up locally.

Tips welcome!

cc: @mantoni @fatso83 @cjohansen @tenzer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 pkg
 node_modules
 coverage/
+.eslintcache
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ sudo: false
 
 cache:
   directories:
+  - .eslintcache
   - node_modules
 
 before_install:
@@ -18,7 +19,7 @@ before_install:
 
 before_script:
   # we only need to run eslint once per build, so let's conserve a few resources
-  - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.12" ]; then $(npm bin)/eslint .; fi'
+  - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.12" ]; then ./scripts/ci-eslint; fi'
 
 script:
   - ./scripts/ci-test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ sudo: false
 
 cache:
   directories:
-  - .eslintcache
   - node_modules
 
 before_install:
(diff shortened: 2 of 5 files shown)


mantoni:
Not sure if you saw it: for speeding up development with an eslinted project, I've made `eslint_d` to eliminate the startup time on subsequent invocations: https://www.npmjs.com/package/eslint_d

Obviously, this doesn't help at all on travis, but you might be able to use it in your editor.

fatso83:
Hmm ... that is mysterious. I have no idea. 
console.log-ing various environment variables like $PATH and $CWD to see if something is different when running on Travis? Sorry, no great ideas here.

mantoni:
Oh, I guess it's not getting any faster with caching on Travis, because it does timestamp comparisons and all the files are checked out freshly on each build.

mroderick:
@mantoni interesting. How did you find out? Perhaps we can improve ESLint to overcome this scenario

mantoni:
@mroderick I don't know. That was my issue debugger background task there. It just popped the solution into my consciousness. I had a look, and eslint uses [file-entry-cache](https://www.npmjs.com/package/file-entry-cache) which checks file size and mtime from stat calls. If one of the two is different from the last check, it marks files as modified. This would have to change into some sort of md5 hash to work for the travis use case.

mroderick:
> This would have to change into some sort of md5 hash to work for the travis use case.

Yeah, I wonder if we can do hashing fast enough to be of real use.

mroderick:
https://github.com/eslint/eslint/pull/4019

mroderick:
This PR should be updated to use latest ESLint before merging.

mantoni:
The issue you linked is just about hashing the configs, isn't it?

mroderick:
I don't think this is worth pursuing right now (I don't have time to contribute to ESLint also).

PR #911, opened 2015-11-02
Set up SauceLabs tests

mroderick:
This is an experiment, and is not exactly ready for review let alone merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,12 @@ script:
 
 git:
   depth: 10
+
+# set up SauceLabs
+env:
+  global:
+    - secure: false
+    - secure: false
+
+addons:
+  sauce_connect: true


dominykas:
@mroderick this is one way to do it:
* https://github.com/h5o/h5o-js/blob/master/.travis.yml#L14
* https://github.com/h5o/h5o-js/blob/master/Gruntfile.js#L86
* https://github.com/h5o/h5o-js/blob/master/Gruntfile.js#L110
* https://www.npmjs.com/package/buster-reporter-sauce

i.e. use `buster-static` to run tests (fire it up on a "well known port", e.g. 8080), use `buster-reporter-sauce` to tell sauce the results. My solution uses `grunt` - but it should be possible to get away without `grunt`. I will also look into the approach `when` are using.

fatso83:
:clap:

dominykas:
`when` are also using `buster-static`: https://github.com/cujojs/when/blob/master/test/sauce.js#L144 except that they are parsing the HTML results themselves, rather than using a reporter and reading the results from sauce.

PR #913, opened 2015-11-06
Prevent stubbed getter from being called during restore()

### LINKED ISSUE #897: Stubbed getter called on restoring
Let's take an example with subbed getter and restoring:
````js
        var obj = {
            get prop() {
                console.log('Getter content called!');
                return 42;
            }
        };
        var stub = sinon.stub(obj, 'prop', {get: function () {
            return 43;
        }});
        console.log('Stubbed output', obj.prop);
        // restoring
        stub.restore();
````
When running this code I'm getting correctly subbed `obj.prop`, but on restore getter content is called!
````
Stubbed output 43
Getter content called!
````
And in my real test I'm getting error as environment for getter is not mocked!
It is possible to not call stubbed getter on `restore()`?

Any help appreciated!

fearphage:
It does what it says on the tin. This fixes #897.

deal-with-it-cat

### DIFF
--- a/lib/sinon/util/core/wrap-method.js
+++ b/lib/sinon/util/core/wrap-method.js
@@ -112,10 +112,18 @@ module.exports = function wrapMethod(object, property, method) {
             Object.defineProperty(object, property, wrappedMethodDesc);
         }
 
+        if (hasES5Support) {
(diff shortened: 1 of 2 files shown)


fatso83:
A good read. Nice!

PR #918, opened 2015-11-09
Fix interaction with spy getters/setters - fixes #917

### LINKED ISSUE #917: Getters/setters don't work with spies
Problem:

The following line throws, because `sinon.wrapMethod` gets [passed the entire property descriptor](https://github.com/sinonjs/sinon/blob/6a79d61d5a51aff4cead9e681303ab19968c0d1b/lib/sinon/spy.js#L29-L34) (`{get, set, enumerable, configurable}`) and attempts to wrap every "method" (read: property) in the descriptor. 

```js
var spy = sinon.spy(object, "property", ["get"]);
```

The first one to throw is `set` which is `undefined`.

Side note: I don't see any documentation about getters/setters spies and stubs anywhere.

fearphage:
Now passing a user -directed limited descriptor to `wrapMethod` so it doesn't blow up.

I also found a missing test (in `test/call-test.js` strangely) that pushes `lib/sinon/spy.js` coverage to 100%.

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -18,6 +18,8 @@ var slice = Array.prototype.slice;
 var callId = 0;
 
 function spy(object, property, types) {
+    var descriptor, i, methodDesc;
+
     if (!property && typeof object === "function") {
         return spy.create(object);
     }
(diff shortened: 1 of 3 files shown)


fearphage:
This fixes #917.

fearphage:
Looking at #88 where this was added, I don't believe this ever worked in spies at least. 

It also wasn't tested when it was added. It would be nice to be able to tell when coverage decreased on a file. It's now back to 100% after this PR.

fatso83:
Great work! I like the thorough investigation :) Any idea on how we could track coverage? We cannot be too pedantic about this, as sometimes coverage can drop a little bit after refactoring, yet still the result

fearphage:
There are services that offer free services to OSS:

coveralls - https://coveralls.io/
codeship - https://codeship.com/

There are more. That's just what comes to mind. One awesome thing about coveralls is it shows updates in the pull requests. So it will post/show in the PR "Coverage dropped 5% in this PR".

fearphage:
> We cannot be too pedantic about this, as sometimes coverage can drop a little bit after refactoring, yet still the result

Well in the case of #88, there were just missing tests. It would be good to catch it going in instead of after the fact.

fatso83:
Thanks for tips: maybe we should invest some time in something like coveralls.

mroderick:
I would love to have long term code coverage monitoring for all the sinonjs repositories

PR #920, opened 2015-11-11
Commonjs-ified sinon.spy

fearphage:
Updated `sinon.spy` to make use of cjs'd modules and to export itself using commonjs styling.

## Notable changes
* moved `sinon.format` to core in the process. Core seems like a placeholder that will one day go away
* sinon.spyCall is exposed solely for testing (both uses of it internally are commonjs-fied)
* Due to a few tests, I could not remove the `sinon/core` include from `spy` (related to `sinon.format`)


lib/sinon.js
```diff
@@ -14,6 +14,8 @@ module.exports = exports = require("./sinon/util/core");
 
 exports.extend = require("./sinon/extend");
 exports.match = match;
+exports.spy = require("./sinon/spy");
+exports.spyCall = require("./sinon/call");
```
fearphage: Exposed solely for testing purposes? It seems like an internal tool only though.

lib/sinon/spy.js
```diff
@@ -8,10 +8,15 @@
...
 
-require("./extend");
-require("./call");
-require("./format");
+var extend = require("./extend");
+var deepEqual = require("./util/core/deep-equal");
+var functionName = require("./util/core/function-name");
+var functionToString = require("./util/core/function-to-string");
+var getPropertyDescriptor = require("./util/core/get-property-descriptor");
 var sinon = require("./util/core");
```
fearphage: Cannot remove/replace due to `sinon.format` coupling in tests. :cry:

test/call-test.js
```diff
@@ -807,8 +805,9 @@
                 assert.equals(object.doIt.getCall(0).toString().replace(/ at.*/g, ""), "doIt() !TypeError(Oh noes!)");
             },
 
+            // these tests are ensuring that call.toString is tightly coupled to sinon.format
```
fearphage: See note.

fearphage:
Just rerunning the tests. :wink:

lib/sinon/spy.js
fatso83: Cannot remove what exactly? You seem to be replacing require calls here, and the tests are running fine, so I am not quite sure what you refer to ...

test/call-test.js
fatso83: Do you see a way we could we rewrite the tests somehow to avoid the tight string formatting coupling?

lib/sinon/spy.js
fearphage: I have to use `sinon.format` from core instead of`var format = require('.../format')`.

mroderick:
@fearphage this is a very nice pull request. I love how easy it is to follow along, when there are separate commits for each step. :+1:

test/call-test.js
fearphage: I don't see an easy way. I'd have to inject `format` as a dependency like [deepEqual + matcher](https://github.com/sinonjs/sinon/commit/5fd9821a3454e4261e5bbca697fd3f528b1b95b3#diff-7ac78866bb2aab0e4513c3e57f189c88R24). The relationship and usage is not nearly as clean though. I'll give it another look after some rest.

fearphage: The solution is for the tests to use the commonjs module format to include exactly what it needs instead of the global sinon object.

fearphage:
@mroderick Thank you. I appreciate it.

lib/sinon.js
mantoni: I think this was used internally. I think it can be removed since it would be released with a major. @cjohansen do you recall?

mantoni:
Very nice, indeed. Are you intending to clear up the "format coupling" within this pull request or is this going to happen separately?

lib/sinon.js
cjohansen: Not a part of the external API at least.

fearphage: It's exposed publicly, because that's how it is tested (see `call-test.js`).

fearphage:
I think the tests need to be refactored in a big way to fix the call+format tests. I think this code is good to merge. I'm always open to suggestions though.

fearphage:
So what's the word, folks?

fatso83:
Sorry for keeping you waiting. Very nice work, indeed.

fearphage:
FYI, you can now import this separately from everything else.

```shell
$ node
> var foo = require('sinon/lib/sinon/spy')();
> foo.callCount
1
```

I haven't used it, but I know it can be included at least. I was wondering if anyone considered pulling out the separate pieces into different repos or at least different npm targets. Imagine `npm install sinon-spy`. Just something to think about.

Thanks everyone.

cjohansen:
Awesome! That's definitely something I've thought about more than once.

PR #923, opened 2015-11-15
Clarify documentation on creating stubs and spies

ichthala:
While learning how to use Sinon, some members of my team struggled a bit in understanding Sinon's spies/stubs. Specifically, it wasn't immediately clear how `sinon.spy()` is different from `sinon.spy(object, "method")`. These are proposed changes to the documentation that make the differences between the method signatures of `sinon.spy` clearer to newcomers.

This is my first open source contribution, so if I'm getting anything wrong about the contribution process, please let me know.

### DIFF
--- a/docs/current/spies.md
+++ b/docs/current/spies.md
@@ -7,15 +7,17 @@
 ### What is a test spy?
 
 A test spy is a function that records arguments, return value, the value of
-`this` and exception thrown (if any) for all its calls. A test spy can be an
-anonymous function or it can wrap an existing function.
+`this` and exception thrown (if any) for all its calls. There are two types of spies:
+Some are anonymous functions, while others wrap methods that already exist in
+the system under test.
 
 
-### When to use spies?
+### Creating a spy as an anonymous function
 
-Test spies are useful to test both callbacks and how certain functions/methods
-are used throughout the system under test. The following simplified example
-shows how to use spies to test how a function handles a callback:
+When the behavior of the spied-on function is not under test, you can use an
+anonymous function spy. The spy won't do anything except record information
+about its calls. A common use case for this type of spy is testing how a function
+handles a callback, as in the following simplified example:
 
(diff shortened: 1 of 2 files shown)


docs/current/spies.md
```diff
@@ -7,15 +7,16 @@
 ### What is a test spy?
 
 A test spy is a function that records arguments, return value, the value of
-`this` and exception thrown (if any) for all its calls. A test spy can be an
-anonymous function or it can wrap an existing function.
+`this` and exception thrown (if any) for all its calls. There are two types of spies:
+Some are anonymous functions, while others wrap previously existing methods.
```
fatso83: Either it exists or it does not exist. A previously existing method has been removed. 

I think this specific change is detrimental, but otherwise it seems fine.

docs/current/stubs.md
```diff
@@ -69,13 +69,10 @@ Creates an anonymous stub function
 
 #### `var stub = sinon.stub(object, \"method\");`
 
-Replaces `object.method` with a stub function.
+Replaces `object.method` with a stub function. An exception is thrown if the property is not already a function, to help avoid typos when stubbing methods.
```
mroderick: I would shorten the addition to "An exception is thrown if the property is not already a function."

We generally try to steer clear of explaining motivation, and only document behaviour.

mroderick:
@ichthala thank you for your pull request!

We're currently in a transition phase with the documentation site for Sinon.JS, so once this PR is merged,  someone will have to also update https://github.com/sinonjs/sinon-docs.

We can create an issue to import the changes from this PR and then someone can pick that up.

docs/current/stubs.md
ichthala: I agree, actually -- that clause was in there before and I didn't feel comfortable ripping it out. I'd be happy to do it though!

Issue #928, opened 2015-11-23
spy: Expose the function being spied upon?


papandreou:
Hi, we have this crazy API in the unexpected-sinon module where you can specify the calls you expected were made via a function that performs the calls: http://unexpected.js.org/unexpected-sinon/assertions/array-like/to-have-calls-satisfying/ (see the bottom example).

That works great as long as calling the spies has no side effects, ie. when created as `sinon.spy()`. However, it doesn't work so great when they do, eg. when spying on real code. For the purpose of "recording" the calls made by the function, we already monkey-patch the spy in all sorts of ways, but the function being spied on (`func`) is not exposed: https://github.com/sinonjs/sinon/blob/master/lib/sinon/spy.js#L73-L86

So, would you be likely to merge a PR that exposes `func` on the spy instance?

papandreou:
I tried to implement it just to verify that it would work, and it does: #929

papandreou:
Withdrawn (see https://github.com/sinonjs/sinon/pull/929#issuecomment-159090923)

PR #929, opened 2015-11-23
Expose proxy.func

papandreou:
See #928


fatso83:
Possibly fine, but the [test suite is complaining](https://travis-ci.org/sinonjs/sinon/jobs/92738821).

papandreou:
Whoopsie-daisy. Went in to find out what `withArgs` is about, and it turns out that what I'm after can already be accomplished by just setting `spy.func` from the outside due to: https://github.com/sinonjs/sinon/blob/master/lib/sinon/spy.js#L177

PR withdrawn. I hope this is a public API :)

PR #908, opened 2015-10-28
Fixed #758 by checking the global variables are set

Serfenia:
By adding a function that checks if the 'window' or 'global' is set, it then picks corresponding one. This makes it work in a browser and Node.js (strict and non-strict). The strict mode caused the this pointer in the module to become undefined.

PR #930, opened 2015-11-25
Ensure support for running inside a WebWorker

fatso83:
Related to the changes in #908.

It turned out that although we support running Sinon inside a WebWorker, we do not actually test this in any of the tests. So although manual inspection from sharp eyes might reveal that some change breaks the support  (thanks @mroderick), some not-so-sharp eyes (me) might not.

This test ensures we check for this by making a WebWorker load the packaged build and exercise some of sinon's functions. The PR is for the master branch, but I also ensured it runs fine on the v1.17 branch as well (where it was needed).

Up for discussion
To avoid loading the test in the `source` config I had to avoid getting the test matched with the `test/*/**.js` pattern, so I named the test `webworker-support-assessment.js`, which is kind of an awkward name, but at least dealt with the problem in a way...

### DIFF
--- a/test/buster-packaged.js
+++ b/test/buster-packaged.js
@@ -17,3 +17,18 @@ config.packaged = {
         "test/**/*-test.js"
     ]
 };
+
+// The sole purpose of this test setup is to
+// ensure that Sinon could be run inside of
+// a WebWorker
+config.webworkerSupport = {
+    environment: "browser",
+    rootPath: "../",
+    sources: [
+        "pkg/sinon.js",
+        "test/webworker-script.js"
+    ],
+    tests: [
+        "test/webworker-support-assessment.js"
+    ]
+};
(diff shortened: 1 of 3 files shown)


fatso83:
Not sure why the commit id in Travis differs from this one. I updated the branch with a rebase, producing `f5c2164`, but Travis fails on commit id `8c5f410`. Weird.

fatso83:
Argh, I was testing this in Phantom JS 2.0. See now that it is failing on Travis CI running PhantomJS 1.9.8. The SlimerJS merge cannot get here soon enough. 

Back to testing why 1.9.8 won't work then ...

mantoni:
I know little about web workers, but this looks good to me :+1:

fatso83:
It's a little test, but thought it might come in handy :-) If you think this looks fine, I'll add this to the 1.17 branch too (where the need for it arose).

mroderick:
Nice!

PR #933, opened 2015-11-29
CommonJSify `sinon.mock`

jonnyreeves:
Following on from #920 and #932, I've extracted `sinon.mock`; and just for fun, `sinon.assert` and `createSpyCall` into standalone CommonJS modules! :rocket: 

## Notable Changes
* Extracted `sinon.expectation` into it's own module, `mock-expectation` to satisfy the tests and keep the `mock` module clean.  Once the tests are migrated to CommonJS we can remove this from the public API.

PR #934, opened 2015-11-29
CommonJSify sinon utils package

jonnyreeves:
This builds upon #933 by migrating the contents of `/lib/sinon/utils` into CommonJS modules (apologies for not cutting these from master).  These modules now export an API whereas before they mutated the `sinon` object directly :rocket: 

## Notable Changes
* `FakeXMLHttpRequest` and `FakeXDomainRequest` both include methods to allow internal mutation which isn't great for encapsulation - however, IMHO we should wait to refactor / address this once the project has been fully migrated.
* Theres a handful of places where the `sinon` object is still imported due to the tight coupling in test-cases; once the tests are migrated these should fall out.

### DIFF
--- a/lib/sinon.js
+++ b/lib/sinon.js
@@ -33,6 +33,11 @@ exports.CustomEvent = event.CustomEvent;
 exports.ProgressEvent = event.ProgressEvent;
 exports.EventTarget = event.EventTarget;
 
+var fakeTimers = require("./sinon/util/fake_timers");
+exports.useFakeTimers = fakeTimers.useFakeTimers;
+exports.clock = fakeTimers.clock;
+exports.timers = fakeTimers.timers;
+
 var fakeXdr = require("./sinon/util/fake_xdomain_request");
 exports.xdr = fakeXdr.xdr;
 exports.FakeXDomainRequest = fakeXdr.FakeXDomainRequest;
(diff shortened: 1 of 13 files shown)


lib/sinon.js
```diff
@@ -12,14 +12,45 @@ var match = require("./sinon/match");
...
 exports.match = match;
 exports.spy = require("./sinon/spy");
 exports.spyCall = require("./sinon/call");
 exports.stub = require("./sinon/stub");
+exports.mock = require("./sinon/mock");
+exports.expectation = require("./sinon/mock-expectation");
 exports.createStubInstance = require("./sinon/stub").createStubInstance;
 exports.typeOf = require("./sinon/typeOf");
 
+exports.log = function () {};
```
fatso83: What is the point of this?

fatso83:
It's great to see the pace at which these cleanups are happening these days. Looks fine to my eyes.

lib/sinon.js
jonnyreeves: sinon.logError uses it; I believe the idea is that consumers can supply their own `sinon.log` function to handle log messages raised by the fakeServer, fakeXhr, etc.  From the docs site:

> ### sinon.log(string)
> Logs internal errors, helpful for debugging. By default this property is a noop function, set it to something that prints warnings in your environment for more help, e.g. (if you are using JsTestDriver):
>
> ```sinon.log = function (message) {  jstestdriver.console.log(message); };```

mantoni:
Looks good to me. Thanks! :+1:

jonnyreeves:
So are we good to land this, or #933?

fatso83:
Saw no reason not to.

PR #936, opened 2015-11-30
Extract `sandbox` into a CommonJS module.

jonnyreeves:
Tests are tightly coupled to `sinon.useFakeTimers` and `sinon.fakeServer`; however we should be able to resolve this once we CJSify the tests. :rocket:

### DIFF
--- a/lib/sinon.js
+++ b/lib/sinon.js
@@ -20,6 +20,7 @@ exports.spy = require("./sinon/spy");
 exports.spyCall = require("./sinon/call");
 exports.stub = require("./sinon/stub");
 exports.mock = require("./sinon/mock");
+exports.sandbox = require("./sinon/sandbox");
 exports.expectation = require("./sinon/mock-expectation");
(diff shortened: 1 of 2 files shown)


jonnyreeves:
OK, so almost there!  Started hacking on the last couple (`test` and `test_case`) however I've run into a little problem with the way they mutate the `sinon.config` object; I seem to recall seeing this mentioned in a previous PR; but I can't find reference to.  

What's the plan here, deprecate the mutation of `sinon.config` for the 2.0 release?

fatso83:
I also remember something about removing sinon.config ... I'll let the others look into this.

mantoni:
I also remember that we planed to remove `sinon.config` entirely, right @cjohansen?

cjohansen:
Yep. `sinon.config` exists solely so you can configure the singleton Sinon instance. Ideally there should be no singleton.

jonnyreeves:
`sinon.config` conversation aside - are we good to merge this PR?

mantoni:
LGTM. Thanks!

Issue #924, opened 2015-11-16
Run source tests in browsers


mroderick:
It would be great if we could somehow implement a `require` function that would understand CommonJS modules, so we can run the source tests in browsers again, and get coverage reports for those also.

```
$ buster-test
Running 1515 tests in 1 runtime ...

test
4 deferred tests
1189 tests, 1993 assertions, 1 runtime ... OK
```

That's some 326 tests skipped, because they rely on browser features. My bet is on fake XHR and friends.

I had a quick experiment with https://github.com/dfcreative/require-stub, but couldn't get it working. I have no idea if that is a good solution for it.

jonnyreeves:
I find the testing structure a little confusing at present; what would the ideal structure look like?  Should we just run a single test-pass in Phantom/SlimerJS and remove the busterjs `node` pass?

mantoni:
This has been addressed a while back.

PR #937, opened 2015-12-01
CJSify Tests Spike

jonnyreeves:
# DO NOT MERGE :)

Uses browserify to create a single bundle from CommonJS test-sources which is executed in PhantomJS via buster.

I've removed all other test cases to make life easier during this spike; if we are to proceed with this approach we would want to remove all environment preconditions from the test-cases.

#### Pros
* Write tests in CommonJS!
* Only one test-pass!
* Quick build (browserify'd bundle is small)

#### Cons
* Lost coverage support (bane module causes a stack-trace with the previous config)
* BusterJS stack-traces on assertion failures reference those from the bundle


fatso83:
I am not quite sure if I get it ... I get the write tests in commonjs bit, but I am just not sure what you mean by "only one test pass". We would still need to test it in both Node and a browser (at least one).

mroderick:
Interesting!

I see "only one test" pass as a con though

We try to have wide support, and even run the tests in multiple node versions for the same reasons.

However, we can still keep running them in node, no problem there

mroderick:
We can generate coverage report for the node run

jonnyreeves:
Thanks for the review gents,

> I see "only one test" pass as a con though

I was basing this on the discussion in #924.

If we want to keep testing two test passes (NodeJS and Browsers), then maybe it would be a good idea to split the test folder up into a clear structure:

```
./sinon/
└── test
    ├── browser  <-- These only run in the Browser pass.
    └── unit  <-- These run in both the Node pass and the Browser pass
```

Let me know what you think and I can modify this spike to accomidate.

fatso83:
Two questions
1. What do you mean by _we would want to remove all environment preconditions from the test-cases_? Are you referring to the `requiresSupportFor: { WebWorker: supportsWorkers },` parts of the tests? 
2. Why would we need to split it into two parts? 

If an environment simply does not support a feature it will not run. Simple as pie. What if a functionality might be missing from Node, but then later on it is adopted; should we then move it into the other directory? This just reminds me of browser sniffing. Better to learn from front-end and do feature detection at runtime. Don't see what splitting brings to the table, other than visual pleasure by having fewer files in a directory.

jonnyreeves:
@fatso83, thanks for clarifying - sounds like I came into this task a little confused as to the outcome :)

It sounds like all we need to do is rewrite all tests to use CommonJS and then browserify the "packaged" tests before passing that bundle.js to buster?  Otherwise everything else remains the same.

jonnyreeves:
Closing this as I'm unsure as to what the desired outcome of 'commonjs-ifying' the tests would be.

PR #938, opened 2015-12-02
Use latest buster from git

dominykas:
Please keep this open - it will spare me the trouble of setting up extra travis builds while I'm upgrading `buster`'s dependencies. 

~~It will also let me use saucelabs once #911 is done.~~ Meh: https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests

Thanks :)


mroderick:
Neat trick ;)

fatso83:
@dominykas : seems like the Travis security features are preventing you from implementing the saucelabs feature as easily. Are you able to work around it?

dominykas:
We'll see. Whenjs expose their sauce credentials without encryption - maybe it's not a problem?

fatso83:
That just means anyone can use their credentials freely, and that can
quickly turn into a problem if saucelabs finds out. I don't think that is a
good strategy.

dominykas:
It would be a problem if the tokens were irrevocable and/or paid private accounts. As this is all open source - all tests and results are public. Not sure it's that big of a deal. I'll look for more info.

dominykas:
Here's some info on access keys from saucelabs: https://support.saucelabs.com/customer/en/portal/articles/2092509-protecting-your-access-key - the only really dangerous thing is if the account is suspended because of non-public tests and also I don't really have an idea what Sauce Storage does and if there's potential for abuse.

I have also found this quote at https://github.com/travis-ci/travis-ci/issues/1946#issuecomment-55500556:

> In either case, Saucelabs and BrowserStack don't seem to mind about credentials leaking

Ofc, that's heresay :) Still - angular keep it in public, seemingly, with approval from saucelabs: https://github.com/angular/angular.js/issues/5596#issuecomment-33776829

So maybe it's just a question of asking them nicely?

fatso83:
What's with all the opening and closing? Some form of "bumping"? :-)

dominykas:
Since I'm not a contributor here, I can't restart a build in Travis - I either need to push a code change, or open/close :)

This results in notifications, doesn't it?...

fatso83:
Ah, so that is why :-) Yeah, it results in some notifications, but I think it's fine when we know why. You know you can just run the ci-tests, yourself, right? You might not get all the various Node versions, but at least you get can test in whatever Node version you are running and the browser (PhantomJS).

dominykas:
Oh, I do that plenty, as I'm fixing the failures :) But I only uncovered some of them, because they _sometimes_ failed on Travis (not sure why, just yet) - you need to run tests a lot of times to find these.

fatso83:
I think the failed Travis builds are failed positives. Everyone gets them from time to time, and I think the consensus is that it is somehow related to PhantomJS being a bit flakey (thus the unfinished slimerjs attempt).

dominykas:
Yes, they are false positives from sinon's perspective - but I suspect there might be some things that can be done on buster's side to make it more sturdy against such issues. I narrowed it down to the hearbeats that are sent from the browser to buster-server - I don't know why they don't arrive in properly equal 250ms intervals as they should - it's likely some CPU blocking either on phantomjs or on buster's server side. Or both.

It's not a blocker problem for release - but I did uncover some unreported exceptions and some other minor improvements that I'm working on. There are also some historical bugs in buster around "All slaves timedout" - which may or may not be related. We'll see :) Still - what everyone wants is to make sure that their test framework is reliable enough to not distract from their actual work.

mroderick:
@dominykas do you still need this PR?

dominykas:
I'm a bit too busy with real life, so not at the moment. I'll reopen later if I need it again.

PR #943, opened 2015-12-08
Get rid of eval in sinon spy

### LINKED ISSUE #710: Question: is it possible to avoid "EvalError unsafe-eval"?
Sinon version: 1.13.0 (master)
Environment: OSX / Atom Package Manager (apm 0.142.0) / node 0.10.35

Context - working an Atom.io package, specifically [a code linter](https://github.com/AtomLinter/Linter) and the tests make liberal use of Sinon.

When using a stub it throws:
```
EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".
at createProxy (/Users/doug/Projects/Sinon.JS/lib/sinon/spy.js:86:86)
at Function.create (/Users/doug/Projects/Sinon.JS/lib/sinon/spy.js:145:29)
at Function.create (/Users/doug/Projects/Sinon.JS/lib/sinon/stub.js:86:42)
at Object.stub (/Users/doug/Projects/Sinon.JS/lib/sinon/stub.js:43:32)
at [object Object].<anonymous> (/Users/doug/Projects/linter/spec/linter-spec.coffee:14:31)
```

Found one attempt at a solution called [loophole](https://github.com/atom/loophole) but it [has issues](https://github.com/atom/loophole/issues/1).

Looking for any advice / suggestions?

ibash:
This gets rid of the eval in sinon spy and replaces it with a nice ~~christmas~~ case tree. Fixes #710

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -74,13 +74,25 @@ function createCallProperties() {
     this.lastCall = this.getCall(this.callCount - 1);
 }
 
-var vars = "a,b,c,d,e,f,g,h,i,j,k,l";
 function createProxy(func, proxyLength) {
     // Retain the function length:
     var p;
     if (proxyLength) {
-        eval("p = (function proxy(" + vars.substring(0, proxyLength * 2 - 1) + // eslint-disable-line no-eval
-            ") { return p.invoke(func, this, slice.call(arguments)); });");
+      switch (proxyLength) {
+        case 1: p = (function proxy(a) { return p.invoke(func, this, slice.call(arguments)); }); break;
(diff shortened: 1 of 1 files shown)


mantoni:
A tear drops my eye. Can you add a short comment to the code stating why it's written in this way? A future myself might feel like this could be done shorter with eval ;-)

ibash:
@mantoni done!

mantoni:
I can't think of any counter arguments for this. @cjohansen Do you have anything to add?

cjohansen:
Not really!

mantoni:
Thank you!

ibash:
:+1:

mroderick:
:clap:

PR #958, opened 2015-12-27
Prep for buster@vNext

dominykas:
Just some small tweaks, to make the transition smoother:

* increased buster runner timeout - will dig deeper into why it was happening later, but there's no harm to increase them
* adding a manual teardown to avoid problems on node 0.x using `Date.now` for some purpose
* ~~removing deprecated `buster-core` - `buster.nextTick` is available since, seemingly, `0.6`~~ breaks node 0.8 in a mysterious fashion - investigating
* adding node 5 to travis while at it


fatso83:
AFAI can tell, this could be merged, @dominykas ? Nothing else going in?

dominykas:
0.8 is still with a beta tag as it is some 2x slower in some cases. I'm still debugging that, although I have a lot less time.

I would stick with 0.7 for now, but sure - go for 0.8 if you're brave, I just can't promise I'll fix bugs reasonably fast.

dominykas:
Oh. Wait. This is a differrent issue - yeah, this is good to go :D

PR #962, opened 2016-01-07
Update dependencies

mroderick:
This PR updates all the dependencies that are trivial updates. I still haven't found the time to figure out why formatio 1.1.2+ makes the tests fail, so we're leaving that one out.

This is ready for review


fatso83:
Seems completely fine.

PR #964, opened 2016-01-15
Fix stub.withArgs unable to recognize sinon matchers

raejin:
I've noticed when converting to commonJS, `sinon.deepEqual` was mistakenly changed to the `deepEqual` function from `util/core/deep-equal.js`. This caused the unit tests of our project failing after upgrading to `v2.0.0-pre`.

I also added a unit test to cover this bug.

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -9,7 +9,6 @@
 "use strict";
 
 var extend = require("./extend");
-var deepEqual = require("./util/core/deep-equal");
 var functionName = require("./util/core/function-name");
 var functionToString = require("./util/core/function-to-string");
 var getPropertyDescriptor = require("./util/core/get-property-descriptor");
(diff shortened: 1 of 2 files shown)


raejin:
The project tests are not fully passed yet, will follow up with a more complete pull request later.

mantoni:
Nice catch! We want to move away from using the `sinon` object as part of the migration towards a more modular Sinon. You should be able to get the new test case passing by injecting the matchers into `deepEqual` [like this](https://github.com/sinonjs/sinon/blob/master/lib/sinon/call.js#L15):

```js
var sinonMatch = require("./match");
var deepEqual = require("./util/core/deep-equal").use(sinonMatch);
```

raejin:
@mantoni I missed that one, commit has updated per suggestion. Thanks!

mantoni:
Thanks! :+1:

PR #967, opened 2016-01-16
Remove the use of `sinon.format` from the codebase

jonnyreeves:
The `test-call` suite now check that the `util/core/format` module was invoked by asserting against the returned string value; this allows us to import `util/core/format` inside both `spy.js` and `call.js` instead of using `sinon.format`


fatso83:
Another one bites the dust :tada:

PR #968, opened 2016-01-16
Re-organise Tests (removes `sinon-test.js`)

jonnyreeves:
Re-organise the contents of the `test` folder so each modules tests are better aligned to their subject's location in `lib/`.

Removes a single test-case for `sinon.log`; I wasn't convinced it was worth keeping :)


fatso83:
Thank you! It's a dirty job, so it's great when someone takes on the task :-)

PR #972, opened 2016-01-24
Remove `sinon.log` and `sinon.logError`

jonnyreeves:
These two functions introduced global, mutable state to the sinon instance; sinon.log has been removed and sinon.logError has been replaced by `sinon.configureLogError` which will return a `logError` function using the supplied options.  Modified FakeXMLHttpRequest and FakeXDR constructors to pass configuration down to the instances they create and fakeServer to allow configuration to be supplied.


lib/sinon/util/fake_server.js
```diff
@@ -80,12 +81,14 @@ var fakeServer = {
 
         return server;
     },
+
     configure: function (config) {
         var whitelist = {
```
fatso83: I was a bit confused by this whitelist, although I know it was there earlier. The values of this object is not used at all, so I assume that the only reason to have this as an object, instead of a list (array) of properties is to make the `hasOwnProperty` symmetrical for both the `whitelist` and the `config` object of purely visual reasons. Am I right in thinking so?

lib/sinon/util/fake_server.js
```diff
@@ -137,13 +142,21 @@ var fakeServer = {
         }
     },
 
+    logger: function () {
+        // no-op; override via configure()
```
fatso83: I do not see how this is overridden by the configure call. `this.logger` has only read-references in this file, and the `this` object is not passed on to the `configure` method of `./core/log_error`. Therefore the check in line 157 makes no sense as it will always be true. What am I missing here?

lib/sinon/util/fake_server.js
jonnyreeves: If I understand the original authors intent correctly The whitelist object is being used as a [Set](https://en.wikipedia.org/wiki/Set_%28abstract_data_type%29) (so the values are un-important).  As well as looking a little cleaner, it also allows O(1) lookup of the values (not that it would make any meaningful difference in this particular situation ;)

lib/sinon/util/fake_server.js
jonnyreeves: The config object is passed to the `fake_server#create` method; this same config object is also passed to the `configureLogError` function in the `fake_server#configure()` method.

If the user passes a `logger` func in their config object it will be set on both the `fake_server` instance (overwriting the default implementation shown here) and also to the `logError` instance created by `fake_server#configure()`.

lib/sinon/util/fake_server.js
fatso83: OK, thanks for asserting my understanding :-)

lib/sinon/util/fake_server.js
fatso83: Ah, now I get it. I was confused by the `Set` construction mentioned further up. I glanced at what values were used there and noticed that it used booleans, and assumed they were not used for that purpose, but now that I know the values are not used things are suddenly looking different :-)

fatso83:
Thanks for your hard work!

PR #973, opened 2016-01-26
Remove `sinon.test()` and `sinon.testCase`.

jonnyreeves:
Functionality has been extracted out into the `sinon-test` module allowing us to remove these two tightly coupled methods from sinon's public API.

Oh, just noticed, this removes the last use of `sinon.config` - we're getting there! :dancers:


jonnyreeves:
Hmm; the Node 4.2 build failed for a seemingly unrelated test; could a mod please trigger a rebuild?  Thanks! :)

mroderick:
Sure

jonnyreeves:
:white_check_mark:

mroderick:
It' didn't work for awhile, restarted several times, asked on IRC ... and suddenly, it works!

mroderick:
lol, now another build has started ... /me gets popcorn out

fatso83:
AFAIK, this could be merged now, as the new `sinon-test` package has been created, but there might be something obvious I am missing?

jonnyreeves:
@fatso83, this should be good to go.  For those importing sinon from master and relying on `sinon.test` or `sinon.testCase`, you will need to start importing the `sinon-test` npm module and the configure these methods with:

```js
var sinon = require('sinon');
var sinonTest = require('sinon-test');

sinon.test = sinonTest.configureTest(sinon  /*, config */ );
sinon.testCase = sinonTest.configureTestCase(sinon  /*, config */ );
```

mroderick:
@jonnyreeves do you think this means that we can also remove the injected config into `sandbox.create()`, which is currently documented in `sandbox.md`?

fatso83:
@jonnyreeves: not sure what you mean by removing the injected config? From where; the API? We cannot remove `sandbox.create(config)`, as this [is being used by `sinon-test` for configuration](https://github.com/sinonjs/sinon-test/blob/master/lib/test.js#L73).

PR #978, opened 2016-02-10
Remove linting errors in switch cases

fatso83:
This pull request removes some linting errors we get in
master when running `npm run lint`.

Refer to the discussion in #898 for background info.


fatso83:
@mroderick This pull request was somewhat of an attempt to revive the discussion on #898 where it halted on missing input from you. This PR is an attempt to just implement what I think you suggested, but it might just serve as input and be closed if we decide for something else. All in all, this is not the most important thing in the world (getting rid of two warnings), which is also why I'd like to close it.

PR #980, opened 2016-02-10
Use PhantomJS 2

fatso83:
PhantomJS was made available on NPM in January, and this makes some hacks in the build script superfluous. This commit removes the hack and adds an explicit dependency on the PhantomJS package that is used in the tests of the packaged browser build.

Also removing older Node versions from Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: node_js
 node_js:
-  - "0.8"
-  - "0.10"
   - "0.12"
-  - "4.2"
+  # LTS release per February 2016
+  - "4.3"
   - "5"
 
 sudo: false
@@ -14,8 +13,6 @@ cache:
 
 before_install:
   - npm config set strict-ssl false
-  # use a more modern npm than ships with 0.8
-  - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.8" ]; then npm install -g npm@1.4; fi'
 
 before_script:
   # we only need to run eslint once per build, so let's conserve a few resources
--- a/build
+++ b/build
@@ -22,27 +22,6 @@ PREAMBLE
   end
 end
 
-# This code is an awful hack that is done here only because
-# PhantomJS 1.x incorrectly handles "use strict" statement in
-# some cases leading to sinon spy executing in a non-strict
-# mode, even though it explicitly says otherwise. Since the
-# people still use PhantomJS 1.x, this has to remain here
-# to add another wrapper with another "use strict" on top
-# of UMD bundle packaged by browserify.
-def add_use_strict(file)
-
-  contents = File.read(file)
-
-  File.open(file, "w") do |f|
-    f.puts("(function () {")
-    f.puts("  'use strict';")
-    f.puts(contents)
-    f.puts("}());")
-  end
-
-  file
-end
-
 Dir.chdir(File.dirname(__FILE__)) do
   version = File.read("package.json").match(/"version":\s+"(.*)"/)[1]
   version_string = ARGV[0] == "plain" ? "" : "-#{version}"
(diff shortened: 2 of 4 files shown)


jonnyreeves:
Nice, but you are no longer running the linting task on any of the builds, not sure if that was intentional?

fatso83:
Wasn't aware of that, but at least this PR isn't touching that in any way.

jonnyreeves:
Ah but I think it does, there was a conditional that resulted in the linting task only being executed under a specific node version which I think you have removed? 

Could be wrong as I am both on mobile and have not had coffee yet 😀

fatso83:
Ah, good catch if so. I am also on mobile. Cannot view the entire file. Thought it was version 0.12 and that is still there.

.travis.yml
```diff
@@ -1,10 +1,8 @@
 language: node_js
 node_js:
-  - "0.8"
-  - "0.10"
   - "0.12"
   - "4.2"
-  - "5"
+  - "5.5"
```
mroderick: By being more specific, we will be locked to that minor version, and we won't be testing in > `5.5`, once those versions become available on the servers. Was that your intention?

fatso83: Not my intention. Changing to major version now. Will do it for version 4 as well.

fatso83:
@jonnyreeves Hmm ... the line that checks for 0.12 is still there, and we are still using 0.12, so if it does not run now it does not normally run either. I am looking into it.

```
  # we only need to run eslint once per build, so let's conserve a few resources
  - 'if [ "x$TRAVIS_NODE_VERSION" = "x0.12" ]; then $(npm bin)/eslint .; fi'
```

fatso83:
@jonnyreeves :  I think you might have been confused by the fact that output from scripts running in `before_script` is suppressed. Thus any linting does not appear to run unless the linting fails. I made [a build with debugging info](https://travis-ci.org/sinonjs/sinon/jobs/108490738) where I send the output to a file and display it. There you can see the output from the linting being run.

.travis.yml
fatso83: Fixed btw.

.travis.yml
```diff
@@ -1,9 +1,7 @@
 language: node_js
 node_js:
-  - "0.8"
-  - "0.10"
   - "0.12"
-  - "4.2"
+  - "4"
```
mroderick: Ehm ... I would very much like to keep `4.2`, as that's the Long Term Support version, so a lot of cautious users will be relying on that.

https://nodejs.org/en/blog/release/v4.2.0/

Perhaps we can have `4` AND `4.2`?

fatso83: ah, crap. sorry about that. was a bit too eager in trying to follow your lead. I think 4.2 is plenty. I don't really see what having more than one run per Node version is giving us. especially now that major versions are coming out as rapidly as they have been.

t1st3: @mroderick you'll probably want 4.3.0 rather than 4.2.0 for now:
https://nodejs.org/en/blog/release/v4.3.0/

fatso83: Good catch, @t1st3 
> Please note that our LTS "Argon" release line has moved from v4.2.x to v4.3.x due to the security fixes enclosed. There will be no further updates to v4.2.x. Users are advised to upgrade to v4.3.0 as soon as possible.

Is there a way one can specify "always use the latest LTS release" in Travis?

t1st3: @fatso83 it seems that when you set `4` in the Travis matrix, Travis will use the latest version of Node.js 4.

Here is an example on `gulp-template`, which has the [following Travis matrix](https://github.com/sindresorhus/gulp-template/blob/master/.travis.yml).
You'll see that [gulp-template latest job on Travis for Node 4](https://travis-ci.org/sindresorhus/gulp-template/jobs/100388923) used the latest version of Node.js 4 (4.2.4) at the time it was run (4.3.0 was not available yet when that job ran). I verified this with other modules, and this behavior is confirmed.

So to *always use the latest LTS release for this major version*, I would just set a major version in Travis matrix, like`4` and/or `5`, but no minor versions like `4.3`

Issue #976, opened 2016-02-02
Fix defake responseType and attribute updates.


randym:
Sinon version: 1.17.2 
Environment OSX Chrome 48.0.2564.97 (64-bit) 
Exmaple URL: http://jsbin.com/soxibo/edit?html,js,console,output

**Bug Description** 
When using sinon.useFakeXMLHttpRequest, it is possible to specify
filters to allow specific requests to be sent via a working XHR.

Unfortunately, the implementation ignores the configured responseType
and the mapping of attribute updates from the working XHR to the
exposed fake XHR on state change are not inline with the specification.

I need this working for a client now, and have forked/fixed this issue here:
https://github.com/randym/sinon/commit/7cca0fbb2a05e3e34d425addd33767d3d4560ed5

Please advise if you would like a PR.

fatso83:
Cannot see why we wouldn't want a working fix for an existing problem. Seems like a great candidate for a pull request

randym:
Is there a particular branch/commit would you like to see a PR against?

Assuming 2.0.0 will come with some non-backwards compatible changes, I would love to see a 1.17.4 release with this fixed. I am also intimately familiar with the work/effort that comes with every release of an open source project and am happy to follow your lead.

fatso83:
Sorry for keeping you waiting...  Preferably two pull requests :
- one against the master branch
- one against the 1.17 branch (backport) 

Squash smaller commits and try to keep commits in as few logical units as makes sense.

randym:
No worries. 

I'll get a PR in for master and do the backport when I have a bit of free time.

PR #982, opened 2016-02-11
Fixed defake responseType and attribute updates.

randym:
https://github.com/sinonjs/sinon/issues/976

When using sinon.useFakeXMLHttpRequests, it is possible to specify
filters to allow specific requests to pass through to a working XHR.

Unfortunately, the implementation did not apply the responseType
configured on the fakeXhr. Thus any filtered request for an
arraybufffer, blobs or json responseType returns the wrong data type.

In addition to applying the responseType from the fakeXhr to the working
XHR, the mapping of attribute updates from the working XHR to the
fakeXhr on state change needed to be brought inline with the
specification.

Implementation details:

1. responseType is applied to the working XHR object when the fakeXhr
recieves 'send'.

This allows consumers to specify responseType any time prior to sending
the request.

Concretely - both of these work for unfaked requests:

var xhr = new XMLHttpRequest();
xhr.open('GET', url, true);
xhr.responseType = 'arraybuffer';

var xhr = new XMLHttpRequest();
xhr.responseType = 'arraybuffer';
xhr.open('GET', url, true);

2. All valid attributes are mapped from the working XHR to the fakeXhr.

Depending on the responseType specified, requesting responseText or
responseXML can result in an InvalidState error.

Concretely:

- responseText throws an InvalidState error if the responseType is not and
empty string or "text"
- responseXML throws an InvalidState error if the responseType is not an
  empty string or "document"

https://www.w3.org/TR/XMLHttpRequest/#the-responsetext-attribute
https://www.w3.org/TR/XMLHttpRequest/#the-responsexml-attribute

Instead of querying readyState and making partial, potentially fatal
updates, I've opted to ensure that all valid attributes of the working
XHR are applied to the fakeXhr on all state changes.

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -213,12 +213,42 @@ FakeXMLHttpRequest.defake = function defake(fakeXhr, xhrArgs) {
         "removeEventListener"
     ], function (method) {
         fakeXhr[method] = function () {
+
+            // responseType can be configures any time
+            // prior to send. This ensure we capture it
+            // at the last possible moment.
+            if (method === "send") {
+                xhr.responseType = fakeXhr.responseType;
+            }
+
             return apply(xhr, method, arguments);
         };
     });
 
-    var copyAttrs = function (args) {
-        each(args, function (attr) {
+    // Map attributes from the workingXHR back to our fakeXhr
+    //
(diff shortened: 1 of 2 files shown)


randym:
Found a flickering spec failure locally when altering the order of specs due to reuse of this.fakeXHR in the defakedXHR test block and will resubmit this PR with less brittle spec coverage.

Issue #947, opened 2015-12-11
Sandbox.reset doesn't work


madSimonJ:
I've found that using the Sandbox the callcounts never reset between tests if I call sandbox.reset().  I have to call the individual resets for each stub I'm using.

fatso83:
Seems important, but I am afraid this won't get much priority until more info is supplied.

[Try this](https://github.com/sinonjs/sinon/blob/master/CONTRIBUTING.md#reporting-an-issue)

fatso83:
@madSimonJ We never got any test data from you, but we have tried looking into this and as far as we can tell, this is not reproducible. We have made many tests in #985, and we'll probably close this unless you can supply some code showing that the problem actually exists in some cases.

madSimonJ:
Hi,
      For sure the problem still exists.  Apologies for the lack of test data, I've not long had a daughter and I'm struggling to find any time for non-day-job IT stuff atm.  I'll try and get something for you asap.
Thanks,
Simon
 49 66 20 59 6F 75 20 43 61 6E 20 52 65 61 64 20 54 68 69 73 2C 20 59 6F 75 20 4E 65 65 64 20 41 20 4C 69 66 65 21 

    On Friday, February 12, 2016 2:54 PM, Carl-Erik Kopseng <notifications@github.com> wrote:
 

 @madSimonJ We never got any test data from you, but we have tried looking into this and as far as we can tell, this is not reproducible. We have made many tests in #985, and we'll probably close this unless you can supply some code showing that the problem actually exists in some cases. —

fatso83:
@madSimonJ Thank you for getting back to us! When you have time try and add  code in this issue that triggers the problem. You can use the tests supplied in #985 as a template/inspiration.

I am closing this _for now_ until we see actual code, but feel free to reopen it as soon as you have some time to spare for this issue :-)

fatso83:
P.S. #965 might be related, and it was reported that it worked fine in Sinon 1.16.1, so you might try out that version to see if we are facing a regression. It might as well not be related, but it could help in pinpointing when this happened.

PR #985, opened 2016-02-11
Add tests for #947

mroderick:
This is a PR for **v1.17** branch. Once merged, I'll cherry pick the changes into a PR for **master**.

Having had the same experience of having to manually reset stubs, I thought I could try to show the issue with some tests. However, to the best of my understanding these tests show that the issue reported as #947 is resolved.

If someone can contribute tests that shows that `sandbox.reset()` **doesn't** reset the `callCount`, that would be great. Until that time, I think we can consider #947 fixed.

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -125,6 +125,102 @@
                     sinon.createStubInstance(A);
                 });
             }
+        },
+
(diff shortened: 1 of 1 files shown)


mantoni:
LGTM. Loving the test data :postal_horn:

PR #987, opened 2016-02-12
Improve Blob support detection logics

nelsonsilva:
Current Blob support detection returns false in Safari 9 for example where `typeof Blob` is `object`.
This change is based on detection logics found in Modernizr.


fatso83:
Seems legit to me.

kriuz:
This also fixes tests run on phantom.

pcardune:
@fatso83 Any chance this change could go into a 1.17.4 release?

fatso83:
@pcardune: sure. Just supply a PR and the likelihood is quite high :-)

pcardune:
@fatso83 I can do that. What steps do you follow for cutting new releases? Seeing as master is well on it's way to a 2.0 version, should I just submit a pull request against the v1.17 branch with just this change? I assume after merging, you'll go ahead and bump the version number in `package.json` and publish a new build to npm?

fatso83:
@pcardune Your assumptions are correct. I'll go ahead and merge it now.

PR #997, opened 2016-03-02
Fix non enumerable methods stub restore

STuFF:
hello,

I had a case where I've stubbed an object with non enumerable methods. It works fine, however, those methods are **not** restored because a simple for-in loop is used in the restore process (ignoring non enumerable methods...) but sinon.walk method is used to actually stub all the methods, including non enumerable ones....

I fixed that and added tests.


fatso83:
Ooh, nice! Have to look at this in the morning, but the first impression is good. thanks!

PR #1010, opened 2016-03-16
Actually use useFakeXMLHTTPRequest for sandbox

bouk:
useFakeServer returns something very different from useFakeXMLHttpRequest so we can't use sandbox as a drop-in replacement for useFakeXMLHttpRequest right now.

This just uses sinon.useFakeXMLHttpRequest and adds it to the sandbox

### DIFF
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -113,6 +113,13 @@ describe("sinon.sandbox", function () {
                     assert(sinon.useFakeXMLHttpRequest.called);
                 });
 
+                it("doesn't secretly use useFakeServer", function () {
+                    this.sandbox.stub(sinon.fakeServer, "create").returns({ restore: function () {} });
+                    this.sandbox.useFakeXMLHttpRequest();
+
+                    assert(sinon.fakeServer.create.notCalled);
+                });
+
                 it("adds fake xhr to fake collection", function () {
                     this.sandbox.useFakeXMLHttpRequest();
                     this.sandbox.restore();
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -75,6 +75,11 @@ extend(sinonSandbox, {
         return this.add(this.server);
     },
 
+    useFakeXMLHttpRequest: function useFakeXMLHttpRequest() {
+        var xhr = sinon.useFakeXMLHttpRequest();
+        return this.add(xhr);
+    },
+
     inject: function (obj) {
         sinonCollection.inject.call(this, obj);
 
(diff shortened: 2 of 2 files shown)


mantoni:
Thank you! I think this should into the upcoming Sinon 2.0 release.
Can I ask you to also add a test for this?

fatso83:
@mantoni The error on the failing Node 5 build was:
"You used username 'sinonjs' and access key 'None' to authenticate, which are not valid Sauce Labs credentials." 

Do you know if/understand why this build differs (i.e. fails) from the other builds?

mantoni:
@fatso83 No. The current master build succeeded. Maybe we should try to run the SauceLabs builds on the "Stable" Node 4 builds instead of "5".

fatso83:
Agreed, we are doing checks on if it works in Node in other places. Use LTS for SauceLabs

bouk:
I've added a test that basically just makes sure we're not called fakeServer.create for useFakeXhr. There was already a test making sure the Xhr mock is set up.

test/sandbox-test.js
```diff
@@ -113,6 +113,13 @@ describe("sinon.sandbox", function () {
                     assert(sinon.useFakeXMLHttpRequest.called);
                 });
 
+                it("doesn't secretly use useFakeServer", function () {
+                    this.sandbox.stub(sinon.fakeServer, "create").returns({ restore: function () {} });
```
fearphage: This seems like a bad test. Did this test fail prior to this change? By default, sinon uses `sinon.fakeServerWithClock` to construct fake servers.

fatso83: Going through old PRs that have stalled to try and get them rolling again ... 

@fearphage : other than simply exchanging `sinon.fakeServer` with `sinon.fakeServerWithClock` to (hopefully) get a test that failed prior to the change, do you have a better way of asserting we are doing the right thing?

@bouk: anything preventing you from changing the test to actually testing the change? The test should fail before your change. I am assuming this is not the case (? - feel free to prove otherwise).

mroderick:
I am going through stalled PRs to get them going again.

@bouk any chance of getting the last few things sorted out on this PR so we can merge it?

fatso83:
I can't just merge this as #1088 removed the block of tests surrounding it, with reference to `sandbox.useFakeXMLHttpRequest()` being deprecated(?). I cannot explicitly remember this deprecation thing, but I remember talk about moving all XHR stubbing out of Sinon and into its own project/repo, which in turn would probably mean removing it from the sandbox as well (injecting it into the sandbox could be one idea).

In any case, ignoring future changes, I think it should be ok to add this test under the old (removed) `describe` block. I can do the rebase and the required change and resubmit this. Thoughts, @mroderick and/or @jonnyreeves (as the last one to touch the code)?

cjohansen:
Manually applied, Github is not coping.

PR #1012, opened 2016-03-19
Pull request and issue templates

fatso83:
This is a first draft. Just having something is better than nothing, I thought. Thoughts, @mantoni or @mroderick?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,17 @@
+> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.
+
+* Sinon version :
+* Environment   :
+* Example URL   :
+* Other libraries you are using:
+
+**What did you expect to happen?**
+
+**What actually happens**
+
+**How to reproduce**
+> Describe *with code* how to reproduce the faulty behaviour,
+> or link to code on JSBin or similar
+
+[See this for an example](https://github.com/sinonjs/sinon/blob/master/CONTRIBUTING.md#bug-report-template)
+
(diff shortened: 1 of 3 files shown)


mantoni:
I like it. Maybe the issue template could be improved by removing the first and last line so that the template only contains things that actually are the template? So that nobody is forced to remove stuff from the template.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+
+**What did you expect to happen?**
+
+**What actually happens**
+
+**How to reproduce**
+> Describe *with code* how to reproduce the faulty behaviour,
+> or link to code on JSBin or similar
+
+[See this for an example](https://github.com/sinonjs/sinon/blob/master/CONTRIBUTING.md#bug-report-template)
```
mroderick: I think we should link to this one from `CONTRIBUTING.md`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,6 @@
+*Making a contribution to the code base?* Please read [this](https://github.com/sinonjs/sinon/blob/master/CONTRIBUTING.md#contributing-to-the-code-base).
```
mroderick: This file repeats a lot of what is said in `CONTRIBUTING.md`. I think it would be more valuable, if instead we could make it easy to copy-paste into a PR and fill in the sections

```markdown
> give a concise (one or two short sentences) description of what what problem is being solved by this PR
>
> Example: Fix issue #123456 by re-structuring the colour selection conditional in method `paintBlue`

> When relevant, give a more thorough description of what the problem the PR is trying to solve. Examples of good topics for this section are:
> * Link to an existing GitHub issue describing the problem
> * Describing the problem in greater detail than the TL;DR section above
> * How you discovered the issue, if it's not already described as an issue on GitHub
> * Discussion of different approaches to solving this problem and why you chose your proposed solution

#### Solution  - optional
> When contributing code (and not just fixing typos, documentation and configuration), please describe why/how your solution works. This helps reviewers spot any mistakes in the implementation.
>
> Example: 
> "This solution works by adding a `paintBlue()` method"
> Then your reviewer might spot a mistake in the implementation, if `paintBlue()` uses the colour red.

2. `npm install`
3. <your-steps-here>
```

PR #1014, opened 2016-03-28
Feature/es6 symbols support

### LINKED ISSUE #1002: spy.calledWithMatch fails for objects containing a Symbol as value.
Sinon version : 1.17.3 Environment : OSX Chrome  48.0.2564.116 Example URL : http://jsbin.com/qeloruvupi/edit?js,console

#### Bug Description ####

Calling a spy with an object containing a `Symbol` as value and then calling `calledWithMatch` on the spy with a similar object makes `Object.sinon.match` throw a `TypeError`.

```
TypeError: Cannot convert a Symbol value to a string
```

havardh:
Adds explicit cast from value to String, when the value is used in error messages. This is to support the es6 Symbol type in these message. The Symbol type cannot be coerced to string. This causes the existing code throw a TypeError when concatenating the error messages containing Symbols. The explicit casts introduced in this PR solves this issue. 

This PR aims to resolve #1002, and goes a little further in resolving similar issues throughout the codebase. 
In addition it adds a matcher for Symbol located at `sinon.match.symbol`.

Note: There are a few more issues like this in `wrapMethod`. These issues arise when a symbol is used as a property key. As far as I know you need to use the `var obj = { [Symbol()]: function() {} };` syntax to write a test to generate this case. Should I include these cases as well?

### DIFF
--- a/docs/current/matchers.md
+++ b/docs/current/matchers.md
@@ -118,6 +118,11 @@ Requires the value to be a regular expression.
 Requires the value to be a `Date object.
 
 
+#### `sinon.match.symbol"
+
+Requires the value to be a `Symbol.
+
+
 #### `sinon.match.same(ref)"
 
 Requires the value to strictly equal `ref`.
@@ -134,8 +139,9 @@ Requires the value to be of the given type, where `type` can be one of
     `"object"`,
     `"function"`,
     `"array"`,
-    `"regexp"` or
-    `"date"`.
+    `"regexp"`,
+    `"date"` or
+    `"symbol"`.
 
 
 #### `sinon.match.instanceOf(type)"
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -83,7 +83,7 @@ function getCallbackError(behavior, func, args) {
 
         if (behavior.callArgProp) {
             msg = functionName(behavior.stub) +
-                " expected to yield to '" + behavior.callArgProp +
+                " expected to yield to '" + String(behavior.callArgProp) +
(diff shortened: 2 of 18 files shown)


mantoni:
Nice work, thanks. To get the build passing in all environments, you will have to add feature detections around the symbol tests.

havardh:
@mantoni: Ah. good idea, I'll do that. Do you have a standarized way of abstracting feature detections or should I just wrap the test implementations in `if (typeof Symbol === "function")`?

In `node 0.12` there seems to be a bug [1] that I would like to work around. It causes `String(Symbol())` to throw a `TypeError`. The workaround is to use `toString` in place of `String`. 

[1] https://github.com/paulmillr/es6-shim/issues/361#issuecomment-142927487

mantoni:
Using simple `if` statements is fine.

havardh:
Looks like my note is incorrect and that you can use symbols as property keys with `var obj = {}; obj[Symbol()] = function () {};`

I see

mantoni:
Thank you!

blakehaswell:
When will this change be released be published to NPM?

We're currently depending on a commit SHA, because there hasn't been a release with this code in it, and we're running into peer dependency difficulties. If we could get a rough idea when this will be released it will help us decide how we'll deal with that issue.

Thanks.

fatso83:
Tonight, sorry for the wait.

blakehaswell:
Thanks – I appreciate the update!

fatso83:
```
npm show sinon dist-tags.next
2.0.0-pre.2
```

PR #1016, opened 2016-03-28
Caught exceptions from async tests are thrown by sinon.test.

### LINKED ISSUE #1015: sinon.test eats errors thrown in the first tick of an async test
Throwing an error in an async function does not get re thrown by sinon.test.

For example, the following mocha test times out instead of failing with an error.
`it('should do something', sinon.test((done) => throw new Error('should have blown up'))`

I'm using
- Node v5.5.0
- Sinon v1.17.3
- Mocha

RobertWHurst:
Fixes #1015

### DIFF
--- a/lib/sinon/test.js
+++ b/lib/sinon/test.js
@@ -48,13 +48,11 @@
                     exception = e;
                 }
 
-                if (typeof oldDone !== "function") {
-                    if (typeof exception !== "undefined") {
-                        sandbox.restore();
-                        throw exception;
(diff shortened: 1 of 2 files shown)


fatso83:
This is great. Thank you. Would you mind pushing the same fix to the sinon-test project? Otherwise it will get lost once people start using sinon 2 with the external sinon-test.

RobertWHurst:
Sure thing :)

PR #1017, opened 2016-03-28
Re-did #861 for v1.17 branch

fearphage:
The diff should look very familiar. This is #861 back for the second time. This is against the v1.17 branch this time.


fatso83:
Great work as always. Thanks, @fearphage!

PR #1023, opened 2016-04-07
Fix capturing of stack traces in Phantom.js.

papandreou:
Apparently Phantom.JS requires an `Error` instance must be thrown before it has its `stack` serialized.

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -207,7 +207,16 @@ var spyApi = {
 
         push.call(this.exceptions, exception);
         push.call(this.returnValues, returnValue);
-        push.call(this.stacks, new Error().stack);
+        var err = new Error();
+        var stack = err.stack;
+        if (!stack) {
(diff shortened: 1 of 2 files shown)


mantoni:
This is interesting. We’re running the Sinon tests in PhantomJS, would you be able to add a test case to verify the issue?

mroderick:
Great contribution!
I too would like to see a unit test for it

papandreou:
Of course -- added a test that fails with `npm run test-headless` on master.

mroderick:
Great work! Would you mind squashing them before we merge it?

papandreou:
Sure! Done.

fatso83:
After merging this the clock tests for [IE 10 and 11 started failing](https://travis-ci.org/sinonjs/sinon/jobs/121782850). Anyone that understands why?

papandreou:
Sounds unrelated, except that this branch could have caused a slight performance degradation, which could cause that timeout. Try to double the timeout.

PR #1029, opened 2016-04-18
Add missing mocaccino and phantomic to package.json

mroderick:
#### Purpose
This is a PR for **master**, it adds two missing development dependencies

#### Background
When running `npm test` from `master`, the tests for web workers do not run.

#### How to verify
2. `npm install`
3. `npm test`
4. Verify that the command exits with zero: `echo $?`

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,11 @@
     "eslint": "^1.7.1",
     "eslint-config-defaults": "^2.1.0",
     "eslint-plugin-mocha": "^1.1.0",
+    "mocaccino": "^1.8.2",
     "mocha": "^2.4.5",
     "mochify": "^2.17.0",
     "mochify-istanbul": "^2.4.1",
+    "phantomic": "^1.4.0",
     "pre-commit": "^1.1.2",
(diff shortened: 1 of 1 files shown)


mantoni:
Aaah, see, that's the reason for [this comment in the travis config](https://github.com/sinonjs/sinon/blob/master/.travis.yml#L25). Those tests implicitly depend on the npm 3 flat dependency tree. I guess you're right and we should better add the dependencies here.

fatso83:
Something seems wrong. We seems to have some persistent `clock` related errors for the IE versions. I have rerun the build 3 times. Fails at the same point each time.

mroderick:
> Something seems wrong. We seems to have some persistent clock related errors for the IE versions. I have rerun the build 3 times. Fails at the same point each time.

Yeah, those have been there for awhile now. I first noticed them ... last Friday? ... wasn't sure if they arrived after I merged a PR, of if they were already there.

I guess it's time to add an IE box to my setup, so I can debug this locally... unless someone else feels like working on this...

PR #1031, opened 2016-04-18
Use event loaded instead of error event for code like 403, 500, etc.

vietnogi:
Codes like 403 and 500 are loaded events not error events.

Axios.js would not reject a promise for 403 error codes.

#### Solution  - optional
Change to use loaded event instead of error event.

2. `npm install`
3. tbd

### DIFF
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -1106,14 +1106,6 @@ if (typeof window !== "undefined") {
             assertEventOrdering("abort", 0, function (xhr) {
                 xhr.abort();
             });
-
-            assertEventOrdering("error", 0, function (xhr) {
-                xhr.respond(500, {}, "");
-            });
-
-            assertEventOrdering("load", 100, function (xhr) {
-                xhr.respond(200, {}, "");
-            });
         });
 
         describe(".response", function () {
@@ -1696,6 +1688,20 @@ if (typeof window !== "undefined") {
                 this.xhr.respond(200, {}, "");
             });
 
+            it("triggers 'load' event on for non-200 events", function (done) {
+                var xhr = this.xhr;
+
+                this.xhr.addEventListener("load", function () {
+                    assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);
+                    assert.equals(xhr.status, 500);
+
+                    done();
+                });
+
+                this.xhr.send();
+                this.xhr.respond(500, {}, "");
+            });
+
             it("triggers 'load' with event target set to the XHR object", function (done) {
                 var xhr = this.xhr;
 
@@ -1723,54 +1729,12 @@ if (typeof window !== "undefined") {
                 this.xhr.respond(200, {}, "");
             });
 
-            it("calls #onload on success for non-200 requests", function (done) {
+            it("calls #onload for non-200 events", function (done) {
                 var xhr = this.xhr;
 
-                this.xhr.addEventListener("error", function () {
+                this.xhr.onload = function () {
                     assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);
-                    refute.equals(xhr.status, 0);
-
-                    done();
-                });
-
-                this.xhr.send();
-                this.xhr.respond(500, {}, "");
-            });
-
-            it("triggers 'error' event on failure", function (done) {
-                var xhr = this.xhr;
-
-                this.xhr.addEventListener("error", function () {
-                    assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);
-                    refute.equals(xhr.status, 0);
-
-                    done();
-                });
-
-                this.xhr.send();
-                this.xhr.respond(500, {}, "");
-            });
-
-            it("triggers 'error' with event target set to the XHR object", function (done) {
-                var xhr = this.xhr;
-
-                this.xhr.addEventListener("error", function (event) {
-                    assert.same(xhr, event.target);
-
-                    done();
-                });
-
-                this.xhr.send();
-                this.xhr.respond(500, {}, "");
-            });
-
-
-            it("calls #onerror on failure", function (done) {
-                var xhr = this.xhr;
-
-                this.xhr.onerror = function () {
-                    assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);
-                    refute.equals(xhr.status, 0);
+                    assert.equals(xhr.status, 500);
 
                     done();
                 };
@@ -1844,7 +1808,7 @@ if (typeof window !== "undefined") {
 
             it("triggers 'loadend' event at the end", function (done) {
                 this.xhr.addEventListener("loadend", function (e) {
-                    assertProgressEvent(e, 0);
+                    assertProgressEvent(e, 100);
                     assert(true);
 
                     done();
@@ -1870,7 +1834,7 @@ if (typeof window !== "undefined") {
 
             it("calls #onloadend at the end", function (done) {
                 this.xhr.onloadend = function (e) {
-                    assertProgressEvent(e, 0);
+                    assertProgressEvent(e, 100);
                     assert(true);
 
                     done();
@@ -1894,6 +1858,10 @@ if (typeof window !== "undefined") {
                     });
                 });
             }
+
+            assertEventOrdering("load", 100, function (xhr) {
+                xhr.respond(200, {}, "");
+            });
         });
 
         describe("xhr.upload", function () {
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -1750,7 +1750,6 @@ if (typeof window !== "undefined") {
                 this.xhr.respond(500, {}, "");
             });
 
-
             it("calls #onerror on failure", function (done) {
                 var xhr = this.xhr;
 
@@ -1763,6 +1762,21 @@ if (typeof window !== "undefined") {
 
                 this.xhr.send();
                 this.xhr.respond(500, {}, "");
+            });
+
+            it("calls #onerror on failure", function (done) {
+                var xhr = this.xhr;
+
+                this.xhr.onerror = function () {
+                    assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);
+                    refute.equals(xhr.status, 0);
+
+                    done();
+                };
+
+                this.xhr.send();
+
+                this.xhr.respond(403, {}, "");
             });
 
             it("does not trigger 'load' event on abort", function (done) {
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -1723,6 +1723,20 @@ if (typeof window !== "undefined") {
                 this.xhr.respond(200, {}, "");
             });
 
+            it("calls #onload on success for non-200 requests", function (done) {
+                var xhr = this.xhr;
+
+                this.xhr.addEventListener("error", function () {
+                    assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);
+                    refute.equals(xhr.status, 0);
+
+                    done();
+                });
+
+                this.xhr.send();
+                this.xhr.respond(500, {}, "");
+            });
+
             it("triggers 'error' event on failure", function (done) {
                 var xhr = this.xhr;
 
@@ -1750,6 +1764,7 @@ if (typeof window !== "undefined") {
                 this.xhr.respond(500, {}, "");
             });
 
+
             it("calls #onerror on failure", function (done) {
                 var xhr = this.xhr;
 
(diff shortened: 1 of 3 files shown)


mantoni:
It's interesting that this didn't break any tests. Apparently this was uncovered.

Could you add a test to make sure this does not break in the future?

fearphage:
> Codes like 403 and 500 are loaded events not error events.

Where'd you get this information from? Can you point to it in the specification? I set up the code to mimic what I read in the spec.

Standard8:
I was just testing this in the context of issue #1040. The code in 1.17.4 is broken in itself for that PR, because `error` isn't an event added as a listener in the constructor of `FakeXMLHttpRequest`.

However, even with that fixed, the behaviour of 1.17.4 is different to that if you debug an XHR in Firefox & Google Chrome. Doing:

`this.server.requests[0].respond(401, { 'Content-Type' : 'application/json' }, JSON.stringify({}));`

with sinon-1.17-4.js, calls the `onerror` handler for the XHR. Doing an XHR which returns a similar response in Firefox & Chrome calls the `onload` handler with `request.status` set appropriately.

> Where'd you get this information from? Can you point to it in the specification? I set up the code to mimic what I read in the spec.
> 
> https://xhr.spec.whatwg.org/#events
> 
>     error - The fetch failed.
>     load - The fetch succeeded.

From the way I read it, `load` and `error` relate to the fetch sequence - obtaining data from the server. The status returned by the server isn't part of the XHR.

Hence, I think this PR is correct & matching sinon to the browser's interpretation.

fearphage:
> with sinon-1.17-4.js, calls the onerror handler for the XHR. Doing an XHR which returns a similar response in Firefox & Chrome calls the onload handler with request.status set appropriately.

Tested and verified. You are correct. This needs test to ensure this behavior remains in tact going forward.

fearphage:
The current fix is too small actually. Some of the code around checking status codes should be removed.

vietnogi:
hey guys, i added a test, first time making changes to sinon so not sure if I added the test in the correct place and if the test itself is sufficient

test/util/fake-xml-http-request-test.js
```diff
@@ -1710,6 +1710,20 @@ if (typeof window !== "undefined") {
                 this.xhr.respond(200, {}, "");
             });
 
+            it("calls #onload on success for 403", function (done) {
```
fearphage: Make the message more generic. This is not about 403 specifically. It's about the non-200 requests.

fearphage:
How would one identify and trigger an `error` at all? It will be impossible after this change.

vietnogi:
good point @fearphage,  perhaps we could add an `error` method, i.e `this.xhr.error();`

Standard8:
> How would one identify and trigger an error at all? It will be impossible after this change.

I believe it was impossible before - I think I remember looking at it a while ago, and then unfortunately forgetting to file an issue for the `error` method.Should we split that out to a different PR/Issue?

fearphage:
> I believe it was impossible before - I think I remember looking at it a while ago, and then unfortunately forgetting to file an issue for the error method.

No, it was not. It triggered the error event for all non-200 response codes. That's the bug this ticket is fixing.

Standard8:
> No, it was not. It triggered the error event for all non-200 response codes. That's the bug this ticket is fixing.

I just tried it in sinon 1.17.3, and responding with the 400s as per my earlier comment would always call the `onload` handler. I could never find a way for the `onerror` to be called - how were you getting `onerror` to be called?

fearphage:
You are correct. Error events are triggering, but `xhr.onerror` is not being called. I misread your message.

```js 
it("should follow request onerror steps", function (done) {                                             
      var expectedOrder = [                                                                             
          "upload:progress",                                                                            
          "upload:error",                                                                               
          "upload:loadend",                                                                             
          "xhr:progress",                                                                               
          "xhr:onerror",                                                                                
          "xhr:error"                                                                                   
      ];                                                                                                
      var eventOrder = [];                                                                              
                                                                                                        
      function observe(name) {                                                                          
          return function (e) {                                                                         
              assertProgressEvent(e, 0);                                                                
              eventOrder.push(name);                                                                    
          };                                                                                            
      }                                                                                                 
                                                                                                        
      this.xhr.open("GET", "/");                                                                        
      this.xhr.send();                                                                                  
                                                                                                        
      this.xhr.upload.addEventListener("progress", observe("upload:progress"));                         
      this.xhr.upload.addEventListener("error", observe("upload:error"));                               
      this.xhr.upload.addEventListener("loadend", observe("upload:loadend"));                           
      this.xhr.addEventListener("progress", observe("xhr:progress"));                                   
      this.xhr.addEventListener("error", observe("xhr:error"));                                         
      this.xhr.onerror = observe("xhr:onerror");                                                        
      this.xhr.addEventListener("loadend", function (e) {                                               
          assertProgressEvent(e, 0);                                                                    
          assert.equals(eventOrder, expectedOrder);                                                     
                                                                                                        
          done();                                                                                       
      });                                                                                               
                                                                                                        
      this.xhr.respond(500, {}, "");                                                                    
  });                                                                                                   
});                             
```

fails:

> AssertionError: [assert.equals] upload:progress,upload:error,upload:loadend,xhr:progress,xhr:error expected to be equal to upload:progress,upload:error,upload:loadend,xhr:progress,xhr:onerror,xhr:error

For comparison:

```
upload:progress,upload:error,upload:loadend,xhr:progress,xhr:error // actual upload:progress,upload:error,upload:loadend,xhr:progress,xhr:onerror,xhr:error // expected
```

`xhr:onerror` is the only thing missing.

Standard8:
I have a possible fix for allowing `onerror` to be called here:

https://github.com/vietnogi/sinon/compare/loadedFix...Standard8:xhr-onerror

basically it allows `response(0)` to simulate the error. The other option would be to create a function `respondWithNetworkError` although tbh, the structure of the code would be the same.

Once this PR gets landed, I'm happy to create a PR for that if it is wanted.

vietnogi:
I also think its a good idea to have a separate PR for simulating errors.  Anything left to do to get this PR in?

fearphage:
> it allows response(0) to simulate the error.

That seems surprising and magical. I vote for something more explicit -- `.error()` or something similar.

Both your and the original PR, need to remove the check for the status codes since they no longer matter. That shouldn't be a part of the forking logic any longer.

> I also think its a good idea to have a separate PR for simulating network errors.

I agree that the fix should be in another PR, but once your PR is merged it becomes impossible to trigger an `error` with fake xhr. That seems like a nonstarter to me.

Standard8:
> I agree that the fix should be in another PR, but once your PR is merged it becomes impossible to trigger an error with fake xhr. That seems like a nonstarter to me.

Can you clarify why this would be a non-starter? We're basically reverting to the pre-1.17.4 state aiui (where `error` couldn't be called in any case, and is now being called wrongly for a lot of cases). The next PR can follow quite quickly as we already have the rough design for it.

Alternately, @vietnogi could do the simplification changes for dropping the status number checking, then I could update my changeset which vietnogi can pull onto the branch for this PR.

fatso83:
I am not totally in with the specific details in this, but if this patch makes `sinon` behave as in `1.17.3` and there is an upcoming patch coming along, then I think this should be fine to merge. The additional functionality to explicitly trigger errors could hold for a new PR, as this is basically just fixing a regression.

fearphage:
> Can you clarify why this would be a non-starter?

I am not using the latest or whatever is broken. You cannot trigger `error`s at all in 1.17.3 (`.respond(500, {}, "")`? Is this correct? Or `onerror` is not triggered?

fearphage:
I'm seeing different behavior and it might be due to my environment. That is probably the disconnect. Sorry for confusing matters.

fearphage:
I think I'm caught up to where everyone has been for a while now. My tests are failing in a somewhat sensible way. There are two distinct issues at play here.

1. This ticket wants to change non-200 requests to trigger `load` events instead of `error` events. Currently `fakeXhr.respond(500, {}, '')` triggers an `error` event. It should trigger `load` based on what browsers do.
2. When an xhr error occurs, `onerror` is not triggered (v1.17 or master). Error in this case means a non-200 response (current behavior which will go away after this ticket). To restate, `xhr.addEventListener('error', ...)` will trigger, but `xhr.onerror` will not. I have a PR out to fix master.

### Missing test cases

```js
"triggers 'error' event on failure": function (done) {                                                  
    var xhr = this.xhr;                                                                                 
                                                                                                        
    this.xhr.addEventListener("error", function () {                                                    
        assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);                                   
        refute.equals(xhr.status, 0);                                                                   
                                                                                                        
        done();                                                                                         
    });                                                                                                 
                                                                                                        
    this.xhr.send();                                                                                    
    this.xhr.respond(500, {}, "");                                                                      
},                                                                                                      
                                                                                                        
"triggers 'error' with event target set to the XHR object": function (done) {                           
    var xhr = this.xhr;                                                                                 
                                                                                                        
    this.xhr.addEventListener("error", function (event) {                                               
        assert.same(xhr, event.target);                                                                 
                                                                                                        
        done();                                                                                         
    });                                                                                                 
                                                                                                        
    this.xhr.send();                                                                                    
    this.xhr.respond(500, {}, "");                                                                      
},                                                                                                      
                                                                                                       
"calls #onerror on failure": function (done) {                                                          
    var xhr = this.xhr;                                                                                 
                                                                                                        
    this.xhr.onerror = function () {                                                                    
        assert.equals(xhr.readyState, sinon.FakeXMLHttpRequest.DONE);                                   
        refute.equals(xhr.status, 0);                                                                   
                                                                                                        
        done();                                                                                         
    };                                                                                                  
                                                                                                        
    this.xhr.send();                                                                                    
    this.xhr.respond(500, {}, "");                                                                      
},
```

* v1.17.3 - all test cases fail
* v1.17.4 - only the last test case fails
* master - only the last test case (in the correct test format) fails here as well

### Important

Once this ticket is merged, it will be _impossible_ to trigger the `error` event at all in any way (even incorrectly). This would make it just as broken as v1.17.3, but it would be a regression (in some ways) going from v1.17.4 to this change. 

> if this patch makes sinon behave as in 1.17.3

It does do that, but this patch is *not* fixing a regression. It is making a big change to what sinon considers to be an error in xhr. The change does align with how browsers treat status codes, but we're going to lose a big piece of functionality in the process.

Did I miss anything?

vietnogi:
That seems about right.  To provide the missing functionality, we would need to expand the fakeServer API

fearphage:
That's a big change though imo. removing the ability to remove trigger a type of event seems like a minor tick as far as semver goes.

Standard8:
> It does do that, but this patch is not fixing a regression. It is making a big change to what sinon considers to be an error in xhr. The change does align with how browsers treat status codes, but we're going to lose a big piece of functionality in the process.
> 
> Did I miss anything?

I disagree with this last statement. Going back in time even further to 1.17.0, I just re-verified: Doing `fakeXhr.respond(500, {}, '')` would call `onload`. In 1.17.4 its calling `onerror`. That's the regression I see and what #1040 is about.

This will mean that onerror won't be able to be called without further fixes, but that's something that was only unofficially available prior to #861 anyway.


@vietnogi Another patch has landed on master, so could you update your one here? As @fearphage was saying earlier, I think the if statement can be simplified to just `if (this.aborted) ...` and drop the error code checking.

vietnogi:
rebased, but no longer passed due to new specs

fearphage:
@vietnogi You should fix the tests to pass in your branch based on your change.

Standard8:
@vietnogi I've just updated my branch: https://github.com/vietnogi/sinon/compare/loadedFix...Standard8:xhr-onerror

There's two commits, one that fixes the tests for this PR, plus I've also added a commit for the `.error()` functionality.

Would you be happy to pull https://github.com/standard8/sinon/commit/0ebaac8ab38dcb90ad759466fed34a9915cf2b2b onto your branch for this PR or do you want me to create a fresh PR?

I'll create a new PR in any case for the second commit for the `.error()` straight after we get this one landed.

vietnogi:
@Standard8 i rebased my loadedFix branch with xhr-onerror branch

Standard8:
Ok great! Now we just need someone to review this :-)

fatso83:
Thanks for picking up on this, guys! From my point of view, this is starting to look good.

@fearphage judging from the discussion, I think you are better equipped (meaning already looked into the details) to have a final say on this.

As a sidenote, @fearphage has just been promoted to co-maintainer based on his previous contributions 🎉

vietnogi:
ping!

test/util/fake-xml-http-request-test.js
```diff
@@ -1106,13 +1106,80 @@ if (typeof window !== "undefined") {
...
+            it("nulls request headers", function () {
+                this.xhr.open("GET", "/");
+                this.xhr.setRequestHeader("X-Test", "Sumptn");
+
+                this.xhr.error();
+
+                assert.equals(this.xhr.requestHeaders, {});
+            });
+
+            it("does not have undefined response headers", function () {
```
fearphage: The next test already handles this. This one isn't needed.

test/util/fake-xml-http-request-test.js
```diff
@@ -1106,13 +1106,80 @@ if (typeof window !== "undefined") {
...
+
+            it("sets responseText to empty string", function () {
+                this.xhr.responseText = "Partial data";
+
+                this.xhr.error();
+
+                assert.same(this.xhr.responseText, "");
+            });
+
+            it("sets errorFlag to true", function () {
```
fearphage: Set `errorFlag` to `false` to confirm that it changed. In theory `this.xhr.errorFlag` could be `true` before `.error()` is called in this test.

vietnogi:
im losing hope in this PR, im probably going to close it

fatso83:
@vietnogi did you update the tests? wasn't really anything else preventing its inclusion.

fatso83:
OK, I will fix up the tests and get this merged. These are some good changes, IMHO.

fatso83:
To include additional commits I re-submitted this as #1069.

fatso83:
This (or rather #1069) has been merged into `master`. Thanks for your contributions, @vietnogi and @Standard8!

Issue #1040, opened 2016-05-02
fakeServer is broken with .respond(500) on 1.17.4


gil:
Hi guys,

I'm using `karma-sinon` and it's always installing Sinon's latest version by default. It seems that the version 1.17.4 broke this for me:

`this.server.requests[0].respond(500, { 'Content-Type' : 'application/json' }, JSON.stringify({}));`

It won't call the error handler on my Ajax call. For some reason, I can't even find the tag for this version here on Github, to help debug the problem. As a workaround, I've downgraded to 1.17.3 and ran shrinkwrap on my project to be safe.

* Sinon version : 1.17.4
* Environment   : OSX
* Other libraries you are using: karma-sinon

**What did you expect to happen?**
Ajax error to be triggered.

**What actually happens**
I won't trigger my Ajax error handler.

**How to reproduce**
```
this.server = sinon.fakeServer.create();
this.server.requests[0].respond(500, { 'Content-Type' : 'application/json' }, JSON.stringify({}));
```

mbarlock:
I can confirm that this happens on 1.17.4 but not 1.17.3. I have a similar setup with karma-sinon.

mbarlock:
My guess is the problem is in this  commit https://github.com/sinonjs/sinon/commit/2cfbacd5cea5b63c014076d3a65b6642b2200793

wlepinski:
The tag `1.17.4` was pushed to npm registry but I didn't found any trace of this tag on this repository. What happened?

mbarlock:
My guess is the tag just hasn't be created yet. It was only released 3 hours ago

wlepinski:
@mbarlock Yeah, probably — nonetheless I think would be better if the tag on GitHub was released first. At least we would take a look and help on a PR or something to fix.

fatso83:
My bad. Forgot `git push --tags`. Thanks for the info on the bug.

Standard8:
I've just confirmed commit 2cfbacd definitely caused the tests to fail for us in mozilla/loop#400.

I applyed the patch from #1031 locally, and it fixes our tests.

overcaffeinated:
If 1.17.4 changes existing behavior, shouldn't it be part of a new marjor release? Currently, it's considered compatible with 1.17.3, so a package.json specifying a sinon dependency as "^1.17.3" will get 1.17.4 and potentially fail tests that used to work.

fearphage:
[...]

PR #1041, opened 2016-05-04
Fix onerror event triggering for fake xhr requests

fearphage:
The `xhr.onerror` event was not firing at all.

I cam across this issue while looking into #1040 and #1031. There are a finite number of `xhr.on__event__` hooks that sinon was checking and `onerror` was not on the list. 

Note: There is talk of errors being triggered at the wrong time in sinon. I did not address that at all. This only makes `xhr.onerror` trigger when an `error` event is triggered. I added additional tests to ensure the order and make sure this stays functional in the future.

#### Solution
Added `error` event to sequence of events looked for on xhr object (`xhr['on' + name]`).

2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -125,7 +125,7 @@ function FakeXMLHttpRequest(config) {
     }
 
     var xhr = this;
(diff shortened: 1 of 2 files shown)


avaly:
Could this be released in a patch version soon? Thanks!

Issue #432, opened 2014-03-11
XMLHttpRequest.abort() fires load event


kumar303:
I'm working with code like this:

```
var server = sinon.fakeServer.create();
server.respondWith(function(xhr, id) {
  xhr.abort();
});
// make the request...
server.respond();
```

It seems that abort() triggers the `load` event first, with data. Eventually it triggers the abort event but the timing doesn't seem right. In reading the code I see that a ready state change before triggering abort is probably the culprit here. I think abort() should not trigger any load events at all.

Maybe I'm setting up an abort call wrong?

I'm using Sinon from node @1.9.0

Standard8:
I believe this has been fixed in the latest code by #861. It isn't released yet, but the fix is there and seems to work.

lddubeau:
I've tried with 1.17.4 which contains PR #861. To get jQuery to recognize an xhr as aborted in 1.17.3, I had to do this:

    xhr.readyState = sinon.FakeXMLHttpRequest.UNSENT;
    xhr.statusText = "abort";
    xhr.abort();

with 1.17.4, I have to do this:

    xhr.statusText = "abort";
    xhr.abort();

fatso83:
There is currently a big discussion related to the changes introduced by 1.17.4 in https://github.com/sinonjs/sinon/issues/1040#issuecomment-216996170. Regarding what fires when, loaded events and errors, you might want to wait until the dust settles for a final solution. Seems like we have a bigger bug in here.

fearphage:
I believe this issue has been resolved. If you disagree, feel free to reopen the issue with any additional details that you can provide.

PR #1043, opened 2016-05-07
Additional testing around XHR events to ensure only expected events are fired

fearphage:

Added additional tests around the events fired from fake XHR. There are no code changes -- just additional tests.

These tests are related to and build on tests added in #1041. There are partially in response to the conversation in #432.

2. `npm install`
3. `npm test`

### DIFF
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -84,17 +84,30 @@ var assertEventOrdering = function (event, progress, callback) {
         this.xhr.send();
 
         this.xhr.upload.addEventListener("progress", observe("upload:progress"));
-        this.xhr.upload.addEventListener(event, observe("upload:" + event));
         this.xhr.upload.addEventListener("loadend", observe("upload:loadend"));
         this.xhr.addEventListener("progress", observe("xhr:progress"));
-        this.xhr.addEventListener(event, observe("xhr:" + event));
-        this.xhr["on" + event] = observe("xhr:on" + event);
         this.xhr.addEventListener("loadend", function (e) {
             assertProgressEvent(e, progress);
-            assert.equals(eventOrder, expectedOrder);
 
-            done();
+            // finish next tick to allow any events that might fire
+            // after loadend to trigger
+            setTimeout(function () {
+                assert.equals(eventOrder, expectedOrder);
+
+                done();
+            }, 1);
         });
+
+        // listen for abort, error, and load events to make sure only
+        // the expected events fire
+        ["abort", "error", "load"].forEach(
+            function (name) {
+                this.xhr.upload.addEventListener(name, observe("upload:" + name));
+                this.xhr.addEventListener(name, observe("xhr:" + name));
+                this.xhr["on" + name] = observe("xhr:on" + name);
+            },
+            this
+        );
 
         callback(this.xhr);
     });


fatso83:
P.S. The Travis build fails when doing the cloud tests, but this is not this test's fault (we think). We have unfortunately have had failing builds for a while ([history](https://travis-ci.org/sinonjs/sinon/builds)) ...

fatso83:
@mantoni any idea on how to get some more info out of `mochify`? The current build just seems to stall and time out without any error messages.

mantoni:
You can try to reproduce locally with a local WebDriver install. It's easy to set up and instruction are [provided here](https://github.com/mantoni/mochify.js#selenium-webdriver-setup).

There is also an option in [min-webdriver](https://github.com/mantoni/min-webdriver#supported-options) to keep the browser open on failures (set `closeOnError` to `false`). This gives you a chance to inspect the browser console output.

fearphage:
I'm on master and fully up to date, but the command run is not even the same:

```shell
➜  Sinon.JS git:(master) npm run test-headless

> sinon@2.0.0-pre test-headless /home/phred/src/git/Sinon.JS
> mochify --recursive -R dot --grep WebWorker --invert test/

# phantomjs:



  ...............................................................................................................................................................................................
  ...............................................................................................................................................................................................
  ...............................................................................................................................................................................................
  ......................................test
.........................................................................................................................................................
  ...............................................................................................................................................................................................
  ...............................................................................................................................................................................................
  ...............................................................................................................................................................................................
  ....................................................................................................................................................

  1477 passing (856ms)
  8 pending
```

Compared to Travis:

```
> sinon@2.0.0-pre test-headless /home/travis/build/sinonjs/sinon
> mochify --recursive -R dot --grep WebWorker --invert test/ "--wd"
```

fearphage:
I found the error in my ways. I should be running `npm run test-cloud` instead.

New problem: `SAUCE_ACCESS_KEY`

I pulled the SAUCE_USERNAME from the Travis logs.

mantoni:
@fearphage The easiest is to create a free SauceLabs account and put your user and access key into your environment. Then you can always run any mochify browser tests from your local console. It doesn't have to be the one used for the Sinon travis builds.

mantoni:
@fearphage @fatso83 I tried to reproduce the build issue on my machine, but I can't reproduce. Travis seems to time out after 10 minutes, and what I see is that IE 10 and 11 take a long time (around 2 minutes) to run. That still does not explain why Travis does not receive any output and fails with a timeout eventually.

However, there are two test cases that currently fail on IE 10 and IE 11:

```
1479 passing (2m)
8 pending
2 failing

1) sinon.clock .tick fires timers in correct order:
   timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


2) sinon.clock .tick passes 2 hours, 34 minutes and 12 seconds:
   timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

There have been a few releases of Mocha recently – that's the only external change that comes to mind.

fatso83:
@mantoni This did not start recently. Those two tests started failing 2 months ago with [this build](https://travis-ci.org/sinonjs/sinon/jobs/121782850) that merged pull request #1023.

fearphage:
If you wait long enough, they do pass:

## Fixes

* `timers in correct order` - Set timeout for test to 5 seconds (`this.timeout(5000);`). It generally takes 2.5 seconds in IE.
* `passes 2 hours, 34 minutes and 12 seconds` - Set test timeout to 50 seconds (`this.timeout(50000)`). I'm seeing it generally takes roughly 25 seconds, but I'm just being safe here. I think this is a bug in `lolex`. I'm investigating.

fatso83:
@fearphage That's great news. I have been scratching my head about this for some time, but haven't looked into it really hard.

PR #1050, opened 2016-05-25
backport hasOwnPropery fix to v1.17.x

developit:
This fixes `deepEqual()`'s reliance on `hasOwnProperty()` being in all objects (and object having prototypes at all) in 0.17.x.  It's something that has already made it's way into V2, but it's a very small change that un-breaks testing objects created with `Object.create(null)`.

### DIFF
--- a/test/sinon-test.js
+++ b/test/sinon-test.js
@@ -506,16 +506,47 @@
                 assert(sinon.deepEqual(obj1, obj2));
             },
 
-            "passes objects without a prototype": function () {
-                function Obj() {}
-                Obj.prototype = Object.create(null);
+            "passes object without prototype compared to equal object with prototype": function () {
+                var obj1 = Object.create(null);
+                obj1.a = 1;
+                obj1.b = 2;
+                obj1.c = "hey";
 
-                var obj = new Obj();
-                obj.foo = "bar";
+                var obj2 = { a: 1, b: 2, c: "hey" };
 
-                var pojo = { foo: "bar" };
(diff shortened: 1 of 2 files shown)


fearphage:
Thank you for the PR. This should be accompanied by tests.

developit:
Indeed! Will update shortly.

fatso83:
0.17.x? That's `1.17.x`, I presume?

developit:
Ah yes - sorry!

test/sinon-test.js
```diff
@@ -504,6 +504,18 @@
                 };
 
                 assert(sinon.deepEqual(obj1, obj2));
+            },
+
+            "passes objects without a prototype": function () {
+                function Obj() {}
+                Obj.prototype = Object.create(null);
+
+                var obj = new Obj();
```
fearphage: Simplify the test a bit:

```js
var obj = Object.create(null);
obj.foo = "bar";
```

Also there a [few variations of this test](https://github.com/sinonjs/sinon/blob/master/test/util/core/deep-equal-test.js#L266-L307) in master. Could you cover all these bases please?

developit:
@fearphage I just replaced my test with the four from that copy.

fearphage:
Looks good to me. Thank you, @developit.

calebboyd:
Awesome. This fixes all my breaking tests after updating to Node 6 (Node 6 `querystring.parse` now returns prototype-less objects)

PR #1052, opened 2016-05-26
ReactNative compatibility. Allow sinon fakeServer to run in React Native

andpor:

1. Create react native sample app. Look at RN getting started
2. Add sinon as dependency in package.json, run npm install
3. Require sinon
4. Code fakeServer test cases in RN app
5. Run all normal test cases to verify no regression.

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -32,8 +32,7 @@ function responseArray(handler) {
     return response;
 }
 
-var wloc = typeof window !== "undefined" ? window.location : {};
-wloc = typeof wloc !== "undefined" ? wloc : { "host": "localhost", "protocol": "http"};
+var wloc = typeof window !== "undefined" ? window.location : { "host" : "localhost", "protocol" : "http"};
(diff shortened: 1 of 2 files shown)


lib/sinon/util/fake_server.js
```diff
@@ -33,6 +33,7 @@ function responseArray(handler) {
 }
 
 var wloc = typeof window !== "undefined" ? window.location : {};
+wloc = typeof wloc !== "undefined" ? wloc : { "host" : "localhost", "protocol" : "http"};
```
fatso83: Line 35 and 36 could be squashed into one line. You are guaranteed `that` wloc is defined.

```javascript
var wloc = typeof window !== "undefined" ? window.location : { "host" : "localhost", "protocol" : "http"};
```

fatso83:
I think this is fine, but could you just compress the two lines mentioned? I am also going to squash these commits into one, unless you do it yourself, as a final step.

lib/sinon/util/fake_server.js
andpor: will do that tomorrow.

andpor:
Change is done. Please squash.

Issue #43, opened 2011-10-03
How do I pronounce 'sinon'


timoxley:
is it a long i or short i sound? eee?

> sigh-non
> sin-on
> see-non?

Please help.

helior:
I'm sure developers everywhere are still debating this. I just looked up the wikipedia entry and cross-referenced with Google translate (https://translate.google.com/#auto/es/%CE%A3%CE%AF%CE%BD%CF%89%CE%BD) and it sounds like "sea-naan".

But if Christian is pronouncing it "sign-on", who am I to argue :)

cjohansen:
I rarely pronounce it in English. In Norwegian I pronounce it as "sea-non". But who am I to be decide how it's pronounced?

MelindaYin:
It's actually pronounced Shinon. Shi sound from shin.
http://swordartonline.wikia.com/wiki/Talk:Sinon

SinonIsQueen:
In regular English language, it's pronounced, (Sea-non), but if you were to say her name in the proper way with the Japanese language, The "S" Becomes a soft "Sh" So it would be pronounced (She-non), hope this helped. c:

howell-chad-volusion:
http://www.dictionary.com/browse/sinon

fatso83:
@MelindaYin and @SinonIsQueen: you are referring to another Sinon. This has nothing to do with sword art :-)  Sinon produces spies and stubs (which are things pretending to be something else). Sinon is named after the [Greek warrior that pretended to be a deserter](https://en.wikipedia.org/wiki/Sinon) and lured the Trojans into taking the horse containing the Greek soldiers into their city. Makes sense?

PR #1056, opened 2016-06-01
Add a line recommending how to pronounce.

AdilKhn:
Adds a line on how to pronounce Sinon.


fatso83:
While the idea is fine, could you please tell me where you got the idea that it is pronounced "see-non"? Referring to #43, you could se Christian give two versions:
- "sea-non", which is a Norwegian pronounciation
- "sigh-non", which is the English pronounciation

Given the global usage of Sinon it seems reasonable to use the English version as a standard.

mantoni:
+1 for "sigh-non".

fatso83:
@AdilKhn Just change the line as per above and this is fine to merge.

fatso83:
Merged as  9bdf832

PR #1059, opened 2016-06-03
Do not invoke getters in walk

mantoni:

When stubbing a prototype, the `walk` implementation accesses getters to obtain the value of each property to pass it to the iterator function. This value is not needed by any of the iterator implementations anymore, but invoking the getter may have side effects or throw exceptions.

#### Solution  - optional

With this change, values are no longer obtained from the objects. It keeps the code to pass a different target object to the iterator which allows to obtain the value in the iterator implementation if necessary.

It turned out that none of the iterator functions passed to `walk` actually used the value.

```js
var redis = require("redis");
var sinon = require("sinon");

sinon.createStubInstance(redis.RedisClient); // throws without this change
```

### DIFF
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -109,7 +109,7 @@ var collection = {
             var col = this;
             var stubbedObj = sinon.stub.apply(sinon, arguments);
 
-            walk(stubbedObj, function (val, prop, propOwner) {
+            walk(stubbedObj, function (prop, propOwner) {
                 if (
                     typeof getPropertyDescriptor(propOwner, prop).value === "function"
                 ) {
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -56,7 +56,7 @@ function stub(object, property, func) {
     }
 
     if (typeof property === "undefined" && typeof object === "object") {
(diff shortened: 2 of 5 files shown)


mantoni:
Note: The browser tests are still failing due to our SauceLabs account being configured with zero concurrent VMs. I’ve email their support and hope to get the issue resolved soon.

fatso83:
Merging this as the browser tests are failing anyway these days.

kkamkou:
Thank you for the commit. fatso83, is it possible to cpeek this commit and release a new minor version?

fatso83:
@mantoni and @fearphage : should we release a new minor version, or try to get #1040 fixed first?

mantoni:
@fatso83 I'm currently linking sinon locally until this fix is released, so no pressure for me personally.

fatso83:
@kkamkou : if you don't want to wait for a minor fix you can just link to the relevant commit in your `package.json`. See the [NPM docs in the Github section](https://docs.npmjs.com/files/package.json#github-urls) for details.

kkamkou:
@fatso83 already. In the same time we're using versioneye, therefore version tracking is broken for one particular repo.

PR #1060, opened 2016-06-03
Added timeouts to ensure tests pass

fearphage:
Tests were timing out. I updated the mocha timeout values to allow them to complete.

This was a result of the discussion in #1043.

### DIFF
--- a/test/util/fake-timers-test.js
+++ b/test/util/fake-timers-test.js
@@ -346,6 +346,8 @@ describe("sinon.clock", function () {
         });
 
         it("fires timers in correct order", function () {
+            this.timeout(5000);
+
             var spy13 = sinon.spy.create();
             var spy10 = sinon.spy.create();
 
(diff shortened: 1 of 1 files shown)


test/util/fake-timers-test.js
```diff
@@ -414,6 +416,8 @@ describe("sinon.clock", function () {
         });
 
         it("passes 2 hours, 34 minutes and 12 seconds", function () {
+            this.timeout(50000);
```
mantoni: This seems a bit excessive. Why does it take almost a minute?

fearphage: I agree. The problem from a quick glance seems to be based on the implementation of [`clock.tick` in lolex](https://github.com/sinonjs/lolex/blob/master/src/lolex-src.js#L479-L524). Here is [some (very little) context](https://github.com/sinonjs/sinon/pull/1043#issuecomment-222141239). I haven't put a lot of time into investigating though.

mantoni:
It’s better to have the tests pass after a long time instead of a failure where nothing is really wrong, so I merged. The browser builds are working again - let’s see whether they turn green now ⌛

mantoni:
As green as can be! 🎉

mroderick:
🎆

mroderick:
Thanks @fearphage!!!

Issue #1061, opened 2016-06-03
Unable to use sinon due to fake server unsafe cookie checks


andpor:
Hi,

Currently sinon seems to have hardcoded logic to check http requests for unsafe cookies and rejects the headers should they match what is considered unsafe. There is no way to disable this check in case there is a need for it which renders the fake server with http request unusable.

It would be great to make this configurable, enabled by default.

I implemented a code change where unsafeCookieCheckEnabled property is added to sinon, defaulting to true but overridable by user code. The code in FakeXMLHtppRequest then checks this property value before engaging in unsafeCookie check.

2 lines of code change. Backward compatible and safe.

Should I go ahead and create a PR for this?

P.S. I have a React Native PR outstanding for a few days. Are you planning on merging it in or you se issues with the code or not finding the functionality particularly useful?

fatso83:
This seems useful, but try to make it a non-global change. In other words, change the factory method creating the server to accept such an option. I haven't looked at the code, so details may be off... 

ReactNative your pull request : this happens. We only do this voluntarily so things can take some time. Your team could always just use your fork while waiting.

PR #1063, opened 2016-06-06
Fix for issue #1061 - make unsafe header check in fake server configurable.

andpor:

There are times when developer wants to test the app with the fake server without it enforcing any cookie rules with regard to their safety. Currently, this is not possible and in such situations sinon becomes unusable because it blindly checks for unsafe headers and rejects any requests that carry them.

#### Solution

-     Make the unsafe headers optionals/configurable
-     Limit the change to local scope - fake server and fake xmlhttprequest
-     Make the change backward compatible i.e. unless overwritten by user, behavior is the same as before

```
let server = sinon.fakeServer.create()
server.unsafeHeadersEnabled = false
```

2.     npm install
3.     run all standard tests - none should fail
4.     create a new basic http request test case and set "cookie" header in the request.
5.     config the fakeServer with unsafeHeaderEnabled = false . test should succeed.
6.     config the fakeServer with unsafeHeaderEnabled = true . test should fail with unsafe header rejection.

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -74,11 +74,11 @@ var fakeServer = {
             this.xhr = sinon.useFakeXMLHttpRequest();
         }
         server.requests = [];
-
-        this.xhr.unsafeHeadersEnabled = function () {
-            return !(server.unsafeHeadersEnabled === false);
-        };
+        
         this.xhr.onCreate = function (xhrObj) {
+            xhrObj.unsafeHeadersEnabled = function () {
+                return !(server.unsafeHeadersEnabled === false)
+            };
             server.addRequest(xhrObj);
         };
 
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -74,10 +74,9 @@ var fakeServer = {
             this.xhr = sinon.useFakeXMLHttpRequest();
         }
         server.requests = [];
-        
         this.xhr.onCreate = function (xhrObj) {
             xhrObj.unsafeHeadersEnabled = function () {
-                return !(server.unsafeHeadersEnabled === false)
+                return !(server.unsafeHeadersEnabled === false);
             };
             server.addRequest(xhrObj);
         };
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -76,9 +76,8 @@ var fakeServer = {
         server.requests = [];
 
         this.xhr.unsafeHeadersEnabled = function () {
-            return !(server.unsafeHeadersEnabled === false)
+            return !(server.unsafeHeadersEnabled === false);
         };
-        
         this.xhr.onCreate = function (xhrObj) {
             server.addRequest(xhrObj);
         };
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -75,6 +75,10 @@ var fakeServer = {
         }
         server.requests = [];
 
+        this.xhr.unsafeHeadersEnabled = function () {
+            return !(server.unsafeHeadersEnabled === false)
+        };
+        
         this.xhr.onCreate = function (xhrObj) {
             server.addRequest(xhrObj);
         };
@@ -88,7 +92,8 @@ var fakeServer = {
             "autoRespondAfter": true,
             "respondImmediately": true,
             "fakeHTTPMethods": true,
-            "logger": true
+            "logger": true,
+            "unsafeHeadersEnabled": true
         };
         var setting;
 
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -12,7 +12,7 @@
 "use strict";
 
 var push = [].push;
-var sinon = require("./core");
+var sinon = require("../../sinon");
 var createInstance = require("./core/create");
 var format = require("./core/format");
 var configureLogError = require("./core/log_error");
@@ -64,6 +64,22 @@ function match(response, request) {
     return false;
 }
 
+function incrementRequestCount() {
+    var count = ++this.requestCount;
+
+    this.requested = true;
+
+    this.requestedOnce = count === 1;
+    this.requestedTwice = count === 2;
+    this.requestedThrice = count === 3;
+
+    this.firstRequest = this.getRequest(0);
+    this.secondRequest = this.getRequest(1);
+    this.thirdRequest = this.getRequest(2);
+
+    this.lastRequest = this.getRequest(count - 1);
+}
+
 var fakeServer = {
     create: function (config) {
         var server = createInstance(this);
@@ -74,6 +90,8 @@ var fakeServer = {
             this.xhr = sinon.useFakeXMLHttpRequest();
         }
         server.requests = [];
+        server.requestCount = 0;
+
         this.xhr.onCreate = function (xhrObj) {
             xhrObj.unsafeHeadersEnabled = function () {
                 return !(server.unsafeHeadersEnabled === false);
@@ -107,6 +125,8 @@ var fakeServer = {
     addRequest: function addRequest(xhrObj) {
         var server = this;
         push.call(this.requests, xhrObj);
+
+        incrementRequestCount.call(this);
 
         xhrObj.onSend = function () {
             server.handleRequest(this);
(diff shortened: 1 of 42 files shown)


fatso83:
Just to make sure I understand: is the behavior you are allowing actually possible in any browsers? Seems weird to introduce non-compliant behavior in that case.

andpor:
It is allowed in React Native. Without disabling the header check, simple thing, the entire package is unusable since we set cookie header...

fatso83:
I am ready to merge this, but it lacks tests for this feature. Could you add a few that shows that the feature is working?

andpor:
So this is a little bit of predicament. Because in browser env where std tests run the feature would not work, which is expected. Not sure how to write test cases for React Native...can you suggest an approach for adding tests for this PR?

fatso83:
I am not sure what you mean. There is nothing in the code changes that is React Native, specific, AFAIK? And the hardware has nothing to do with this. It is a pure unit test that's wanted.

I would assume that there is a test there already testing for the opposite (that it crashes when the setting is not enabled). So just create a new test after the existing one, create a server object, passing it a config object where you enable the unsafe headers setting, create a XHR object, and see it pass without any error being thrown. I am quite sure the existing tests will guide you just fine :-)

lib/sinon/util/fake_server.js
```diff
@@ -74,8 +74,10 @@ var fakeServer = {
             this.xhr = sinon.useFakeXMLHttpRequest();
         }
         server.requests = [];
-
         this.xhr.onCreate = function (xhrObj) {
+            xhrObj.unsafeHeadersEnabled = function () {
```
jonnyreeves: Why does `unsafeHeadersEnabled` need to be a function - why can't it just be a bool?

jonnyreeves:
As @fatso83 says, this would benefit from test-coverage, please can you also document this configuration option in the [fakeServer docs](https://github.com/sinonjs/sinon/blob/master/docs/current/fake-xhr-and-server.md)?  Thanks.

lib/sinon/util/fake_server.js
andpor: With function approach it is a much cleaner approach as the property value is maintained in one single place - server. Otherwise you introduce a disconnect between the server and http request for no reason. What if decide to change the value of this property at server level during the lifecycle of the app? If I propagate the property value down to the http request level, the request will never see the new server level value but rather use the now invalid original property?

Makes sense ?

andpor:
Yes, i am planning on adding the test case and will update docs as you suggest as well.

lib/sinon/util/fake_xml_http_request.js
```diff
@@ -54,22 +54,22 @@ sinonXhr.supportsCORS = isReactNative ||
 var unsafeHeaders = {
     "Accept-Charset": true,
     "Accept-Encoding": true,
-    Connection: true,
+    "Connection": true,
```
fearphage: Why are you adding unnecessary quotes here? What purpose are they serving?

fatso83: I think it is just to be consistent - a visual thing. Some people object to using quotes for just some of the keys. Not seeing a big problem :-)

andpor: cosmetics...looks more consistent with all keys in quotes...

andpor:
i hope to be able to get to these test cases this week finally...

fatso83:
@andpor that's great news. just ping us if you get stuck somehow, but I think you should be able to figure it out from the info above. just a quick tip: try to run the test with your changes commented out and see it fail, just to make sure you are testing the right thing :-)

andpor:
Been awfully busy ever since I wrote my last msg on 07/26. I hope to get back to finishing up this PR within the next week. Feel bad for delaying but priorities...

andpor:
@fatso83 - hi - I finally managed to add the proper test cases to PR but now i have a conflict with one of the files. Can you tell me the procedure to resolve it ? have never done it in GitHub settings...

andpor:
@fatso83 - I have merged the latest master into my local and trying to commit. Get an error that I do not recognize ... seems something wrong with project config??

Now using node v4.2.6 (npm v3.9.3)
Done
/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config/config-validator.js:102
        throw new Error(message.join(""));
        ^

Error: /Users/sylwia/Documents/Projects/sinon/.eslintrc:
	Configuration for rule "keyword-spacing" is invalid:
	Severity should be one of the following: 0 = off, 1 = warning, 2 = error (you passed "error").

    at validateRuleOptions (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config/config-validator.js:102:15)
    at /Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config/config-validator.js:148:13
    at Array.forEach (native)
    at Object.validate (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config/config-validator.js:147:35)
    at Object.load (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config/config-file.js:390:19)
    at loadConfig (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config.js:74:33)
    at getLocalConfig (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config.js:177:23)
    at Config.getConfig (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/config.js:272:22)
    at processText (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/cli-engine.js:203:27)
    at processFile (/Users/sylwia/Documents/Projects/sinon/node_modules/eslint/lib/cli-engine.js:270:18)
pre-commit: 
pre-commit: We've failed to pass the specified git pre-commit hooks as the `eslint-pre-commit`
pre-commit: hook returned an exit code (1). If you're feeling adventurous you can
pre-commit: skip the git pre-commit hooks by adding the following flags to your commit:
pre-commit: 
pre-commit:   git commit -n (or --no-verify)
pre-commit: 
pre-commit: This is ill-advised since the commit is broken.
pre-commit:

andpor:
@fatso83 Carl - test cases added to the PR - should be ready to go. Please review.

mroderick:
If you rebase this on to `master`, so the merge commits go away, then I would say this is good to go

andpor:
@mroderick - how do I do that exactly? Are these the right steps you are looking for?

```
git fetch origin            
git rebase origin/master   
```

fatso83:
@andpor: 
Given the following assumptions are correct, that seems correct to me :) 

- `origin` is pointing to `git@github.com/sinonjs/sinon` (or the https version)
- `myRepo` is pointing to `git@github.com/andpor/sinon` (or the https version)

Remember to run `npm test` afterwards in case anything should go awry. If you mess it up, you can always reset the branch to what you have pushed by doing `git reset --hard myRepo/master`. 

When you are finished, and things look good, just do `git push -f` to overwrite your previously pushed commits.

andpor:
So I am clear....what is wrong with the PR and how is rebasing going to fix this? I want to understand before I start playing with the repo...

fatso83:
Avoiding merge commits makes the history far clearer and makes using tools like `git bisect` to find regressions far simpler. I won't go into details, but you can check out this article from another open source project [on the background and how to do this](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request). On a general note, just use `git pull --rebase` whenever you usually would do a `git pull` and you won't have any issues. I have set that as the default (`pull.rebase=true`).

Anyway, seeing that there is basically just one commit message of any importance, I rebased your branch, fixed the conflicts, and squashed all your changed and pushed them to a branch on my own repo: [1063-rebased__fix-for-1061-unsafe-headers](https://github.com/fatso83/sinon/tree/1063-rebased__fix-for-1061-unsafe-headers). 

The tests are still running, so I guess it worked out. You can reset your branch pointer to point to mine by doing
```
git checkout master
git remote add fatso83 git@github.com:fatso83/sinon.git
git fetch fatso83
# optionally have a look at it by checking it out at this point
git reset --hard fatso83/1063-rebased__fix-for-1061-unsafe-headers
```

andpor:
Thank you for doing this Carl. I will take a look at the reference to learn the process. Assume you will merge from your branch to origin..

fatso83:
Wrong assumption, as the process sketched out above would let your reset the branch pointer to mine, and then you could have force pushed the changes back into your published branch. But to make this a bit quicker, I have just recreated this PR as #1168 to speed it up. So I am closing this and merging that.

andpor:
Cool thanks again. Need to play with GIT to learn this process.

Issue #988, opened 2016-02-17
exposing wrappedMethod


tonylukasavage:
I have a use case where I need to stub a function, but only under a certain circumstance. Essentially, I need a function to execute normally until a certain condition is met, and then I need to stub it. I could do this easily if the `wrappedMethod` was exposed, rather than internal to the module. 

So basically I'm asking if there's a fundamental reason that `wrappedMethod` is not exposed? If not, would you be open to a PR that exposed it on spies, stubs, and mocks?

fatso83:
I am having a problem understanding why you cannot do this already?

```
var originalFunction = myLib.functionFoo;
var stub = sinon.stub(myLib, 'functionFoo');

// ...
originalFunction(1,2,3);

if(caseIsMet)
 stub(1,2,3);

originalFunction(4,5,6);
```

tonylukasavage:
Some code would probably better explain. It's not a big deal, I was more wondering if there was an explicit reason _not_ to expose it.

```js
// a.js
module.exports = function() {
  api.func(1);
  api.func('badvalue');
  api.func(2);
}

// test (how I do it now)
var _func = api.func;
sinon.stub(api, 'func', function(val) {
  if (val === 'badvalue') { 
    throw new Error('some_error'); 
  } else {
    _func.apply(api, arguments);
  }
});

// test (how I'd like to do it)
var stub = sinon.stub(api, 'func', function(val) {
  if (val === 'badvalue') { 
    throw new Error('some_error'); 
  } else {
    stub.wrappedFunction.apply(api, arguments);
  }
});
```

fatso83:
OK, I get what you want, and I have a feeling this comes up from time to time, but I don't remember where I saw the last discussion on this...  For a more recent example, instance #989 seems like essentially a duplicate of this. It is not a direct duplicate, but the suggested API would probably achieve what you intend to accomplish. Am I right? That would be a cleaner API IMHO.

fatso83:
To elaborate, your example would then be written:
```javascript
sinon.stub(api,'func')
     .withArgs('badvalue')
     .throws(Error).callThrough()
```

It should be mentioned that this would only apply to a subset (probably the majority) of possible conditions as it just filters on equality of arguments passed, and you cannot pass arguments to the error constructor.

lucasfcosta:
Hi friends!

[...]

PR #1064, opened 2016-06-08
expose wrappedMethod on the wrapped method (#988)

tonylukasavage:
#### Purpose

Add #988 which asks for the `wrappedMethod` variable to be exposed as a property
#### Background

All relevant background is in #988 
#### Solution

This solution works by exposing the `wrappedMethod` property on the sinon method object.
#### How to verify

``` js
var API = {
    foo: function() {}
};
var func = API.foo;
console.log(sinon.stub(API, 'foo').wrappedMethod === func); // prints true
```

### DIFF
--- a/lib/sinon/util/core/wrap-method.js
+++ b/lib/sinon/util/core/wrap-method.js
@@ -76,7 +76,7 @@ module.exports = function wrapMethod(object, property, method) {
 
         var types = objectKeys(methodDesc);
         for (i = 0; i < types.length; i++) {
-            wrappedMethod = wrappedMethodDesc[types[i]];
+            method.wrappedMethod = wrappedMethod = wrappedMethodDesc[types[i]];
             checkWrappedMethod(wrappedMethod);
         }
 
@@ -86,7 +86,7 @@ module.exports = function wrapMethod(object, property, method) {
         }
         Object.defineProperty(object, property, methodDesc);
     } else {
-        wrappedMethod = object[property];
+        method.wrappedMethod = wrappedMethod = object[property];
         checkWrappedMethod(wrappedMethod);
         object[property] = method;
         method.displayName = property;


fatso83:
Thanks for the PR. This needs documentation tests for it to be allowed in. Please also add a test for how this behaves with chained stub creation, i.e. using `withArgs()`, `returns()`, etc. I am unsure of whether it would work in the same way.

tonylukasavage:
I saw `wrappedMethod` as kind of a combo breaker here. I wouldn't expect to be chaining beyond it, as its use implies that you are not referencing the stub anymore. If this is an unacceptable approach, it might make sense to just close out this PR since I personally don't think chaining when referencing `wrappedMethod` makes any sense. If the absence of chaining on `wrappedMethod` is ok, I'll work on the documentation tests.

fatso83:
that's not what I meant. I was just wondering if these two would return the same:

```javascript
var orgFunction = sinon.stub(API, 'foo').wrappedMethod;
```

```javascript
var stillTheOrgFunction /*?*/ = sinon.stub(API, 'foo').withArgs(100).returns(2).withArgs(10).returns(4).wrappedMethod;
```

fatso83:
Just tested, and while it works for the first, it seems to break for the second example:

```javascript
var API = { foo : function() { console.log("hi!") } }
var stillTheOrgFunction /*?*/ = sinon.stub(API, 'foo')
.withArgs(100)
.returns(2)
.withArgs(10)
.returns(4)
.wrappedMethod;
stillTheOrgFunction() // BOOM => TypeError: stillTheOrgFunction is not a function
```

Needs some more work?

fatso83:
Ping - did you read my comment? There seems to be a bug.

fatso83:
@tonylukasavage Hi, Tony. There seems to be some bugs in your PR. Any chance of fixing them?

fatso83:
Closing due to lack of response. Anyone willing to pick this up and fix the remaining issues are welcome to do so.

PR #1068, opened 2016-06-20
Add `sandbox.reset()` description

jackbrown:

Add docs for `sandbox.reset()` which was previously undocumented

2. open `docs/current/sandbox.md`
3. review `sandbox.reset()` documentation

### DIFF
--- a/docs/current/sandbox.md
+++ b/docs/current/sandbox.md
@@ -104,6 +104,10 @@ Access requests through `sandbox.requests` and server through `sandbox.server`
 
 Restores all fakes created through sandbox.
 
+#### `sandbox.reset();`
+
+Resets the internal state of all spies created through sandbox.
+
 
 ## Test methods
 


mantoni:
You want to document `sandbox.restore()`, right? The description also says "Restores ...".

mantoni:
Ah, no - you just copied the `restore` documentation. They're not the same thing.

jackbrown:
what would be a better way of documenting this? maybe something like
`resets the internal state of all spies created through a sandbox`?

mantoni:
@jackbrown Yes, that sounds more like it.

jackbrown:
@mantoni thanks, updated.

mantoni:
Great! I know it's a pain for such a small change, but can I ask you to squash the commits?

jackbrown:
@mantoni no worries, done.

mantoni:
Thanks! :+1:

fatso83:
@mantoni : don't know if you noticed, but GH now allows you to select if you want to squash the commits when pressing the merge button. so we can do it ourselves in cases like this, unless you explicitly require the merge commit (which I don't think adds much for one-commit PRs).

mantoni:
@fatso83 Yes, I know. I wasn't sure whether we want to use the feature here. I agree that it's a good idea for simple changes like these where we don't really need the merge commit.

PR #1071, opened 2016-06-23
Fix issue 283 tests

### LINKED ISSUE #283: Fake timers do not work when asked to tick() too far ahead
Taking a program that sets an interval x and then asks Sinon to tick() y milliseconds. It seems that if y is much greater than x (double or more), then the interval will never be called.

I have reproduced this problem with the small mocha/sinon test case below and with sinon 1.4 and 1.7.

````
var assert = require('assert'),
    sinon = require('sinon');

function testSinonFakeTimersWith(interval, ticks) {
    var clock = sinon.useFakeTimers();

    var spy = sinon.spy();
    setInterval(spy, interval);
    assert(! spy.calledOnce);
    clock.tick(ticks);
    assert(spy.calledOnce);

    clock.restore();
}

describe("Testing sinon useFakeTimers", function() {
  // FAILS
  it('10/101', function() {
    testSinonFakeTimersWith(10, 101);
  });
  // PASS
  it('99/101', function() {
    testSinonFakeTimersWith(99, 101);
  });

  // FAILS
  it('100/200', function() {
    testSinonFakeTimersWith(100, 200);
  });
  // PASS
  it('199/200', function() {
    testSinonFakeTimersWith(199, 200);
  });

  // FAILS
  it('500/1001', function() {
    testSinonFakeTimersWith(500, 1001);
  });
  // PASS
  it('1000/1001', function() {
    testSinonFakeTimersWith(1000, 1001);
  });
});
````

````
$ mocha -R spec test/sinon-test.js                                                                                   ±[●●][error-notopened]


  Testing sinon useFakeTimers
    1) 10/101
    ✓ 99/101
    2) 100/200
    ✓ 199/200
    3) 500/1001
    ✓ 1000/1001
````

lucasfcosta:
#### Purpose
This fixes #283.
Basically the tests were wrong, we expect `calledOnce` to be true only if something got called exactly once, but with `setInterval` we will trigger the callback function multiple times if the `ticks` are more than double the interval set.

[**Please see this post for more details and a complete explanation**](https://github.com/sinonjs/sinon/issues/283#issuecomment-227921227)

#### How to verify
To check this out you can:
* Replace the spy by any function with a `console.log`, for example, and see it getting called multiple times (`ticks / interval`)
* Use `console.log` to show `callCount`
* Take a look at [this line](https://github.com/sinonjs/sinon/blob/master/lib/sinon/spy.js#L67) and confirm `calledOnce` will be true only if the spy was called exactly once

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -16,7 +16,7 @@ describe("issues", function () {
         this.sandbox.restore();
     });
 
-    it.skip("#283", function () {
+    it("#283", function () {
         function testSinonFakeTimersWith(interval, ticks) {
             var clock = sinon.useFakeTimers();
 
@@ -24,28 +24,17 @@ describe("issues", function () {
             var id = setInterval(spy, interval);
             assert(!spy.calledOnce);
             clock.tick(ticks);
-            assert(spy.calledOnce);
+            assert(spy.callCount === Math.floor(ticks / interval));
 
             clearInterval(id);
             clock.restore();
         }
 
-        // FAILS
         testSinonFakeTimersWith(10, 101);
-
-        // PASS
         testSinonFakeTimersWith(99, 101);
-
-        // FAILS
         testSinonFakeTimersWith(100, 200);
-
-        // PASS
         testSinonFakeTimersWith(199, 200);
-
-        // FAILS
         testSinonFakeTimersWith(500, 1001);
-
-        // PASS
         testSinonFakeTimersWith(1000, 1001);
     });
 


mantoni:
Great, thanks! You can also remove the "FAILS" / "PASSES" comments below since they all pass now :sunglasses:

lucasfcosta:
@mantoni
Ah, I totally forgot these, thanks for noticing, I'll do it :smile:

lucasfcosta:
@mantoni It's done.
Let me know if you guys need anything else.
Keep up the excellent work you've been doing on Sinon 👍

mantoni:
Thank you!

PR #1074, opened 2016-06-24
Add resetBehavior and resetHistory to sandbox API

### LINKED ISSUE #1072: Sandbox resetBehaviour and resetHistory
My understanding of issue #806 is that the behaviour of `reset` for stubs was changed to reset both the call history and the stub behaviour. It was decided that `reset` would reset both, `resetHistory` would reset just the call history, and `resetBehaviour` would reset just the stub behaviour.

It would be very convenient for me if these two functions, `resetHistory` and `resetBehaviour`, were made available through the sandbox API. That is, I would like them to be added to `lib/sinon/collections.js` alongside `reset` so that you could reset the history or behaviour of an entire sandbox at once.

Thoughts?

Gudahtt:
Adds resetBehavior and resetHistory to the sandbox API. Closes #1072 

I also noticed that a couple of the inline function names were incorrect - probably just accidentally misnamed - so there's a second commit to fix those.

### DIFF
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -44,7 +44,7 @@ function compact(fakeCollection) {
 }
 
 var collection = {
-    verify: function resolve() {
+    verify: function verify() {
         each(this, "verify");
     },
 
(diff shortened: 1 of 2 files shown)


fatso83:
Nice! LGTM.

mantoni:
Thank you!

PR #1085, opened 2016-07-08
Feature/cjsify typeof extend tests

jonnyreeves:
CJSify `extend` and `typeOf` tests

Removes all `sinon.*` references from the test suites


mantoni:
👍

PR #1088, opened 2016-07-08
Feature/cjsify sinon sandbox tests

jonnyreeves:
Remove `sinon.*` references from `sandbox` and its test suite! 🚀 

This PR represents the last "core" sinonJS module that needed to be migrated 🎉   There were a few of minor casualties in the test-suite, namely  3 tests around the deprecated(?) `sandbox.useFakeXMLHttpRequest()` API (which just call through to the `useFakeServer()` API anyway...)

### DIFF
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -5,6 +5,15 @@ var samsam = require("samsam");
 var sinon = require("../lib/sinon");
 var assert = referee.assert;
 var refute = referee.refute;
+var fakeServerWithClock = require("../lib/sinon/util/fake_server_with_clock");
+var sinonSandbox = require("../lib/sinon/sandbox");
+var sinonCollection = require("../lib/sinon/collection");
+var createInstance = require("../lib/sinon/util/core/create");
+var sinonSpy = require("../lib/sinon/spy");
+var sinonStub = require("../lib/sinon/stub");
+var sinonConfig = require("../lib/sinon/util/core/get-config");
+var sinonMatch = require("../lib/sinon/match");
+var sinonAssert = require("../lib/sinon/assert");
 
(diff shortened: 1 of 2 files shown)


test/sandbox-test.js
```diff
@@ -72,65 +82,41 @@ describe("sinon.sandbox", function () {
         it("uses restorable clock", function () {
             this.sandbox.useFakeTimers();
 
-            assert.isFunction(this.sandbox.clock.restore);
+            assert.clock(this.sandbox.clock);
         });
 
-        it("passes arguments to sinon.useFakeTimers", function () {
-            var stub = sinon.stub(sinon, "useFakeTimers").returns({ restore: function () {} });
+        // Not sure how we can test this without monkey-patching or testing the implementation of lolex.
```
mantoni: What is the problem with the commented out code? Looks like that should do ...

jonnyreeves: When I stubbed `useFakeTimers` it cause a bunch of assertion failures in other modules, almost as if the clock stub was not being restored.  Let me try it again and see if I can get to the root of that issue.

fatso83: There have been a few fixes in `lolex` related to `Date` and timers recently, so maybe try upgrading to `1.5.1` and see if that helps? I could also help out if you are short on time.

jonnyreeves:
@fatso83, @mantoni managed to resolve this by simply using a stub instead of a spy; before doing this I tried installing 1.5.0 of lollex (which didn't help) and I rewrote the 2 failing `fake-timers-tests` to cache `window.setTimeout` before the suite ran (this fixed the issue, but I have no idea why!)

My hope is that this kind of weird flakiness falls out from the suite once the last of the modules are cjsified... :)

fatso83:
LGTM

mantoni:
🚢

Issue #966, opened 2016-01-16
Sinon 2.0 Release


jonnyreeves:
What do we want to achieve in preparation for releasing a 2.0 release candidate of Sinon?

@mantoni  @fatso83  @cjohansen Here are a handful of proposed tasks; please edit this Issue, or comment below so we can get a list of tasks together and get 2.0 shipped :rocket: 

### CommonJS Migrations
- [x] Migrate `sinon.spy` #920 
- [x] Migrate `sinon.stub` #932 
- [x] Migrate `sinon.mock` #933 
- [x] Migrate fake_server and friends (Bulk of work done in #934, `useFakeXMLHttpRequest` still referenced, see #1118)
- [x] Migrate `sinon.sandbox` (Bulk of work done in #936) #1088
- [x] Migrate `sinon.format` (Tightly coupled in tests) #967
- [x] Migrate `sinon.collection` #1084

### Test Suite CommonJS Migrations
- [x] Migrate `assert` suite #1078 
- [x] Migrate `call` suite #1079 
- [x] Migrate `collection` suite #1084
- [x] Migrate `extend` suite #1085
- [x] Migrate `match` suite #1086
- [x] Migrate `mock` suite #1087
- [x] Migrate `sandbox` suite #1088
- [x] Migrate `spy` suite #1001
- [x] Migrate `stub` suite #1001
- [x] Migrate `typeOf` suite #1085
- [x] Migrate `util/core` suites #998, #1081
- [x] Migrate `util/event` suite #1115
- [x] Migrate `util/fake-timers` suite #1116
- [x] Migrate `util/fake-server` suite #1118
- [x] Migrate `util/fake-server-with-clock` suite #1118
- [x] Migrate `util/fake-xdomain-request` suite
- [x] Migrate `util/fake-xml-http-request` suite #1125

### Cleanup Tasks
- [x] Break down `test/sinon-test.js` suite #968
- [x] Remove `sinon.config` usage (Decision: [#936](https://github.com/sinonjs/sinon/pull/936#issuecomment-172175661). Removed entirely in #973)
- [x] Remove `sinon.logError` and `sinon.log` [#972]
- [x] Use CommonJS imports in tests (No more `sinon` global access, allows us to remove internal helpers from the public API) #996
- [x] Document API changes from 1.17 -> 2.0 and provide upgrade advice. #1090 

### Public API Changes
_tasks with a `?` require clarification from maintainers_
[...]

PR #1090, opened 2016-07-10
Add v2.0 Migration Guide

jonnyreeves:
- [x] Add deprecation guide.
- [x] Mark core utils as deprecated.
- [x] Mark `sinon.Event` and friends as deprecated.
- [x] Call out deprecation of legacy browsers

----

Went through all the pull requests opened since the initial kick off of the CJSify work and tried to reason about the changes to the public API.

Wasn't 100% where to the put the migration guide, but the `docs` folder felt like a good place :)

### DIFF
--- a/lib/sinon/util/core/deprecated.js
+++ b/lib/sinon/util/core/deprecated.js
@@ -1,0 +1,28 @@
+/*eslint no-console: 0 */
+"use strict";
+
+// wrap returns a function that will invoke the supplied function and print a deprecation warning to the console each
+// time it is called.
+exports.wrap = function (func, msg) {
+    var wrapped = function () {
+        // Watch out for IE7 and below! :(
+        if (typeof console !== "undefined") {
+            if (console.info) {
+                console.info(msg);
+            } else {
+                console.log(msg);
+            }
+        }
+        return func.apply(this, arguments);
+    };
+    if (func.prototype) {
+        wrapped.prototype = func.prototype;
+    }
+    return wrapped;
+};
+
+// defaultMsg returns a string which can be supplied to `wrap()` to notify the user that a particular part of the
+// sinon API has been deprecated.
+exports.defaultMsg = function (funcName) {
+    return "sinon." + funcName + " is deprecated and will be removed from the public API in a future version of sinon.";
+};
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
-# Sinon.JS
-
-[![Join the chat at https://gitter.im/sinonjs/sinon](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/sinonjs/sinon?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-
-[![Build status](https://secure.travis-ci.org/sinonjs/sinon.svg?branch=master)](http://travis-ci.org/sinonjs/sinon) [![bitHound Score](https://www.bithound.io/github/sinonjs/sinon/badges/score.svg)](https://www.bithound.io/github/sinonjs/sinon)
+# Sinon.JS [![npm version](https://img.shields.io/npm/v/sinon.svg?style=flat)](https://www.npmjs.com/package/sinon) [![Join the chat at https://gitter.im/sinonjs/sinon](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/sinonjs/sinon?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) [![Build status](https://secure.travis-ci.org/sinonjs/sinon.svg?branch=master)](http://travis-ci.org/sinonjs/sinon) [![bitHound Score](https://www.bithound.io/github/sinonjs/sinon/badges/score.svg)](https://www.bithound.io/github/sinonjs/sinon) [![Sauce Test Status](https://saucelabs.com/buildstatus/sinonjs)](https://saucelabs.com/u/sinonjs)
 
 Standalone and test framework agnostic JavaScript test spies, stubs and mocks (pronounced "sigh-non").
+
+[![Sauce Test Status](https://saucelabs.com/browser-matrix/sinonjs.svg)](https://saucelabs.com/u/sinonjs)
 
 ## Installation
 
@@ -20,11 +18,11 @@ See the [sinon project homepage](http://sinonjs.org/) for documentation on usage
 
 If you have questions that are not covered by the documentation, please post them to the [Sinon.JS mailing list](http://groups.google.com/group/sinonjs) or drop by <a href="irc://irc.freenode.net:6667/sinon.js">#sinon.js on irc.freenode.net:6667</a> or the [Gitter channel](https://gitter.im/sinonjs/sinon).
 
-### Important: AMD needs pre-built version
+### Important: Sinon v1.x does not work with AMD/CommonJS Bundlers!
 
-Sinon.JS *as source* **doesn't work with AMD loaders** (when they're asynchronous, like loading via script tags in the browser). For that you will have to use a pre-built version. You can either [build it yourself](CONTRIBUTING.md#testing-a-built-version) or get a numbered version from http://sinonjs.org.
+Sinon.JS v1.x *as source* **doesn't work with AMD loaders / RequireJS / Webpack / Browserify**. For that you will have to use a pre-built version. You can either [build it yourself](CONTRIBUTING.md#testing-a-built-version) or get a numbered version from http://sinonjs.org.
 
-This might or might not change in future versions, depending of outcome of investigations. Please don't report this as a bug, just use pre-built versions.
+This has been resolved in Sinon v2.x; Please don't report this as a bug.
 
 ## Goals
 
--- a/docs/current/migrating-to-2.0.md
+++ b/docs/current/migrating-to-2.0.md
@@ -1,0 +1,48 @@
+# Migrating to v2.0
+
+Sinon v2.0 is the second major release, we have made several breaking changes in this release as a result of modernising the internals of Sinon.  This guide is intended to walk you through the changes.
+
+## sinon.log and sinon.logError Removed
+`sinon.log` and `sinon.logError` were used in Sinon v1.x to globally configure `FakeServer`, `FakeXMLHttpRequest` and `FakeXDomainRequest`; these three functions now allow the logger to be configured on a per-use basis.  In v1.x you may have written:
+
+```js
+sinon.log = function (msg) { // your logging impl };
+```
+
+You would now individually import and configure the utility upon creation:
+
+```js
+var sinon = require("sinon");
+
+var myFakeServer = sinon.fakeServer.create({
+ logger: function (msg) { // your logging impl }
+});
+```
+
+## sinon.test, sinon.testCase and sinon.config Removed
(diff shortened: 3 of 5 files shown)


mroderick:
Very nice! 👍

fatso83:
Very nice, but we were talking of removing `mock` as well, right? None of us had any interest in maintaining it, knew it well and thought it could be removed in a version 3, so deprecating for v2 could be a start.

P.S. Regarding the deprecated methods should we try to wrap the methods in some kind of deprecation closure that will output a warning? All internal use should by now be removed so that leaves external uses. I was thinking of something like

```javascript
function deprecated(func, name) {
    var callCount = 0;
    var warn = typeof console !== "undefined" && (console.warn || console.log);
    return function() {
        if(callCount++ === 0){
            warn(util.format("[deprecated] The use of sinon.%s has been deprecated. Refer to the migration docs", name));
        }
        func.apply(this, arguments);
    }
}

exports.mock = deprecated( require("./sinon/mock"), "mock") );
```

jonnyreeves:
> ...removing mock as well, right?

If we have concensus on this I'm happy to add it to the notes above.

> ...wrap the methods in some kind of deprecation closure...

Excellent suggestion @fatso83; I'll update #966 to ensure we deprecate those methods as part of v2

mantoni:
We can use a node util function instead of writing our own: https://nodejs.org/api/util.html#util_util_deprecate_function_string

fatso83:
@jonnyreeves : not sure which response you are awaiting :)? We have talked several times about removing `mock` so just add it to the list and merge/squash this in.

jonnyreeves:
All methods called out in the migration guide as deprecated now write a message to the console when they are invoked via the `sinon` API with no message logged when sinon uses them internally.

I have chosen to *not* mark `sinon.mock`as deprecated in this pull as I don't feel we should deprecate methods until we have a migration plan in place; which makes me wonder if we should revisit the decision to remove `sinon.test` and `sinon.testCase` from the 2.0 API...

Please let me know your thoughts @fatso83, @mantoni

jonnyreeves:
So in looking at the failing tests I can see my approach for wrapping the Event constructors is flawed as they are going to be invoked via `new` and therefore can't be applied against null.

What's strange is that the sandbox tests are now failing; I'll have to take a deeper look into that as marking the utils as deprecated should not have had any other side-effects 😖

fatso83:
@jonnyreeves : What do you mean by migration plan? An actual document in `/docs` or just some informal sketch of a plan? For `test` and `testCase` they can just start using the extracted NPM package. It is perhaps inconsistent not to have deprecated those methods as well, but now have been away for a while, so ... 

Re the `sandbox` thing, I suspect this is similar to `Event` (not looked into it, though): you might have interfered with/wrapped some prototype  ...

jonnyreeves:
@fatso83 I mean that we probably shouldn't mark `mock` as deprecated in 2.0 unless we are willing to extract it into its own npm module before 2.0 get published otherwise users are not going to have any way to 'un-deprecate' themselves other than removal all mocks from their codebase (which feels a tad extreme :)

IMHO we should ship 2.0 with `mock`, extract it to a new repo at a later date and then mark `mock` as deprecated (with a migration plan) for `2.1+` - YMMV.

jonnyreeves:
OK tracked down the failure here, 2 parts:

1. `/src/sinon/util/fake-server.js` imports `./core/index.js` directly and is expecting that module's `exports` object to have been mutated by `src/sinon.js` (in particular, `useFakeTimers`.  This is just plain ugly, but was an easy fix - just had to repoint it to import `../../sinon.js` which is what it actually wanted ;)

2. The sandbox tests still crap out, but the good news is that cherry picking #1088 onto this branch magically fixes that issue(!)  So I'll push a change up once that has landed.

3. Emitting a `console.warn` causes the phantomJS tests to fail... I don't know why... could be an issue in mochify.js?  For now I have just downgraded our deprecation message to a `console.info` which works fine 😁 

Just need to patch up deprecating `sinon.Event` and friends and getting closure on how to deal with `sinon.mock`...

mantoni:
@jonnyreeves Yes, I think mochify will fail the test if there is a `console.warn` or `console.error` call. This is actually somewhere downstream in phantomic to figure out what the exit code is. I should figure out a way to ignore this when using a `process.on('exit', ...)` listener, like the one installed by mocaccino. 💡

fatso83:
@jonnyreeves : the migration plan you proposed for `mock` seems fine, I forgot for a minute (month, actually) we actually have minor releases as well :-)

Re the wrap function: why did you end up _not_ using `util.deprecate()`?

mantoni:
@jonnyreeves  I think there needs to be an additional option for phantomic to not exit with 1 if an error was logged ([it is caused here](https://github.com/mantoni/phantomic/blob/master/lib/runner.js#L46)).

lib/sinon/util/core/deprecated.js
```diff
@@ -0,0 +1,24 @@
+/*eslint no-console: 0 */
+"use strict";
+
+// wrap returns a function that will invoke the supplied function and print a deprecation warning to the console each
+// time it is called.
+exports.wrap = function (func, msg) {
+    return function () {
+        // Watch out for IE7 and below! :(
+        if (typeof console !== "undefined") {
+            if (console.warn) {
```
fatso83: Not seeing much trace of the `console.info` instead of `console.warn` thing you talked about for PhantomJS?

jonnyreeves: Ah sorry I have not pushed the changes mentioned above as I was waiting on
the sandbox cjs work to land first.

fatso83:
This needs to be rebased to include #1088, no?

jonnyreeves:
Yep, I'll sort it out this evening / this weekend free time depending :)

jonnyreeves:
> why did you end up not using util.deprecate()?

a) it doesn't copy the `prototype` property, so we still need our own util
b) it writes to `console.error` without first testing for the presence of the console (IE7) and if the console has a func named `error`.
c) `util.deprecate()` will only warn the user once; this implemented warns them each time the deprecated func is invoked (not sure which behaviour is desirable - happy to update our custom impl if we decide to stick with it to only warn once as well).

If you feel strongly I can swith it over to use `util.deprecate()` with this caveats in place (although it does mean we need to fix / update `phantomic` first).

jonnyreeves:
OK; all functions marked as deprecated in the `.md` file now write a message to the console when invoked via the `sinon` object.  

I believe the only outstanding questions are around how the `deprecate` util should work - once we've cleared that up we are good to land this and starting thinking about a cleaner API / internals refactor for sinon v3!!

jonnyreeves:
@fatso83 I was not aware we were dropping support for ES3 (legacy) browsers in 2.x; I'm more than happy to revert this back to using `Object.keys` and `Array.forEach` if you want?

fatso83:
@jonnyreeves: maybe this highlights the fact that we should explicitly define which browser versions we are supporting going forwards? I am quite sure @mroderick said something about not wanting to support < IE10 (or was it IE11?) in some issue, and I replied that I agreed, and that those needing to support older browsers could use the 1.x branch. I was unable to find that discussion right now using the issue tracker search engine right now. I really don't think we should support anything not supporting ES5 going forwards. Preferably, I'd drop everything pre IE11 as that means we could drop `sinon-ie` in `master` and stop having issues like #1091.

Dropping browser support should be part of a migration document, I suppose.

fatso83:
One hint to browser support in `master` is that our [Sauce Labs tests](https://github.com/sinonjs/sinon/blob/master/.min-wd) only cover IE 10 and IE 11.

jonnyreeves:
Good call @fatso83; I've updated the README to add both a Saucelabs build status badge and a Saucelabs Browser Matrix badge - you can preview this [here](https://github.com/jonnyreeves/Sinon.JS/blob/5770acd8ae02c2ecb4c5ab844f95e91c0393c8cf/README.md).

Looks like the status is not being pulled through at the moment, we may have to tinker.  According to [the saucelabs guide](https://wiki.saucelabs.com/display/DOCS/Using+Status+Badges+and+the+Browser+Matrix+Widget+to+Monitor+Test+Results#UsingStatusBadgesandtheBrowserMatrixWidgettoMonitorTestResults-SettingUpStatusBadgesforTestResults), we need to:

1. Add the badge (done)
2. Run your tests.
3. Make sure to set a build number, a pass/fail status, and a public, share, or public restricted visibility for every test that runs using the Test Configuration Options.
4. You'll know that these are set correctly if your tests have a status of Pass or Failed instead of Finished on your dashboard, and that a build number is also displayed.

I don't have a login for saucelabs - would you or @mroderick be able to take a look for me please?

fatso83:
@jonnyreeves: In my car, on my way home from vacation, so I just sent you the credentials by email.

jonnyreeves:
Cheers @fatso83; added I've pushed a commit to master which exposes the `TRAVIS_BUILD_ID` to SauceLabs which should in theory make the badges work; they're still looking broken to me - but I'll see if it magically fixes itself over night...

fatso83:
Sure it's been pushed? Last commit was 4 hours ago.

mantoni:
@jonnyreeves There was [a pull request on min-webdriver](https://github.com/mantoni/min-webdriver/pull/7/files) that should have made this work. The property it's using is `TRAVIS_BUILD_NUMBER` though while you mentioned `TRAVIS_BUILD_ID`. Maybe this should be improved?

mantoni:
@jonnyreeves Also please message me / private twitter so I can send you our SauceLabs account.

jonnyreeves:
No idea why the badges aren't showing in the README preview, but if you visit the links directly you will see them in all their glory!

- https://saucelabs.com/browser-matrix/sinonjs.svg
- https://saucelabs.com/buildstatus/sinonjs

I think that's everything for this particular PR; @fatso83 @mantoni - let me know if you think there's anything else that needs to be addressed?

mantoni:
Nice! But your empty commit is not actually empty. I would also prefer to rebase this commit away before merging. As a Sinon member you should be able to trigger rebuilds in Travis.

mantoni:
@jonnyreeves I mean, the content of the commit seems fine, just the commit name makes no sense 😄

fatso83:
@jonnyreeves : the badges _are_ [showing in the README](https://github.com/jonnyreeves/Sinon.JS/blob/66fe46d67d30222a471d461886471e7e6d74b693/README.md), btw :)

LGTM, too, btw, so just remove the last commit and merge it at will.

jonnyreeves:
Whoops, sorry about that last commit; accidentally look a non-ruby Publish script I was hacking around on along for the ride! 💃 

All commits squashed; could someone give it a quick once over and then make it so, please?

fatso83:
Boom.

mantoni:
👍 @fatso83 You’ve been 5 seconds faster than me 🚀

sephie:
The migration guide is clear about the use of sinon-test. However, The docs for sinon-test do not explain what `testCase()` actually is intended for, nor can I find any info in any version of the sinon docs (checked the 1.17 docs too)

This gist gives me my best guess: something about using `sinon-test` from a `describe()`
https://gist.github.com/jclevine/a603c0814fbe12649ac9

If someone could explain the exact function I'd be happy to write some docs over at the sinon-test repo.

fatso83:
@sephie You are totally correct about that omission, and one I was pondering last Friday :smile: I found the previous description for you [on the WayBack Machine](https://web.archive.org/web/20150205020347/http://sinonjs.org/docs/#sinon-testCase). AFAIK, you basically use it as a wrapper around the `describe` block instead of the `it` blocks and it supports typical pre/post setup.

sephie:
That's another way to do it. 😆 PR is up over at sinon-test

Issue #1093, opened 2016-07-11
calledOnce should fail given extra arguments (to reduce accidental belief of calledWith behavior)


ClintFMullins:
Not really a bug, more of an issue that causes us to create bugs.
* Sinon version : latest
* Environment   : running through node with karma test runner inside sinon-chai
* Example URL   :  -
* Other libraries you are using: - 

**What did you expect to happen?**

We've had the following accident happen many times over the past couple years in our codebase:
`assert.calledOnce(someStub, wrongArg1) // will pass when given wrong args.`

Proposal:
`assert.calledOnce()` should not accept extra arguments, and should fail when they are given like so:
'calledOnce only accepts one argument, did you mean to use calledWith?'

ideally we'd also have a
`calledOnceWith`
`calledTwiceWith`
etc

with the behavior of asserting the stub is called X many times, with the given arguments.

**What actually happens**
`calledOnce(someStub, wrongArg1, wrongArg2)` silently fails given wrong arguments, as currently expected from sinon. 

**How to reproduce**

assert.calledOnce(someStub, wrongArg1) // will pass when given wrong args.

<details>
   none necessary
</details>

I would be happy to add an asserting function that would get called inside function mirrorPropAsAssertion. This function would go to some case statement keying off the method name and given arguments. If you are interested in a PR let me know. If you have a better idea of how to fix (if you agree this is broken) feel free to let me know or fix yourself. I'd like to contribute though :)

ClintFMullins:
`function mirrorPropAsAssertion(name, method, message) {
    if (arguments.length === 2) {
        message = method;
        method = name;
    }

    assert[name] = function (fake) {
        verifyIsStub(fake);

        var args = slice.call(arguments, 1);
        var failed = false;

        if (typeof method === "function") {
            failed = !method(fake);
        } else {
            failed = typeof fake[method] === "function" ?
                !fake[method].apply(fake, args) : !fake[method];
        }

        if (failed) {
            failAssertion(this, (fake.printf || fake.proxy.printf).apply(fake, [message].concat(args)));
        } else {
            assert.pass(name);
        }
    };
}`

fatso83:
Just to clear up some confusion, I think you have mixed up fail and pass in your descriptions.

You *expected* this:
> assert.calledOnce(someStub, wrongArg1) // will pass when given wrong args.

This is what _is_ happening, so nothing wrong.

What actually happens (_according to you_):
[...]

PR #1097, opened 2016-07-12
Add assertion check for too many args with calledOnce/Twice/Thrice

ClintFMullins:
Add assertion to make sure calledOnce, calledTwice, calledThrice don't get called with any extra arguments. This will prevent the incorrect assumption that calledOnce works like calledWith

Incorrect case that will be prevented:
calledOnce(stub, arg1, agr2); // dev thinks we are asserting with arguments, but this just checks that the stub was called

https://github.com/sinonjs/sinon/issues/1093

This works by adding a check to verify the arguments of an assertion. This goes to a function that could be extended to include other checks for different cases, but for now just includes the single case for the three assertions outlined above. 

2. `npm install`
3. `npm test`

running the tests should tell you all you need to know here.

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -43,6 +43,23 @@ function verifyIsStub() {
     }
 }
 
+function verifyIsValidAssertion(assertionMethod, assertionArgs) {
+    switch (assertionMethod) {
+        case "notCalled":
+        case "called":
+        case "calledOnce":
+        case "calledTwice":
+        case "calledThrice":
+            if (assertionArgs.length !== 0) {
+                assert.fail(assertionMethod +
+                            " takes 1 argument but was called with " + (assertionArgs.length + 1) + " arguments");
+            }
+            break;
+        default:
+            break;
+    }
+}
+
 function failAssertion(object, msg) {
     object = object || global;
     var failMethod = object.fail || assert.fail;
@@ -60,6 +77,8 @@ function mirrorPropAsAssertion(name, method, message) {
 
         var args = slice.call(arguments, 1);
         var failed = false;
+
+        verifyIsValidAssertion(name, args);
 
         if (typeof method === "function") {
             failed = !method(fake);
--- a/test/assert-test.js
+++ b/test/assert-test.js
@@ -131,6 +131,15 @@ describe("assert", function () {
             assert(sinonAssert.fail.called);
         });
 
+        it("fails when called with more than one argument", function () {
+            var stub = this.stub;
+            stub();
+
+            assert.exception(function () {
+                sinonAssert.called(stub, 1);
+            });
+        });
+
         it("does not fail when method was called", function () {
             var stub = this.stub;
             stub();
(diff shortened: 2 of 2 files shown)


test/assert-test.js
```diff
@@ -293,6 +304,18 @@ describe("assert", function () {
...
+        it("fails when called with more than one argument", function () {
+            var stub = this.stub;
+            this.stub();
+            this.stub();
+
+            assert.exception(function () {
+                sinonAssert.calledTwice(stub, 1);
+            });
+
+            assert(sinonAssert.fail.called);
```
fatso83: Not sure this is needed?

test/assert-test.js
```diff
@@ -331,6 +354,19 @@ describe("assert", function () {
...
+            var stub = this.stub;
+            this.stub();
+            this.stub();
+            this.stub();
+
+            assert.exception(function () {
+                sinonAssert.calledThrice(stub, 1);
+            });
+
+            assert(sinonAssert.fail.called);
```
fatso83: Ditto - wouldn't the first assertion be enough?

fatso83:
Not sure about the `assert(sinonAssert.fail.called);`, but otherwise this seems fine.

fatso83:
LGTM :)

lib/sinon/assert.js
```diff
@@ -43,6 +43,21 @@ function verifyIsStub() {
     }
 }
 
+function verifyIsValidAssertion(assertionMethod, assertionArgs) {
+    switch (assertionMethod) {
+        case "calledOnce":
```
TheSavior: Should this include `notCalled` and `called` as well? http://sinonjs.org/docs/#assert-pass

ClintFMullins: @TheSavior (love the name, btw). Yup that makes sense to me, I'll add.

ClintFMullins:
Okay, added notCalled/called which also revealed I was using a variable that was sometimes the method_name and other times the method (woops). Now we are always using the name in the switch statement.

ClintFMullins:
@fatso83 First of all, thanks for reviewing and replying - really appreciate it! Second, I'd like to get this merged soon so our codebase enforces these new assertions. Are there further steps I can / need to take in order to get these changes merged?

fatso83:
Until a new pre-2 release out you can just depend directly on the `master` branch. Just add `"sinon": "sinonjs/sinon#master" in your `package.json`.

ClintFMullins:
@fatso83 Thanks for the quick turn around on that.

ClintFMullins:
hey @fatso83, we are currently using the latest 1.x.x version of sinon and it'd be nice to have these changes in there. Is that a possibility? If that could happen, do you have a time estimate of when that release would come? Or do you know when you guys are thinking about releasing the non-pre 2.x.x?

fatso83:
OK, then re-submit it as a PR against the v1.17 branch. I don't have any estimate on when a 1.17.5 version will be out as we surprised a few people with some network changes in 1.17.4, and what to do with them has yet to be decided. I will be off-grid in the mountains for a couple of weeks, so someone else needs to handle this.

Issue #1046, opened 2016-05-16
localStorage stub doesn't work on 1.17.3


ilgooz:
in version 1.12.2 everything is good on Chrome http://jsbin.com/noqigilome/edit?html,js,console,output
but on 1.17.3 it doesn't seem to work http://jsbin.com/fufoyomuja/edit?html,js,output

ubuntu 14.10 - Chrome 48.0.2564.109 (64-bit)

thank you.

```
"TypeError: Attempted to wrap string property getItem as function
    at checkWrappedMethod (http://sinonjs.org/releases/sinon-1.17.3.js:1355:29)
    at Object.wrapMethod (http://sinonjs.org/releases/sinon-1.17.3.js:1398:21)
    at Object.stub (http://sinonjs.org/releases/sinon-1.17.3.js:3465:26)
    at qulovapigu.js:1:12
    at https://static.jsbin.com/js/prod/runner-3.35.12.min.js:1:13891
    at https://static.jsbin.com/js/prod/runner-3.35.12.min.js:1:10820"
```

fatso83:
Thanks for reporting this. Seems to be a bug, but not sure what causes it after inspecting the source.

Adding output to your report.

fatso83:
@ilgooz Would you mind trying [my approach](https://github.com/sinonjs/sinon/issues/821#issuecomment-139280107) of finding where this regression was introduced? You just need to modify the [test script](https://raw.githubusercontent.com/fatso83/Sinon.JS/setImmediate-bug/test-script.sh) a little bit to make git do all the work for you in finding where this bug was introduced.

It is a quite fun exercise and you could reuse it in other projects :)

jonnyreeves:
@ilgooz please could you provide an example which reproduces this issue?  Thanks.

fatso83:
@jonnyreeves he has :-) He just used JSBin to make it runnable in the browser, as a simple Node test would not show it. I verified it. 

I can perhaps try to follow my own instructions above to find where it was introduced.

jonnyreeves:
@fatso83 @ilgooz yep sorry, saw the jsbin example just after commenting!  I've since deleted my comment, I'll delete your and this one as well to keep the log for this issue on topic - sorry for the spam.

fatso83:
[...]

PR #1098, opened 2016-07-12
Fix regression in #692 when stubbing browser natives

fatso83:
#### Purpose
Fix regression in #692 by falling back to simple property assignment.

#### Background
Spying and stubbing of localStorage in Chrome started failing when a regression was introduced in #692. This issue was reported in #1046.

Spying on the methods of the localStorage object failed to work in Chrome when Sinon started using `Object.defineProperty`. Instead, the effect was to add a key with the name of the method
(say `"getItem"`) to the disk storage, instead of on the actual `localStorage` object.

This seems to only affect Chrome.

#### Solution 
Falls back to the ES3 simple assignment method when the assignment fails.

#### How to verify
New tests were added, so just run `npm test`.

You can also verify that it works through this external script used to find the regression:
```sh
# first checkout this branch ... of course
curl -o test-script.sh https://rawgit.com/fatso83/7f135d2de168a07582b075b0fa46e9a9/raw/ceedbd7694c93aef52e146bc9701125f67e34730/test-script.sh
./test-script.sh
```

### DIFF
--- a/lib/sinon/util/core.js
+++ b/lib/sinon/util/core.js
@@ -94,6 +94,13 @@
 
             var error, wrappedMethod, i;
 
+            function simplePropertyAssignment() {
+                wrappedMethod = object[property];
+                checkWrappedMethod(wrappedMethod);
+                object[property] = method;
+                method.displayName = property;
+            }
+
             // IE 8 does not support hasOwnProperty on the window object and Firefox has a problem
             // when using hasOwn.call on objects from other frames.
             var owned = object.hasOwnProperty ? object.hasOwnProperty(property) : hasOwn.call(object, property);
@@ -126,11 +133,17 @@
                     mirrorProperties(methodDesc[types[i]], wrappedMethodDesc[types[i]]);
                 }
                 Object.defineProperty(object, property, methodDesc);
+
+                // catch failing assignment
+                // this is the converse of the check in `.restore` below
+                if ( typeof method === "function" && object[property] !== method ) {
+                    // correct any wrongdoings caused by the defineProperty call above,
+                    // such as adding new items (if object was a Storage object)
+                    delete object[property];
+                    simplePropertyAssignment();
+                }
             } else {
-                wrappedMethod = object[property];
-                checkWrappedMethod(wrappedMethod);
-                object[property] = method;
-                method.displayName = property;
+                simplePropertyAssignment();
             }
 
             method.displayName = property;


fatso83:
Not totally sure about the new test I added. Is it in the right place and with the right name? And any objections to the fix?

Of course this also needs to be merged into `master` ...

test/spy-test.js
```diff
@@ -240,7 +240,26 @@
             }
         },
 
-        call: {
+        // test for regression introduced in #692. See #1046 for details
+        "localStorage support": {
```
fatso83: Maybe this name is a bad idea? It basically says that we support stubbing localStorage, which is impossible in some browsers (like Firefox). I would still like to document the fix, though ...

test/spy-test.js
```diff
@@ -240,7 +240,26 @@
...
+
+            "can spy on localStorage": function () {
+                var spy = sinon.spy(localStorage, "getItem");
+
+                localStorage.clear();
+                localStorage.setItem("foo", "fooval");
+                localStorage.getItem("foo");
+
+                assert(typeof localStorage.getItem.restore, "function" );
+                assert(spy.firstCall.returnValue === "fooval");
```
fearphage: I believe both of these assertions should be `assert.equals`.

test/spy-test.js
```diff
@@ -240,7 +240,26 @@
...
+            }
+        },
+
+        "call": {
```
fearphage: Why'd you add quotes?

test/spy-test.js
fatso83: Probably, right, but I am actually just trying out these changes, and it turned out that it (as assumed) breaks in Firefox (as stubbing localStorage cannot be done), so both the test and the heading is wrong. 

Do you have any idea how I can make a decent test that will check my changes? While this does revert the problems introduced in #692 it certainly has little to do with localStorage explicitly - that was just a side effect. But perhaps this is one area where unit tests add little value, as we are crossing the boundary over to browser internals, where browser specifics come into play. So perhaps I should remove the test?

test/spy-test.js
fatso83: Consistency. All other test names have them, so while it is technically fine, I don't see why it should be treated specially.

fatso83:
@fearphage : for the reasons mentioned in the comments on the test file, I decided to remove the additional test altogether. A pure unit test for this seemed impossible, and the other logic dealing with this did not have such either. So I am merging it in without any changes to the test code.

PR #1099, opened 2016-07-12
Revert #1017 fix-fake-xhr-compliance-again"

fatso83:
Revert the networking changes introduced in #1017 for the v1.17 branch

#### Background 
The changes in the networking api introduced in #1017 proved to be too big of a change for version 1.17, and was decided to be kept only for the upcoming version 2 in the master branch to lower the noise and not introduce major breaking changes, even though the implementation predating this was incorrect per the spec.

See [this discussion](https://github.com/sinonjs/sinon/issues/1040#issuecomment-231727271) for more background info.

#### How to verify
Hmm ... well the tests still pass, so ...

### DIFF
--- a/lib/sinon/util/event.js
+++ b/lib/sinon/util/event.js
@@ -41,8 +41,8 @@ if (typeof sinon === "undefined") {
 
         sinon.ProgressEvent = function ProgressEvent(type, progressEventRaw, target) {
             this.initEvent(type, false, false, target);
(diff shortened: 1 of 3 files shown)


fatso83:
@fearphage, are you ok with this? Seeing that the 1.17.4 release broke a lot of stuff, I feel that we could rather keep the current implementation and just have the breaking (but correct) changes in v2.

fearphage:
I fixed the code to mirror master (in #1102) instead of reverting. Reopen as needed.

PR #1101, opened 2016-07-14
moving assertion from 2x to 1x, resolving conflicts

ClintFMullins:
same as: https://github.com/sinonjs/sinon/pull/1097

but for a 1.x

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -45,6 +45,24 @@
             }
         }
 
+        function verifyIsValidAssertion(assertionMethod, assertionArgs) {
+            switch (assertionMethod) {
+                case "notCalled":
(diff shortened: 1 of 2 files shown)


lib/sinon/assert.js
```diff
@@ -12,6 +12,26 @@
  *
  * Copyright (c) 2010-2013 Christian Johansen
  */
+
+
+//  function verifyIsValidAssertion(assertionMethod, assertionArgs) {
+//     switch (assertionMethod) {
```
fatso83: This should go. Guessing it's not intentional

jonnyreeves:
LGTM; once we resolve #1040 we can pull a 1.17 release together.

PR #1102, opened 2016-07-16
Stopped treating non-200 responses as error - fixes #1040

### LINKED ISSUE #1040: fakeServer is broken with .respond(500) on 1.17.4

fearphage:
* non-200 requests trigger `load` event (fixes regression)
* added `xhr.error()` to simulate network errors

Tests and fixes are ported from `master`.

### DIFF
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -64,6 +64,58 @@
         assert.equals(event.lengthComputable, !!progress);
     };
 
+    var assertEventOrdering = function (event, progress, callback) {
+        return function (done) {
+            var expectedOrder = [
+                "upload:progress",
+                "upload:" + event,
+                "upload:loadend",
+                "xhr:progress",
+                "xhr:on" + event,
+                "xhr:" + event
+            ];
+            var eventOrder = [];
+
+            function observe(name) {
+                return function (e) {
+                    assertProgressEvent(e, progress);
+                    eventOrder.push(name);
+                };
+            }
+
+            this.xhr.open("GET", "/");
+            this.xhr.send();
+
+            this.xhr.upload.addEventListener("progress", observe("upload:progress"));
+            this.xhr.upload.addEventListener("loadend", observe("upload:loadend"));
+            this.xhr.addEventListener("progress", observe("xhr:progress"));
+            this.xhr.addEventListener("loadend", function (e) {
+                assertProgressEvent(e, progress);
+
+                // finish next tick to allow any events that might fire
+                // after loadend to trigger
+                setTimeout(function () {
+                    assert.equals(eventOrder, expectedOrder);
+
+                    done();
+                }, 1);
+            });
+
+            // listen for abort, error, and load events to make sure only
+            // the expected events fire
+            ["abort", "error", "load"].forEach(
+                function (name) {
+                    this.xhr.upload.addEventListener(name, observe("upload:" + name));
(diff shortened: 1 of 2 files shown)


fearphage:
This fixes #1040.

fatso83:
👍 The code LGTM, on a casual (by-mobile) look :-) Re the actual implementation, I won't revisit the W3 spec, as I think the previous discussions has covered this in detail, so I'll just take your word for it doing what it should 😌

fearphage:
It more closely emulates what browsers do now based on these simplified test cases.

http://people.mozilla.org/~mbanner2/sinonXHRBrowserTest.html

fearphage:
This introduced a new bug (regression) that already exists in `master`.

Now progress events from non-200 responses don't have the right properties - `loaded`, `total`, `lengthComputable`.

I've filed it.  #1105

fearphage:
I think this should ship broken (just like master). I'm working on a fix though.

Thoughts? 

Also as I am new here, do I need to wait for a few people or anyone in particular to weigh in on this fix before merging/releasing?

test/util/fake-xml-http-request-test.js
```diff
@@ -1797,7 +1887,7 @@
 
             "triggers 'loadend' event at the end": function (done) {
                 this.xhr.addEventListener("loadend", function (e) {
-                    assertProgressEvent(e, 0);
```
fearphage: This test was correct and had to change based on the regression.

test/util/fake-xml-http-request-test.js
```diff
@@ -1821,7 +1911,7 @@
 
             "calls #onloadend at the end": function (done) {
                 this.xhr.onloadend = function (e) {
-                    assertProgressEvent(e, 0);
```
fearphage: This test was correct and had to change based on the regression.

fearphage:
I'm reconsidering now. Maybe we shouldn't exchange one regression for another one. I'm torn.

I have a possible fix (fearphage/Sinon.JS#1 -- ignore the repo and all), but I think maybe there's a middle ground there too.

fatso83:
@fearphage Regarding merging: it depends. We haven't created any hard and fast rules. For minor fixes, we sometimes submit a PR (just to notify others), and then merge it ourselves. For bigger stuff, we usually wait for some input from the others.

Since you already have a nice fix, I don't see why we couldn't merge both this and your fix before releasing 1.17.5? @mantoni?

fatso83:
@fearphage Merged. Care to make a PR with fearphage/Sinon.JS#1? I think it should be included in a 1.17.5 release. I could also do it of course, if you are unavailable, since I have a few days vacation that's not off-grid :-)

PR #1108, opened 2016-07-24
V1.17 progress event compliance - fixed #1105

### LINKED ISSUE #1105: Progress events from non-200 responses should have `e.loaded = 0`
Non-200 requests now trigger the `load` event now (in `master` and maybe soon in `v1.17.*` - #1102). However the progress events triggered have several incorrect properties -- `loaded`, `total`, and `lengthComputable`. Here is an explanation in code:

## Expected progress event from non-200 response

```js
var progressEvent = {
  loaded: 0,
  total: 0,
  lengthComputable: false
};
```

## Actual progress event from non-200 response

```js
var progressEvent = {
  loaded: 100,
  total: 100,
  lengthComputable: true
};
```  

I spotted the bug here:

https://dl.dropboxusercontent.com/u/2400/tc/sinon/xhr-browser-test.htm

We've got the events correct now but not the properties of the triggered event.

Related: #1040

fearphage:
This changes the data that progress events send to more closely mimic browsers. This fix affects the state and properties of the `ProgressEvent` fired in ready state 4 (done).

### Change Affects the following properties for ProgressEvent

* loaded
* total
* lengthComputable

### State Before this Branch

* `progressEvents` fired with only two progress `loaded`/`total` values -- `0` or `100`
* All `error` and `abort` events triggered a `0` progress event
* All `load` events triggered a `100` progress event

### Changes in this Branch

* `load` events can fire with `0` progress events (no body was returned)
* `error` events can fire with non-0 progress events (error with `lengthComputable`)
* All events can fire with any size body (0+) -- not limited to `0` or `100` like before

This is a major change I believe. This fixes #1105.

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -459,12 +459,20 @@
                 }
 
                 if (this.readyState === FakeXMLHttpRequest.DONE) {
-                    if (this.aborted || this.status === 0) {
-                        progress = {loaded: 0, total: 0};
-                        event = this.aborted ? "abort" : "error";
+                    // ensure loaded and total are numbers
+                    progress = {
+                      loaded: this.progress || 0,
+                      total: this.progress || 0
+                    };
+
+                    // order of checking is important here
+                    if (this.aborted) {
+                        event = "abort";
+                    }
+                    else if (this.status === 0) {
+                        event = "error";
                     }
                     else {
-                        progress = {loaded: 100, total: 100};
                         event = "load";
                     }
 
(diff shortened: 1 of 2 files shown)


fearphage:
I used this to spot and attempt to fix the error:

https://dl.dropboxusercontent.com/u/2400/tc/sinon/xhr-browser-test.htm

@fatso83 Here's the PR.

fatso83:
Great!

fearphage:
You're quick! I was going to revert a part of that.

fatso83:
ah, sorry, but there is no functional change in the revert, is there? I find the explicit if/else easier to read, anyhow.

fearphage:
I'm fine either way. It's not pressing, but I think it makes the code a little harder to break.

fatso83:
Now that v1.17 is fine, should we try and include these fixes (#1108 and #1109) into `master` as well? No rush, but just to get feature parity between the two. I can do it if you are busy with real work :-)

PR #1109, opened 2016-07-24
Simplifies code (reverts part of #1108)

fearphage:
This just reverts a small part of the code/logic introduced in #1108. It makes the code a little harder to break (less order dependent).

@fatso83 Not pressing, but here it is.

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -465,12 +465,8 @@
                       total: this.progress || 0
                     };
 
-                    // order of checking is important here
-                    if (this.aborted) {
-                        event = "abort";
-                    }
-                    else if (this.status === 0) {
-                        event = "error";
+                    if (this.aborted || this.status === 0) {
+                        event = this.aborted ? "abort" : "error";
                     }
                     else {
                         event = "load";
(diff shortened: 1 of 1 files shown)


fatso83:
Huh, I tried reading that revert twice, and I did not see that it was not exactly the same. I guess I need to look a third time to spot the difference ...

fatso83:
On third reading I still cannot see how revert is not exactly the same. Can you give an example input that will be processed differently?

fearphage:
It doesn't change the code or the way it works. It just makes it less order-dependent. There is no code that would execute differently.

fatso83:
I might be a tad bit slower at nighttime, but I don't get how clumping conditions into one statement is better/less fragile than splitting them into two explicit checks. I tried googling "order-dependent" to see if there was an obvious point I did not get - to no avail. 

The code to be reverted
- reads better
- does the same

How can one version be more (or less) order-dependent than the other? Is there an overlapping case (for instance `this.aborted && this.status === 0`) that you somehow feels is better handled in the combined checking?

Sorry for expanding on this super-small change, but I did _try_ contacting you in a hangout :-) So far it seems the change is purely cosmetic.

fatso83:
Thanks, Phred!

PR #1110, opened 2016-07-25
Add prepublish steps to avoid including extra files

fatso83:
Avoid including files that are not in the index by cleaning the build directory and refusing to publish to NPM when `git status` detects any changes. This should avoid release problems of the past where older built versions sometimes were bundled along new versions by mistake when publishing to NPM.

1. `touch f`
2. `npm run prepublish #should fail`
3. `rm f`
4. `npm run prepublish #should execute without error`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,12 @@
 language: node_js
 node_js:
-  - "0.8"
-  - "0.10"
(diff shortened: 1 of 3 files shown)


fatso83:
🚢 Ready for 1.17.5

PR #1114, opened 2016-07-28
ESLint upgrade to version 3

fatso83:
#### Purpose
Updates ESLint to current major version

Had to change code in a few places to avoid failing the `consistent-return` rule. Turned out we deviated a tiny bit from W3C spec on EventTarget#removeEventListener by returning the listeners instead of void. Disabled the rule in the two cases it was intentional. Added explicit returns elsewhere.


mantoni:
:+1:

PR #1116, opened 2016-08-05
CJSify fake-timer tests

jonnyreeves:
🚀

### DIFF
--- a/test/util/fake-timers-test.js
+++ b/test/util/fake-timers-test.js
@@ -1,19 +1,22 @@
 "use strict";
 
 var referee = require("referee");
-var sinon = require("../../lib/sinon");
+var fakeTimers = require("../../lib/sinon/util/fake_timers");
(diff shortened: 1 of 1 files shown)


mantoni:
Wrestling them down, one by one :facepunch: 
Very nice!

jonnyreeves:
aye, slow going - but only a handful to do now - 2.0 is almost there (although I'm slightly concerned that we are going to get blocked by the docs website?)

fatso83:
@jonnyreeves : I don't think the website should hold us back. We'll release 2.0 when 2.0 is finished, not when the revamped site is finished. That being said, me and Roderick had a appear.in session talking about this a week ago, and we have gotten the basic site generation up, so things are rolling. I'll be adding some articles as well so we can ship it with some new content too.

jonnyreeves:
@mantoni good to land?

mantoni:
:+1:

PR #1118, opened 2016-08-08
Feature/cjsify fake server tests

jonnyreeves:
Another straight forward port.

`fake-server-test.js` is still pulling in the `sinon` module as it tests that the global `useFakeXMLHttpRequest` method is invoked upon construction of the fakeServer instance.  I'll add a note in #966 as it will result in an API change.

### DIFF
--- a/test/util/fake-server-test.js
+++ b/test/util/fake-server-test.js
@@ -2,15 +2,24 @@
 
 var referee = require("referee");
 var sinon = require("../../lib/sinon");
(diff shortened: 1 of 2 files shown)


fatso83:
@jonnyreeves, one of the few last bits remaining for Sinon 2.0 is "Migrate fake_server and friends", where you mention "useFakeXMLHttpRequest still referenced"

> as it will result in an API change.

I am not quite sure which API will change? I would like to give this a go, but unsure what you are referring to here.

PR #1123, opened 2016-08-08
Fixing a Markdown syntax error

JamesMGreene:
Fixing a Markdown syntax error

The Markdown syntax error results in incorrect rendering

#### Solution  - optional
Fix the syntax error in the Markdown.

View this rendered Markdown file to ensure that the last line is rendered as code.


fatso83:
Thanks!

PR #1128, opened 2016-08-11
sinon.test: restore sandbox when promise is fulfilled/rejected

### LINKED ISSUE #1119: sinon.test restores the sandbox before a promise-based async test is completed
- Sinon version : 1.17.5
- Environment   : node.js

**What actually happens**

**How to reproduce**
Run this snippet in mocha: 

``` js
    var sandbox = this;
    return Promise.resolve()
        .then(function () {
        });
}));
```

joelmukuthu:
#### Purpose (TL;DR)
Fixes #1119 

1. Check out this branch
2. `npm install`
3. `npm test`

#### TODOs
- [ ] Apply fix to sinon v1.x
- [ ] Port fix to [sinon-test](https://github.com/sinonjs/sinon-test)

### DIFF
--- a/lib/sinon/util/core.js
+++ b/lib/sinon/util/core.js
@@ -94,6 +94,13 @@
 
             var error, wrappedMethod, i;
 
+            function simplePropertyAssignment() {
+                wrappedMethod = object[property];
+                checkWrappedMethod(wrappedMethod);
+                object[property] = method;
+                method.displayName = property;
+            }
+
             // IE 8 does not support hasOwnProperty on the window object and Firefox has a problem
             // when using hasOwn.call on objects from other frames.
             var owned = object.hasOwnProperty ? object.hasOwnProperty(property) : hasOwn.call(object, property);
@@ -126,11 +133,17 @@
                     mirrorProperties(methodDesc[types[i]], wrappedMethodDesc[types[i]]);
                 }
                 Object.defineProperty(object, property, methodDesc);
+
+                // catch failing assignment
+                // this is the converse of the check in `.restore` below
+                if ( typeof method === "function" && object[property] !== method ) {
+                    // correct any wrongdoings caused by the defineProperty call above,
+                    // such as adding new items (if object was a Storage object)
+                    delete object[property];
+                    simplePropertyAssignment();
+                }
             } else {
-                wrappedMethod = object[property];
-                checkWrappedMethod(wrappedMethod);
-                object[property] = method;
-                method.displayName = property;
+                simplePropertyAssignment();
             }
 
             method.displayName = property;


joelmukuthu:
@fatso83 PTAL. I'm having trouble running the packaged tests due to this error `ReferenceError: Can't find variable: require` caused by https://github.com/sinonjs/sinon/pull/1128/files#diff-3053afa1330e8ab359155a638bbc23cbR9. I assume it should be an easy fix but I couldn't figure out how to fix it, could you point me towards the right direction?

joelmukuthu:
I fixed the issues with making the `Promise` global available in browser environments for buster. The failing build seems to be something related to phantomjs. All tests tests ran in development with both node v0.12.15 and v4.4.7.

lib/sinon/test.js
```diff
@@ -55,6 +55,23 @@
                     sandbox.verifyAndRestore();
                 }
 
+                if (result) {
+                    if (typeof Promise === "function") {
```
fatso83: don't see the point in this check. we don't really care if the global environment supports Promises: we care about the actual returned object. So just use the thenable check on line 62 and drop this. This could fail too:
Say, a promise was returned that was not native (pinkySwear, for instance). Then this test would say `isPromise == false`. That's wrong. So just focus on if it is a Thenable or not.

package.json
```diff
@@ -33,6 +33,7 @@
     "samsam": "1.1.2"
   },
   "devDependencies": {
+    "bluebird": "3.4.1",
```
fatso83: No points in adding this to the environment. We have checks to see if the environment supports Promises and if not the tests will not execute.

test/test-test.js
```diff
@@ -233,6 +233,63 @@
             }).call({}, "arg1", {}, done);
         },
 
+        "async promise-based test restores the sandbox when the promise is fulfilled": function (done) {
```
fatso83: Prefix this section with a guard clause a la what you can see done [here](https://github.com/joelmukuthu/sinon/blob/2047100d5b26155c9879ac14e2d3c765e8700e57/test/sandbox-test.js#L10) and [here](https://github.com/joelmukuthu/sinon/blob/2047100d5b26155c9879ac14e2d3c765e8700e57/test/sandbox-test.js#L100).

fatso83:
I'll have to try and look into the PhantomJS thing to see what's going on there.

joelmukuthu:
Thanks for the review @fatso83! I reverted the commits in question but alternatively I could prune them out and force push if you'd rather keep a clean history :)

fatso83:
Ha, seems like whatever you did did the trick for PhantomJS as well! I guess there was some reference to `Promise` in there that failed (or something like that). It would be great if you could do just what you proposed: do an interactive rebase to clean up the log. We try to keep a clean history around here :neckbeard:

joelmukuthu:
All cleaned up. Crossing fingers for successful build :)

test/test-test.js
```diff
@@ -233,6 +233,69 @@
...
+                var spy = sinon.spy();
+                sinon.test(function (callback) {
+                    return Promise.reject(new Error("dummy"))
+                        .catch(function (err) {
+                            spy();
+                            callback(err);
+                        });
+                }).call({}, function (err) {
+                    assert.equals(spy.called, true);
+                    assert.equals(err instanceof Error, true);
```
fearphage: Assert that it's the same error instead of similar to the originally thrown error.

fatso83: @joelmukuthu: yeah, did not see that myself. just create the error in the outer scope and throw it before comparing. promise I'll merge it after that. This is otherwise good to go.

joelmukuthu: Yeah, good catch.

lib/sinon/test.js
```diff
@@ -55,6 +55,15 @@
                     sandbox.verifyAndRestore();
                 }
 
+                if (result && typeof result.then === "function" && typeof result.catch === "function") {
```
fearphage: If you simplify this, it will work with more Promise libraries like jQuery (before v3). If you only use `.then(onResolve, onReject)`. I do realize that old jQuery was not Promises/A+ compliant. It's just a thought and almost functionally the same.

The way this is currently written, it will actually catch the failure if something were to throw in the `onResolve` of the `then` which seems unnecessary. 

I apologize for the string of comments instead of all at once.

fearphage: This change will only make our Promise duck-typing more permissive. It also conforms to the Promise/A+ checking for promises. 

https://promisesaplus.com/

fearphage: It also feels like this should be an `else-if` of the previous `if` statement. Looking at the code now, a synchronous test that returns a promise will have restore called multiple times.

I'd add new tests to ensure that that is not the case.


If you move your if statement before the `typeof OldDone` check, you can actually make it so returning a promise is all that is necessary to make tests async instead of return a promise and taking a callback.

These are just my thoughts from looking at the code. I looked at issue this is fixing #1119 and that sample code looks synchronous but returns a promise. This doesn't yet fix that case. You now have a weird case where you're double-y saying it's synchronous.

Imagine this code:

```js
sinon.test(function(callback) {
  callback();
  // do lots of work here
  return Promise.reject(new Error);
});
```

I feel like we're enabling a bad situation the more I look at it. Mocha just [fixed this issue](/mochajs/mocha/blob/master/CHANGELOG.md#boom-breaking-changes) (see #4 on the list). I believe we should enforce the same restrictions. 

The end user's test should fall into one of these categories: 

* synchronous
* synchronous but returns a promise
* asynchronous

test/test-test.js
```diff
@@ -233,6 +233,69 @@
...
+
+            "restores the sandbox when the promise is fulfilled": function (done) {
+                var method = function () {};
+                var object = { method: method };
+                var spy = sinon.spy();
+                sinon.test(function (callback) {
+                    var stub;
+                    var sandbox = this;
+                    return Promise.resolve()
+                        .then(function () {
```
fearphage: I would greatly simplify these tests. Simply return `Promise.resolve();`. There is no need to test the functionality of the promise library. Remove spies from the equation. We're only testing the sandbox functionality.

test/test-test.js
```diff
@@ -233,6 +233,69 @@
...
+                });
+            },
+
+            "restores the sandbox when the promise is rejected": function (done) {
+                var method = function () {};
+                var object = { method: method };
+                sinon.test(function (callback) {
+                    var sandbox = this;
+                    return Promise.resolve()
+                        .then(function () {
```
fearphage: Same here. Just return `Promise.reject()`.

test/test-test.js
joelmukuthu: > Remove spies from the equation. We're only testing the sandbox functionality.

+1

> Simply `return Promise.resolve();`.

Does that guarantee that the sandbox was restored after the Promise was resolved? I think it's important to keep the `sandbox.stub()` (or whatever sandbox method) in the `then` callback.

lib/sinon/test.js
joelmukuthu: > If you simplify this, it will work with more Promise libraries like jQuery (before v3). If you only use .then(onResolve, onReject). I do realize that old jQuery was not Promises/A+ compliant. It's just a thought and almost functionally the same.

+1

>  It also feels like this should be an else-if of the previous if statement. Looking at the code now, a synchronous test that returns a promise will have restore called multiple times.

Speaking strictly with regards to mocha, a test that returns a promise is regarded as asynchronous. I was of the impression that `sandbox.restore` was idempotent but sure, I can move the code around to ensure that doesn't happen, and add a test for it as you suggested.

> If you move your if statement before the typeof OldDone check, you can actually make it so returning a promise is all that is necessary to make tests async instead of return a promise and taking a callback.

I'm not sure I follow. If the user supplied a "done" callback and also returned a Promise, how will that case be handled then?

fearphage: > If the user supplied a "done" callback and also returned a Promise, how will that case be handled then?

That case should throw an error (like Mocha). The user is giving mixed signals.

test/test-test.js
fearphage: If you stub `sandbox.restore`, wouldn't that be the assertion you want? You're confirming the side effects of restoring now instead of the actual functionality you want to confirm.

fearphage: Even if you create the `stub`, you don't need to call it and confirm it was called. That is tested elsewhere. At most I would stub it and confirm it was no longer stubbed at the end, but that seems unnecessary when you can confirm `sandbox.restore` was called. I guess it's hard to confirm when it was called though. I'm fine with leaving the stub in, but no need to call it and test that stubs work in this file.

fearphage:
@joelmukuthu If you would [allow me to](https://github.com/blog/2247-improving-collaboration-with-forks), I wouldn't mind finishing this PR out for you.

joelmukuthu:
@fearphage done :)

fearphage:
FYI, I got stuck on some tests and busterjs fighting. I walked away for a bit to regain my sanity. I haven't forgotten about this.

fearphage:
I can't figure out how to push back into this branch/commit/PR. The remote branch has the same name as the local branch so I'm not sure if that's the problem. The only option I have is to push *by force* and I don't know what that will do. Any pointers would be appreciated.

 I've gotten the tests passing and I refactored a lot. I'm not 100% happy with the way it turned out and I'm open to suggestions. I modelled it very closely to the way mocha.js handles the sitatuation. However, there is a big difference between mocha tests and sinon tests. In mocha, you know the moment you see the `fn`/`callback` whether the test is sync or async (based on the `fn.length > 0`). In sinon, you don't know until you're inside the test itself.

```js
function test1(stub, mock) // looks async because of fn.length
function test2(done) // looks async because of fn.length
function test3() // known synchronous
```

This made the abstraction not as clean as I would have hoped.

fearphage:
Here's a manual diff of the changes until I figure out what I'm doing.

https://github.com/sinonjs/sinon/compare/v1.17...pr/1128

jonnyreeves:
Any reason this is going into 1.x instead of the standalone sinon-test
library?

fearphage:
> Any reason this is going into 1.x instead of the standalone sinon-test library?

That is simply where the PR began.

fearphage:
I broke something with a forced push. I don't know how to recover. It would be easiest for me to open a new PR giving @joelmukuthu full credit of course. I'll look into that this evening.

joelmukuthu:
@fearphage To be fair I haven't had too much input in the solution you came up with (and it looks good), but I'll take full credit for posting the bug :) If you need any help from me, let me know.

fearphage:
@joelmukuthu I opened the PR against `sinon-test` instead of sinon v1.17. It has fewer tests to fight with. Thank you again.

sinonjs/sinon-test#20

PR #1132, opened 2016-08-30
Support returning promises in sinon.test. Fixes #1119.

### LINKED ISSUE #1119: sinon.test restores the sandbox before a promise-based async test is completed

STRML:
Fix issue #1119 by calling thenables if present.

This is a backport of the [accepted solution in sinon-test](https://github.com/sinonjs/sinon-test/issues/6).

The problem is fully described in #1119.

Interestingly, this only bit us when https://github.com/petkaantonov/bluebird/issues/1170 was fixed.

#### Solution  - optional
This solution is simply a backport with additional tests.

2. `npm install`
3. `npm test`.


STRML:
Apologies - incorrect base branch.

PR #1133, opened 2016-08-30
Support returning promises in sinon.test. Fixes #1119.

### LINKED ISSUE #1119: sinon.test restores the sandbox before a promise-based async test is completed

STRML:
#### Purpose (TL;DR) 
Fix issue #1119 by calling thenables if present.

This is a backport of the [accepted solution](https://github.com/sinonjs/sinon-test/commit/c93063c18e33fe064d619f3d9b1d2603cd2c0157) ([issue](https://github.com/sinonjs/sinon-test/issues/6)) in sinon-test into v1. This issue is fixed in v2.

The problem is fully described in #1119.

Interestingly, this only bit us when https://github.com/petkaantonov/bluebird/issues/1170 was fixed.

#### Solution
This solution is simply a backport with additional tests.

#### How to verify
2. `npm install`
3. `npm test`.

### DIFF
--- a/lib/sinon/test.js
+++ b/lib/sinon/test.js
@@ -51,7 +51,19 @@
                 if (typeof exception !== "undefined") {
                     sandbox.restore();
                     throw exception;
-                } else if (typeof oldDone !== "function") {
+                }
+
+                if (result && typeof result === "object" && typeof result.then === "function") {
+                    return result.then(function sinonHandlePromiseResolve(val) {
+                        sandbox.verifyAndRestore();
+                        return val;
+                    }, function sinonHandlePromiseReject(err) {
+                        sandbox.restore();
+                        throw err;
+                    });
+                }
+
+                if (typeof oldDone !== "function") {
                     sandbox.verifyAndRestore();
                 }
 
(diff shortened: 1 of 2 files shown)


fatso83:
I think this duplicates the effort in #1128?

STRML:
Yes - recognized that after submission.

That PR appears to be delayed on review of the method of detecting thenables, which assumes a `catch` function. This should be much easier to merge as it uses the already-merged [implementation in sinon-test](https://github.com/sinonjs/sinon-test/commit/c93063c18e33fe064d619f3d9b1d2603cd2c0157).

fearphage:
This has some of the same problems as the other PR, but there is more context there. Let's try to finish it in the previous thread. Closing as a dupe.

STRML:
Does this have the same issues as the other PR? The other PR uses a method that is different than what is merged into sinon-test. If #1128 is eventually reworked, it will simply look like this PR.

Issue #914, opened 2015-11-08
Improve testing process/structure


fearphage:
Are there any plans to simplify the test running process? It's incredibly slow and convoluted. I have no idea how to run a single test (think TDD). In the majority of my other projects, test runs are measured in milliseconds. This test suite is very heavy (2 background processes among other things) and thus very slow. 
- Do you have any plans to switch to something lighter/faster? 
- Are you open to PR's for something like this?

mantoni:
Related to the discussion here: https://github.com/sinonjs/sinon/issues/912
Yes, your help would be appreciated. The test suite should ideally run in headless browsers, real browsers and on Node. We're open for suggestions on how to move forward.

mroderick:
Use the focus rocket to run fewer tests: http://docs.busterjs.org/en/latest/overview/

PR #1134, opened 2016-09-01
Add chrome to SauceLabs tested browser list

### LINKED ISSUE #912: Run tests in real browsers
It would be very good if we could run tests in multiple, real browsers in the same build on Travis, that runs tests in node and PhantomJS / SlimerJS

Some browsers to consider (the usual suspects):

* IE9+
* Firefox
* Safari
* Chrome

I imagine that we could use

* SauceLabs
* Browserstack

or similar service that has free options for open source projects.

mantoni:
Run the test suite in Chrome on SauceLabs again.

I finally managed to investigate further on the Chrome issue. An attempt to fix the issue was made in [min-webdriver 2.9.1](https://github.com/mantoni/min-webdriver/blob/master/CHANGES.md#291).

This PR should (hopefully) demonstrate that Chrome testing now works as expected.

This will finally allow us to close #912 and #914

### DIFF
--- a/.min-wd
+++ b/.min-wd
@@ -1,6 +1,8 @@
 {
   "sauceLabs": true,
   "browsers": [{
+    "name": "chrome"
+  }, {
     "name": "firefox"
   }, {
     "name": "internet explorer",


jonnyreeves:
🎉

Issue #1138, opened 2016-09-06
Sinon mock expects verify is setting my method to 'undefined' on restore (hasOwnProperty conflict)


launchcg-ztonia:
- Sinon version : _please verify that the bug exists in the latest Sinon release_
  sinon@1.17.5 
- Environment   :
  node 4.5.0
  OSX El Capitain 10.11.5
- Example URL   :
- Other libraries you are using:

"chai": "^3.4.0",
"mocha": "^2.3.3",
"mockery": "^1.4.1"

**What did you expect to happen?**
When my mocked object is 'verify'd, the original method must be restored.

**What actually happens**
The method is set as undefined

**How to reproduce**

> Describe _with code_ how to reproduce the faulty behaviour,
> or link to code on JSBin or similar

```
var objectInstance = require('./MODULE');
var objectInstance2 = require('./MODULE2');
describe ("testincoming", function() {
  var mockObject, mockObject2;
  beforeEach(function() { mockObject = sinon.mock(objectInstance); })
  afterEach(function() { mockObject.restore() })

  describe("nested", function() {
  beforeEach(function() { mockObject2 = sinon.mock(objectInstance2); })
  afterEach(function() { mockObject2.restore() })
  it("test", function() {
    mockObject.expects('method').once().withArgs(5).returns(Promise.resolve());
    mockObject2.expects('method').once().withArgs(5).returns(Promise.resolve());
    return invokeSomethingElsePromise()
      .then(function() {
        mockObject2.verify(); // This is destroying my method
        mockObject.verify(); // This is not destroying my method. I can't tell what's different.
      });
  });
});
})
```

Please note, there are two nearly identical objects which I am mocking in this same way. Only ONE is exhibiting this behavior and the other is not. I cannot paste source code, but I can answer follow up questions.

launchcg-ztonia:
I have found the source of the error.
The mocked method has implemented the method 'hasOwnProperty'
Which is conflicting in sinon/util/core and causing the restore method to fail.

launchcg-ztonia:
I now see related issue
#887
But it is marked closed resolved.

PR #1140, opened 2016-09-07
Has Own Property Check in Core now compares vs Object prototype.

launchcg-ztonia:

Fix Issue #1138 Mocked objects which override method 'hasOwnProperty' fail to restore(). Methods were being deleted instead.

Issue arises because of a conflict between browser needs. IE does not implement hasOwnProperty, but Firefox cannot blindly rely on Object.prototype.hasOwnProperty.call.
#### Solution  - optional

Not enough to check that hasOwnProperty is implemented. Necessary to also check that it matches Object.prototype.hasOwnProperty
2. `npm install`
3. Run testing suite.

### DIFF
--- a/lib/sinon/util/core.js
+++ b/lib/sinon/util/core.js
@@ -103,7 +103,8 @@
 
             // IE 8 does not support hasOwnProperty on the window object and Firefox has a problem
             // when using hasOwn.call on objects from other frames.
-            var owned = object.hasOwnProperty ? object.hasOwnProperty(property) : hasOwn.call(object, property);
+            var owned = (object.hasOwnProperty && object.hasOwnProperty === hasOwn) ?
+                object.hasOwnProperty(property) : hasOwn.call(object, property);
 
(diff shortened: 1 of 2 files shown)


launchcg-ztonia:
The only thing I would worry about is if someone wrote the browser test for the Firefox issue.

mroderick:
> Not enough to check that hasOwnProperty is implemented. Necessary to also check that it matches Object.prototype.hasOwnProperty

The code uses `object.hasOwnProperty`, not `Object.prototype.hasOwnProperty` as in your description. Was the description wrong, or am I missing something?

launchcg-ztonia:
That's the problem.

Please see the attached issue. If you mock an object which overrides 'hasOwnProperty', sinon will delete properties on restore.

fatso83:
Thanks for your work. This seems fine. Would you mind filing a backport to `master`? I'll see to merging it asap.

PR #1145, opened 2016-09-09
Help coveralls separate the code from the tests

fearphage:
Just flagging the `lib` dir as code.


fearphage:
This didn't make it better, but it didn't make it worse either. I'm hoping merging to master will do the trick.

PR #1150, opened 2016-09-19
Throw when sandbox.restore is given arguments (Fixes #1149)

### LINKED ISSUE #1149: sandbox.restore with args should throw
* Sinon version: 1.17.5
* Environment: node / browser
* Example URL: https://runkit.com/5701d7f537a13d1200a55a1b/57ddaf833ffa1f1400543e75
* Other libraries you are using: none

I expected this code to throw since `sandbox.restore` doesn't take any arguments.

```js
const sinon = require("sinon")

const sandbox = sinon.sandbox.create();
sandbox.restore('foo', 'bar');
```

I've found that in a couple places in our code base, engineers think they can restore an individual stub by calling:
```js
sandbox.restore(stub);
```
Which actually restores **all** stubs

Whereas they should be calling:
```js
stub.restore();
```

I've been thinking of how I should address this problem, possibly by using a linter or by monkey patching the library. However, I figure that more than just our team has run into this and this is a thing sinon itself can protect engineers from doing.

TheSavior:
Fix issue #1149 by throwing when sandbox.restore is given arguments

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -95,6 +95,10 @@ extend(sinonSandbox, {
     },
 
     restore: function () {
+        if (arguments.length) {
+            throw new Error("sandbox.restore() does not take any parameters. Perhaps you meant stub.restore()");
+        }
+
         sinonCollection.restore.apply(this, arguments);
         this.restoreContext();
     },
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -243,6 +243,28 @@ describe("sinonSandbox", function () {
         }
     });
 
+    describe(".restore", function () {
+        it("throws when passed arguments", function () {
+            this.sandbox = sinonSandbox.create();
+
+            var failed = false;
+            var exception;
+
+            try {
+                this.sandbox.restore("args");
+                failed = true;
+            } catch (e) {
+                exception = e;
+            }
+
+            assert.isFalse(failed);
+            assert.equals(
+                exception.message,
+                "sandbox.restore() does not take any parameters. Perhaps you meant stub.restore()"
+            );
+        });
+    });
+
     describe("configurable sandbox", function () {
         beforeEach(function () {
             this.requests = [];
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -245,23 +245,13 @@ describe("sinonSandbox", function () {
 
     describe(".restore", function () {
         it("throws when passed arguments", function () {
-            this.sandbox = sinonSandbox.create();
+            var sandbox = sinonSandbox.create();
 
-            var failed = false;
-            var exception;
-
-            try {
-                this.sandbox.restore("args");
-                failed = true;
-            } catch (e) {
-                exception = e;
-            }
-
-            assert.isFalse(failed);
-            assert.equals(
-                exception.message,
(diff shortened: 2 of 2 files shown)


lib/sinon/sandbox.js
```diff
@@ -95,6 +95,10 @@ extend(sinonSandbox, {
     },
 
     restore: function () {
+        if (arguments.length) {
+            throw new Error("sandbox.restore() does not take any parameters. Perhaps you meant stub.restore()");
+        }
+
         sinonCollection.restore.apply(this, arguments);
```
TheSavior: This line is weird to me. collection.restore also doesn't take any arguments, so it seems weird to be passing through this functions arguments when neither of them take any. 

I considered removing it here but decided that I didn't have enough context and didn't want to make this PR any more risky to you guys by changing existing code.

test/sandbox-test.js
```diff
@@ -243,6 +243,28 @@ describe("sinonSandbox", function () {
         }
     });
 
+    describe(".restore", function () {
+        it("throws when passed arguments", function () {
+            this.sandbox = sinonSandbox.create();
+
+            var failed = false;
```
TheSavior: I was surprised to see that referree doesn't have an `assert.throws` function. I found this pattern elsewhere in your test suite.

TheSavior:
Coveralls seems to be complaining about the coverage of the `catch` clause in my test. I'm surprised by that because it is definitely going into that branch to set the exception variable.

https://coveralls.io/builds/7943286/source?filename=..%2Ftest%2Fsandbox-test.js#L256

Any ideas?

lib/sinon/sandbox.js
fatso83: Good point, this does seem a bit weird. That line has been there since restore was first introduced for the sandbox in e1bcec820, so maybe @marcushg remembers if this was intentional or just done by convention.

test/sandbox-test.js
fatso83: Close. You can use [assert.threw](http://docs.busterjs.org/en/latest/modules/referee/#threw) or optionally [assert.exception](http://docs.busterjs.org/en/latest/modules/referee/#exception) (for callbacks).

fatso83:
@TheSavior No, I am not breaking any sweat over a 5 thousands of a percent's change in coverage, and neither should you. Don't worry about it. It's there to catch sudden drops. This is not it.

test/sandbox-test.js
TheSavior: The only keys on assert seem to be:

```
[ 'toString',
  'same',
  'equals',
  'greater',
  'less',
  'defined',
  'isNull',
  'match',
  'isObject',
  'isFunction',
  'isTrue',
  'isFalse',
  'isString',
  'isBoolean',
  'isNumber',
  'isNaN',
  'isArray',
  'isArrayLike',
  'keys',
  'exception',
  'near',
  'hasPrototype',
  'contains',
  'tagName',
  'className',
  'spy',
  'stub',
  'mock',
  'fakeServer',
  'clock',
  'fakeServerWithClock' ]
```

`assert.exception` was exactly what I was looking for.

fatso83: Ah, we might need to update the dependencies for that package. Maybe it's not included in that version. I'll try to keep that info stored somewhere :-)

fatso83:
Thank you!

PR #1151, opened 2016-09-19
Throw when sandbox.restore is given arguments (Fixes #1149)

### LINKED ISSUE #1149: sandbox.restore with args should throw

TheSavior:
Backporting https://github.com/sinonjs/sinon/pull/1150 to the 1.17 branch as specified in #1149.

@fatso83

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -93,6 +93,10 @@
             },
 
             restore: function () {
+                if (arguments.length) {
+                    throw new Error("sandbox.restore() does not take any parameters. Perhaps you meant stub.restore()");
+                }
+
                 sinon.collection.restore.apply(this, arguments);
                 this.restoreContext();
             },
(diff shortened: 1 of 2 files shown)


fatso83:
Thank you!

PR #1162, opened 2016-10-06
XHR spec conformance: abort() should not dispatch readystatechange event in DONE state

dottedmag:

Fix XHR spec conformance by not issuing "load" event if abort() is called on XHR in DONE state.

FakeXHR.readyStateChange is issuing "load" event if called on XHR with this.status != 0. FakeXHR.abort() calls FakeXHR.readyStateChange() unconditionally, so don't call it if XHR is in DONE state. This also avoids spurious "progress" and "loadend" events.
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -545,7 +545,8 @@ extend(FakeXMLHttpRequest.prototype, sinonEvent.EventTarget, {
         this.requestHeaders = {};
         this.responseHeaders = {};
 
-        if (this.readyState > FakeXMLHttpRequest.UNSENT && this.sendFlag) {
+        if (this.readyState !== FakeXMLHttpRequest.UNSENT && this.sendFlag
+            && this.readyState !== FakeXMLHttpRequest.DONE) {
             this.readyStateChange(FakeXMLHttpRequest.DONE);
             this.sendFlag = false;
         }
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -1266,6 +1266,17 @@ if (typeof window !== "undefined") {
                 assert.isFalse(this.xhr.onreadystatechange.called);
             });
 
+            it("does not dispatch readystatechange event if readyState is done", function () {
+                this.xhr.open("GET", "/");
+                this.xhr.send();
+                this.xhr.respond();
+
+                this.xhr.onreadystatechange = sinonStub();
+                this.xhr.abort();
+
+                assert.isFalse(this.xhr.onreadystatechange.called);
+            });
+
             assertEventOrdering("abort", 0, function (xhr) {
                 xhr.abort();
             });
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -1164,7 +1164,50 @@ if (typeof window !== "undefined") {
                 assert.equals(this.xhr.responseHeaders, {});
             });
 
-            it("sets state to DONE if sent before", function () {
+            it("keeps readyState unsent if called in unsent state", function () {
+                this.xhr.abort();
+
+                assert.equals(this.xhr.readyState, FakeXMLHttpRequest.UNSENT);
+            });
+
+            it("resets readyState to unsent if it was opened", function () {
+                this.xhr.open("GET", "/");
+
+                this.xhr.abort();
+
+                assert.equals(this.xhr.readyState, FakeXMLHttpRequest.UNSENT);
+            });
+
+            it("resets readyState to unsent if it was opened with send() flag sent", function () {
+                this.xhr.open("GET", "/");
+                this.xhr.send();
+
+                this.xhr.abort();
+
+                assert.equals(this.xhr.readyState, FakeXMLHttpRequest.UNSENT);
+            });
+
+            it("resets readyState to unsent if it headers were received", function () {
+                this.xhr.open("GET", "/");
+                this.xhr.send();
+                this.xhr.setResponseHeaders({});
+
+                this.xhr.abort();
+
+                assert.equals(this.xhr.readyState, FakeXMLHttpRequest.UNSENT);
+            });
+
+            it("resets readyState to unsent if it was done", function () {
+                this.xhr.open("GET", "/");
+                this.xhr.send();
+                this.xhr.respond();
+
+                this.xhr.abort();
+
+                assert.equals(this.xhr.readyState, FakeXMLHttpRequest.UNSENT);
+            });
+
+            it("signals onreadystatechange with state set to DONE if sent before", function () {
                 var readyState;
                 this.xhr.open("GET", "/");
                 this.xhr.send();


fatso83:
@dottedmag : do you have any reference to the W3 spec that this is actually how it is supposed to work?

@fearphage, I think this is your territory after the work you did this spring :-)

dottedmag:
Here's the link: https://xhr.spec.whatwg.org/#the-abort()-method

Mea culpa, I thought it was included in commit message. Apparently it was not.

fearphage:
I thought I'd fixed this once already in #861:

> bug fix: 'load' was fired before 'abort'

I don't know if I/someone reverted or broke it or what.  I'll look into this when I get a chance.

fearphage:
@dottedmag This is what I've been using lately to compare to browser implementations:

https://dl.dropboxusercontent.com/u/2400/tc/sinon/xhr-browser-test.htm

dottedmag:
@fearphage Definitely none of the popular browsers "in wild" emit `load` on `abort()` – I've stumbled upon this issue by mocking XHR on the code which is in widely production.

Also I have noticed that the code in question changed significantly between the time previous bugfix was applied and current `master`.

fearphage:
@dottedmag This one got away from me. I'll be unavailable for a few weeks so I won't be working on it then either. 

Anyone feel free to move this forward. Again, I'm sorry I missed this one.

fatso83:
@fearphage, I investigated this a bit and you are indeed right in that [the tests you included](https://github.com/sinonjs/sinon/pull/861/files#diff-6c977700ac09a0bbc2131d24ac071244R1730) in #861 should have covered this. 

So why did they not? I checked that the tests were still there and I ran the tests myself. Nothing happened. All green. Strange. Then I forcibly tried to make the test fail. Still green... 👎 

In the end I found out that our headless PhantomJS test runner is not working, and has not done so ever since converting the headless runner over to mochify in 931da2b467b. Good thing we haven't released an official 2.0, given that no tests for DOM or XHR related have been run for half a year 😓 

@mantoni: could you look into what makes the runner fail? I really don't (want to) know the internals of `mochify`. This is all I get out of it:
```
$ $(npm bin)/mochify --recursive -R dot test
# phantomjs:

$ echo 'what to do?'
```

mantoni:
@fatso83 I just ran our npm test script, made a change in the xhr tests and ran again and tests fail as expected. The script is running this: `mochify --recursive -R dot --grep WebWorker --invert test`, because we need to exclude the WebWorker tests. Did I misunderstand what the issue is?

fatso83:
@mantoni: Hmm ... maybe this is a case of "it works on my machine"? No change I do to tests (either before or after the `window` check) can trigger any errors:
```
$ npm run test-headless

> sinon@2.0.0-pre.3 test-headless /Users/carl-erik.kopseng/dev-priv/sinon
> mochify --recursive -R dot --grep WebWorker --invert test/

# phantomjs:

carl-erik.kopseng at MWNOOSLMACLT103-2 in ~/dev-priv/sinon (master)
$ git diff
diff --git a/test/util/fake-xml-http-request-test.js b/test/util/fake-xml-http-request-test.js
index 6d9ec34..e240842 100644
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -119,6 +119,7 @@ var assertEventOrdering = function (event, progress, callback) {
     });
 };

+    throw new Error('are we here yet');
 if (typeof window !== "undefined") {
     describe("FakeXMLHttpRequest", function () {
```

PhantomJS simply does not do anything.

mantoni:
Which version are you using? I have `2.1.1` installed.

fatso83:
You might be on to something (humiliating for me, most probably) ... Running `phantomjs` gave absolutely no output, so it might have been the local system-wide installation that had stopped working (OS X Sierra upgrade breaking Homebrew kegs, perhaps?). 

...

Yep. That was it. 1501 tests passing ... 

Removing the `node_modules` dir and `npm i` fixed the issue. Damn. Sorry for making a racket.

test/util/fake-xml-http-request-test.js
```diff
@@ -1063,18 +1063,39 @@ if (typeof window !== "undefined") {
...
+                this.xhr.onloadend = sinonStub();
+                this.xhr.onloadstart = function () {
+                    this.abort();
                 };
+                this.xhr.send();
+
+                assert.isTrue(this.xhr.onloadend.called);
+            });
+
+            it("sets state to DONE if sent before", function () {
```
fatso83: I don't see how the test description matches the content? You are not referring to DONE anywhere.

test/util/fake-xml-http-request-test.js
```diff
@@ -1086,14 +1107,15 @@ if (typeof window !== "undefined") {
                 assert.isFalse(this.xhr.sendFlag);
             });
 
-            it("dispatches readystatechange event if sent before", function () {
+            it("does not dispatch readystatechange event if sent before", function () {
                 this.xhr.open("GET", "/");
```
fatso83: By "sent" you mean the actual event, not the `xhr.send()` call being triggered twice, right?

fatso83:
@fearphage, your test still passes, but I think the key to this might be in this statement from @dottedmag:

> FakeXHR.readyStateChange is issuing "load" event if called on XHR with this.status != 0. 

I won't pretend I have any clue on this, but I did notice that `xhr.status === 0` throughout the entire test, so if this happens when `xhr.status !== 0`, then your test won't cover it. No idea if that makes any sense to you.

test/util/fake-xml-http-request-test.js
dottedmag: Well, the name was there before, so I don't really know what it was supposed to mean. I have updated it to match the spec though.

test/util/fake-xml-http-request-test.js
dottedmag: You're right, I forgot to change the name when I fixed the test.

test/util/fake-xml-http-request-test.js
fatso83: OK, that's fine, but could you please change the name to whatever makes sense to you and/or the casual reader? Just to avoid dummies like myself wondering what the test is supposed to mean in the future :-)

fatso83:
@dottedmag : just update the test names and this is good to go from my perspective 🙂 Thanks!

fearphage:
Hello folks, I'm just catching up. Is the description of for this ticket still accurate? If not, can we fix it. It would make it much easier to discuss I feel.

dottedmag:
@fearphage Yes and "does not dispatch readystatechange event in DONE state".

I have updated the patch.

fearphage:
@dottedmag Can you edit the title of the ticket to reflect that?

lib/sinon/util/fake_xml_http_request.js
```diff
@@ -545,7 +545,7 @@ extend(FakeXMLHttpRequest.prototype, sinonEvent.EventTarget, {
         this.requestHeaders = {};
         this.responseHeaders = {};
 
-        if (this.readyState > FakeXMLHttpRequest.UNSENT && this.sendFlag) {
+        if (this.readyState > FakeXMLHttpRequest.UNSENT && this.sendFlag && this.readyState < FakeXMLHttpRequest.DONE) {
```
fearphage: This could be `this.readyState !== FakeXMLHttpRequest.DONE`. This seems a bit more meaninful to me. This is inline with the readystateCHANGE event since it can't change from 4 to 4. That's not a change and thus it shouldn't fire.

dottedmag: I see no reason why `UNSENT` test has to be `>`, but `DONE` has to be `!==`.

dottedmag:
The changed tests had assertions which matched implementation, but not what XHR spec requests. IOW, they were wrong.

"does not dispatch readystatechange event in DONE state" put XHR into DONE state (by waiting for loadend).

lib/sinon/util/fake_xml_http_request.js
fearphage: It's more explicit in my opinion. Do you care that you're in the done state or do you care that you're in a state that is less than the done state?

dottedmag: Then I'll fix `UNSENT` test too.

fearphage:
@dottedmag When I revert your test changes only, all the tests still pass. Could you be more specific about what the tests were doing that didn't confirm to spec? I'm mostly curious. Looking at them side by side, the error in the tests aren't jumping out at me.

lib/sinon/util/fake_xml_http_request.js
fearphage: :+1:

dottedmag:
Weird, I'll check it out again.

dottedmag:
Now the tests should be better.

I have found another issue (readystatechange event is not firing on abort() if XHR is synchronous), but I'll deal with it in another PR.

dottedmag:
The test in 7787389 is the one which fails without the fix.

fatso83:
@fearphage, I think this just needs your approval, as I think your objections have been addressed?

fearphage:
@dottedmag Thanks for the work.

PR #1170, opened 2016-10-21
fix some printf typos

brad:

This fixes some issues with the `printf` docs
- missing opening `"`
- visible `<var>` tags

### DIFF
--- a/docs/current/spies.md
+++ b/docs/current/spies.md
@@ -399,7 +399,7 @@ Returns the passed format string with the following replacements performed:
     <dt><code>%t</code></dt>
     <dd>a comma-delimited list of <code>this</code> values the spy was called on</dd>
 
-    <dt><code>%</code><var>n</var></dt>
+    <dt><code>%</code><var><code>n</code></var></dt>
     <dd>the formatted value of the <var>n</var>th argument passed to <code>printf</code> <br/>FIXME: needs example</dd>
 
     <dt><code>%*</code></dt>
--- a/docs/current/spies.md
+++ b/docs/current/spies.md
@@ -383,7 +383,7 @@ Resets the state of a spy.
 Replaces the spy with the original method. Only available if the spy replaced an existing method.
 
 
-#### `spy.printf(format string\", [arg1, arg2, ...])`
+#### `spy.printf(\"format string\", [arg1, arg2, ...])`
 
 Returns the passed format string with the following replacements performed:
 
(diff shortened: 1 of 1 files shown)


fatso83:
awaiting response.

brad:
@fatso83 Sorry, I have no idea if the code works. I just noticed the documentation problems and wanted to help fix them. Can you be more specific about breaking formatting? I testing the `var` change by going to the [docs](http://sinonjs.org/docs/) page in chrome and using developer tools to change the HTML in the browser. It looks good to me, no code is visible on the page and the `n` is italics as it should be. I see now that the `n` isn't monospace, I'll make another commit now to fix that.

Regarding your suggestion on the first fix, I agree that would be sensible. However, all other similar examples in the doc do escape the quotes. I assumed there was a good reason for that and thought it would be best to follow the same style

docs/current/spies.md
```diff
@@ -399,7 +399,7 @@ Returns the passed format string with the following replacements performed:
     <dt><code>%t</code></dt>
     <dd>a comma-delimited list of <code>this</code> values the spy was called on</dd>
 
-    <dt><code>%<var>n</var></code></dt>
+    <dt><code>%</code><var><code>n</code></var></dt>
```
mroderick: If I read the HTML spec correctly, then `<var>` is a perfectly valid child of `<code>`.

https://www.w3.org/TR/html5/text-level-semantics.html#the-code-element
https://www.w3.org/TR/html5/dom.html#phrasing-content-1

Do you agree? Is this change necessary?

mroderick:
**Background**: the documentation published to http://sinonjs.org/docs/ exists a separate repository, which is really difficult to maintain, let alone contribute to. Therefore, we are working on a new documentation website to replace the old one.

@brad your PR is for the docs in `master`, which is just as it should be. Thank you!

fatso83:
As long as the docs in `master` are yet to be published as the new site, and these changes are based on changing code that is on a site which is soon discontinued I think we'll pass on this one. Maybe we'll revisit once the docs are up and running. It's a bit hard to know what is actually being changed and how it will look as is. The rendered Markdown seems fine without the changes, so I think this is just a byproduct of bad rendering on the sinonjs.org site.

@mroderick: any chance we'll get a XMas present in the form of a sinon.github.io (with possible CNAME to sinonjs.org) site based on the great new docs? 🙏

PR #1173, opened 2016-10-24
Expose readyState constants on XHR instances

dottedmag:

Expose readyState constants on XHR instances to allow mocking code which uses them.
- Browsers implement readyState constants on XHR instances.
- XHR spec says to do so too: https://xhr.spec.whatwg.org/#states
- There is code which uses it and it breaks under Sinon's XHR mock.
2. `npm install`
3. `npm test`


fatso83:
Big up! Nice and clear.

PR #1178, opened 2016-11-07
Expose XHR.setStatus to simplify asynchronous answers

dottedmag:

Previously the only way to do a two-stage XHR response (first headers, then body) was by setting `.status`/`.statusText` manually, as the only method which sets them is `.respond`, which also sets body.

This patch adds `.setStatus`, which allow scenarios like

    xhr.setStatus(200);
    xhr.setResponseHeaders(...);
    // ... sleep for a specified amount of time
    xhr.setResponseBody(...)

2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -492,6 +492,14 @@ extend(FakeXMLHttpRequest.prototype, sinonEvent.EventTarget, {
         }
     },
 
+    setStatus: function setStatus(status) {
+        var sanitizedStatus = typeof status === "number" ? status : 200;
+
+        verifyRequestOpened(this);
+        this.status = sanitizedStatus;
+        this.statusText = FakeXMLHttpRequest.statusCodes[sanitizedStatus];
+    },
+
     // Helps testing
     setResponseHeaders: function setResponseHeaders(headers) {
         verifyRequestOpened(this);
@@ -629,8 +637,7 @@ extend(FakeXMLHttpRequest.prototype, sinonEvent.EventTarget, {
     },
 
     respond: function respond(status, headers, body) {
-        this.status = typeof status === "number" ? status : 200;
-        this.statusText = FakeXMLHttpRequest.statusCodes[this.status];
+        this.setStatus(status);
         this.setResponseHeaders(headers || {});
         this.setResponseBody(body || "");
     },
--- a/test/util/fake-xml-http-request-test.js
+++ b/test/util/fake-xml-http-request-test.js
@@ -705,6 +705,58 @@ if (typeof window !== "undefined") {
             });
         });
 
+        describe(".setStatus", function () {
+            beforeEach(function () {
+                this.xhr = new FakeXMLHttpRequest();
+            });
+
+            it("cannot be set on closed request", function () {
+                assert.exception(function () {
+                    this.xhr.setStatus();
+                });
+            });
+
+            it("cannot be set on unsent request", function () {
+                this.xhr.open("GET", "/");
+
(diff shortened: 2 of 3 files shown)


lib/sinon/util/fake_xml_http_request.js
```diff
@@ -492,6 +492,12 @@ extend(FakeXMLHttpRequest.prototype, sinonEvent.EventTarget, {
         }
     },
 
+    setStatus: function setStatus(status) {
+        verifyRequestOpened(this);
+        this.status = typeof status === "number" ? status : 200;
+        this.statusText = FakeXMLHttpRequest.statusCodes[this.status];
```
mroderick: Not a big deal, but you can shorten this a bit and just use `status` when looking up the `statusText`.

```javascript
this.statusText = FakeXMLHttpRequest.statusCodes[status];
```

test/util/fake-xml-http-request-test.js
```diff
@@ -705,6 +705,44 @@ if (typeof window !== "undefined") {
...
+                assert.equals(this.xhr.statusText, "OK");
+            });
+
+            it("sets status and status text", function () {
+                this.xhr.open("GET", "/");
+                this.xhr.send();
+                this.xhr.setStatus(206);
+
+                assert.equals(this.xhr.status, 206);
+                assert.equals(this.xhr.statusText, "Partial Content");
```
mroderick: Try to avoid using literal values for expected results, as they have no explanation of why they're expected or how we arrive at them.

A better approach would be to use variable names (or function calls) to show why this is the expected result.

The necessity is a bit exaggerated in this example, as the values are near-trivial. However, it's a good habit to build, to make code easier to understand. 

Consider questions like these:

* "why is this value expected to be `3`?"
* "is this `3` meant to be the same `3` as above, or is that an unhappy coincidence?"

Anyway, here's how I would restructure this test:

* split up the two expectations into two `it()` calls
* use variable names to explain how we arrive at the expected value

```javascript
it("sets status", function () {
    var expectedStatus = 206;
    var xhr = this.xhr;

    xhr.open("GET", "/");
    xhr.send();
    xhr.setStatus(expectedStatus);

    assert.equals(xhr.status, expectedStatus);
});

it("sets status text to the value from FakeXMLHttpRequest.statusCodes", function () {
    var status = 206;
    var expectedStatusText = FakeXMLHttpRequest.statusCodes[]
    var xhr = this.xhr;

    xhr.open("GET", "/");
    xhr.send();
    xhr.setStatus(status);

    assert.equals(xhr.statusText, expectedStatusText);
});
```

lib/sinon/util/fake_xml_http_request.js
dottedmag: Can't do, the line above does the sanitizing, so `undefined` is treated as 200.

BTW, this is a piece of code moved over from `respond`.

test/util/fake-xml-http-request-test.js
dottedmag: I have updated the tests.

dottedmag:
Updated.

mroderick:
Excellent! 👍

mroderick:
Perhaps I was a bit hasty merging this, as we haven't updated the documentation.

@dottedmag would you be up for updating the documentation in a separate PR?

dottedmag:
Not sure what do you mean: https://github.com/sinonjs/sinon/commit/c4580474e3be32dd472725d43099a9d709f5c6ea#diff-685bf9e071f73c3dec0a76f57b3a40e7

mroderick:
Wow, I really can't GitHub when I am tired. Sorry about that

Issue #1147, opened 2016-09-16
sinon.assert.match documentation missing


robertknight:
sinon has a useful `match()` assertion method to test whether a value matches another using `sinon.match()`. This was added in https://github.com/sinonjs/sinon/pull/410 but is not documented in http://sinonjs.org/docs/#assertions-api

@ropez I assume this was intended to be a public API?

A secondary issue is that when sinon's assertions are exposed on the same `assert` object as chai's assertions, this conflicts with chai's `assert.match` which is a regexp match. Perhaps it should be renamed before making it public?

fatso83:
@robertknight : the state of documentation is less than optimal at the moment, but hopefully we will push a new site soon. Documentation is kept up-to-date in the `docs` directory in this repo. You can find [plenty of docs on matchers](https://github.com/sinonjs/sinon/blob/master/docs/current/matchers.md) there.

As to what other people/libraries are doing with sinon's assertions, that is up to them, and if some naming collision arises, that is frankly for them to fix. It is out of scope for us IMHO, unless I am misunderstanding what you describe.

robertknight:
Thanks @fatso83 . Although there is decent documentation for matchers themselves, `sinon.assert.match` is not mentioned on https://github.com/sinonjs/sinon/blob/master/docs/current/assertions.md

mroderick:
@robertknight I've created a PR to add the missing documentation: https://github.com/sinonjs/sinon/pull/1180

robertknight:
Thanks @mroderick , much appreciated.

PR #1180, opened 2016-11-09
Add documentation for sinon.assert.match

mroderick:
This is a PR for **master**, it is a solution to #1147.

### DIFF
--- a/docs/current/assertions.md
+++ b/docs/current/assertions.md
@@ -143,6 +143,22 @@ If only one argument is provided, the assertion passes if `spy` ever threw any e
 
 Like above, only required for all calls to the spy.
 
+#### `sinon.assert.match(actual, expectation);`
+
+Uses [`sinon.match`](./matchers.md) to test if the arguments can be considered a match.
+
+```javascript
+var sinon = require('sinon');
+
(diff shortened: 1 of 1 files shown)


docs/current/assertions.md
```diff
@@ -143,6 +143,26 @@ If only one argument is provided, the assertion passes if `spy` ever threw any e
...
+#### `sinon.assert.match(actual, expectation);`
+
+Uses [`sinon.match`](./matchers.md) to test if the arguments can be considered a match.
+
+```javascript
+var sinon = require('sinon');
+
+describe('example', function(){
+    it('should fail for mismatching strings', function() {
+        sinon.assert.match('hello', 'world');
```
robertknight: Perhaps it would be worth using an example involving fuzzy matching, since this is what makes `sinon.match` useful compared to say, node or chai assertions?

eg:

```js
// Test value of `x` property, ignore `y`
sinon.assert.match({x:1,y:2}, sinon.match({x:1}))
```

mroderick: I've pushed a new commit with an improved example based on your feedback @robertknight

PR #1183, opened 2016-11-10
Use last defined withArgs match when multiple conditions are met

### LINKED ISSUE #1053: withArgs(sinon.match()) doesn't use last defined return definition
- Sinon version : 1.17.4
- Environment   : OS X
- Example URL   : 

I would expect that the last withArgs return declaration to to have precedence when multiple declarations would match the invocation.

**What actually happens**

The first one seems to have precedence opposed to when you are using strict values instead of sinon.match

**How to reproduce**

```
var x = sinon.stub().returns(1);
x.returns(2);

x() // -> 2

var x = sinon.stub();
x.withArgs(1).returns(1);
x.withArgs(1).returns(2);

x(1) // -> 2

var x = sinon.stub();
x.withArgs(sinon.match('1')).returns(1);
x.withArgs(sinon.match('1')).returns(2);

x('1') // -> 1 (wat?)
```

Probably due to that "strict" declarations actually overwrite the withArgs precondition, and that is not possible with match-objects. However, if they were applied on the opposite order (last defined, first matched), it would match the normal behavior, I think.

jishi:
This hopefully fixes #1053

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -53,13 +53,11 @@ function matchingFake(fakes, args, strict) {
         return undefined;
     }
 
-    for (var i = 0, l = fakes.length; i < l; i++) {
-        if (fakes[i].matches(args, strict)) {
-            return fakes[i];
-        }
-    }
+    var matchingFakes = fakes.filter(function (fake) {
+        return fake.matches(args, strict);
+    });
 
-    return undefined;
+    return matchingFakes.pop();
 }
 
 function incrementCallCount() {
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -1345,6 +1345,35 @@ describe("stub", function () {
             assert.equals(stub(), 23);
             assert.equals(stub({ foo: "bar", bar: "foo" }), 99);
         });
+
+        it("ensure stub uses last matching arguments", function () {
+            var unmatchedValue = "d3ada6a0-8dac-4136-956d-033b5f23eadf";
+            var firstMatchedValue = "68128619-a639-4b32-a4a0-6519165bf301";
+            var secondMatchedValue = "4ac2dc8f-3f3f-4648-9838-a2825fd94c9a";
+            var expectedArgument = "3e1ed1ec-c377-4432-a33e-3c937f1406d1";
+
+            var stub = createStub().returns(unmatchedValue);
+
+            stub.withArgs(expectedArgument).returns(firstMatchedValue);
+            stub.withArgs(expectedArgument).returns(secondMatchedValue);
+
+            assert.equals(stub(), unmatchedValue);
+            assert.equals(stub(expectedArgument), secondMatchedValue);
+        });
+
+        it("ensure stub uses last matching sinonMatch arguments", function () {
+            var unmatchedValue = "0aa66a7d-3c50-49ef-8365-bdcab637b2dd";
+            var firstMatchedValue = "1ab2c601-7602-4658-9377-3346f6814caa";
+            var secondMatchedValue = "e2e31518-c4c4-4012-a61f-31942f603ffa";
+            var expectedArgument = "90dc4a22-ef53-4c62-8e05-4cf4b4bf42fa";
+
+            var stub = createStub().returns(unmatchedValue);
+            stub.withArgs(expectedArgument).returns(firstMatchedValue);
+            stub.withArgs(sinonMatch(expectedArgument)).returns(secondMatchedValue);
+
+            assert.equals(stub(), unmatchedValue);
+            assert.equals(stub(expectedArgument), secondMatchedValue);
+        });
     });
 
     describe(".callsArgAsync", function () {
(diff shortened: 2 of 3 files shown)


test/util/core/called-in-order-test.js
```diff
@@ -33,13 +33,13 @@ describe("util/core/calledInOrder", function () {
         it("returns true, if stubs were called in given order", function () {
             assert(calledInOrder([testObject1.someFunction, testObject2.otherFunction]));
             assert(calledInOrder([testObject1.someFunction, testObject2.otherFunction,
-                                        testObject3.thirdFunction]));
```
mroderick: I agree with the changes in this file, but they add noise. In the future, someone is going to land on this commit using `git bisect`, and the clearer each commit is, the faster they can get on with their day.

Would you mind splitting this commit, and put the white-space only changes in this file into a separate commit?

jishi: I splitted the linting fix into individual commit as suggested.

lib/sinon/spy.js
```diff
@@ -53,9 +53,11 @@ function matchingFake(fakes, args, strict) {
         return undefined;
     }
 
-    for (var i = 0, l = fakes.length; i < l; i++) {
-        if (fakes[i].matches(args, strict)) {
-            return fakes[i];
+    var fakesIndex = fakes.length;
+
+    while (fakesIndex--) {
```
mroderick: We don't yet have style and compatibility guidelines for the effort of creating v2.x. I'll try to get that discussion going in a separate thread.

For this PR, I would suggest using [`Array.prototype.filter`][filter] instead of a while loop. `filter` is defined in ES5.1, and we're already using `forEach`, so we have committed ourselves it ES5.1 by now.

```javascript
function matchingFake(fakes, args, strict) {
    if (!fakes || fakes.length === 0) {
        return undefined;
    }

    var matchingFakes = fakes.filter(function (f) {
        return f.matches(args, strict);
    });

    return matchingFakes.pop() || undefined;
}
```

I am aware that the solution iterates over ALL the fakes ... but there's not likely to be so many that we can actually measure the difference under anything other than truly extraordinary circumstances.

I think this refactoring of your changes works (all the tests pass). What do you think?

[filter]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

test/stub-test.js
```diff
@@ -1345,6 +1345,24 @@ describe("stub", function () {
             assert.equals(stub(), 23);
             assert.equals(stub({ foo: "bar", bar: "foo" }), 99);
         });
+
+        it("ensure stub uses last matching arguments", function () {
+            var stub = createStub().returns(23);
```
mroderick: Try to avoid using literal values for tests, but label them using function names and variable names. This removes a lot of confusion situations and questions, like "is this `2` value the same `2` value as above?".

I realise that it's not practiced throughout the Sinon.JS codebase much yet, it's something I am hoping to change, so that it'll be easier for people to do more casual contributions.

Using [v4 UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) values for return values, where we don't care about the exact value has the benefit of making it trivial to search for them in a large codebase, unlike values like `2`, `"username"`, etc.

I've tried applying the advice to the first test.

```javascript
it("ensure stub uses last matching arguments", function () {
    var unmatchedValue = "d3ada6a0-8dac-4136-956d-033b5f23eadf";
    var firstMatchedValue = "68128619-a639-4b32-a4a0-6519165bf301";
    var secondMatchedValue = "4ac2dc8f-3f3f-4648-9838-a2825fd94c9a";
    var expectedArgument = "3e1ed1ec-c377-4432-a33e-3c937f1406d1";
    
    var stub = createStub().returns(unmatchedValue);
    
    stub.withArgs(expectedArgument).returns(firstMatchedValue);
    stub.withArgs(expectedArgument).returns(secondMatchedValue);

     assert.equals(stub(), unmatchedValue);
     assert.equals(stub(expectedArgument), secondMatchedValue);
});
```

What do you think?

test/util/core/called-in-order-test.js
mroderick: 👍

mroderick:
Thank you for your pull request! I just have a few of minor comments

lib/sinon/spy.js
jishi: Yeah, the filter is probably more intuitive, I wasn't aware which target we we're aiming for so I stuck with a known approach since I knew it supposed to run in a browser.

The last `|| undefined` seems unecessary though since pop() returns undefined by default. 

Do you want me to make those changes and update the PR?

test/stub-test.js
jishi: makes sense to me.

lib/sinon/spy.js
mroderick: You're right, `|| undefined` is not necessary.

It'd be great to get the changes in, thanks!

mroderick:
Good stuff, thanks! 👍

BitForger:
Came across this trying to figure out why my chained `.withArgs` statements were always returning the last one. I'm using sinon/mocha with typescript and have a model that i'm trying to mock with static methods. The problem is i need to switch the value returned based on the arguments passed.
```
public static findOne = sinon.stub()
    .withArgs({slug: 'test'}).returns({
      exec: sinon.stub().resolves(new MockFeature()),
    });
```
How would I do that?

fatso83:
Not sure I understand the problem. In any case, this is a usage problem. could you please post this to the sinon tag on the Stackoverflow forum? Lots more people following that than this issue tracker

Issue #1034, opened 2016-04-24
Attempting to (incorrectly) stub a method with an object silently fails and throws no error


serhalp:
- Sinon version: 1.17.3, 2.0.0-pre
- Environment: node 0.10, 0.12, 4, 5 on OS X 10.10.5 (I don't think this matters)
#### What did you expect to happen?

Given

``` javascript
var obj = {foo: function () {}};
```

I'd expect the same error thrown for all these cases:

``` javascript
sinon.stub(obj, 'foo', 3)
sinon.stub(obj, 'foo', 'bar')
sinon.stub(obj, 'foo', true)
sinon.stub(obj, 'foo', {})
```
#### What actually happens

The first three throw `Custom stub should be a function or a property descriptor`:

```
node_modules/sinon/lib/sinon/stub.js:22
                throw new TypeError("Custom stub should be a function or a property descriptor");
                ^
TypeError: Custom stub should be a function or a property descriptor
    at Object.stub (node_modules/sinon/lib/sinon/stub.js:22:23)
```

while the object form does not throw at all, and does not stub the method.
#### How to reproduce

``` javascript
var obj = {foo: function () {}};
sinon.stub(obj, 'foo', {})
```

Here are [a pair of tests, one failing for objects and one passing for primitives](https://github.com/goodeggs/sinon/commit/06cdc022d7dedb50ae88a43c6397d75c98a08a0b).
#### Discussion

As far as I'm able to grok from reading the code, it seems this is because the third argument to `sinon.stub` can be either an actual function or a property descriptor (e.g. `{ value: function () {}, writable: true, enumerable: true, configurable: true }`), but this latter form is [not documented](https://github.com/sinonjs/sinon/blob/36f571f/docs/current/stubs.md#var-stub--sinonstubobject-method-func), so this behavior is unexpected from a user's standpoint.  At the very least, perhaps an error could be thrown when invalid property descriptors are passed in?

cc @demands

PR #1182, opened 2016-11-10
Add documentation for accessor method support for stubs and spies

mroderick:
This is a PR for **master**.

It is a solution for #1095, by documenting Sinon's accessor method support with stubs and spies.

#### Verify the examples

I've created a repository with the code samples used:

1. `git clone https://github.com/mroderick/sinon-demo-getter-setter.git`
1.  `cd sinon-demo-getter-setter`
1. `npm install`
1. `node spy-demo.js`
1. `node stub-demo.js`

PR #1184, opened 2016-11-10
Fail on empty property descriptor

mroderick:
This is a PR for **master**, it is a solution for #1034


mroderick:
Between these changes, and the documentation changes in #1182 I think we can consider #1034 fixed

Issue #1124, opened 2016-08-09
Stubbing and then restoring `getters` causes the wrapped method to be invoked


bradbarrow:
**_Details**_
- Sinon version : 1.17.5
- Other libraries you are using: mocha, chai
- Environment: Browser and Node v5.5.0
- Load method: Script tag and CDN hosted Sinon and / or webpack bundled sinon npm module 

**What did you expect to happen?**

When stubbing a getter on a class, and then restoring the stub, I expect that the stubbed getter (property) will never be invoked/accessed.

**What actually happens**

When I restore the stubbed getter, the original wrapped getter is invoked as a side effect.

**How to reproduce**

Stub a getter, then restore the stub and witness that the original getter is invoked:

https://codepen.io/foxwisp/pen/LkXXzd?editors=0010

**The cause**

https://github.com/sinonjs/sinon/blob/d24a1f1ac7ea500d3d185adb3821786d321f6a92/lib/sinon/util/core.js#L159

When restoring a wrapped method, there's a last ditch attempt to put the original method back that uses strict equality to compare the current value for `object[property]` against the stubbing value `method`. 

When `object[property]` is evaluated, it invokes the getter that was stubbed because it's already been restored using `Object.defineProperty` here:

https://github.com/sinonjs/sinon/blob/d24a1f1ac7ea500d3d185adb3821786d321f6a92/lib/sinon/util/core.js#L154

**Proposed Solutions**
- Don't do the strict check for `hasES5Support` cases...trust that the `Object.defineProperty` will work?
- For `hasES5Support` cases, don't compare the `object[property]` itself, but rather use `getPropertyDescriptor` and strictly compare those to avoid invoking the stubbed getter.

**Why is this an issue**

Often we stub to stub away some complexity that we don't want to have to setup / cost we don't want to incur. If the original method is going to be called by `stub.restore()` we're still going to incur the cost / require the setup to avoid exceptions.

I know there's an argument that getters should be simple and shouldn't even need to be stubbed, but if we can - they should work.

fearphage:
@bradbarrow Thank you for the great level of detail. We appreciate it.

I believe this was already fixed on master. #1059

Someone just needs to port the fix.

mroderick:
While investigating #1177, I created a runnable example for this issue: https://github.com/mroderick/sinon-issue-1177

This still fails in `master`.

mroderick:
[...]

PR #1194, opened 2016-11-21
restoring etters does not invoke the getter #1124

zamnuts:
#### Purpose
Fix issue #1124 "Stubbing and then restoring `getters` causes the wrapped method to be invoked" by reading the "value" descriptor property for equality comparison when checking if the property has been restored. Originally a direct comparison was made which would invoke the getter; this is an alternate way to compare that mitigates the problem, albeit slower.  Only supports ES5; further work with __lookupSetter__ / __lookupGetter__ should be implemented to support lower ECMAScript versions.

1. Check out this branch
2. `npm install`
3. `npm test`

The following unit tests were added to `test/sinon-test.js`:
- "does not invoke getter on restore"
- "does not invoke setter on restore"

#### Warning
I'm not entirely convinced this fixes all scenarios. I was not able to test the logic branch if the comparison is truthy, i.e. `object[property] = wrappedMethod`, but this is a step in the right direction and addresses the immediate need.

### DIFF
--- a/lib/sinon/util/core.js
+++ b/lib/sinon/util/core.js
@@ -167,9 +167,17 @@
                     Object.defineProperty(object, property, wrappedMethodDesc);
                 }
 
+                // this only supports ES5 getter/setter, for ES3.1 and lower
+                // __lookupSetter__ / __lookupGetter__ should be integrated
+                if (hasES5Support) {
+                    var checkDesc = sinon.getPropertyDescriptor(object, property);
+                    if (checkDesc.value === method) {
+                        object[property] = wrappedMethod;
+                    }
+
                 // Use strict equality comparison to check failures then force a reset
                 // via direct assignment.
-                if (object[property] === method) {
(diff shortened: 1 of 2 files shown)


lib/sinon/util/core.js
```diff
@@ -167,9 +167,17 @@
                     Object.defineProperty(object, property, wrappedMethodDesc);
                 }
```
fatso83: ES5 is sufficient. The 1.17.x branch is in maintenance and any new bigger work should go into the `master` branch, and we won't support pre-ES5 environments in that anyway.

test/sinon-test.js
```diff
@@ -193,6 +193,38 @@
                 sinon.stub(object, "method");
 
                 assert.isFalse(object.method.called);
+            },
```
fatso83: Nice and readable. Thanks!

fatso83:
Thanks for your work.  Much appreciated!

Issue #1179, opened 2016-11-08
Expected arguments to be called with diff


jdgreenberger:
Using sinon 1.17.6 in Node.JS with sinon-chai. When expecting function arguments (i.e. expect(stub).to.have.been.calledWith) if a test fails, the thrown error contains the entire set of function arguments. If these arguments are big, it can be difficult to trace down the diff. Would be happy to submit my own PR, but was wondering if there is intention to improve this error by iterating through function arguments and specifying the exact diff which causes the expectation to fail.

Example sinon response: https://gist.github.com/jdgreenberger/b7ade89cda2762a2c02944e81457f62e
Ideal sinon response: https://gist.github.com/jdgreenberger/4f84c3803bc376c55a4b772fb68455a6

mroderick:
In to help reviewers to better understand what you're asking, could you add a few examples to this issue?

jdgreenberger:
Updated @mroderick

mroderick:
I think I better understand what you're after now.

Check out http://unexpected.js.org, they have really excellent diffs.

If I read their `package.json` correctly, then they're using https://github.com/kpdecker/jsdiff#readme

A contribution that uses a stable module like that would be most welcome.

PR #1203, opened 2016-12-07
Add expectation diff logs

jdgreenberger:
#### Purpose

Addresses issue #1179 by using npm libraries "diff" and "colors" to display colored diffs when arguments to `sinonSpy.calledWith` do not match the expectation.

#### Background
Took a stab at implementing diffs for the base case of calledWith. In order to limit changes in this PR, there's a temporary hack to use the original formatting when arguments include sinon matchers. Would appreciate feedback on a better way to implement this, and any other architectural improvements.

#### Solution  

Creates a new formatting type `%D` on spyApi.formatters in  `spy.js` to take the diff of expected and called arguments and created a useful colored message.

Example:

```diff
expected doSomething to be called with arguments 
[{
  first: "a",
-  mismatchKey: true,
+  mismatchKeyX: true,
  second: { nest: true },
-  third: [{ fourth: { nest: true } }]
+  third: [{ fourth: { nest: false } }]
}, "fifth"] 
```
#### How to verify -
Check the added tests in assert-test.js
1. Check out this branch 
2. `npm install`
3. mocha test/assert-test.js

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -199,7 +199,7 @@ mirrorPropAsAssertion(
 );
 mirrorPropAsAssertion("calledWithNew", "expected %n to be called with new");
 mirrorPropAsAssertion("alwaysCalledWithNew", "expected %n to always be called with new");
-mirrorPropAsAssertion("calledWith", "expected %n to be called with arguments %*%C");
+mirrorPropAsAssertion("calledWith", "expected %n to be called with arguments %D");
 mirrorPropAsAssertion("calledWithMatch", "expected %n to be called with match %*%C");
 mirrorPropAsAssertion("alwaysCalledWith", "expected %n to always be called with arguments %*%C");
 mirrorPropAsAssertion("alwaysCalledWithMatch", "expected %n to always be called with match %*%C");
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -79,6 +79,7 @@ function mirrorPropAsAssertion(name, method, message) {
         var failed = false;
 
         verifyIsValidAssertion(name, args);
+
         if (typeof method === "function") {
             failed = !method(fake);
         } else {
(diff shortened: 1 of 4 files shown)


jdgreenberger:
@mroderick excited to get this in! Let me know what needs to be changed 😃

lib/sinon/assert.js
```diff
@@ -78,6 +78,11 @@ function mirrorPropAsAssertion(name, method, message) {
         var args = slice.call(arguments, 1);
         var failed = false;
 
+        // Hack to be replaced
```
fatso83: This seems like a TODO statement. No one ever does anything to a TODO statement. Could you rather expand on your comment? I am not totally sure what this does.

jdgreenberger: Lines 82-84 check for one of the sinon match objects, and if found, revert to using the original message formatting without colored diffs. I did this to avoid making a lot of changes in a single PR, and because the choice to print a diff for a sinon match comparison is a little bit more ambigious than comparing two plain objects (What do I print for `expect({x: true}).to.be.calledWith(match.falsy)`?). I was hoping one of the contributors could suggest a less hacky way to reserve the diff formating just for plain object comparison.

startswithaj:
@jdgreenberger great work on this. LGTM.

lib/sinon/assert.js
fearphage: 1. If this is explicitly to catch matchers, can you use the built-in `sinonMatch.isMatcher(...)` on line 82? 
2. Can you update the comment to say why you're doing this (like some of the posted comment above) instead of this one?

lib/sinon/assert.js
```diff
@@ -199,7 +204,7 @@ mirrorPropAsAssertion(
 );
 mirrorPropAsAssertion("calledWithNew", "expected %n to be called with new");
 mirrorPropAsAssertion("alwaysCalledWithNew", "expected %n to always be called with new");
-mirrorPropAsAssertion("calledWith", "expected %n to be called with arguments %*%C");
+mirrorPropAsAssertion("calledWith", "expected %n to be called with arguments %D");
```
fearphage: Why only `calledWith`? Why not `calledWithExactly` and others?

jdgreenberger:
@fearphage knowing about `.isMatcher`, it wasn't much harder to use colored diffs for the matchers. Updated PR to apply formatting to other relevant `calledWith` methods

fatso83:
This now looks to be in great IMHO! Could I ask you just to clean up the commit history before merging? Just do an interactive rebase, or whatever you prefer, to get rid of the cleanup log commits. 

PS. I am not asking you to squash _all_ the commits. They tell a good story.

jdgreenberger:
@fatso83 updated!

fatso83:
Perfect! Thanks for the work on this.

jdgreenberger:
No problem! Thanks for the review 😄

PR #1204, opened 2016-12-14
Fix being able to spy Error

estobbart:
`node -e "require('sinon').spy(global, 'Error');"` caused a recursion.

When simulating a PhantomJS/QtWebKit environment, Error's behave differently. To achieve this simulated environment when testing, Error could be stubbed.

2. `npm install`
3. `node -e "require('sinon').spy(global, 'Error');"`
Should return as expected, also see tests.


fatso83:
Thanks!

PR #1205, opened 2016-12-15
Fix spying on *etter methods

### LINKED ISSUE #1018: Stub of an object property with getter/setter doesn't work
If there's a method that was defined with `defineProperty` with a `value` it works, but not with `get`/`set`.

Error is "Attempted to wrap undefined property fnGetterSetter as function".

``` js

function factory() {
  let object = {}

  Object.defineProperty(object, 'fnValue', { configurable: true, value() { return 'fnValue' } })
  let fnGetterSetter = () => 'fnGetterSetter'
  Object.defineProperty(object, 'fnGetterSetter', {
    get() { return fnGetterSetter },
    set(newValue) { fnGetterSetter = newValue }
  })

  return object
}

let o1 = factory()

console.log('o1.fnValue: ' + o1.fnValue()) // => "fnValue"
console.log('o1.fnGetterSetter: ' + o1.fnGetterSetter()) // => "fnGetterSetter"
o1.fnGetterSetter = () => 'new fn'
console.log('o1.fnGetterSetter: ' + o1.fnGetterSetter()) // => "new fn"

let o2 = factory()

sinon.stub(o2, 'fnValue').returns('stubbed fnValue')
console.log('o2.fnValue: ' + o2.fnValue()) // works, => "stubbed fnValue"

sinon.stub(o2, 'fnGetterSetter').returns('stubbed fnGetterSetter') // fails - Attempted to wrap undefined property fnGetterSetter as function
console.log('o2.fnValue: ' + o2.fnGetterSetter())
```

Runnable code: https://tonicdev.com/56e1cbdb104098110026e06c/56fd90d539b35a13000ede79

lucasfcosta:
#### Purpose (TL;DR)
This PR aims to fix #1018 on `v1.17`.
Basically I just needed to pass a new descriptor with only the property from the propDescriptor which is going to be spied on.
I also added tests to prove both stubbing and spying work as @mroderick did on the issue itself.

The problem happened because the property descriptor being passed to be spied on had non-function property values (because it was a complete property descriptor taken from that property) and then that caused an error.

#### Solution
As I've said above I just needed to passa a property descriptor with only the property which was going to be spied instead of the whole property descriptor

#### How to verify
2. `npm install`
3. `npm run test` -> This will run the new tests I've added

--------------------------------------------

Thanks for the awesome work you have been doing on Sinon, I'm looking forward to contribute more with you, please let me know if you need anything!

**I've read the contribution guidelines but I'm not sure this PR meets your commit/coding patterns, please let me know if I've done anything wrong or missed anything and I'll happily fix it**. You rock ❤️  

EDIT: The build seems to be failing on CI for an unknown reason. I've read the log and I think it might be related to Travis itself and not this PR. Do you have any idea why this is happening?

### DIFF
--- a/test/spy-test.js
+++ b/test/spy-test.js
@@ -2226,6 +2226,34 @@
 
                 assert.equals(spy.length, 3);
             }
+        },
+
+        "getters and setters": {
+            "it is possible to spy getters": function () {
+                var object = {
+                    get myProperty() {
+                        return "fb41a645-24fb-4580-b4c0-17d71ecc1c74";
+                    }
+                };
+
+                var descriptor = sinon.spy(object, "myProperty", ["get"]);
+                var value = object.myProperty; // eslint-disable-line no-unused-vars
+
+                assert.equals(descriptor.get.callCount, 1);
+            }
+        },
+
+        "it is possible to spy setters": function () {
+            var object = { // eslint-disable-line accessor-pairs
+                set myProperty(val) {
+                    this.otherProperty = val * 2;
+                }
+            };
+
+            var descriptor = sinon.spy(object, "myProperty", ["set"]);
+            object.myProperty = 10; // eslint-disable-line no-unused-vars
+
+            assert.equals(descriptor.set.callCount, 1);
         }
     });
 }(this));
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -31,11 +31,15 @@
             }
 
             if (types) {
+                // A new descriptor is needed here because we can only wrap functions
+                // By passing the original descriptor we would end up trying to spy non-function properties
+                var descriptor = {};
                 var methodDesc = sinon.getPropertyDescriptor(object, property);
+
                 for (var i = 0; i < types.length; i++) {
-                    methodDesc[types[i]] = spy.create(methodDesc[types[i]]);
+                    descriptor[types[i]] = spy.create(methodDesc[types[i]]);
                 }
-                return sinon.wrapMethod(object, property, methodDesc);
+                return sinon.wrapMethod(object, property, descriptor);
             }
 
             return sinon.wrapMethod(object, property, spy.create(object[property]));
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -1796,6 +1796,47 @@
 
                 assert.equals(stub.length, 3);
             }
+        },
+
+        "getters and setters": {
+            "it is possible to stub getters": function () {
+                var object = {
+                    get myProperty() {
+                        return "4fb82903-6222-45e4-bfeb-a536bfc9734a";
+                    }
+                };
+
+                function fakeGet() {
+                    return "faked myProperty";
+                }
+
+                var descriptor = sinon.stub(object, "myProperty", {
+                    get: fakeGet
+                });
+
+                assert.equals(object.myProperty, "faked myProperty");
+                assert.equals(descriptor.get.callCount, 1);
+            },
+
+            "it is possible to stub setters": function () {
+                var object = { // eslint-disable-line accessor-pairs
+                    set myProperty(val) {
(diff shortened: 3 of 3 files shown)


fatso83:
Sorry for the slow followup ... you are right about the failing build. Restarting the Travis build seems to have fixed the failing tests.

fatso83:
@mroderick objections to merging this? I could like a little comment or two, other than that ... 👍

lucasfcosta:
@fatso83 I made some comments while studying this code and doing the fix, but since the code didn't have any comments before I assumed those pieces of information would be considered unnecessary.
I'd be happy to add a few more comments to this code if you want me to, all you gotta do is let me know 😄 

Thanks for the feedback!

test/spy-test.js
```diff
@@ -2226,6 +2226,34 @@
 
                 assert.equals(spy.length, 3);
             }
+        },
+
+        "etters": {
```
fearphage: I don't like the naming here. Either `getters and setters` or a `getters` section separate from the `setters` section.

fatso83:
@lucasfcosta, as long as comments document the intent, not the steps of the implementation, I think it usually helps the understanding of the reader. If you can add a small comment I think it would help future readers understanding the code. I also wasn't a fan of the "etters" naming, but it's not a big thing. Just change it per fearphage's comment and this should be good to merge.

test/spy-test.js
lucasfcosta: You're right! It really looks better, thanks for the feedback.

lucasfcosta:
Here it is!

I made sure to add two concise comment lines explaining why we needed a new `descriptor` object whenever trying to spy on property descriptor's properties.

I also renamed both of those tests containing the `getter` and `setter` specific tests to `getters and setters`, it really looks better than `*etters`.

Thanks for the feedback @fatso83 and @fearphage, let me know if I missed anything.

fatso83:
Great work, and thanks for listening to the input. Much appreciated.

muuki88:
Hi,

Is there a migration path on how to implement this in the new `2.x.x` version series? I see that [the tests ignore the deprecation warnings](https://github.com/lucasfcosta/sinon/blob/master/test/stub-test.js#L967-L986).

Thanks a lot :)

lucasfcosta:
Hi @muuki88, this is only a fix for the v1.17 version.
The implementation for this feature in `v2` has happened at #1297.

These are the docs on how to implement getter and setters spies and stubs:

#### `stub.get(getterFn)`

Replaces a new getter for this stub.

```javascript
var myObj = {
    prop: 'foo'
};

createStub(myObj, 'prop').get(function getterFn() {
    return 'bar';
});

myObj.example; // 'bar'
```

#### `stub.set(setterFn)`

Defines a new setter for this stub.

```javascript
var myObj = {
    example: 'oldValue',
    prop: 'foo'
};

createStub(myObj, 'prop').set(function setterFn(val) {
    myObj.example = val;
});

myObj.prop = 'baz';

myObj.example; // 'baz'
```

muuki88:
Thanks a lot @lucasfcosta . This works for me. The confusing was due to the missing typescript definitions for the `get` and `set` methods on the stub. I tried the `callsFake`, which didn't worked as expected.

Should I open an issue for adding the type signatures to the `SinonStub` interface?

test/spy-test.js
```diff
@@ -2226,6 +2226,34 @@
...
+                };
+
+                var descriptor = sinon.spy(object, "myProperty", ["get"]);
+                var value = object.myProperty; // eslint-disable-line no-unused-vars
+
+                assert.equals(descriptor.get.callCount, 1);
+            }
+        },
+
+        "it is possible to spy setters": function () {
```
fearphage: The indentation is off here. This should have the same indentation as the test above it and the preceding closing brace should be moved below this test.

PR #1207, opened 2016-12-17
Replace `stub(o, 'm', fn)` with `stub(o, 'm').callsFake(fn)`

hurrymaplelad:
Replace
```
var stub = sinon.stub(foo, 'bar', myCustomFunc); // does not return a real stub
```
with
```
var stub = sinon.stub(foo, 'bar').callsFake(myCustomFunc); // returns a real stub
```
as discussed [here](https://github.com/sinonjs/sinon/pull/823).

Discussed [here](https://github.com/sinonjs/sinon/pull/823).

#### Solution  - optional
Adds a new `callsFake()` behavior, stashed on the internal field `fakeFn` of stubs.

2. `npm install`
3. `npm test`

### DIFF
--- a/docs/current/migrating-to-2.0.md
+++ b/docs/current/migrating-to-2.0.md
@@ -22,6 +22,18 @@ var myFakeServer = sinon.fakeServer.create({
 ## sinon.test, sinon.testCase and sinon.config Removed
 `sinon.test` and `sinon.testCase` have been extracted from the Sinon API and moved into their own node module, [sinon-test](https://www.npmjs.com/package/sinon-test). Please refer to the [sinon-test README](https://github.com/sinonjs/sinon-test/blob/master/README.md) for migration examples.
 
+## stub.callsFake replaces stub(obj, 'meth', fn)
+`sinon.stub(obj, 'meth', fn)` return a spy, not a full stub. Behavior could not be redefined. `stub.callsFake`
+now returns a full stub. Here's a [codemod script](https://github.com/hurrymaplelad/sinon-codemod) to help you migrate.
+See [discussion](https://github.com/sinonjs/sinon/pull/823).
+
+```js
+// Old
+sinon.stub(obj, 'meth', fn);
+// New
+sinon.stub(obj, 'meth').callsFake(fn);
+```
+
 ## Deprecation of internal helpers
 The following utility functions are being marked as deprecated and are planned for removal in Sinon v3.0; please check your codebase for usage to ease future migrations:
 
@@ -45,4 +57,3 @@ The following utility functions are being marked as deprecated and are planned f
 * `sinon.ProgressEvent`
 * `sinon.typeOf`
 * `sinon.extend`
-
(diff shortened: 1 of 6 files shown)


hurrymaplelad:
As noted in docs changes, codemod script here:
https://github.com/hurrymaplelad/sinon-codemod

fatso83:
This PR is going into generate some noise in the future when stuff breaks due to the non-compatibility. I think the new interface is better, but are we better of leaving the old stuff in with a deprecation warning or should we just keep it clean? 

Other than that, thanks, @hurrymaplelad!

hurrymaplelad:
A deprecation warning seems reasonable to me. Are there others who should chime in? Do we have a usual channel for such a warning? `typeof console !== 'undefined' && console.warn`?

fatso83:
We put a deprecation warning on some methods when @jonnyreeves did a refactoring of the project to common js. You can probably use that util somehow [by looking into what he did](https://github.com/sinonjs/sinon/blob/master/lib/sinon.js#L46).

lib/sinon/stub.js
```diff
@@ -19,12 +19,15 @@ var functionToString = require("./util/core/function-to-string");
 var valueToString = require("./util/core/value-to-string");
 var wrapMethod = require("./util/core/wrap-method");
 
-function stub(object, property, funcOrDescriptor) {
-    if (!!funcOrDescriptor && typeof funcOrDescriptor !== "function" && typeof funcOrDescriptor !== "object") {
-        throw new TypeError("Custom stub should be a function or a property descriptor");
```
fatso83: This should probably not throw an error at this point in time, and rather print a deprecation warning using `lib/sinon/util/core/deprecated`. We can do the more drastic bit later on.

docs/current/migrating-to-2.0.md
```diff
@@ -22,6 +22,18 @@ var myFakeServer = sinon.fakeServer.create({
 ## sinon.test, sinon.testCase and sinon.config Removed
 `sinon.test` and `sinon.testCase` have been extracted from the Sinon API and moved into their own node module, [sinon-test](https://www.npmjs.com/package/sinon-test). Please refer to the [sinon-test README](https://github.com/sinonjs/sinon-test/blob/master/README.md) for migration examples.
 
+## stub.callsFake replaces stub(obj, 'meth', fn)
+`sinon.stub(obj, 'meth', fn)` return a spy, not a full stub. Behavior could not be redefined. `stub.callsFake`
```
fatso83: This is great. Thanks for making the mod script!

lib/sinon/stub.js
hurrymaplelad: Done!

fatso83:
I can also create a separate PR with the suggested `deprecated.printWarning` if you feel it doesn't belong in this?

hurrymaplelad:
`printWarning` extracted. Agree it's cleaner!

hurrymaplelad:
🎉

mroderick:
Great PR, thanks @hurrymaplelad

shaunc:
NB the deprecation warning is misleading for getters & setters. For instance, rather than `foo(obj, 'meth', {get: fakeGetter})` => `foo(obj, meth).callsFake(fakeGetter)` the proper translation is `foo(obj, 'method').get(fakeGetter)`. Also I think the code mod will fail for such cases.

gabrieledarrigo:
Mhh to be honest the API is quite ugly : P
But thanks!

Issue #1113, opened 2016-07-28
Arrays matched with deepEqual


Elberet:
Consider:

``` javascript
var actual = ['foo'];
actual.hasCustomProperty = true.

var match = sinon.match(['foo']);
match.test(actual); // => false
```

According to the docs, this test should succeed: `actual` contains and matches all of the properties defined on the expectation object, yet the matcher uses `deepEqual` to compare arrays and stumbles over the additional property on `actual`.

PR #316 to compare arrays without `deepEqual` had been accepted way back when in 2013, but that behaviour is missing from the current code. Is this a regression or was this reverted intentionally?

mantoni:
Can you please state which Sinon version you're using? `1.17.5` or `2.0.0-pre`?

Elberet:
I've just upgraded from `v1.17.4` to `v1.17.5`, but the behaviour is the same as before (as expected from a quick glance at the commits between those two releases).

fatso83:
@Elberet: looked into this now. That test case started failing with the code from #316 (e4fdd41). It worked before that. I [automated a test](https://gist.github.com/fatso83/7f135d2de168a07582b075b0fa46e9a9) to find this, if you want to verify.

fatso83:
@Elberet : To which documentation are you referring? The [Sinon homepage](http://sinonjs.org/docs/#sinon-match-api) does not seem to mention either `deepEquals` nor `matcher#test`, so I am curious to find out where you got this impression from.

PR #316 does not avoid using `deepEquals` on arrays, it _makes sure_ `deepEquals` is being used for _both_ array elements _and_ attributes. Both the code, behavior and tests that were introduced then are still present.

`deepEquals` is a two-way equality, which you can see from the fact that [the amount of props need to be equal](https://github.com/sinonjs/sinon/blob/v1.17/lib/sinon/util/core.js#L255) (after checking equality one-way first). If `b` has props that `a` does not have the length will simply not match. It as been like that since the first usable version of Sinon back in April 2010 (84dbb22).

So it seems like what you want is to have `matcher#test` be a one-way equality testing of sorts. That seems more like a feature request than a bug, unless I am misunderstanding you somehow?

Elberet:
You are right about PR #316. I was too hasty and only noticed that a call to deepEqual got removed.

Considering the doc:
> **sinon.match(object)**
> Requires the value to be not null or undefined and have at least the same properties as `expectation`.

[...]

PR #1208, opened 2016-12-19
Array matchers

lucasfcosta:
#### Purpose (TL;DR)
This adds the following array matchers as indicated by @mroderick at #1113:

> * `match.array.equals()`: matches when the arrays are equal (unstrict)
> * `match.array.startWith()`: matches sequence against start of array
> * `match.array.endsWith()`: matches sequence against end of array
> * `match.array.contains()`: matches sequence anywhere in the array

The `equals` matcher has also been renamed to `deepEquals`, because as @fearphage suggested the `deepEquals` name is more explicit since we're not doing a `strict` comparison between array elements.

#### Solution

I added a new matcher for each one of the suggested items and tests for each one of them too.

Here go a few considerations about the implementation:

* At the [deepEquals](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:array-matchers?expand=1#diff-562e004061f0ab49a70d3728469de642R229) matcher I've added a `length` check to avoid iterating through array items in case they don't have the same length. This makes it faster to spot differences in most cases and also allows me to iterate through every item in `actual` later to finish the check
* At the [endsWith](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:array-matchers?expand=1#diff-562e004061f0ab49a70d3728469de642R246) matcher I've used an offset in order to know from where we should start comparing items in order to match the last items in the `actual` array. I've added a comment to make that clear but let me know if you think that comment is unnecessary
* At the [contains](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:array-matchers?expand=1#diff-562e004061f0ab49a70d3728469de642R254) matcher I'm using a simple `indexOf` call to find items into the `actual` array. I thought about mapping the `actual` array's items using an object (`O(n)`) in order to traverse it only once since we would then be able to find each element by an `O(1)` factor. Since we would also have to traverse the whole `expectation` array I believe that would be `O(n * m)` where `n` is `actual`'s length and `m` is `expectation`'s length. Let me know if I said anything wrong or if I'm overthinking. I avoided that solution because I was not sure it would be too much over engineering.
* When creating a `message` for each one of the new matchers I've just used `<nameOfMatcher>([x,y,z])`, please let me know if you don't like that output format.
* I also invoked `Array.prototype`'s methods by using `call` to avoid incompatibilities with fake arrays or overwritten methods

#### How to verify
2. `npm install`
3. `npm run test` will run the tests I added to `test/match-test.js`.

Please let me know if I missed anything or if you disagree with any of these changes and I'll be more than happy to fix this PR.

Thanks everyone for reading this. You rock, thanks for maintaining this awesome project ❤️

### DIFF
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -14,6 +14,9 @@ var functionName = require("./util/core/function-name");
 var typeOf = require("./typeOf");
 var valueToString = require("./util/core/value-to-string");
 
+var every = Array.prototype.every;
+var arrayToString = Array.prototype.toString;
+
 function assertType(value, type, name) {
     var actual = typeOf(value);
     if (actual !== type) {
@@ -225,11 +228,12 @@ match.array = match.typeOf("array");
 
 match.array.deepEquals = function (expectation) {
     return match(function (actual) {
-        // Comparing the length of both is the fastest way to spot a difference before iterating through every item
-        return actual.length === expectation.length && actual.every(function (element, index) {
+        // Comparing lengths is the fastest way to spot a difference before iterating through every item
+        var sameLength = actual.length === expectation.length;
+        return sameLength && every.call(actual, function (element, index) {
             return expectation[index] === element;
         });
-    }, "deepEquals([" + Array.prototype.toString.call(expectation) + "])");
+    }, "deepEquals([" + arrayToString.call(expectation) + "])");
 };
 
 match.array.startsWith = function (expectation) {
@@ -259,6 +263,8 @@ match.array.contains = function (expectation) {
     }, "contains([" + Array.prototype.toString.call(expectation) + "])");
 };
 
+=======
+>>>>>>> Add deepEquals array matcher
 match.bool = match.typeOf("boolean");
 match.number = match.typeOf("number");
 match.string = match.typeOf("string");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -238,33 +238,12 @@ match.array.deepEquals = function (expectation) {
 
 match.array.startsWith = function (expectation) {
     return match(function (actual) {
-        return expectation.every(function (expectedElement, index) {
+        return every.call(expectation, function (expectedElement, index) {
             return actual[index] === expectedElement;
         });
-    }, "startsWith([" + Array.prototype.toString.call(expectation) + "])");
+    }, "startsWith([" + arrayToString.call(expectation) + "])");
 };
 
-match.array.endsWith = function (expectation) {
-    return match(function (actual) {
-        // This indicates the index in which we should start matching
-        var offset = actual.length - expectation.length;
-
-        return expectation.every(function (expectedElement, index) {
-            return actual[offset + index] === expectedElement;
-        });
-    }, "endsWith([" + Array.prototype.toString.call(expectation) + "])");
-};
-
-match.array.contains = function (expectation) {
-    return match(function (actual) {
-        return expectation.every(function (expectedElement) {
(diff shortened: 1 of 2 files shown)


fatso83:
This is great. Just want to have it browsed by someone else than me :-)

test/match-test.js
```diff
@@ -623,6 +623,70 @@ describe("sinonMatch", function () {
...
+                var deepEquals = sinonMatch.array.deepEquals([1, 2, 3]);
+
+                assert(sinonMatch.isMatcher(deepEquals));
+                assert.equals(deepEquals.toString(), "deepEquals([1,2,3])");
+            });
+
+            it("matches arrays with the exact same elements", function () {
+                var deepEquals = sinonMatch.array.deepEquals([1, 2, 3]);
+                assert(deepEquals.test([1, 2, 3]));
+                assert(!deepEquals.test([1, 2]));
```
fearphage: The negation here is easy to overlook. Could you change these to `assert.isFalse` throughout?

lib/sinon/match.js
```diff
@@ -221,12 +221,50 @@ match.hasOwn = createPropertyMatcher(function (actual, property) {
     return actual.hasOwnProperty(property);
 }, "hasOwn");
 
+match.array = match.typeOf("array");
+
+match.array.deepEquals = function (expectation) {
+    return match(function (actual) {
+        // Comparing lengths is the fastest way to spot a difference before iterating through every item
+        var sameLength = actual.length === expectation.length;
+        return sameLength && Array.prototype.every.call(actual, function (element, index) {
```
fearphage: Any reason you didn't use `actual.every(....)`? Same question for all the array methods here (`indexOf`, `toString`, etc').

lucasfcosta: Sometimes people end up creating iterables without those methods, so if we just used `actual.every` or other methods directly on `actual` they might fail due to the lack of them being implemented. However they would still work when called this way. An example for this would be `arguments`. In order to make it an array and call these methods on it we need to use `Array.prototype.slice.call` before, since it doesn't have the other methods implemented.

Also, I think this approach is more functional and explicit, but this second reason is just a matter of personal opinion.

lucasfcosta: However, I was just thinking if we should act upon those iterables as if they were arrays, after all this marcher has the word `array` in it.

Please let me know what you think, I can change this without problems if you want me to :smile:

Also, thank you both for the feedback, it's always good to have such a careful review. I usually learn a lot with it.

fatso83: Well, you could still cache the references to the functions to make it a bit less verbose. 

```
const proto = Array.prototype;
const every = proto.every;
```

lucasfcosta: @fatso you're right!
I've just seen you doing that on the `behavior` module for `push` and it looks much better.
I think I'll start doing this too the next time I have to call `Array` methods like this. Thanks for the tip, I'm gonna update this PR as soon as I get home :smile:

lucasfcosta:
Hi friends!
Here are the changes you have requested.

Thanks for the review, your considerations were great, specially the idea of assigning `Array.prototype`'s methods to other variables in order to make the code less verbose. I tried to follow the same pattern you used on [the `behavior.js` file](https://github.com/sinonjs/sinon/blob/master/lib/sinon/behavior.js#L16-L17) for doing that.

Let me know if I missed something and I'll happily fix it, thanks again for reviewing 😄

fearphage:
@lucasfcosta Thank you for the contribution.

Issue #817, opened 2015-08-25
One-liner conditional stub returns an unconditional stub


searls:
When a conditional stub is setup in a one-liner (such that the value that's retained by the test is whatever gets returned by `returns()`), then the stub is in fact _unconditional_, which is very surprising and almost definitely wrong.
## Expected

When I do this in a test:

```
stub = sinon.stub().withArgs("foo").returns("bar")`
```

I expect:

```
expect(stub("not foo")).toBe(undefined)
```

But it is actually `"bar"`.
## Workaround

Apparently, what every Sinon user has been doing to this point is:

```
stub = sinon.stub()
stub.withArgs("foo").returns("bar")`
```

Which behaves as you'd expect, but is (IMHO) needlessly verbose.
## Contributor guideline notes
- Env: Affects all versions
- Version: 1.16.1
- How: Using console in the homepage, using npm, using the download script
- Other libs: no

cjohansen:
This has probably snuck in with the `withArgs` modifier. I agree that the current behavior is less than optimal and should probably be changed. However, that would be a _very_ breaking change, and is not likely to happen anytime soon.

The reason for this behavior is the following:

```js
sinon.stub().withArgs("foo").returns("bar").withArgs("fooz").returns("ball")
```

This is a great illustration of the pitfalls of highly chainable APIs, and in this case,Sinon has unfortunately opted to support the wrong case IMO.

Anyway, as I said, too many people depend on this functionality at this point to allow us to just "fix" it, but we certainly should consider fixing it long term.

searls:
I understand the tension and appreciate your candor (agreeing but also favoring stability).

I think in the short-term what I'll do with my own project using Sinon.js is to wrap this API with functions that configure the narrow set of doubles I want to see, which will work around this issue.

The only recommendation I have for you as a maintainer is perhaps to consider an alternative API for setting up a one-and-done stub configuration.

jasonkarns:
Could this be addressed in a major version bump? I have many of these two-liner instances throughout my test suite that I would much prefer to make one-liners.

And I'll toss in my *suspicion* that there are more people who would prefer `withArgs`/`returns` to return the narrow stub than there are people who do multiple `withArgs` chaining. I don't think I've ever seen multiple `withArgs` in a single chain in the wild, though I encounter the two-line method frequently.

mroderick:
[...]

PR #1209, opened 2016-12-19
Avoid unnecessary empty stub creation

lucasfcosta:
#### Purpose (TL;DR)
While investigating #817 I found a little typo and I also found out that whenever we don't have neither an `object` nor a `property` we were still invoking `stub.create` unnecessarily due to the `else` clause of the `if (func)` statement.

#### Solution
In order to avoid that I moved the empty stub creation immediately before that `if` statement.

#### How to verify
2. `npm install`
3. `npm run test` all tests will still be passing

Please let me know if I missed anything and therefore this improvement is not valid.

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -289,7 +289,7 @@ var spyApi = {
 
         var original = this;
         var fake = this.instantiateFake();
-        fake.matchingAguments = args;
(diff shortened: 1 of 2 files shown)


fatso83:
Seems like a nice, little improvement. Thanks!

PR #1210, opened 2016-12-20
Add docs for new array matchers

lucasfcosta:
#### Purpose (TL;DR)

This aims to add docs for each one of the changes made at #1208 (which adds new Array Matchers).

#### Solution

I've read [the `CONTRIBUTING.md` file in the `docs` folder](https://github.com/sinonjs/sinon/blob/master/docs/CONTRIBUTING.md) and I think this is the correct place to add the new docs, but I'm not 100% sure, so please correct me if I'm wrong and I'll add these pieces of information to the right place.

Also, I'm not a native english speaker so please let me know if you find any spelling/grammar mistakes. It would be much appreciated 😄 

PS: I've made this in a single commit, but I can split it up into multiple commits, one for each matcher, if you find it to be better.

#### How to verify

2. Go to the docs folder by running `cd docs` on Sinon's root folder
3. Ensure you have bundler by running `gem install bundler`
4. Install dependencies by running `bundle install`
5. Run a local server by running `bundle exec jekyll serve`
6. Go to `http://localhost:4000/releases/v2.0.0-pre.4/matchers/` and check the new docs

### DIFF
--- a/docs/release-source/release/matchers.md
+++ b/docs/release-source/release/matchers.md
@@ -112,6 +112,26 @@ Requires the value to be a `Function`.
 Requires the value to be an `Array`.
 
 
+#### `sinon.match.array.deepEquals(arr)`
+
+Requires an `Array` to be deep equal another one.
+
+
+#### `sinon.match.array.startsWith(arr)`
+
+Requires an `Array` to start with the same values as another one.
+
+
+#### `sinon.match.array.endsWith(arr)`
+
+Requires an `Array` to end with the same values as another one.
+
+
+#### `sinon.match.array.contains(arr)`
+
+Requires an `Array` to contain each one of the values the given array has.
+
+
 #### `sinon.match.regexp"
 
 Requires the value to be a regular expression.
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -239,7 +239,7 @@ match.array.deepEquals = function (expectation) {
 
 match.array.startsWith = function (expectation) {
     return match(function (actual) {
-        return every.call(expectation, function (expectedElement, index) {
+        return typeOf(actual) === "array" && every.call(expectation, function (expectedElement, index) {
             return actual[index] === expectedElement;
         });
     }, "startsWith([" + arrayToString.call(expectation) + "])");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -250,7 +250,7 @@ match.array.endsWith = function (expectation) {
         // This indicates the index in which we should start matching
         var offset = actual.length - expectation.length;
 
-        return every.call(expectation, function (expectedElement, index) {
+        return typeOf(actual) === "array" && every.call(expectation, function (expectedElement, index) {
             return actual[offset + index] === expectedElement;
         });
     }, "endsWith([" + arrayToString.call(expectation) + "])");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -258,7 +258,7 @@ match.array.endsWith = function (expectation) {
 
 match.array.contains = function (expectation) {
     return match(function (actual) {
-        return every.call(expectation, function (expectedElement) {
+        return typeOf(actual) === "array" && every.call(expectation, function (expectedElement) {
             return indexOf.call(actual, expectedElement) !== -1;
         });
     }, "contains([" + arrayToString.call(expectation) + "])");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -231,7 +231,7 @@ match.array.deepEquals = function (expectation) {
     return match(function (actual) {
         // Comparing lengths is the fastest way to spot a difference before iterating through every item
         var sameLength = actual.length === expectation.length;
-        return sameLength && every.call(actual, function (element, index) {
+        return typeOf(actual) === "array" && sameLength && every.call(actual, function (element, index) {
             return expectation[index] === element;
         });
     }, "deepEquals([" + arrayToString.call(expectation) + "])");
--- a/test/match-test.js
+++ b/test/match-test.js
@@ -638,6 +638,11 @@ describe("sinonMatch", function () {
                 assert.isFalse(deepEquals.test([1, 2]));
                 assert.isFalse(deepEquals.test([3]));
             });
+
+            it("fails when passed a non-array object", function () {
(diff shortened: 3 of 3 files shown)


docs/_releases/v2.0.0-pre.4/matchers.md
```diff
@@ -112,6 +112,26 @@ Requires the value to be a `Function`.
 Requires the value to be an `Array`.
 
 
+#### `sinon.match.array.deepEquals(iterable)`
+
+Requires an iterable value to be deep equal another one.
```
fearphage: Once you see it written down, this is weird. You're checking inalterable from the array matcher. I would expect the array namespace to contain array matchers.

lucasfcosta: @fearphage I agree, I also raised the same question in [a comment](https://github.com/sinonjs/sinon/pull/1208#discussion_r93198894) at my PR at #1208.

> However, I was just thinking if we should act upon those iterables as if they were arrays, after all this matcher has the word array in it.

Since we're calling methods from `Array.prototype` and passing a value to be used as `this` it would work if users had their own array implementations or with things like `arguments`.

I can change this if you want me too, I'd like to do whatever you prefer. If we want to be more strict it would be good to accept `Arrays` only, but if we want to allow other similar implementations this is the way to go.

Maybe we could also have an `match.iterable` or something like that.

What do you think it's better? Let me know and we can fix this if needed 😄

fearphage: I understand the benefits and usage. I'm just saying when you document it, it looks weird/out of place. I would expect an *array* matcher to throw/error if I didn't pass it an array honestly. I was considering an `iterable` matcher. I'd wait to see what other thing about it though.

lucasfcosta: @fearphage Yes, a new `iterable` matcher would be the most adequate option IMO.

Thanks for your input, I totally agree with everything you said.

fatso83: Changing array to iterable is fine with me, as long as we do it now, before a new release. 

But do we have a definition of iterable? Objects are iterable in a sense too (`for ... in`) ... Is it anything that has a length property?

fatso83: If it is basically anything that is array like I wouldn't mind keeping array.

lucasfcosta: This is a tricky question, but according to MDN this is the definition of an `iterable`:

>In order to be iterable, an object must implement the `@@iterator` method, meaning that the object (or one of the objects up its prototype chain) must have a property with a `@@iterator` key which is available via constant `Symbol.iterator`

Also there are lots of objects other than arrays that can be considered iterables, such as Strings, TypedArrays, Maps and Sets. The `arguments` object is also considered an `iterable`.

I think that maintaining these matchers under the `array` property will cause confusion and will not leave it explicit that it can be used with other objects (Maps, Sets, etc...).

I'd like to change this as suggested by @fearphage.

Please let me know what you think about this subject. Also, if we agree on changing this, would you like me to address that in a separate PR and them amend a commit for fixing this one or would you like me to rename this PR and work on this branch?

fatso83: Ah, jeez, this is harder. So we are either to choose between 
- expanding the namespace through `match.maps.*`, `match.sets.*`, etc

or 
- expanding the code under `match.iterable` for Maps, Sets, etc. 

I guess for the latter it would mean we should throw for all types we do not explicitly support (like for now: those missing a `length` property).

I need to get some input on this.

fatso83:
Btw, the docs have been put in the wrong folder. `docs/_releases` is for docs on previous, released versions. `docs/release-source` is for what is will be released.

docs/_releases/v2.0.0-pre.4/matchers.md
lucasfcosta: I'd vote for the first option.

Let's get some more input on this so we can take the optimal choice 😄

fatso83: I think we should go for the first option too. I think it would make for clearer tests as well, and we won't come in the position of always being late with implementing all the variations of what `iterable` could constitute => simpler implementation, less branching. As a minor thing, it also means we don't have to change the existing code.

Agree/disagree, @fearphage?

fearphage: > expanding the namespace through match.map.*, match.set.*, etc

This is going to be a lot of duplicate or (hopefully) resused code, but I think that is the most explicit option.

> Agree/disagree, @fearphage?

So the first option being to make this only work for arrays since it's in the `array` namespace? If that is the case, I would agree. I think we should confirm that though. 

For instance, `matcher.array.startsWith(...)` should throw when you don't pass an array to it. It seems like you're doing 2 things at once here. `matcher.array()` && `array.startsWith()`. So if I pass `{length: 1, 0: 'foo'}`, it would fail saying the test expected an array. Are we in agreement or did I miss the mark?

fatso83: So no duck typing then? Strictly `Array.isArray()`? I am not super sure what it really brings to the table, since it would leave out `arguments` and other array-like objects, but people can always convert these to proper arrays if need be, so it should be ok. 

And since the test is for an array, I guess it should break on non-arrays.

lucasfcosta: @fatso83 @fearphage I totally agree with that. People using this matcher should be aware it will work for `Array` objects only and anything else should be converted to that.

fatso83: Ok, seems all are pro. Care to update the docs, Lucas? As long as you don't change the interface you document you can also add any strengthening assertions you like.

lucasfcosta: Of course!
I'm already working on the `Map` and `Set` matchers, for now I'll make sure to add commits allowing arrays only on the matchers we already have and update the docs.

lucasfcosta:
Here it is 😄 
These matchers will now fail whenever `actual` is not an `Array` and I also updated the docs (commited w/ `--amend`).
Let me know if I forgot something.

I have already started working on matchers for the `Map` and `Set` objects and I'll submit another PR for that later this week.

fatso83:
Great work. Thanks for the collaboration :-)

PR #1211, opened 2016-12-22
Add .resolves and .rejects to stub

Floby:
#### Purpose (TL;DR)
This commit adds the possibility for a stub
to be configured to return a promise which will
either resolve or reject

This was already possible, but this adds a shorthand for it

##### resolving
```javascript
stub.returns(Promise.resolve({}))`
```
becomes 
```javascript
stub.resolves({})
```

##### rejecting
```javascript
stub.returns(Promise.reject(new Error("bad luck")))
```
becomes
```javascript
stub.rejects("bad luck")
```

Since Promises are now part of the standard ES6 feature set, this
commit does not make use of polyfills or third party library.
This is problematic for some browsers and generally for contexts where
the standard is not implemented.
**EDIT** now it does.

#### Solution  - optional
This solution works by adding a `resolves` and `rejects` method to the possible behaviors.

#### How to verify
2. `npm install`
3. `npm test`

### DIFF
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -130,7 +130,7 @@ describe("stub", function () {
             });
         });
 
-        it("should return stub", function () {
+        it("should return the same stub", function () {
             var stub = createStub.create();
 
             assert.same(stub.resolves(""), stub);
@@ -166,7 +166,7 @@ describe("stub", function () {
             });
         });
 
-        it("returns stub", function () {
+        it("should return the same stub", function () {
             var stub = createStub.create();
 
             assert.same(stub.rejects({}), stub);
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -119,6 +119,94 @@ describe("stub", function () {
         });
     });
 
+    describe(".resolves", function () {
+        it("returns a promise to the specified value", function () {
+            var stub = createStub.create();
+            var object = {};
+            stub.resolves(object);
+
+            return stub().then(function (actual) {
+                assert.same(actual, object);
+            });
+        });
+
+        it("should return stub", function () {
+            var stub = createStub.create();
+
+            assert.same(stub.resolves(""), stub);
+        });
+
+        it("supersedes previous throws", function () {
+            var stub = createStub.create();
+            stub.throws().resolves(1);
+
+            refute.exception(function () {
+                stub();
+            });
+        });
+
+        it("supersedes previous rejects", function () {
+            var stub = createStub.create();
+            stub.rejects(Error("should be superseeded")).resolves(1);
+
+            return stub().then();
+        });
+    });
+
+    describe(".rejects", function () {
+        it("returns a promise which rejects for the specified reason", function () {
+            var stub = createStub.create();
+            var reason = new Error();
+            stub.rejects(reason);
+
+            return stub().then(function () {
+                referee.fail("this should not resolve");
+            }).catch(function (actual) {
+                assert.same(actual, reason);
+            });
+        });
+
+        it("returns stub", function () {
+            var stub = createStub.create();
+
+            assert.same(stub.rejects({}), stub);
+        });
+
+        it("specifies exception message", function () {
+            var stub = createStub.create();
+            var message = "Oh no!";
+            stub.rejects("Error", message);
+
+            return stub().then(function () {
+                referee.fail("Expected stub to reject");
+            }).catch(function (reason) {
+                assert.equals(reason.message, message);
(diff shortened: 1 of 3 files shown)


Floby:
I would need guidance from the maintainers in order to know which strategy to use for contexts where the global `Promise` object does not exist.

fatso83:
@Floby, thanks for this. We have previously talked about just integrating `sinon-as-promised` into Sinon, and this basically provides that functionality. With regards to missing Promise implementations, you could just take a gander at what [`sinon-as-promised` does:](https://github.com/bendrucker/sinon-as-promised/blob/master/index.js#L3). Require the polyfill, meaning the native object will be used if it exists, or it will be provided for. See the docs on [native-promise-only](https://github.com/getify/native-promise-only)

fatso83:
You can see that tests are failing when running on PhantomJS in the Travis build log, so this needs to be addressed somehow.

Floby:
haha, thanks. I probably should have looked it up. If integrating sinon-as-promised is in the plans, then it may be best to disregard this PR.

fatso83:
Nonono, if you look at sinon-as-promised, you have basically reimplemented 90% of the same code. It's not rocket science and this is _fine_. What counts is that someone does it. Mind integrating that polyfill as well?

Floby:
Alright, I'll give it a go =)

I was also thinking of allowing user to set their own Promise object to a sinon.Promise thing, so that they can use Bluebird, RSVP, etc.

Floby:
This one seems to work =) using native-promise-only

test/stub-test.js
```diff
@@ -119,6 +119,94 @@ describe("stub", function () {
...
+            var stub = createStub.create();
+            var object = {};
+            stub.resolves(object);
+
+            return stub().then(function (actual) {
+                assert.same(actual, object);
+            });
+        });
+
+        it("should return stub", function () {
```
fatso83: I had to read the code to understand this. Could you change the test name to say you are returning the same stub?

fatso83: Sorry for not catching this on first read :-/

test/stub-test.js
```diff
@@ -119,6 +119,94 @@ describe("stub", function () {
...
+            stub.rejects(reason);
+
+            return stub().then(function () {
+                referee.fail("this should not resolve");
+            }).catch(function (actual) {
+                assert.same(actual, reason);
+            });
+        });
+
+        it("returns stub", function () {
```
fatso83: Same here. It's the _same_ stub.

fatso83:
Basically ready to merge this if you fix the test naming. Thanks!

test/stub-test.js
Floby: sure thing =).
For what it's worth, I pasted those titles from the tests for `.returns` and `.throws` ^^ I don't mind changing them as well, but I'm not sure that's how you want to organise your refactor commits.

what about `it("is chainable")`. It seems to me that this tells better what is expected of it.

fatso83: Sorry, did not see this until I merged ... oh well, no biggie, although I like your suggestion :)

Floby:
Thank you :) when can one expect this to land on a version on npm?

By the way, I love sinon.js, please continue making it great!

lucasfcosta:
Great PR! Can't wait to start using this! 😄

fatso83:
We can get this out pretty soon, but I would have liked to have closed #1210 first, as the changes are already in.

vatula:
sinon-as-promised does it better. Current solution as a syntactic sugar inherits same issue with Promise.reject being unhanded which then throws unhanded rejection error.

fatso83:
@vatula, it's pretty much the same code, so if there's something missing I'm sure you can fix it in just a few lines. So, where's the pull request? ;-)

fatso83:
PS. If there's something wrong with the current implementation please submit a proper issue report with a reproducible code sample. No issue = no ❤️ = no fix.

Floby:
@vatula https://github.com/sinonjs/sinon/commit/0074521c18d615a87b49116161ff03427e529c82 read up ;)

vatula:
@Floby Haha sorry 😊 Didn't check latest version. Thanks for heads up!

Floby:
No worries. I'm also waiting for 2.0 to be released. But no rush is needed =) I'm using sinon-as-promised in the meantime.

alejoamiras:
@Floby great work dude ! I'll be using this feature so much in the next release ! 😀

Issue #1107, opened 2016-07-24
Use path-to-regexp for fake server urls


jmeas:
Sinon's fake server supports writing regex for the URL. This is powerful, but routing DSLs are becoming increasingly common in JS. [path-to-regexp](https://github.com/pillarjs/path-to-regexp) provides a powerful & familiar one.

tl;dr:

Current:

`server.respondWith(/\/books\/(\d+)/, ...)`

with DSL:

`server.respondWith('/books/:bookId/', ...)`

It allows you to embed [mini-regexes](https://github.com/pillarjs/path-to-regexp#custom-match-parameters) as custom params for when ya need it.

Small change but I think it'd go a long way :v:

fatso83:
AFAIK this seems to be a non-breaking feature, am I right? If this just expands on our options I think this makes for quite readable tests where you can reuse strings from your routing logic instead of rewriting them as regexes. That's an improvement IMHO.

jmeas:
I *think* it'd be non-breaking, unless there's any situations where a "regular" URL passed through `path-to-regexp` would get transformed. I can't think of any such URLs atm, but I'd need to look closer at it to be sure.

mantoni:
I like this suggestion. Readability FTW.

fatso83:
Ship a PR, and we can take it from there 👍

mroderick:
Great suggestion! I'd love to see a PR for this

lucasfcosta:
Hello, friends, so, you'd expect this PR to be made using the `path-to-regexp` module or would you like to reimplement the whole feature?

fatso83:
Just reuse the existing `path-to-regexp` module. More lines in a project only means it gets harder to manage.

lucasfcosta:
Hi everyone, sorry to bother you but I was starting to work on this early this morning and I noticed that you do not run the `fake-server` tests neither in the local environment nor on Travis due to [this line](https://github.com/sinonjs/sinon/blob/master/test/util/fake-server-test.js#L21) which makes it run on browsers only.

I can easily set up Karma and start testing this on multiple browsers, but I'm more worried about how you could test it without having to do all this work again.

I've taken a look at your testing scripts and the whole environment but I couldn't find out how to run these tests without adding karma to the whole structure.

I tried to install the buster and phantomjs modules but I also noticed the config files mentioned [in this script](https://github.com/sinonjs/sinon/blob/master/scripts/ci-test.sh) do not exist anymore.

[...]

PR #1212, opened 2016-12-22
Accept route DSLs

lucasfcosta:
#### Purpose (TL;DR)

This pull request adds support for common route DSLs as requested in #1107.

In a nutshell, now we are able to do this:

* `server.respondWith('/books/:bookId/', ...)`
* `server.respondWith('/foo*', ...)`

#### Solution 

In order to solve this I've added the [`path-to-regexp`](https://github.com/pillarjs/path-to-regexp) module as a dependency as suggested by @fatso83 to avoid reinventing the wheel and adding more code to maintain (I totally agree with it).

Then I always convert any non-empty strings to regular expressions using the `path-to-regexp` module.

Since it already deals with capture groups it worked right out of the box and the old tests still passed with zero changes.

I also added tests for both simple URL matching and for checking route parameters.

#### How to verify
2. `npm install`
3. `npm run test-headless`

---------------------------------

I'd also like to thank @fatso83 for the help with the browser testing environment. Sorry for not noticing how it worked before 😅 

Let me know if I missed anything or if you disagree with the coding style or anything else and I'll happily fix it. Thanks for reading this.

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -16,6 +16,7 @@ var sinon = require("../../sinon");
 var createInstance = require("./core/create");
 var format = require("./core/format");
 var configureLogError = require("./core/log_error");
+var pathToRegexp = require("path-to-regexp");
 
 function responseArray(handler) {
     var response = handler;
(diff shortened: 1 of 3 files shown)


fatso83:
Very useful addition! Thanks!

jmeas:
Great work here. Thanks @lucasfcosta !

Issue #810, opened 2015-08-17
Document sinon.restore()


mroderick:
As [pointed out](https://github.com/cjohansen/Sinon.JS/issues/270#issuecomment-127802233) in #270, the documentation doesn't mention `sinon.restore()` at all.

nadrane:
Would anyone be interested in me making this PR?

mroderick:
I think this was done in #1213

PR #1213, opened 2016-12-27
Added documentation for sinon.restore()

tiemevanveen:
Fix issue #810 by adding docs for `sinon.restore()`

Couldn't find this in the docs, really need to be in there!

- Added in all 3 files, hope that's correct
- Not sure if I've added it to the right section
- Not sure if you want the examples

### DIFF
--- a/docs/_releases/v1.17.6/utils.md
+++ b/docs/_releases/v1.17.6/utils.md
@@ -8,6 +8,26 @@ Sinon.JS has a few utilities used internally in `lib/sinon.js`. Unless the metho
 
 ## Utils API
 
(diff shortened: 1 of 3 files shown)


mroderick:
Thank you for the PR @tiemevanveen!

I've looked it over and compared the examples with the tests in `test/util/core/restore-test.js` in `master`. I couldn't find any tests for `restore()` in `v1.17` branch.

As far as I can tell, the new documentation is correct, so I am merging this.

tiemevanveen:
Awesome, and thanks for the quick response!

PR #1030, opened 2016-04-18
Add test for issue #1026

mroderick:
#### Purpose

This is a PR for **v1.17** branch, it adds a failing test case for #1026 
#### How to verify - mandatory
1. Check out this branch (see github instructions below)
2. `npm install`
3. `npm test`
4. Observe the magnificence of the failing test case

PR #1214, opened 2016-12-27
Fix 1026 in 2.x

### LINKED ISSUE #1026: sinon.stub an object with 'watch' function doesn't stub the watch function
- Sinon version : 1.12.2 [Edit, 1.17.3]
- Environment   : JS test framework with Intern

`sinon.stub( { watch: function() {} } )` to stub the object and the watch method. 

**What actually happens**

watch is not stubbed in FIREFOX.

**How to reproduce**

```
var stubbedObject = sinon.stub({ watch: function() {} });
stubbedObject.watch.args[1];
```

stubbedObject.watch does not have property 'args'

This only fails in Firefox. The problem is that when sinon.stub iterates over the functions of an object to stub them, it makes sure it doesn't stub the same function twice. It uses an object to keep track of all of the functions seen already. Object should have a prototype method called 'watch' (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/watch) but right now, only Firefox implements this. 

When .stub walks through the object's functions, and compares them to what it has 'seen' (in /lib/sinon/util/core/walk.js line 19) but the seen object has methods on it already.

Currently, the check if the function is seen is simply '!seen[k]'. I think it should be 'seen[k] !== true' seeing as that directly corresponds to the way the function records if an object is seen (with seen[k] = true). I have tested this and it works.

mroderick:
This is a PR for `master`, it fixes #1026.

It is just the commits from #1030 ported to `master`.

However, there has been some changes that causes `collection-test` to fail.

I will investigate later, unless someone beats me to it

### DIFF
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -120,7 +120,8 @@ var collection = {
 
             walk(stubbedObj, function (prop, propOwner) {
                 if (
-                    typeof getPropertyDescriptor(propOwner, prop).value === "function"
+                    typeof getPropertyDescriptor(propOwner, prop).value === "function" &&
+                    stubbedObj.hasOwnProperty(prop)
                 ) {
                     col.add(stubbedObj[prop]);
                 }
--- a/lib/sinon/util/core/walk.js
+++ b/lib/sinon/util/core/walk.js
@@ -16,7 +16,7 @@ function walkInternal(obj, iterator, context, originalObj, seen) {
     }
 
     Object.getOwnPropertyNames(obj).forEach(function (k) {
-        if (!seen[k]) {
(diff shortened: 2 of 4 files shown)


fatso83:
Fails a test in all Node versions

mroderick:
> Fails a test in all Node versions

Yeah, I am still trying to figure out why

mroderick:
@fatso83 I've fixed the tests by making sure that `collection.stub` does not get a new behaviour, when we're updating the behaviour of `walk`.

Do you agree with the changes?

fatso83:
Seems fine

mroderick:
Great! I am merging this!

PR #1215, opened 2016-12-27
Maps matchers

lucasfcosta:

As discussed on #1210, this adds the following matchers for the `Map` type:

* `sinon.match.map(map)` - Requires an object to be a `Map`
* `sinon.match.map.deepEquals(map)` - Requires a `Map` to be deep equal another one
* `sinon.match.map.contains(map)` - Requires a `Map` to contain each one of the items the given map has.

**This also adds tests for each matcher and updates the docs.**

Due to the fact that the `order` of elements should not mean anything in a `Map` (and we can't really have any item's index), I didn't include the `startsWith` and `endsWith` matchers for Maps.

Since the old matchers should work for `Array` objects only, the new ones for `Map` and `Set` objects will be strict as well.

This PR adds matchers for Maps only, I'll be working on the Sets matchers as soon as this gets merged. I'm submitting this one first because I want to make sure we all agree on the spec before going further.

2. `npm install`
3. `npm run test` - This will run the newly created tests

Let me know if I missed something or if there's anything to improve and I'll happily update this PR. It's always great to hear your feedback.

Thanks for reading this 😄

### DIFF
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -264,6 +264,8 @@ match.array.contains = function (expectation) {
     }, "contains([" + iterableToString(expectation) + "])");
 };
 
+match.map = match.typeOf("map");
+
 match.bool = match.typeOf("boolean");
 match.number = match.typeOf("number");
 match.string = match.typeOf("string");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -12,11 +12,11 @@ var create = require("./util/core/create");
 var deepEqual = require("./util/core/deep-equal").use(match); // eslint-disable-line no-use-before-define
 var every = require("./util/core/every");
 var functionName = require("./util/core/function-name");
+var iterableToString = require("./util/core/iterable-to-string");
 var typeOf = require("./typeOf");
 var valueToString = require("./util/core/value-to-string");
 
 var indexOf = Array.prototype.indexOf;
-var arrayToString = Array.prototype.toString;
 
 function assertType(value, type, name) {
     var actual = typeOf(value);
@@ -234,7 +234,7 @@ match.array.deepEquals = function (expectation) {
         return typeOf(actual) === "array" && sameLength && every(actual, function (element, index) {
             return expectation[index] === element;
         });
-    }, "deepEquals([" + arrayToString.call(expectation) + "])");
+    }, "deepEquals([" + iterableToString(expectation) + "])");
 };
 
 match.array.startsWith = function (expectation) {
@@ -242,7 +242,7 @@ match.array.startsWith = function (expectation) {
         return typeOf(actual) === "array" && every(expectation, function (expectedElement, index) {
             return actual[index] === expectedElement;
         });
-    }, "startsWith([" + arrayToString.call(expectation) + "])");
+    }, "startsWith([" + iterableToString(expectation) + "])");
 };
 
 match.array.endsWith = function (expectation) {
@@ -253,7 +253,7 @@ match.array.endsWith = function (expectation) {
         return typeOf(actual) === "array" && every(expectation, function (expectedElement, index) {
             return actual[offset + index] === expectedElement;
         });
-    }, "endsWith([" + arrayToString.call(expectation) + "])");
+    }, "endsWith([" + iterableToString(expectation) + "])");
 };
 
 match.array.contains = function (expectation) {
@@ -261,7 +261,7 @@ match.array.contains = function (expectation) {
         return typeOf(actual) === "array" && every(expectation, function (expectedElement) {
             return indexOf.call(actual, expectedElement) !== -1;
         });
-    }, "contains([" + arrayToString.call(expectation) + "])");
+    }, "contains([" + iterableToString(expectation) + "])");
 };
 
 match.bool = match.typeOf("boolean");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -276,6 +276,14 @@ match.map.deepEquals = function mapDeepEquals(expectation) {
     }, "deepEquals(Map[" + iterableToString(expectation) + "])");
 };
 
+match.map.contains = function mapContains(expectation) {
+    return match(function (actual) {
+        return typeOf(actual) === "map" && every(expectation, function (element, key) {
+            return actual.has(key) && actual.get(key) === element;
+        });
+    }, "contains(Map[" + iterableToString(expectation) + "])");
+};
+
 match.bool = match.typeOf("boolean");
 match.number = match.typeOf("number");
 match.string = match.typeOf("string");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -10,11 +10,11 @@
 
 var create = require("./util/core/create");
 var deepEqual = require("./util/core/deep-equal").use(match); // eslint-disable-line no-use-before-define
+var every = require("./util/core/every");
 var functionName = require("./util/core/function-name");
 var typeOf = require("./typeOf");
 var valueToString = require("./util/core/value-to-string");
 
-var every = Array.prototype.every;
 var indexOf = Array.prototype.indexOf;
 var arrayToString = Array.prototype.toString;
 
@@ -231,7 +231,7 @@ match.array.deepEquals = function (expectation) {
     return match(function (actual) {
         // Comparing lengths is the fastest way to spot a difference before iterating through every item
         var sameLength = actual.length === expectation.length;
-        return typeOf(actual) === "array" && sameLength && every.call(actual, function (element, index) {
+        return typeOf(actual) === "array" && sameLength && every(actual, function (element, index) {
             return expectation[index] === element;
         });
     }, "deepEquals([" + arrayToString.call(expectation) + "])");
@@ -239,7 +239,7 @@ match.array.deepEquals = function (expectation) {
 
 match.array.startsWith = function (expectation) {
     return match(function (actual) {
-        return typeOf(actual) === "array" && every.call(expectation, function (expectedElement, index) {
+        return typeOf(actual) === "array" && every(expectation, function (expectedElement, index) {
             return actual[index] === expectedElement;
         });
     }, "startsWith([" + arrayToString.call(expectation) + "])");
@@ -250,7 +250,7 @@ match.array.endsWith = function (expectation) {
         // This indicates the index in which we should start matching
         var offset = actual.length - expectation.length;
 
-        return typeOf(actual) === "array" && every.call(expectation, function (expectedElement, index) {
+        return typeOf(actual) === "array" && every(expectation, function (expectedElement, index) {
             return actual[offset + index] === expectedElement;
         });
     }, "endsWith([" + arrayToString.call(expectation) + "])");
@@ -258,7 +258,7 @@ match.array.endsWith = function (expectation) {
 
 match.array.contains = function (expectation) {
     return match(function (actual) {
-        return typeOf(actual) === "array" && every.call(expectation, function (expectedElement) {
+        return typeOf(actual) === "array" && every(expectation, function (expectedElement) {
             return indexOf.call(actual, expectedElement) !== -1;
         });
     }, "contains([" + arrayToString.call(expectation) + "])");
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -266,6 +266,16 @@ match.array.contains = function (expectation) {
 
 match.map = match.typeOf("map");
 
+match.map.deepEquals = function mapDeepEquals(expectation) {
+    return match(function (actual) {
+        // Comparing lengths is the fastest way to spot a difference before iterating through every item
+        var sameLength = actual.size === expectation.size;
+        return typeOf(actual) === "map" && sameLength && every(actual, function (element, key) {
+            return expectation.has(key) && expectation.get(key) === element;
+        });
+    }, "deepEquals(Map[" + iterableToString(expectation) + "])");
+};
+
 match.bool = match.typeOf("boolean");
 match.number = match.typeOf("number");
 match.string = match.typeOf("string");
--- a/lib/sinon/util/core/every.js
+++ b/lib/sinon/util/core/every.js
@@ -1,0 +1,19 @@
+"use strict";
+
+// This is an `every` implementation that works for all iterables
+module.exports = function every(obj, fn) {
+    var pass = true;
+
+    try {
+        obj.forEach(function () {
+            if (!fn.apply(this, arguments)) {
+                // Throwing an error is the only way to break `forEach`
+                throw new Error();
+            }
+        });
+    } catch (e) {
+        pass = false;
+    }
+
+    return pass;
+};
--- a/lib/sinon/util/core/iterable-to-string.js
+++ b/lib/sinon/util/core/iterable-to-string.js
@@ -1,0 +1,34 @@
+"use strict";
+var typeOf = require("../../typeOf");
+
+module.exports = function iterableToString(obj) {
+    var representation = "";
+
+    function stringify(item) {
+        return typeof item === "string" ? "'" + item + "'" : String(item);
+    }
+
+    function mapToString(map) {
+        map.forEach(function (value, key) {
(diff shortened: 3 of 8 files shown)


fatso83:
Damn, you're productive around XMas! I am busy having holidays so will need to check back.

mroderick:
AFAICT, [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) wasn't implemented in IE until IE11.

Will the [saucelabs tests](https://saucelabs.com/u/sinonjs) error out in IE10 when merged into `master`?

I would suggest using https://github.com/medikoo/es6-map in the test file, to ensure that the tests can run, even when `Map` is not implemented in the JavaScript engine.

lucasfcosta:
Thanks for the feedback @mroderick, that's a great idea indeed! It will be a lot better to test these functions individually just to make sure they're working whenever we need to make changes on methods that use them. It makes the whole code a lot safer to change.

@fatso83 hahahaha I'm a bit lonely this year so I'm having a lot of time to work on Open Source and study things. Don't worry about reviewing this fast, I'm already very grateful for you all to be reading these PRs and giving me feedback.

Make sure you spend quality time with your friends and loved ones on these holidays 😄

lucasfcosta:
I have just move those methods to the core and added tests for them.
That also helped me spot some improvement possibilities such as stopping the `forEach` loop by using a `throw` if any of the elements is `false`. This is the only way to stop that loop since `forEach` is the only method available for `Map`, `Set` and `Array` objects.
Let me know if I missed anything!

fatso83:
Adding code that patches the run system in Sinon is not a good idea. It will lead to false negatives: running tests using Sinon will then give the impression that code works on platforms that does not support those features. But as soon as one removes Sinon the code starts breaking. That reminds me that I made a mistake in accepting the `resolves` and `rejects` functionality, since that included a polyfill. 

I don't mind adding a polyfill in the *test code*, mind you.

lib/sinon/util/core/iterable-to-string.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+var typeOf = require("../../typeOf");
+
+module.exports = function (obj) {
```
mroderick: To aid in debugging, especially in less capable JavaScript engines, it is a good practice to name essential functions. I would suggest naming this one `iterableToString`.

lib/sinon/util/core/iterable-to-string.js
```diff
@@ -0,0 +1,23 @@
...
+    var representation = "";
+
+    function stringify(item) {
+        return typeof item === "string" ? "'" + item + "'" : String(item);
+    }
+
+    if (typeOf(obj) === "map") {
```
mroderick: In my eyes `iterableToString` is a higher abstraction function that covers two input scenarios, `Map` or generic input.

I would suggest de-structuring the `iterableToString` function to call lower abstraction functions, perhaps something like this.

```
"use strict";

var typeOf = require("../../typeOf");

function stringify(item) {
    return typeof item === "string" ? "'" + item + "'" : String(item);
}

function mapToString(map) {
    var representation = "";

    map.forEach(function (value, key) {
        representation += "[" + stringify(key) + "," + stringify(value) + "],";
    });

    representation = representation.slice(0, -1);

    return representation;
}

function genericIterableToString(obj) {
    var representation = "";

    obj.forEach(function (value) {
        representation += stringify(value) + ",";
    });

    representation = representation.slice(0, -1);

    return representation;
}


module.exports = function iterableToString(obj) {
    if (typeOf(obj) === "map") {
        return mapToString(obj);
    }

    return genericIterableToString(obj);
};
```

test/util/core/iterable-to-string-test.js
```diff
@@ -0,0 +1,41 @@
...
+describe("util/core/iterable-to-string", function () {
+    it("returns an String representation of Array objects", function () {
+        var arr = [1, "one", true, undefined, null];
+        var expected = "1,'one',true,undefined,null";
+
+        assert.equals(iterableToString(arr), expected);
+    });
+
+    it("returns an String representation of Map objects", function () {
+        if (typeof Map === "function") {
```
mroderick: Wouldn't it be better to wrap the `it` statement, like you did in `match-test.js`?

test/util/core/iterable-to-string-test.js
```diff
@@ -0,0 +1,41 @@
...
+                "[true,true]," +
+                "[undefined,undefined]," +
+                "[null,null]";
+
+            assert.equals(iterableToString(map), expected);
+        }
+    });
+
+    it("returns an String representation of Set objects", function () {
+        if (typeof Set === "function") {
```
mroderick: Same question as above regarding wrapping the `it`statement with the conditional

test/util/core/iterable-to-string-test.js
lucasfcosta: Yes it would be better since then this test would not count as passed, it would just be ignored.
Well noticed, thanks!

lib/sinon/util/core/iterable-to-string.js
lucasfcosta: Excellent idea, it really looks better.
Thanks for the help!

lib/sinon/util/core/iterable-to-string.js
lucasfcosta: I totally agree, I always name my functions whether they get assigned to a variable or are used only once, I just thought you prefer these functions to not have names due to what I've seen in other files.

Thanks for the feedback!

lucasfcosta:
Thanks everyone for your feedback, I've just learned a lot by reading your considerations! 😄 
I'll make the suggested changes as soon as I get home.

lib/sinon/util/core/every.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+// This is an `every` implementation that works for all iterables
+module.exports = function (obj, fn) {
```
fatso83: Name anonymous function to aid debugging (ref [previous review](https://github.com/sinonjs/sinon/pull/1215#discussion_r94131858)).

lib/sinon/match.js
```diff
@@ -231,37 +231,57 @@ match.array.deepEquals = function (expectation) {
...
+        return typeOf(actual) === "array" && every(expectation, function (expectedElement) {
             return indexOf.call(actual, expectedElement) !== -1;
         });
-    }, "contains([" + arrayToString.call(expectation) + "])");
+    }, "contains([" + iterableToString(expectation) + "])");
+};
+
+match.map = match.typeOf("map");
+
+match.map.deepEquals = function (expectation) {
```
fatso83: Name anonymous function to aid debugging (ref [previous review](https://github.com/sinonjs/sinon/pull/1215#discussion_r94131858)).

lib/sinon/util/core/iterable-to-string.js
```diff
@@ -0,0 +1,34 @@
+"use strict";
+var typeOf = require("../../typeOf");
+
+module.exports = function iterableToString(obj) {
```
fatso83: Name anonymous function to aid debugging (ref [previous review](https://github.com/sinonjs/sinon/pull/1215#discussion_r94131858)).

lib/sinon/util/core/every.js
lucasfcosta: Yup, I'm doing this right now, I'm a bit of an anxious guy so I have this bad habit of pushing commits as soon as I do any rebase.

lucasfcosta:
I've just finished this, here are all the changes 😄 

I definitely gotta stop doing multiple pushes to my fork while working, I always forget about this coveralls bot 😅 

I've named all the anonymous functions added in this PR and I've also improved `iterableToString` according to @mroderick suggestions.

As for the `Set` and `Map` polyfills, I think they should not be added to this PR because since we're using the `typeOf` utility function and it uses `Object.prototype.toString` in order to get the type of an object, we would need a polyfill there too, because the `Map` and `Set` polyfills also polyfill the `Symbol.toStringTag` symbol and therefore it does not get used by older platforms unless we also try to use it globally in order for `toString` to be aware of it.

Anyway, would you like to add it? What do you think its better?

Also, thanks for the feedback, it was really valuable.

fatso83:
@lucasfcosta, sorry for jumping the gun. I just got the email about it being updated, so jumped on it. Please do add the polyfills if you can find some decent ones, but add it to the *test* suites, not the main library. See 942fc53970a9962c309764c971b2571cf340cfaf for why and how.

lucasfcosta:
@fatso83 I've been trying to use some symbol Polyfills but none of them seem to reimplement `Object.prototype.toString` in order for it to use `Symbol.toStringTag` so all tests that use `typeOf` end up returning `object`.
Any ideas on how to workaround that?

fatso83:
@lucasfcosta Well, we could fix the polyfills, submit a PR, and use a github branch reference while we might for it to be merged upstream? Might be a bit of a detour of course ... 

On the other hand, if I understand you correctly, the polyfill for Symbol at polyfill.io [seems to implement `toString()`](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Symbol/polyfill.js#L212). Is that what you need?

lucasfcosta:
@fatso83 Yeah, I thought about that, but it would definitely be a detour, however, if that's the only way to fix this, we've gotta do it.

I tried a couple more possibilities including the service you listed and none of them seemed to work.

Apparently `polyfill.io` is geared towards browsers and requires `user-agent` strings in order to target specific browser builds. Also, whenever using it with some old UA strings or nothing it just failed because of some `YAML Parser` module.

I also tried to use [core-js](https://github.com/zloirock/core-js) and the only way to override `Object.prototype.toString` would be to use a global polyfill or call `Map.prototype.toString` instead of `Object.prototype.toString`, which is something we can't do too because of how `typeOf` is expected to work. ([Reference](https://github.com/zloirock/core-js/issues/199))

All the other individual polyfills I've tried didn't address this too.

I'm running out of ideas, I guess I'll just have to implement it somewhere.

fatso83:
You know what, let's just leave it for now. I will merge this, as the tests have been implemented and are being run and that is the main thing. Upping the coverage report is not "up there" on the priority list IMHO. Any objections? Otherwise I'll merge this ASAP as every other item has been checked off the review list.

lucasfcosta:
@fatso83 I'll work on another PR for `Sets` soon and I'll try to tackle this problem again in it.

If I can't find a way to fix it in that PR I'll just reimplement `Object.prototype.toString` in order for it to read the fake Symbols these polyfills use.

For now, I'd just like to thank everyone for the feedback and for your work, it helped a lot. 😄

fatso83:
Huh, [it seems to fail on IE11](https://travis-ci.org/sinonjs/sinon/jobs/187915052)

<details>

  1) sinonMatch .map map.deepEquals has a .deepEquals matcher:
     AssertionError: [assert.equals] deepEquals(Map[1,2,3]) expected to be equal to deepEquals(Map[['one',1],['two',2],['three',3]])
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at ctx.fail (node_modules/referee/lib/referee.js:90)
      at assertion (node_modules/referee/lib/referee.js:98)
      at referee[type][name] (node_modules/referee/lib/referee.js:118)
      at Anonymous function (test/match-test.js:732)
  2) sinonMatch .map map.deepEquals matches maps with the exact same elements:
     AssertionError: [assert] Expected false to be truthy
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at assert (node_modules/referee/lib/referee.js:164)
      at Anonymous function (test/match-test.js:751)
  3) sinonMatch .map map.contains has a .contains matcher:
     AssertionError: [assert.equals] contains(Map[1,2,3]) expected to be equal to contains(Map[['one',1],['two',2],['three',3]])
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at ctx.fail (node_modules/referee/lib/referee.js:90)
      at assertion (node_modules/referee/lib/referee.js:98)
      at referee[type][name] (node_modules/referee/lib/referee.js:118)
      at Anonymous function (test/match-test.js:795)
  4) sinonMatch .map map.contains matches maps containing the given elements:
     AssertionError: [assert] Expected false to be truthy
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at assert (node_modules/referee/lib/referee.js:164)
      at Anonymous function (test/match-test.js:817)
  5) util/core/iterable-to-string returns an String representation of Map objects:
     AssertionError: [assert.equals] 1,'one',true,undefined,null expected to be equal to [1,1],['one','one'],[true,true],[undefined,undefined],[null,null]
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at ctx.fail (node_modules/referee/lib/referee.js:90)
      at assertion (node_modules/referee/lib/referee.js:98)
      at referee[type][name] (node_modules/referee/lib/referee.js:118)
      at Anonymous function (test/util/core/iterable-to-string-test.js:29)
  6) util/core/iterable-to-string returns an String representation of Set objects:
     AssertionError: [assert.equals]  expected to be equal to 1,'one',true,undefined,null
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at ctx.fail (node_modules/referee/lib/referee.js:90)
      at assertion (node_modules/referee/lib/referee.js:98)
      at referee[type][name] (node_modules/referee/lib/referee.js:118)
      at Anonymous function (test/util/core/iterable-to-string-test.js:38)
</details>

lucasfcosta:
It seems like these tests are failing because `typeOf` can't detect the type `Map` properly so it always ends up not entering the right `if` statements.

In Chai we've got [this awesome module for type-detection](https://github.com/chaijs/type-detect) and it works pretty well. Maybe we could start using it instead of this `typeOf` implementation. What do you think? For now I'll try to see what can be done regardin what have already got here.

lucasfcosta:
I managed to discover the reason of this error:

```js
  6) util/core/iterable-to-string returns an String representation of Set objects:
     AssertionError: [assert.equals]  expected to be equal to 1,'one',true,undefined,null
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at ctx.fail (node_modules/referee/lib/referee.js:90)
      at assertion (node_modules/referee/lib/referee.js:98)
      at referee[type][name] (node_modules/referee/lib/referee.js:118)
      at Anonymous function (test/util/core/iterable-to-string-test.js:38)
```

This happens because the `Set` constructor in IE11 doesn't use the array passed to it to define its elements, so we've gotta do this instead:

```js
        it("returns an String representation of Set objects", function () {
            var set = new Set();
            set.add(1);
            set.add("one");
            set.add(true);
            set.add(undefined);
            set.add(null);

            var expected = "1,'one',true,undefined,null";

            assert.equals(iterableToString(set), expected);
        });
```

However, when running the other tests in that stack trace on my IE11 VM they didn't fail.

```
31 12 2016 16:48:02.334:INFO [karma]: Karma v1.3.0 server started at http://localhost:9876/
31 12 2016 16:48:02.335:INFO [launcher]: Launching browser IE11 - Win7 with concurrency 3
31 12 2016 16:48:02.364:INFO [launcher]: Starting browser IE11 - Win7
31 12 2016 16:48:02.592:INFO [launcher.ievms]: Opening VM IE11 - Win7
31 12 2016 16:48:06.371:INFO [IE 11.0.0 (Windows 7 0.0.0)]: Connected on socket /#u43qju6LZsUJ7dbxAAAA with id 81320953
IE 11.0.0 (Windows 7 0.0.0) LOG:

IE 11.0.0 (Windows 7 0.0.0): Executed 9 of 1577 SUCCESS (0.035 secs / 0.008 secs)
31 12 2016 16:48:07.152:INFO [launcher.ievms]: Force stopping VM IE11 - Win7
31 12 2016 16:48:07.152:INFO [launcher.ievms]: Closing VM
31 12 2016 16:48:07.155:INFO [launcher.ievms]: Force stopping VM IE11 - Win7
31 12 2016 16:48:07.155:INFO [launcher.ievms]: Closing VM
```

fatso83:
Hmm weird. I still think your type library is interesting. We are not doing this because we like maintaining code, so if it is better and well maintained; all the better. Maybe a later PR ;)

PR #1225, opened 2016-12-31
Resolve/reject promise only upon invoke

lucasfcosta:
### Purpose (TL;DR) - mandatory

This aims to make `Promise` `resolves` and `rejects` happen only when a stub is `invoked` instead of when the behavior is added to the stub.

Lots of Unhandled Promise Rejection Errors on Tests

### Background (Problem in detail)

In order to add `resolves` and `rejects` to the `stub` object the PR #1211 ended up replacing the `returnValue` property of it with `Promise.resolve(val)` and `Promise.reject(err)`. This ended up resolving or rejecting a promise right when the behavior was added to the `stub` instead of only when the `stub` was invoked.

**This was causing `UnhandledPromiseRejection` errors on our tests, even when we were just defining `rejects` and then superseding it with `resolves`. In the future this could have caused our build (and our user's) to fail on newer Node versions, since unhandled promise rejections will be terminating the Node process with non-zero status.**

### Solution

In order to fix this I added the properties `resolve` and `reject`. These properties indicate whether the stub should `resolve` or  `reject` a promise **when invoked**.

I also added tests to prove that these methods from `Promise` were being called only when the stub is `invoked`.

#### How to verify
2. `npm install`
3. `npm test` --> This will run the newly added tests

Please let me know if I need anything or if you need any further changes.
Thanks for reading this 😄

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -148,6 +148,10 @@ var proto = {
             return context;
         } else if (this.fakeFn) {
             return this.fakeFn.apply(context, args);
+        } else if (this.resolve) {
+            return Promise.resolve(this.returnValue);
+        } else if (this.reject) {
+            return Promise.reject(this.returnValue);
         }
         return this.returnValue;
     },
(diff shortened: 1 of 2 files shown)


fatso83:
Great addition and good catch.

Issue #1220, opened 2016-12-28
Task: Publish new website


mroderick:
In order to publish the new website to https://sinon.js.org, the tasks listed below need to be completed

**Note: this issue will be used as an umbrella for tracking all the remaining tasks. As such, it is not complete and will be modified as new tasks emerge**

I have enabled GitHub Pages for the `master/docs` and have fixed a couple of minor issues. You can preview the new site at https://sinonjs.github.io/sinon/

Once we are happy with what we have there, we should publish it to https://sinon.js.org

#### Remaining work

- [x] Merge #1216 - A PR to improve documentation for stubs and sandboxes
- [x] Implement #1218 - Download page for packaged releases
- [x] Implement #1219 - Publish `Changelog.txt` on new site

#### Publication tasks

Once the tasks in section above have been completed, these are the tasks needed to actuall publish the site to https://sinon.js.org

- [x] Update `docs/_config.yml` to not use the `baseurl` option (fixed in ba0a0ee3ab3c7f002607dac97e42019511ac8547)
- [x] [Configure GitHub to use custom domain](https://help.github.com/articles/adding-or-removing-a-custom-domain-for-your-github-pages-site/)
- [x] Update DNS entry for `sinonjs.org`

mroderick:
Ping @fatso83 @mantoni @fearphage @jonnyreeves @cjno

It would be great if you would look over the site and point out any mistakes or omissions that are essential to publishing the new site. 

Any ideas that are not essential for publishing the new site should be captured in separate issues and marked as ideas.

mantoni:
I think it looks really good. If the plan is to release the site before Sinon 2, the big "FIXME: Out of date" in the API Documentation page could be removed.
👍🍾✨

lucasfcosta:
Hi everyone!

First of all, thanks for the awesome job you have been doing with Sinon.

Regarding the website, I'd like to point a few improvements that me, as a user, would expect:

* The menu items on the top right corner are to close to each other, which makes it difficult to read them. I think that increasing their margins would make it look much better.
  * The `API Documentation` item, for example, looks like two separate items because the words have the same distance between them as the other items have from each one
* Headers are too close of the text immediately before them. Increasing the top-margin for headers would make it easier to distinguish between the start and end of each section

For now I think that's all. Let me know if you agree and if want these changes and I'll be opening a PR for them.

Thanks 😄

mroderick:
@lucasfcosta thank you for your feedback.

[...]

PR #1226, opened 2016-12-31
Improve site readability

lucasfcosta:
### Purpose (TL;DR) - mandatory

This aims to improve the readability of the new site as I've mentioned at #1220.

### Background 

* The menu items on the top right corner are to close to each other, which makes it difficult to read them. I think that increasing their margins would make it look much better.
  * The `API Documentation` item, for example, looks like two separate items because the words have the same distance between them as the other items have from each one
* Headers are too close of the text immediately before them. Increasing the top-margin for headers would make it easier to distinguish between the start and end of each section

#### Solution

I've added the `page-link` class to the navbar `a` elements which adds spacing between them and I've also added a small top margin to header components.

-------------------------------------------

Navbar Before:

screen shot 2016-12-31 at 11 52 03

Navbar After:

screen shot 2016-12-31 at 11 50 52

-------------------------------------------

Headers Before:

screen shot 2016-12-31 at 11 52 14

Headers After:

screen shot 2016-12-31 at 11 53 15

-------------------------------------------

2. `cd docs`
3. `gem install bundler`
4. `bundle install`
5. `bundle exec jekyll serve`
6. Go to `localhost:4000/sinon/`

Let me know if you disagree with any of these changes or if you'd like to see any further improvements.
Thanks 😄

### DIFF
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -2,10 +2,10 @@
     <div class="wrapper">
         <a class="site-title" href="{{ site.baseurl }}/">{{ site.title }}</a>
         <nav class="site-nav">
-            <a href="{{ site.baseurl }}/">Home</a>
-            <a href="{{ site.baseurl }}/releases/{{site.sinon.current_release}}">API Documentation</a>
-            <a href="{{ site.baseurl }}/releases">Releases</a>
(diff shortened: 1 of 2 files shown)


fatso83:
At this rate I think we'll see the new site up before long :-)

Issue #1100, opened 2016-07-13
stub().onCall().returns() should return stub function


zuzusik:
```es6
let a = sinon.stub().returns(12); // a is a stub function
let b = sinon.stub().onCall(0).returns(12); // b is not a stub function (but some object)
```
* Sinon version : 1.17.4

fatso83:
Not sure what this is. It certainly is not a bug, because those chainable methods have never returned a stub AFAIK, so perhaps this is a feature request? If so, I would suggest you try modifying this yourself and submit a PR that keeps the [current API](http://sinonjs.org/docs/#stubs).

Maybe some of the other maintainers know something about the history of this api, and if doing so is a bad idea. 

You basically want to avoid having to write the first line in this code:
```
let b = sinon.stub();
b.onCall(0).returns(12).onCall(1).returns(3);
```
Not sure if this is worth keeping open just to save you a line, so I am going to close this until further.

mvcds:
But I'm sure it has worked in the past, I've use it in another project.

fatso83:
If either @zuzusik or @mvcds can report a version of Sinon where a stub was being returned we have something to work with, as that means we have a regression on our hands. I'll keep it closed until so.

fatso83:
This might be related to #817 and the PR in #1209.

zuzusik:
@fatso83 just investigated a bit, seems like this behavior exists from the very beginning of the implementation of `.onCall()` (v 1.8)

Also I noticed that this code works as expected:

```js
let c = sinon.stub().returns(12); // a is a stub function
let b = sinon.stub().onCall(0).returns(12).stub.onCall(1).returns(13).stub;
```

this means that probably `returns` (and other behavior methods) should  return not `this`, but `this.stub || this` - as far as I see this will guarantee consistency of this methods when being called either directly from stub, or in `.onCall` chain

what do you think?

mvcds:
@fatso83 I am using `sinon@1.17.6` where it behaved as @zuzusik reported.

Separating the code into two lines did the trick though I personally think it does not make as much sense as chaining.

Let me know if I can help you guys with something, I'm new to open source and I'm kind of lost on what or how I can do.

fatso83:
[...]

PR #1227, opened 2017-01-01
Always chain behavior with stub

zuzusik:
#### Purpose (TL;DR)
Fix issue #1100

#### How to verify
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -42,7 +42,7 @@ function throwsException(error, message) {
         this.exception = error;
     }
 
-    return this;
+    return this.chain();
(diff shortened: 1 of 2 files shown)


lucasfcosta:
Very wise manner to allow chaining. I liked it.

fatso83:
Hmm ... of course I missed a thing. It seems as this chaining does not apply to `withArgs`? 

This stub will always return `baz`:
`stub=sinon.stub().returns("moo").withArgs("foo").returns("bar").withArgs("boo").returns("baz")`

This has always been a bug btw, not something new.

PR #1230, opened 2017-01-02
Fix set iterableToString test on IE11

lucasfcosta:
#### Purpose (TL;DR)

This fixes the failing tests on IE11 introduced in #1215 .

#### Background

Our implementation of `typeOf` was too simple and didn't handle older implementations of `Map`s and `Set`s.

Even with the `if` clauses guarding the `Map` and `Set` related tests they would fail, because IE11 did implement them, however, IE11's implementation wasn't following the ECMA standard. This caused our type detection to fail without recognizing these object's types correctly.

Also, IE11's `Set` constructor was a bit weird and didn't accept an array of elements as argument, so I had to use multiple `.add` calls to add elements to a `Set`.

#### Solution

I added `type-detect` to solve the `typeOf` related problem.

I'm not sure everyone agrees on that. So, if you don't want that please let me know and I'll dedicate more time to rewrite `typeOf` and make it compatible with older browsers implementations of `Map`s and `Set`s. However, this library is maintained by everyone at `Chai` and other awesome contributors and it's pretty stable and has good performance.

IMO, having less code to maintain is always good and maybe we can join efforts on that library making it better for everyone.

I have also changed the `Set` constructor on a test in order to use an IE11 compliant way of creating Sets.

2. `npm install`
3. `npm run test`

Let me know if you disagree with anything and I'll happily fix it.

PS.: Sorry for breaking your build! 😅

### DIFF
--- a/lib/sinon/typeOf.js
+++ b/lib/sinon/typeOf.js
@@ -8,12 +8,8 @@
  */
 "use strict";
 
+var type = require("type-detect");
+
 module.exports = function typeOf(value) {
-    if (value === null) {
-        return "null";
-    } else if (value === undefined) {
-        return "undefined";
-    }
-    var string = Object.prototype.toString.call(value);
-    return string.substring(8, string.length - 1).toLowerCase();
+    return type(value).toLowerCase();
 };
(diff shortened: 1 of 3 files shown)


fatso83:
Great PR descriptons btw! Making the life of maintainers much easier.

Issue #956, opened 2015-12-22
Way To Create "Piggyback" Stub


machineghost:
99% of the time when I want to stub something I don't want to call the original method.  But every once in awhile I want to create a stub that enhances, rather than replaces, an existing method.  In other words, I want to create a sort of  "piggyback" stub: a stub that acts like a spy, but lets you add logic before/after the stubbed method.

For example, consider the following class/test:

```
var MyClass = function() {
     this.initialize();
     this.doer.do('something');
};
MyClass.prototype.initialize = function() {
     this.doer = {do: doSomething};
     // initialize some other important stuff
});

var myTest = function() {
    var myObject = new MyClass();
    expect(myObject.doer.do.calledOnce).to.be.true;
});
```

To make the above work, I need to stub the `doer` property, which I can only do by stubbing `initialize`.  But if I stub `initialize` I lose all of the "other stuff" initialization that I need.  Thus, I have to do the following:

```
var myTest = function() {
    var originalInitialize = MyClass.prototype.initialize;
    sandbox.stub(MyClass.prototype, 'initialize', function() {
        originalInitialize.apply(this, arguments);
        sandbox.stub(this.doer, 'do');
    });
    var myObject = new MyClass();
    expect(myObject.doer.do.calledOnce).to.be.true;
});
```

While that's not the end of the world, it would be nice if Sinon could save me from that complexity, using the info it already has.  Imagine if Sinon kept the original version for me:

```
var myTest = function() {
    sandbox.stub(MyClass.prototype, 'initialize', function() {
        this.__PRE_STUB_VERSION.apply(this, arguments);
       sandbox.stub(this.doer, 'do');
    });
}
```

or what if instead Sinon gave me a way to provide before/after functions when I create a stub: 

```
var myTest = function() {
    sandbox.stub(MyClass.prototype, 'initialize', {
        after: function() {
            sandbox.stub(this.doer, 'do');
        }
        // (there could also be a "before" property for running code before the stub)
    });
}
```

Admittedly this isn't the most common case, but something like either of the above would be very handy when it comes up.

fatso83:
Some kind of before/after should be doable. These could also give you access to the original function, avoiding the need to add props on the stub. A la `after : function(originalFunction) { ...} `

You can have a stab at it and submit a pull request.

machineghost:
Will do ... as soon as I can find the time :)

fearphage:
This is loosely related to #988 and #989.

[...]

PR #1234, opened 2017-01-07
Allow calling original function from stub. Closes #989

### LINKED ISSUE #989: Allow calling original function from stub
Seen related ticket #668 about spies, it would still be useful to stub a function only when called with certain parameters, allowing the call to reach to the original function otherwise.

```
sinon.stub(object, 'method')

  .withArgs('1')
  .returns('one')

  .callThrough() // default action, for any other arguments
```

```
sinon.stub(object, 'method')

  .withArgs('1')
  .returns('one')

  .withArgs(sinon.any.arguments) // same, with some syntactic sugar
  .callThrough()
```

One specific use case is to stub individual application configuration variables where they are accessed with a getter function (e.g.: `nconf.get()`).

Configuration objects can get quite large, so we would like to verify different behaviour in integration tests depending on a single configuration variable, but leaving the rest of the configuration untouched, whatever it is, as the rest of the application obviously still depends on it.

lucasfcosta:
## Purpose (TL;DR)

This aims to add the feature requested in #989 and tests for it.

Here's an example of the new feature's usage:

```js
var obj = {
  method: function foo() {
    return 'bar';
  }
};

var myStub = sinon.stub(object, 'method');

myStub.withArgs(1).returns('baz');
myStub.callThrough();

object.method(1); // 'baz'
object.method('not 1'); // 'bar'
```

Due to #817 it's not working in one-liners, but as soon as #1227 gets merged it will work, since that PR allows chaining.

## Solution

In order to implement this I had to change the `wrapMethod` module in order for it to keep a reference for the old method in the new one. I used the key `wrappedMethod` to hold the old method because I thought it was the most semantically valid name possible.

Then I created a new behavior called `callThrough` which adds a property called `callsThrough` to the behavior with the value `true`. 

Whenever the stub can't find any matching fakes for the provided arguments, it will call behavior's `invoke`, as you can see [here](https://github.com/sinonjs/sinon/blob/a6d8e1460e21e0b9cc9a28926b8a130c423cd09e/lib/sinon/spy.js#L195). The `callsThrough` property will then be used by the behavior's [invoke](https://github.com/sinonjs/sinon/blob/a6d8e1460e21e0b9cc9a28926b8a130c423cd09e/lib/sinon/behavior.js#L140) method in order to determine if it should invoke the original `wrappedMethod`.

Please notice that the `wrappedMethod` is invoked with `.apply` in order to keep bot the `this` reference and the arguments correct.

**The tests ensure that the following requirements are met when using `callThrough()`:**

* The original function should not be called if we've got arguments that match any fakes
* The original function should be invoked if the arguments do not match any fakes
* The original function should be invoked with the same arguments when the arguments do not match any fake
* The original function should be invoked with the same `this` reference when the arguments do not match any fake

## How to verify
2. `npm install`
3. `npm run test`

Please let me know if you desire any further changes or if you disagree with anything. 😄 
Also, after this one I'd also like to solve #956, which is partially related (if you don't mind).

Thanks for reading! ❤️

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -152,6 +152,8 @@ var proto = {
             return Promise.resolve(this.returnValue);
         } else if (this.reject) {
             return Promise.reject(this.returnValue);
+        } else if (this.callsThrough) {
+            return this.stub.wrappedMethod.apply(context, args);
         }
         return this.returnValue;
     },
@@ -349,6 +351,11 @@ var proto = {
         this.exception = undefined;
         this.fakeFn = undefined;
 
+        return this;
+    },
+
+    callThrough: function callThrough() {
+        this.callsThrough = true;
         return this;
     }
 };
--- a/lib/sinon/util/core/wrap-method.js
+++ b/lib/sinon/util/core/wrap-method.js
@@ -141,6 +141,8 @@ module.exports = function wrapMethod(object, property, method) {
         }
     };
 
+    method.wrappedMethod = wrappedMethod;
+
     method.restore.sinon = true;
 
     if (!hasES5Support) {
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -2187,4 +2187,92 @@ describe("stub", function () {
         });
     });
 
+    describe(".callThrough", function () {
+        it("does not call original function when arguments match conditional stub", function () {
+            // We need a function here because we can't wrap properties that are already stubs
+            var callCount = 0;
+            var originalFunc = function increaseCallCount() {
+                callCount++;
+            };
+
+            var myObj = {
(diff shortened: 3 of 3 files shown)


RodH257:
Just wanted to say this is great, really helped me out - thanks @lucasfcosta

lucasfcosta:
That's great to hear, thanks @RodH257!
It's this kind of thing that makes me work on open source projects.
Have a nice day buddy 😃

sakovias:
Appreciate this change!

Attrash-Islam:
+1 :heart:

karrtikr:
@lucasfcosta Really helped me out, thanks a lot! Hope you're doing well

Issue #1237, opened 2017-01-09
Wishlist: Have fakeServer record requests


tarjei:
It would be great if Sinon fakeServer stored incoming so that it would be possible to do assertions against them.

* Sinon version : 0.17.6
* Environment   : Browsers
* Example URL   : n.a.
* Other libraries you are using: n.a.

**What did you expect to happen?**
// run test

expect(server.requests).to.have.length(0)

expect(server.requests[0].url).to.eql("xxx")

**What actually happens**

There is no server.requests!

Kind regards,
Tarjei

fatso83:
It looks like [this should already be happening](https://github.com/sinonjs/sinon/blob/master/test/util/fake-server-test.js#L128). Will consider this a non-issue until you can supply us with some code that proves this is not the case, or that can clear up whatever misunderstanding I have created in my head 😃

tarjei:
I think you are correct. I was planning to close this issue but you beat me to it. 

It is still a documentation issue, isn't it? I tried to find out which file I should file a PR against, but failed to find out. If you can point me towards it I'll cook up a PR.

Regards,
Tarjei

fatso83:
Improvements are welcome. We are in the process of building a new site for the Sinon, so the online docs and the source docs might differ. Relevant file [here](https://github.com/sinonjs/sinon/blob/master/docs/release-source/release/fake-xhr-and-server.md).

PR #1238, opened 2017-01-09
Document server.requests

tarjei:
#### Purpose: Document the great server.requests attribute.

See #1237


mroderick:
Thank you for improving documentation!

PR #1239, opened 2017-01-10
Refactor stub methods

mroderick:
This PR refactors `sinon.stub()` and `sandbox.stub()` (internally `collection.stub()` methods to improve legibility.

Feedback on naming of files and functions as well as locations of these in the tree is most welcome.

The aim was to not make *any* changes to functionality, but make it easier to understand the `stub` methods and to highlight differences between them for future improvements.

During refactoring this, I discovered an inconsistency in the API.

#### Proposal: fix inconsistency of stubbing non-function properties

As you can see in `collection.stub` and its tests, it supports stubbing non-function properties, which `sinon.stub` does not.

The documentation for `sandbox.stub` reads:

> The sandbox `stub` method can also be used to stub any kind of property. This is useful if you need to override an object's property for the duration of a test, and have it restored when the test completes

In my opinion, stubs should be used for replacing methods and objects with fakes. Instead of stubbing out a property on an object for the duration of a test, the author should use a stub in the right place to return an entirely fake object that has the (fake) methods and properties that is needed for the test.

I propose that we remove stubbing of non-function properties feature from `collection.stub`, so that it's behaviour matches that of `sinon.stub` (since it would then defer everything to `sinon.stub`).

I am happy to do that as part of this PR.

What do you think?

### DIFF
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -11,12 +11,11 @@
 var sinonSpy = require("./spy");
 var sinonStub = require("./stub");
 var sinonMock = require("./mock");
-var valueToString = require("./util/core/value-to-string");
 var throwOnFalsyObject = require("./throw-on-falsy-object");
 var collectOwnMethods = require("./collect-own-methods");
+var stubNonFunctionProperty = require("./stub-non-function-property");
 
 var push = [].push;
-var hasOwnProperty = Object.prototype.hasOwnProperty;
 
 function getFakes(fakeCollection) {
     if (!fakeCollection.fakes) {
@@ -91,39 +90,22 @@ var collection = {
         return this.add(sinonSpy.apply(sinonSpy, arguments));
     },
 
-    stub: function stub(object, property, value) {
+    stub: function stub(object, property/*, value*/) {
         throwOnFalsyObject.apply(null, arguments);
 
         var isStubbingEntireObject = typeof property === "undefined" && typeof object === "object";
-
-        if (property) {
-            var original = object[property];
-
-            if (typeof original !== "function") {
-                if (!hasOwnProperty.call(object, property)) {
-                    throw new TypeError("Cannot stub non-existent own property " + valueToString(property));
-                }
-
-                object[property] = value;
-
-                return this.add({
-                    restore: function () {
-                        object[property] = original;
-                    }
-                });
-            }
-        }
+        var isStubbingNonFunctionProperty = property && typeof object[property] !== "function";
+        var stubbed = isStubbingNonFunctionProperty ?
+                        stubNonFunctionProperty.apply(null, arguments) :
+                        sinonStub.apply(null, arguments);
 
         if (isStubbingEntireObject) {
-            var col = this;
-            var stubbedObj = sinonStub.apply(null, arguments);
-
-            collectOwnMethods(stubbedObj).forEach(col.add.bind(this));
-
-            return stubbedObj;
+            collectOwnMethods(stubbed).forEach(this.add.bind(this));
+        } else {
+            this.add(stubbed);
         }
 
-        return this.add(sinonStub.apply(null, arguments));
+        return stubbed;
     },
 
     mock: function mock() {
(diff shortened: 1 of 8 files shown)


lib/sinon/collection.js
```diff
@@ -91,46 +90,22 @@ var collection = {
...
-                    typeof getPropertyDescriptor(propOwner, prop).value === "function" &&
-                    stubbedObj.hasOwnProperty(prop)
-                ) {
-                    col.add(stubbedObj[prop]);
-                }
-            });
-
-            return stubbedObj;
+        if (isStubbingEntireObject) {
+            collectOwnMethods(stubbed).forEach(this.add.bind(this));
```
mantoni: Not that it matters here, but you know that you can pass a custom `this` to `forEach` as the second argument?

mantoni:
This is quite a bit of cleanup! I like how you made things more readable by moving the complex decision making logic into separate modules. Looks all good to me.

Regarding the API inconsistency you spotted: I remember running into this as well a while back, but I think the only reason it wasn't cleaned up was laziness. It makes sense to fold this into the stub implementation instead of a collection-only feature.

lib/sinon/collection.js
mroderick: I did, don't use it much, as I try to avoid using `this` as I find it reduces problems when I do :)

mantoni: Absolutely, but since you're using `this` here already ... 😆

mroderick:
> It makes sense to fold this into the stub implementation instead of a collection-only feature.

My proposal was to remove it from `collection.stub`. Are you proposing that we add it to `sinon.stub` instead?

mantoni:
Oh, I didn't read careful enough then. I now remember that it's only implemented for sandboxes because there wouldn't be a possibility to restore the previous value, because the `restore` function is a property of the stub. I think it's a nice feature. We should keep it.

lib/sinon/collection.js
```diff
@@ -11,12 +11,11 @@
 var sinonSpy = require("./spy");
 var sinonStub = require("./stub");
 var sinonMock = require("./mock");
-var walk = require("./util/core/walk");
-var getPropertyDescriptor = require("./util/core/get-property-descriptor");
-var valueToString = require("./util/core/value-to-string");
+var failOnFalsyObject = require("./fail-on-falsy-object");
```
dazld: how about `throwOnFalsyObject` - as this tells you explicitly what this will do. `fail` could mean many different things.

mroderick: I agree, that is a better name. I've renamed the file, the function and all the variables in new commits.

Thank you for the feedback 👍

lib/sinon/stub-descriptor.js
```diff
@@ -0,0 +1,43 @@
...
+var spy = require("./spy");
+var wrapMethod = require("./util/core/wrap-method");
+
+// This is deprecated and will be removed in a future version of sinon.
+// We will only consider pull requests that fix serious bugs in the implementation
+function stubDescriptor(object, property, descriptor) {
+    var wrapper;
+
+    deprecated.printWarning(
+      "sinon.stub(obj, 'meth', fn) is deprecated and will be removed from" +
```
lucasfcosta: Hi, friends, I noticed that the third argument for the stub function is no longer going to be valid. While this seems to be okay when stubbing functions I'd like to know how could we stub the `get/set` properties of descriptors.

Are we going to have a specific API for that or users should just use `Object.getOwnPropertyDescriptor` and then stub the `get` and `set` functions in there?

Also, if you want to take a look at how it works right now, #692 is the PR responsible for introducing it.

fatso83: Good call, although I am not sure.

mroderick: @lucasfcosta that's an interesting discussion. I am not sure it should block merging of this PR. If you agree, would you mind opening a separate issue for the discussion, so we can merge this?

lucasfcosta: Yeah, I totally agree.

Let's discuss that in a separate issue, I've got some ideas for the API and how to implement it, I'd love to help with this. Also, this PR LG(reat)TM.

Awesome work, sir 😄

mroderick:
OK, since there are no objections, I am merging this

PR #1242, opened 2017-01-11
Add documentation for .rejects() and .resolves()

Floby:
#### Purpose (TL;DR)

This PR adds documentation for the functionality provided by #1211.
It has the mention "since sinon@2.0.0".

I completely forgot to document the functionality in the first PR. Sorry =)

2. open `docs/release-source/release/stubs.md`
3. Look for ".resolves" and ".rejects"

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -226,7 +226,7 @@ Causes the stub to return its <code>this</code> value.
 
 Useful for stubbing jQuery-style fluent APIs.
 
-#### `stub.resolves(obj);`
+#### `stub.resolves(value);`
 
 Causes the stub to return a Promise which resolves to the provided value.
 
@@ -268,7 +268,7 @@ Causes the stub to return a Promise which rejects with an exception of the provi
 *Since `sinon@2.0.0`*
 
 
-#### `stub.rejects(obj);`
+#### `stub.rejects(value);`
 
 Causes the stub to return a Promise which rejects with the provided exception object.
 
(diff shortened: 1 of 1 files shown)


docs/release-source/release/stubs.md
```diff
@@ -226,6 +226,15 @@ Causes the stub to return its <code>this</code> value.
 
 Useful for stubbing jQuery-style fluent APIs.
 
+#### `stub.resolves(obj);`
```
mroderick: This is minor: would you mind renaming the argument to `value`, so that users will understand that it can be any value?

docs/release-source/release/stubs.md
```diff
@@ -242,6 +251,30 @@ Causes the stub to throw an exception of the provided type.
...
+
+
+#### `stub.rejects("TypeError");`
+
+Causes the stub to return a Promise which rejects with an exception of the provided type.
+
+*Since `sinon@2.0.0`*
+
+
+#### `stub.rejects(obj);`
```
mroderick: Similar to above, this could be any value. I propose that you rename  the argument to be `value`.

docs/release-source/release/stubs.md
Floby: agreed.
Modeled it after `.returns(obj)` though. I can change that as well.

docs/release-source/release/stubs.md
Floby: agreed

Floby:
PR updated with latest comments.

PR #1253, opened 2017-01-30
Add error equality to deepEqual

BenBrostoff:
#### Purpose
Add `Error` equality to `deepEqual`. This change allows tests to be written like the below, which currently fail:

```javascript
 it('should throw a specific error', () => {
  const err = new Error('very specific error');
  const notSpecificErr = new Error();

  const func = e => e;
  const funcSpy = sinon.spy(func);
  funcSpy(err);

  expect(funcSpy).to.have.been.calledWithExactly(err);
  expect(funcSpy).not.to.have.been.calledWithExactly(notSpecificErr);
});
```

#### Solution

Modify `deepEqual` such that it identifies error comparisons and uses the `===` operator to check equality.

2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/util/core/deep-equal.js
+++ b/lib/sinon/util/core/deep-equal.js
@@ -51,6 +51,10 @@ var deepEqual = module.exports = function deepEqual(a, b) {
             (a.ignoreCase === b.ignoreCase) && (a.multiline === b.multiline);
     }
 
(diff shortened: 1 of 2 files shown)


mroderick:

I think this is a great addition. Code looks ready to merge to me

fatso83:
thanks!

Issue #1221, opened 2016-12-29
Discussion: improve guidelines with JS engine support and target ES version


mroderick:
Hello folks!

Disclaimer: I think that the topics discussed in this issue are closely related and it makes sense to discuss them together. If it turns out to be too confusing, then we can split this into several issues and discuss topics separately.

Solid guidelines makes it much easier to contribute to projects.

As we are gaining traction with more contributors, it is becoming even more important to give good guidelines and explain our aims.

#### ES version in source

ES5  has been out since 2009, it is backwards compatible, easy to polyfill and generally has very good support in popular JS engines.

I think it is time to commit to using ES5 as the official version for Sinon's source.

We should update contribution guidelines to state that we prefer ES5 and ES5 idioms. We should update the ESLint settings (and eventually add plugins) to support this choice as best it can.

If we run into issues with specific runtimes when using ES5 (hello Rhino), we can come up with strategies to support the troublesome engines, like we have in the past.

http://kangax.github.io/compat-table/es5/

Are there any objections to using ES5 in strict mode as our official target for the source?

#### JavaScript engine support

On http://sinonjs.org we claim that the following engines are supported

* IE6+
* Firefox 2+
* Chrome 7+
* Opera 10.10+
* Safari 3.2+
* Mobile Safari 3.2+
* Android 2.1+
* NodeJS 0.2.x+

If we decide on going with ES5 as suggested above, then we need to bump minimum supported versions of most of these runtimes.

In `master` we are using some ES5 features already, which leaves at least old IE out in the cold. I suspect that several of the other browsers are also not supported by the code in `master`.

For the first stable release of Sinon 2, we should revise the supported engines list and try to run tests in the engines that we claim to support.

What JavaScript engines and versions do you think we should target?
How do you think we should communicate this?
Shall we make a compatibility page on the new site, showing the different support levels by Sinon major releases?

mroderick:
Ping @fatso83 @mantoni @fearphage @jonnyreeves @cjno

mantoni:
👍 ES5. I don't think its worth the effort of a compatibility page. Maybe just a short note that Sinon 2 requires ES5. If anyone needs old IE support, they can stick with Sinon 1.17.x. Let's move on.

I didn't follow browser market share developments lately, because I'm mainly doing backend work. However, given that we want ES5 as the baseline, we can still support "old" Node versions without additional work. To reduce testing effort, we could claim to support Node down to 0.10 or 0.12?

[...]

PR #1255, opened 2017-02-01
Remove traces of legacy IE

fatso83:
#### Purpose
Remove code and docs referring to unsupported versions of Internet Explorer. If this is accepted we can remove/tick off most of the remaining checkboxes of #966 (like `Migrate util/fake-xdomain-request suite`).

#1221 pretty much concludes that we support IE11+ for Sinon 2, while #966 still mentions the XDR and XHR code for earlier IE versions. [That seems unnecessary](https://github.com/sinonjs/sinon/issues/966#issuecomment-274250567). To get these points out of the way for a release, let's just delete the code we are not supporting in any case.

2. `npm install`
3.  export SAUCE_USERNAME=sinonjs
4. export SAUCE_ACCESS_KEY=<secret - available for sinonjs maintainers>
5. `npm run test-cloud`


fatso83:
Btw, creating this PR and running the tests on Saucelabs made me aware that the test never run on IE10. Don't know if they ever have, but if we focus on explicitly only supporting IE11 as the last IE for Sinon 2 then it's of no importance. The tests run on IE 11, Edge, Chrome and Firefox.

mroderick:
The PR needs a rebase. Otherwise I think it is ready to merge 👍

PR #1260, opened 2017-02-06
Fire onload event on non-2xx HTTP statuses in FakeXDomainRequest - fixes #1259

### LINKED ISSUE #1259: FakeXDomainRequest still triggers ajax error event instead of load
**Purpose (TL;DR)**
#1040 Brings up the issue of when an error event should be triggered, and when it shouldn't. The conclusion was that a non-2xx status should trigger the load event, _not_ the error event. #1102 fixed this problem in `FakeXMLHttpRequest`, but the problem persists in `FakeXDomainRequest `.

**Minimum Info**
* Sinon version: 1.17.7
* Environment: NodeJS v7.4.0
* Example URL: [none]
* Other libraries you are using: jQuery

What did you expect to happen?
I would expect a status 500 response to trigger the load event.

What actually happens
The status 500 response triggers the error

How to reproduce
I'm leaving out a test case for the time being. I think the problem is evident enough without one. Will consider either creating one as part of a PR or adding one here later.

**In Depth Description**
This is line number in question:
https://github.com/sinonjs/sinon/blob/ac3ca5a5353e68d21523d04a35ef99efce54f011/lib/sinon/util/fake_xdomain_request.js#L85

This line of code is in the `FakeXDomainRequest.readyStateChange` method:
```JavaScript
            case FakeXDomainRequest.DONE:
                if (this.isTimeout) {
                    eventName = "ontimeout";
                } else if (this.errorFlag || (this.status < 200 || this.status > 299)) {
                    eventName = "onerror";
                } else {
                    eventName = "onload";
                }
                break;
```
#1040 Was about the error handler not being called at all; this is still fixed despite this bug. 

The result of this bug is that any libraries (i.e. jQuery) receiving the fake object from SinonJS for testing will also receive the error event when they should be receiving the load event. In the case of jQuery, this means that properties from the XMLHttpRequest object (i.e responseText) won't be copied over to jQuery's jqXHR object meaning the error handler won't receive the body of the response.

I think that jQuery's decision to leave the response body out on an error event is fine because an error event is only supposed to be triggered on network errors. The correct event to trigger here is load, and the discussion on #1040 arrives at the same conclusion.

JoshuaCWebDeveloper:
Fixes #1259 by triggering load event on non-2xx HTTP statuses in FakeXDomainRequest.

 - The issue of when to trigger error vs. load events was originally brought up in #1040. 
 - #1102 fixed the original issue, but FakeXDomainRequest wasn't updated to reflect the decision.
 - This adds a test to ensure that the load event is fired, and updates FakeXDomainRequest.

#### Solution  - optional
See line: https://github.com/JoshuaCWebDeveloper/sinon/blob/802cee6a9c71cef35d7a0d543c5aab670289dda2/lib/sinon/util/fake_xdomain_request.js#L85

Only a status of 0 or a set errorFlag will trigger an error, any other condition will result in a load event during a DONE transition. This is similar to changes introduced in #1102.

2. `npm install`
3. `npm test` (the new test should pass; it will fail if added and run outside of this PR)

### DIFF
--- a/lib/sinon/util/fake_xdomain_request.js
+++ b/lib/sinon/util/fake_xdomain_request.js
@@ -82,7 +82,7 @@ extend(FakeXDomainRequest.prototype, event.EventTarget, {
             case FakeXDomainRequest.DONE:
                 if (this.isTimeout) {
                     eventName = "ontimeout";
-                } else if (this.errorFlag || (this.status < 200 || this.status > 299)) {
+                } else if (this.errorFlag || this.status === 0) {
                     eventName = "onerror";
                 } else {
                     eventName = "onload";
--- a/test/util/fake-xdomain-request-test.js
+++ b/test/util/fake-xdomain-request-test.js
@@ -272,6 +272,11 @@ if (typeof window !== "undefined") {
                 xdr.send();
                 xdr.respond(200, {}, "");
             });
+            it("fire onload event on a non-2xx HTTP status", function () {
+                this.xdr.onload = sinon.spy();
+                this.xdr.respond(500, {}, "");
+                assert.equals(this.xdr.onload.callCount, 1);
+            });
             it("calls readystate handler with readyState DONE once", function () {
                 this.xdr.respond(200, {}, "");
 


fatso83:
Thanks! Just for the record, given that #1255 removes explicit support for anything older than IE11, including XDR, this will not make it into Sinon 2+, but as anyone who needs support for ancient browsers may still continue using Sinon 1, there's not much of a loss.

JoshuaCWebDeveloper:
@fatso83  Understood, this issue was actually encountered on the latest version of NodeJS (I guess `sinon.xhr.supportsCORS` is false in that environment*), but I see that #1255 is going to have Sinon always use XHR, which has already had this issue fixed in #1102.

*https://github.com/sinonjs/sinon/pull/1255/commits/0e42ac3efc1756a2767d04306ce12c4c178cc8e7#diff-2e4c83d83d3d96b1f41ec5b83a73881eL88

Issue #1261, opened 2017-02-08
Use something else besides colors


duclet:
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

* Sinon version : 2.0.0-pre.5
* Environment   : Nodejs
* Example URL   : NA
* Other libraries you are using: NA

**What did you expect to happen?**
Running tests with WebPack should run without issues or warnings.

**What actually happens**
The tests runs fine however, I get the following warnings:

```
WARNING in ./~/colors/lib/colors.js
138:29-43 Critical dependency: the request of a dependency is an expression

WARNING in ./~/colors/lib/extendStringPrototype.js
101:31-45 Critical dependency: the request of a dependency is an expression
```

The reason is because that module is dynamically including in dependencies. The module also looks like it hasn't been updated in a long time and it modify the String prototype which I don't think is good practice. Can we switch to using some other module like "chalk"?

fatso83:
You may if you push a pull request ;-)

duclet:
![bitmoji](https://render.bitstrips.com/v2/cpanel/10120168-229307696_3-s1-v1.png?transparent=1&palette=1&width=246)

PR #1266, opened 2017-02-13
Switch to using "chalk" from "colors"

duclet:
The "colors" library hasn't been updated in over 2 years. It also
uses dynamic imports which causes warnings for those that is using
WebPack 2+ since it isn't allowed.

Resolves: #1261


mroderick:
Very nice, thank you!

mroderick:
I think this is good to merge :

fatso83:
Thanks!

PR #1271, opened 2017-02-19
Refactoring: use ES5.1 features

mroderick:

Refactor code to use ES5.1 features and avoid many `for` and `for..in` loops

2. `npm install`
3. `npm test`

#### Note

This PR is not ready to be merged yet, I haven't removed all the loops yet.

I would like to group the changes better into fewer commits (related changes together), but my brain is too tired to finish it tonight.


lib/sinon/match.js
```diff
@@ -109,7 +107,7 @@ var TYPE_MAP = {
 };
 
 function match(expectation, message) {
-    var m = create(matcher);
+    var m = Object.create(matcher);
```
lucasfcosta: The whole code looks much better and less error prone by using a native function instead of reimplementing it.
Well done!

lib/sinon/mock-expectation.js
```diff
@@ -164,35 +166,36 @@ var mockExpectation = {
...
+            args.length !== expectedArguments.length) {
             mockExpectation.fail(this.method + " received too many arguments (" + format(args) +
-                "), expected " + format(this.expectedArguments));
+                "), expected " + format(expectedArguments));
         }
 
-        for (var i = 0, l = this.expectedArguments.length; i < l; i += 1) {
-
-            if (!verifyMatcher(this.expectedArguments[i], args[i])) {
+        expectedArguments.forEach(function (expectedArgument, i) {
```
lucasfcosta: This looks a lot cleaner too. 👍

lucasfcosta:
Awesome work!
The code looks a lot more clear, especially when you substituted the old implementations for the native ones.
Its impressive to see how cleaner it gets when we start using functions instead of regular control structures. It gets even more interesting when you compare it to languages like Go.

Anyway, philosophical reflections aside, this looks awesome.

lib/sinon/match.js
```diff
@@ -8,7 +8,6 @@
  */
 "use strict";
 
-var create = require("./util/core/create");
 var deepEqual = require("./util/core/deep-equal").use(match); // eslint-disable-line no-use-before-define
 var every = require("./util/core/every");
```
fatso83: Why can't we remove every use of `every` as well in favor of `Array#every`? Are they not doing the same thing?

mroderick: I am working on it :)

mroderick: It'll require a bit of refactoring of the code, as the `every` function is designed to also work with `Object`, because someone decided that they are "iterables"....

lib/sinon/behavior.js
```diff
@@ -377,11 +377,11 @@ function createAsyncVersion(syncFnName) {
 }
 
 // create asynchronous versions of callsArg* and yields* methods
-for (var method in proto) {
+Object.keys(proto).forEach(function (method) {
     // need to avoid creating anotherasync versions of the newly added async methods
     if (proto.hasOwnProperty(method) && method.match(/^(callsArg|yields)/) && !method.match(/Async/)) {
```
mantoni: You can remove the `hasOwnProperty` check here.

lib/sinon/call.js
```diff
@@ -35,32 +34,31 @@ var callProto = {
     },
 
     calledWith: function calledWith() {
-        var l = arguments.length;
-        if (l > this.args.length) {
+        var self = this;
+        var calledWithArgs = Array.prototype.slice.call(arguments);
```
mantoni: There is a `slice` variable at the top of this file because it's used so frequently. Do you prefer using the lengthy version to be more explicit?

lib/sinon/behavior.js
mroderick: I've pushed a review commit that removes it. Once we're happy with the PR, I'll rebase and squash this commit

lib/sinon/call.js
mroderick: I've pushed a commit that uses the `slice` convenience variable

lib/sinon/match.js
lucasfcosta: Yeah, I remember we came to this conclusion when we talked about Array-like objects and Maps and Sets.

In [this comment](https://github.com/sinonjs/sinon/pull/1215#issuecomment-269551320) I talk about it. Actually this method is really useful when it comes to iterating through `Map` and `Set` objects in order to be able to check all of their elements for our related matchers.

mroderick:
I have rebased and squashed the changes from code review into the relevant commits so the history reads better. I think this is good to merge.

I would like to get this merged before the weekend, so we will have a cleaner codebase to work with.

mroderick:
Thanks

erikpukinskis:
Thank you everyone for your hard work on Sinon these years! It's been great for me, but now that you are leaving Node for IO.js, I must part ways!

fatso83:
@erikpukinskis what are you referring to? Doesn't make much sense to me. All code in Sinon should run on pretty much any JS engine for the last 7 years as we only rely on ES5.1 constructs. And IO.js was merged into Node and no longer exists so...

erikpukinskis:
Hm. I think I got confused because sinon-chai hasn't been ported to sinon-2.0 yet, and I'm used to being dropped for runtime reasons! Looks like I can live a little longer on my old systems!

I am still on old pre-IO.js versions of Node. I don't support the direction the ES6 community is going. I think it's important to maintain support for the old JavaScript runtimes.

PR #1273, opened 2017-02-22
Update text-encoding version

### LINKED ISSUE #1272: sinon@next fails on text-encoding import
* Sinon version : `sinon@2.0.0-pre.5`
* Environment   : Mac OS X
* Other libraries you are using: `jest@18.1.0`, `babel-core@6.23.1`

A fix for #878

**What actually happens**
test run fails.

screen shot 2017-02-21 at 1 46 35 pm

**How to reproduce**
```
import sinon from 'sinon'
```

Fails immediately with the above traceback.

**potential fix?**
It seems like this could be potentially fixed by upgrading sinon's text-encoding dependency, which is currently hard-coded to 0.6.2 (current release being 0.6.4). Don't know what the story is there, but they did change their export structure between those releases.

melinath:
Resolved #1272. Updated hard-coded version of text-encoding dependency to resolve import issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "native-promise-only": "^0.8.1",
     "path-to-regexp": "^1.7.0",
     "samsam": "^1.1.3",
-    "text-encoding": "0.6.2",
+    "text-encoding": "0.6.4",
     "type-detect": "^4.0.0"
   },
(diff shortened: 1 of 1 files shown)


fatso83:
Thanks!

PR #1294, opened 2017-02-25
Use Node instead of Ruby in the build script

mantoni:
This replaces the Ruby script with a Node version, adjusted to the new Browserify setup.

Since we now include sourcemaps in Sinon and the resulting file is 2.1 MB, I also create a `no-sourcemaps` version which is less than 900 KB.

Finally, `browserify` is bumped to `^14.1` to align with the test frameworks.


mroderick:
I've tried running it locally, works well 👍

Issue #1258, opened 2017-02-04
Stub get/set on property descriptors with the new API (after deprecating stubDescriptor)


lucasfcosta:
Hi friends, as I have commented [here](https://github.com/sinonjs/sinon/pull/1239#discussion_r96125760), when the [`stubDescriptor`](https://github.com/sinonjs/sinon/blob/master/lib/sinon/stub-descriptor.js#L10) gets deprecated we will not be able to stub `getters` and `setters` on property descriptors anymore since [the function it calls](https://github.com/sinonjs/sinon/blob/master/lib/sinon/stub-descriptor.js) will not be available anymore (the one that allows stubbing descriptors).

I'd either suggest we do not deprecate it since IMO its API is already good or start using `get` and `set` methods chained with the stub. Like this:

```js
// This will not be available anymore
var stub = createStub(obj, "prop", {
  get: function () {
    return 43;
  }
});

// We would write this instead:
var stub = createStub(obj, "prop").get(function () {
  return 43;
});
```

The same goes for `setters`.

What do you think?

fatso83:
Hi, @lucasfcosta, sorry for leaving this hanging for so long. It's a crucial API design bit I think we should decide on what to do with ASAP. I at least forgot this bit when looking at #1239 since I so seldomly stub properties. Hopefully we can decide before the release hackathon we are planning, or what @mroderick and @mantoni? Hopefully we have decided on most bigger decisions before that so we can focus on the lower hanging fruits.
 
I think your proposal looks good. Are you thinking it should be chainable, or what do you intend to do with properties that have both getters and setters?

lucasfcosta:
@fatso83 No problem, sometimes life gets really busy and we have to worry about things other than open source. I too haven't had much time available to send more PRs to Sinon in the last month as I was doing on last year's holidays, I'm looking forward to contributing more next month. Thank you and all the team for your dedication to the project!

------------------------

Regarding the proposal:
I think that both `get` and `set` methods should return the stub itself so we can have both one-liners and we will still be able to write `get` and `set` in a single statement. I think this is the optimal decision for this case because there is not anything else that is interesting to do with `getter/setter` functions (at least I can't remember any).

Also, I think that it is really important for this to be this way because of what I've read in #817. Given that issue and the user's comments we've got there, I think this is what they would expect Sinon to do.

This is an example of what this would allow us to do:

```js
// Since `get` returns the stub itself we don't need to assign in a line and call `get` in another
[...]

PR #1297, opened 2017-02-26
Add getters/setters stub behaviors new API

lucasfcosta:
**This pull request is now complete. Please see [this comment](https://github.com/sinonjs/sinon/pull/1297#issuecomment-285877735) to read about the new changes.**

## Purpose (TL;DR)

This aims to add `get/set` stubbing features for #1258.

I also added some tests for it (supporting only the current state of this PR, not the whole feature).

## Background

As I've said in the previous section, given the new API and deprecation of `stubDescriptor` (#1239) we should have a new API for setting getter and setter stubs.

Right now it's only allowing us to define getters and setters for non-function properties. Which I admit is not that useful as if it allowed stubbing getters and setters for functions too.

**The reason I'm opening an incomplete PR is both to have your feedback on the current work and also because I'll be traveling tomorrow so I won't have time to work on this for the next two weeks so I will leave this here in case someone wants to continue this work (sorry for that!).**

## Solution 

Since this solution is partial I'll explain what I've done until now and the problems I've faced.

First of all I needed to store a reference to the root object and the original property name which we're adding this stub to [in this line](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:getters-setters-new-api?expand=1#diff-048fb1194de4b43f7f7e45758f900749R43). I only do that when someone passes a non-function property in order to avoid calling `wrapMethod` which ends up throwing an error when we pass non-function properties to it. We might want to look forward to allow it to wrap non-function properties too in order to fully accomplish this PR's goal.

The reason I needed to import [`getPropertyDescriptor`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:getters-setters-new-api?expand=1#diff-048fb1194de4b43f7f7e45758f900749R16) is to avoid triggering existing getters when doing our property type checks.

Then finally I have just added two new behaviors which you can see [here](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:getters-setters-new-api?expand=1#diff-90c61de3967bc714354fe10dcfa51b57R353) and [here](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:getters-setters-new-api?expand=1#diff-90c61de3967bc714354fe10dcfa51b57R363). These are the methods responsible for adding the getter/setter replacements to the desired property on the object we're stubbing. It simply gets the reference to the root object and the property's name and defines a new descriptor for it.

#### Known Issues With This Approach:
* We cannot stub getters/setters for functions
* We cannot `restore` these getters/setters

Also, @fatso83 talked to me about not having the stub available anymore when retrieving a property if we stub its property descriptor. I haven't thought about this when I was talking to him, but since the behavior methods return `this.chain()` it would not be a problem, we would just need to assign the result of calling the `get` or `set` method to a variable and then use it to manage our stub.

#### Things we might wanna do for this to be complete:
* Allow `wrapMethod` to wrap non-function properties in order for us to be able to stub them
* Allowing it to wrap non-function properties would also be cool because then we could just call it on our behaviors passing property descriptors and then we would be able to restore them and we also get to reuse more code

In a nutshell: if we make `wrapMethod` work for this use case it would be perfect because then we could use it and accomplish our goal without too much branching.

2. `npm install`
3. `npm test` -> This will run the newly created tests

Feel free to clone my fork and continue your work from there if you want to, this is just a proof of concept and a way of sharing what I've thought as I was talking to @fatso83 this morning.

Sorry again for not being able to finish this, I hope this PR helps you deciding what to do.

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -448,3 +448,38 @@ var stub = sinon.stub().returnsNum(42);
 
 assert.equals(stub(), 42);
 ```
+
+#### `stub.get(getterFn)`
+
+Replaces a new getter for this stub.
+
+```javascript
+var myObj = {
+    prop: 'foo'
+};
+
+createStub(myObj, 'prop').get(function getterFn() {
+    return 'bar';
+});
+
+myObj.example; // 'bar'
+```
+
+#### `stub.set(setterFn)`
+
+Defines a new setter for this stub.
+
+```javascript
+var myObj = {
+    example: 'oldValue',
+    prop: 'foo'
+};
+
+createStub(myObj, 'prop').set(function setterFn(val) {
+    myObj.example = val;
+});
+
+myObj.prop = 'baz';
+
+myObj.example; // 'baz'
+```
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -163,6 +163,26 @@ module.exports = {
 
     callThrough: function callThrough(fake) {
         fake.callsThrough = true;
+    },
+
+    get: function get(fake, getterFunction) {
+        var rootStub = fake.stub || fake;
+
+        Object.defineProperty(rootStub.rootObj, rootStub.propName, {
+            get: getterFunction
+        });
+
+        return fake;
+    },
+
+    set: function set(fake, setterFunction) {
+        var rootStub = fake.stub || fake;
+
+        Object.defineProperty(rootStub.rootObj, rootStub.propName, { // eslint-disable-line accessor-pairs
+            set: setterFunction
+        });
+
+        return fake;
     }
 };
 
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -44,12 +44,11 @@ function stub(object, property, descriptor) {
     var s = stub.create(arity);
     s.rootObj = object;
     s.propName = property;
+    s.restore = function restore() {
+        Object.defineProperty(object, property, actualDescriptor);
+    };
 
-    if (isStubbingNonFuncProperty) {
-        return s;
-    }
-
-    return wrapMethod(object, property, s);
+    return isStubbingNonFuncProperty ? s : wrapMethod(object, property, s);
 }
 
 stub.createStubInstance = function (constructor) {
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -5,6 +5,7 @@ var behaviors = require("./default-behaviors");
 var spy = require("./spy");
 var extend = require("./util/core/extend");
 var functionToString = require("./util/core/function-to-string");
+var getPropertyDescriptor = require("./util/core/get-property-descriptor");
 var wrapMethod = require("./util/core/wrap-method");
 var stubEntireObject = require("./stub-entire-object");
 var stubDescriptor = require("./stub-descriptor");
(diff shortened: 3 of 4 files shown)


fatso83:
Good stuff. I implemented [most of the stuff you mentioned in 1258](https://github.com/sinonjs/sinon/issues/1258#issuecomment-282487849) yesterday, which included most of this, but _not_ the part where you modify the stub in the outer module to keep the reference to the original object. That was the missing part of the puzzle I was too tired to find yesterday :-)

Using your other suggestions I guess most things are covered, and AFAIK this should also work for sandboxes. I think we could easily modify this to also support restoring, if we store the original property definition (`actualDescriptor `) as well. I might have some time to look at this in the coming week, integrating your changes.

mroderick:
@lucasfcosta this branch needs a rebase. How far would you guess you are with this? Any chance of completing it soon?

fatso83:
As said above he will be unable to complete this until he gets back in another two weeks. I did some of it, but won't have time until next weekend.

mroderick:
> As said above he will be unable to complete this until he gets back in another two weeks.

My apologies, I was clearly not paying attention

cjohansen:
Do we want to wait another two weeks for the 2.0 release? Can't this be in 2.1? Or is it breaking?

fatso83:
I'd say we ship this MF. And bump the minor version once we have this.

7. mar. 2017 1.44 p.m. skrev "Christian Johansen" <notifications@github.com

mroderick:
> Do we want to wait another two weeks for the 2.0 release? Can't this be in 2.1? Or is it breaking?

If I understood it correctly, then it we have introduced a breaking change, that has the potential of getting fixed by this PR. While I am keen to ship, I don't think waiting a week or two is going to make that big a difference.

lucasfcosta:
Hi friends, I just got back from London, I'll be able to work on this again. If you have any related PRs or issues that happened in this time but were not linked to this issue, please let me know.

Also, when it comes to releasing this change, I think we should do this ASAP, since this PR adds a feature that has been removed due to the deprecation of the old API (even though it still exists on the codebase).

I'll try to work on this tonight and tomorrow, but I can't guarantee I'll be able to finish it fast since we've got lots of edge cases to cover and checks to remove. I'd like to be careful when removing some of these checks because I want to enable this feature without damaging other restrictions and user feedback. My main concern are the `wrapMethod` restrictions.

lucasfcosta:
**Hi friends, this one is done!**

I will not edit the first post to keep the history of changes and previous explanation. Here goes a description of the changes I've just done.

## Background

So, as I've explained on the first post of this PR we still needed to decide how we were going to wrap function properties. The old approach tried to detect non-function properties and then trigger specific behavior for it. Due to that fact we were not able to wrap function properties, because then we would still call the old `wrapMethod` and then we wouldn't have access to the properties added when we detected non-function properties.

We also needed a restore method to get the old descriptor back in its place if needed.

## Solution

Instead of changing `wrapMethod` and then having to deal with lots of special edge cases, I tried to reuse it and just add new behavior to the stub we are using as the new method. As you can see [here](https://github.com/sinonjs/sinon/pull/1297/files#diff-048fb1194de4b43f7f7e45758f900749R43) I always create a stub with the `rootObj` and `propName` properties doesn't matter if what we're wrapping is a `function` or not. The difference with this new approach is that [we will avoid calling `wrapMethod` if we're not wrapping a method](https://github.com/sinonjs/sinon/pull/1297/files#diff-048fb1194de4b43f7f7e45758f900749R50) and we will just return [the newly created stub itself](https://github.com/sinonjs/sinon/pull/1297/files#diff-048fb1194de4b43f7f7e45758f900749R43) instead.

Also, there's no problem in [always using `arity` as the argument to `stub.create`](https://github.com/sinonjs/sinon/pull/1297/files#diff-048fb1194de4b43f7f7e45758f900749R43) because [it is `0` if `object[property]`'s length is `falsy`](https://github.com/sinonjs/sinon/pull/1297/files#diff-048fb1194de4b43f7f7e45758f900749R29).

Finally [I added a simple restore method to the `stub` we're creating](https://github.com/sinonjs/sinon/pull/1297/files#diff-048fb1194de4b43f7f7e45758f900749R46). It is just a closure that has access to `actualDescriptor`, `object` and `property` so that we can call `Object.defineProperty` on the property that has been wrapped.

I also added a few new tests for that. They ensure this change is compliant with the following specs:
* allows users to stub getter/setter functions for properties
* allows users to stub getter/setter functions for function properties
* allows users to replace old getters/setters without calling them
* can set getters/setters for non-existing properties
* can restore stubbed setters/getters for functions
* can restore stubbed setters/getters for properties

## This Pull Request also includes a commit updating docs with the new API syntax and examples

Sorry for taking so long to continue this. If you need any further changes I'll be more than happy to do them.

fatso83:
We are doing this for free. Don't feel bad :-) The code looks fine, so I am merging this. Coveralls does not come with any good docs, so I am not sure why it fails, but I am guessing it simply fails the build if the coverage drops beyond some threshold.

Looking at some of the non-covered code Coveralls complains about, it seems that it has to do with non-ES5 code which is not triggered. We no longer support non-ES5 browsers, so [that code](https://coveralls.io/builds/10546106/source?filename=sinon%2Futil%2Fcore%2Fwrap-method.js) could well be scrapped.

fatso83:
And with that merge done we are now in ready-to-publish-Sinon-2-land, @sinonjs/sinon-core 🎉

lucasfcosta:
@fatso83 yayyy! Congrats 😃 

Also, I was thinking about refactoring `wrapMethod`, would you mind if I do it and remove the code to handle non-ES5 browsers?

fatso83:
@lucasfcosta sure, go ahead, but it won't stop us from releasing 2.0.

mroderick:
> And with that merge done we are now in ready-to-publish-Sinon-2-land, @sinonjs/sinon-core 🎉

@cjohansen will you do the honours?

mroderick:
> Also, I was thinking about refactoring wrapMethod, would you mind if I do it and remove the code to handle non-ES5 browsers?

Improvements are always welcome 👍

cjohansen:
@mroderick Just did!

https://github.com/sinonjs/sinon/commit/02a6ff6908bf64921967b319d4bfeb0bc9439e5a
https://www.npmjs.com/package/sinon

PR #1300, opened 2017-02-26
Optimize performance of call-stack getting

Gvozd:
call stack not actually needed at most cases, but got always
get(actually only first get) call stack from error is so slow operation, and may be moved on-demand getter

2. `npm install`
3. Put in root of repo this file https://gist.github.com/Gvozd/291e4f0adf247ff70ed72150059e4977
4. Run this file with `node` or `babel-node` command
5. Performance of running this example is better, than withount my fix
x5 times at Node.js
x15 times at babel-node

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -174,10 +174,17 @@ var callProto = {
         return callStr;
     }
 };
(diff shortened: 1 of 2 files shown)


cjohansen:
👍 Nice work!

PR #1302, opened 2017-02-26
addBehavior

### LINKED ISSUE #993: Re-expose lib/sinon/behaviour as a top-level export in Sinon 2
Without an official method for extending Sinon's stub API, a few modules—particularly [sinon-as-promised](https://github.com/bendrucker/sinon-as-promised)—extend the available stub behaviours by patching methods onto `sinon.stub` and `sinon.behavior` directly.

The Sinon 2 beta version doesn't expose behavior as a top-level export anymore (rather it imports it as an internal dependency for lib/sinon/stub), which breaks these modules. While monkey-patching sinon isn't the greatest way to extend the library, it does work, and work nicely.

I attempted to solve this issue at the monkey-patch level using optional imports, but that breaks ES6/CommonJS bundlers.

Since Sinon 2 is required for Webpack/Babel module support, it'd be nice if sinon/lib/behavior was re-exposed as a top-level export to maintain compatibility with these modules.

cjohansen:

Add a proper way to feed in custom behaviors, for people modifying `sinon.behavior` in 1.x

There's docs and a test explaining how this all works. Sinon now uses this mechanism to add its own behaviors, so I'm pretty confident this works as intended. 

This PR fixes #993


mantoni:
Awesome work, mate! 👍

PR #1309, opened 2017-02-28
Fix invalid markdown syntax in fake-timers.md

mroderick:
#### Purpose 

This is the same fix as 7942a15be60a48650e033922172ebc1ecfc13e88 but for 1.17.6 release docs


mantoni:
👍

Issue #1299, opened 2017-02-26
Create redirects for top hits in Google


fatso83:
To avoid losing too much of the hard-earned current SEO score for Sinonjs.org we should quickly get in place some redirects for the top hits on Google. These are mainly `/docs` and `/downloads`. We could quickly just [add redirects to the `config.yml`](https://github.com/jekyll/jekyll-redirect-from) that just points to /releases, but that info would differ too much I guess. Preferably the docs link should always point to `/releases/current`, which does not yet exist, but I am not sure where the download page should go. Possibly just `/releases`?
![skjermbilde 2017-02-26 15 51 16](https://cloud.githubusercontent.com/assets/618076/23340713/7a2f8e3e-fc3b-11e6-8103-7aa9dc8ec13b.png)

fatso83:
This now works after merging #1311. It is just browser redirects, but it works until someone makes a better thing.

PR #1311, opened 2017-03-01
Remove SSL part of base url

fatso83:
Attempt at removing the redirect sending visitors to https. ref [discussion](https://twitter.com/kopseng/status/836326874947158017)

Not possible to test this without deploying, apparently, as it works in either case locally.

refs #1299


fatso83:
Works!

PR #1312, opened 2017-03-02
Fix xhr UploadProgress.eventListerners to handle "on" + event

Boshen:
event listeners should be "on" + event
https://xhr.spec.whatwg.org/#xmlhttprequesteventtarget

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -58,67 +58,14 @@ var unsafeHeaders = {
     "Via": true
 };
 
-// An upload object is created for each
-// FakeXMLHttpRequest and allows upload
-// events to be simulated using uploadProgress
-// and uploadError.
-function UploadProgress() {
-    this.eventListeners = {
-        abort: [],
-        error: [],
-        load: [],
-        loadend: [],
-        progress: []
-    };
-}
 
-UploadProgress.prototype.addEventListener = function addEventListener(event, listener) {
-    this.eventListeners[event].push(listener);
-};
-
-UploadProgress.prototype.removeEventListener = function removeEventListener(event, listener) {
-    var listeners = this.eventListeners[event] || [];
-    var index = listeners.indexOf(listener);
-
-    if (index === -1) {
-        return;
-    }
-
-    listeners.splice(index, 1);
-};
-
-UploadProgress.prototype.dispatchEvent = function dispatchEvent(event) {
-    var listeners = this.eventListeners[event.type] || [];
-
-    listeners.forEach(function (listener) {
-        listener(event);
-    });
-};
-
-// Note that for FakeXMLHttpRequest to work pre ES5
-// we lose some of the alignment with the spec.
-// To ensure as close a match as possible,
-// set responseType before calling open, send or respond;
-function FakeXMLHttpRequest(config) {
-    this.readyState = FakeXMLHttpRequest.UNSENT;
-    this.requestHeaders = {};
-    this.requestBody = null;
-    this.status = 0;
-    this.statusText = "";
-    this.upload = new UploadProgress();
-    this.responseType = "";
-    this.response = "";
-    this.logError = configureLogError(config);
-    if (sinonXhr.supportsCORS) {
-        this.withCredentials = false;
-    }
-
-    var xhr = this;
-    var events = ["loadstart", "load", "abort", "error", "loadend", "progress"];
+function EventTargetHandler() {
+    var self = this;
+    var events = ["loadstart", "progress", "abort", "error", "load", "timeout", "loadend"];
 
     function addEventListener(eventName) {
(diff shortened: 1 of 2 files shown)


fatso83:
You seem to be misreading the spec and the code, and the is no test provided that shows any error with the current code.

The spec says `XMLHttpRequest : XMLHttpRequestEventTarget`, meaning the `XMLHttpRequest` should support the methods in the interface `XMLHttpRequestEventTarget`. That means
`var xhr = new XMLHttpRequest` should result in it being possible to do 
```javascript
xhr.onload = eventHandler1;
xhr.ontimeout = eventHandler2;
// etc
```

Not that the actual events themselves are called `onload`, `onerror`, etc.

If I am misunderstanding you, please supply a testcase that is failing with the _current_ code that should have worked per the spec.

Boshen:
screenshot from 2017-03-02 13-53-50

I am just changing the eventListerners to the correct name, i.e. "on" + event, so we we can manually call `xhr.upload.onprogress` etc.

xhr itself is correct, but the xhr.upload interface is not.

mantoni:
I understand what you want to achieve, but you're adding events that are not event names. The event names never start with "on", they are actually "load", "timeout", etc. The event _handlers_ are prefixed with `on`.

Anyway, if you want to invoke the handlers manually, you should consider using `dispatchEvent` instead as it works with both types of listeners being added.

Boshen:
Well, looking closely at the spec, xhr and xhr.upload both implement `XMLHttpRequestEventTarget`
```
interface XMLHttpRequestEventTarget : EventTarget {
  // event handlers
  attribute EventHandler onloadstart;
  attribute EventHandler onprogress;
  attribute EventHandler onabort;
  attribute EventHandler onerror;
  attribute EventHandler onload;
  attribute EventHandler ontimeout;
  attribute EventHandler onloadend;
};
```

maybe we should just refactor them into one, for example https://github.com/pretenderjs/FakeXMLHttpRequest/blob/master/src/fake-xml-http-request.js#L126-L142

---

Also, I'm not the one who is calling xhr.upload.onprogress, it's in [another library code](https://github.com/ReactiveX/rxjs/blob/master/src/observable/dom/AjaxObservable.ts#L325). Which means we should check for `typeof xhr.upload.onprogress === 'function'` in dispatchEvent...

mantoni:
@Boshen That makes sense.

Boshen:
I reverted my changes and added a failing test case. Hope this helps.

Boshen:
Added a refactoring commit. Both xhr and xhr.upload event handlers are the same now.

Boshen:
This should be correct now. Should I make a another PR (with squashed commits) to filter out all the noise in this PR?

lib/sinon/util/fake_xml_http_request.js
```diff
@@ -58,25 +58,29 @@ var unsafeHeaders = {
...
+    xhr.addEventListener(eventName, function (event) {
+        var listener = xhr["on" + eventName];
+        if (listener && typeof listener === "function") {
+            listener.call(xhr, event);
+        }
+    });
 }
 
-UploadProgress.prototype.addEventListener = function addEventListener(event, listener) {
+function EventTargetHandlers() {
```
fearphage: This should be singular. This is an instead of a single event target handler.

test/util/fake-xml-http-request-test.js
```diff
@@ -2304,9 +2317,11 @@ if (typeof window !== "undefined") {
 
             it("event listeners can be removed", function () {
                 var callback = function () {};
+
+                var eventListenersCount = this.xhr.upload.eventListeners.load.length;
                 this.xhr.upload.addEventListener("load", callback);
                 this.xhr.upload.removeEventListener("load", callback);
-                assert.equals(this.xhr.upload.eventListeners.load.length, 0);
+                assert.equals(this.xhr.upload.eventListeners.load.length, eventListenersCount);
```
fearphage: Could you add a comment explaining this change? My immediate reaction was "This is making the test less explicit". 

Also since we can no longer count on zero listeners, can you change the callback and trigger the event to confirm it is not called?

lib/sinon/util/fake_xml_http_request.js
```diff
@@ -58,25 +58,29 @@ var unsafeHeaders = {
...
-// and uploadError.
-function UploadProgress() {
-    this.eventListeners = {
-        abort: [],
-        error: [],
-        load: [],
-        loadend: [],
-        progress: []
-    };
+function addDefaultEventListener(eventName, xhr) {
```
fearphage: Since we're not reusing this (that I can see), I'd put it inline.

Boshen:
There is already an EventTarget object we can use in util/event ...
Therefore I added another commit to make things even simpler.
This also means we can delete the removeEventListener test case since it's already tested in util/event.

Boshen:
Done and force pushed.

mroderick:
:

PR #1313, opened 2017-03-02
Fix #1274 callCount property is incorrect

### LINKED ISSUE #1274: callCount property is incorrect
* Sinon version :  1.17.7
* Environment   : node v6.9.2
* Example URL   : None
* Other libraries you are using: mocha

I would expect that spy.callCount is number of times the target function was called.
And, spy.withArgs(arg1).callCount is number of times the target function was called with specific arguments.
But, the callCount property is often incorrect.
**What actually happens**
First of all, `f(1)` and `f(1, 1)` and `f(1, 2)` is calling.
As this time, `spy.withArgs(1).callCount` is `3`,
But `spy.withArgs(1, 1).callCount` is often `0`.

This may be affected by calling order `spy.withArgs()`.
Following test code has 2 case pattern, and change the order in which `withArgs(1)` and `withArgs(1, 1)` are called.

**How to reproduce**
```js
const sinon = require('sinon'); // sinon@v1.17.7
const assert = require('referee').assert;

describe("#my-issue", function () {
  it("case1", function () {
    var obj = {};
    obj.f = function () {};

    var spy = sinon.spy(obj, "f");

    // withArgs(1) => withArgs(1, 1)
    assert.equals(spy.callCount, 0);
    assert.equals(spy.withArgs(1).callCount, 0);
    assert.equals(spy.withArgs(1, 1).callCount, 0);

    obj.f();
    obj.f(1);
    obj.f(1, 1);
    obj.f(1, 2);

    assert.equals(spy.callCount, 4);
    assert.equals(spy.withArgs(1).callCount, 3); // passed
    assert.equals(spy.withArgs(1, 1).callCount, 1); // assertion-error
    assert.equals(spy.withArgs(1, 2).callCount, 1); // passed
  });
  it("case2", function () {
    var obj = {};
    obj.f = function () {};

    var spy = sinon.spy(obj, "f");

    // change order, withArgs(1, 1) => withArgs(1)
    assert.equals(spy.callCount, 0);
    assert.equals(spy.withArgs(1, 1).callCount, 0);
    assert.equals(spy.withArgs(1).callCount, 0);

    obj.f();
    obj.f(1);
    obj.f(1, 1);
    obj.f(1, 2);

    assert.equals(spy.callCount, 4);
    assert.equals(spy.withArgs(1).callCount, 3); // assertion error
    assert.equals(spy.withArgs(1, 1).callCount, 1); // passed
    assert.equals(spy.withArgs(1, 2).callCount, 1); // passed
  });
});
```
<details>
    <summary>Result of running above code.</summary>
% mocha test.js

  #my-issue
    1) case1
    2) case2

  0 passing (21ms)
  2 failing

  1) #my-issue case1:
     AssertionError: [assert.equals] 0 expected to be equal to 1
      at referee.fail (node_modules/referee/lib/referee.js:193:25)
      at Object.ctx.fail (node_modules/referee/lib/referee.js:90:21)
      at assertion (node_modules/referee/lib/referee.js:98:21)
      at Function.referee.(anonymous function).(anonymous function) [as equals] (node_modules/referee/lib/referee.js:118:23)
      at Context.<anonymous> (test.js:23:12)

  2) #my-issue case2:
     AssertionError: [assert.equals] 2 expected to be equal to 3
      at referee.fail (node_modules/referee/lib/referee.js:193:25)
      at Object.ctx.fail (node_modules/referee/lib/referee.js:90:21)
      at assertion (node_modules/referee/lib/referee.js:98:21)
      at Function.referee.(anonymous function).(anonymous function) [as equals] (node_modules/referee/lib/referee.js:118:23)
      at Context.<anonymous> (test.js:43:12)
</details>

takasmiley:
#### Purpose (TL;DR)
Fix issue #1274, Rewrite matchingFake in spy.js

`spy.withArgs` function makes fake object and store `this.fakes`, depending on the combination of arguments.
And `spy.invoke` function makes use of `this.fakes`.

`matchingFake` function called by`spy.invoke`, returns the only one fake if arguments are matched.
And `spy.invoke` update the fake properties like `callCount`.
In this time, if some fakes in `this.fakes` are matched arguments, It make this issue happen.

For example
```
spy.withArgs(1, 1);
spy.withArgs(1); // spy.fakes = [ fake(1, 1), fake(1) ]
spy.invoke(1, 1); // invoke increment callCount either fake(1) or fake(1, 1), but not both.
```

#### Solution
Modify matchingFake to matchingFakes, and return array of all fakes matched.
`invoke` function updates properties of this and all matched fakes.
`spy.withArgs` and `stub` are affected by the above fix.
`spy.withArgs` Fixed to fit conventional working.

#### Additional
I investigate around the `matchingFake` function like stub.
And found case3 of #1274.
Conventionally, matchingFake return last fake (pop()).
I Fixed to fit conventional working, but case3 is not solved.
So, I commit 4e68e4d to make use of the most matched argument.

#### How to verify
2. `npm install`
3. `npm test` # include test for this issue


lib/sinon/spy.js
```diff
@@ -308,6 +314,16 @@ var spyApi = {
         return fake;
     },
 
+    matchingFakes: function (args, strict) {
+        if (!this.fakes) {
+            return undefined;
```
mroderick: This could return an empty array, which would make it easier to write elegant consumers of the returned value

lib/sinon/spy.js
```diff
@@ -170,25 +158,30 @@ var spyApi = {
...
+        var matchings = this.matchingFakes(args);
+        var currentCallId = callId++;
+        var exception, returnValue, i;
 
         incrementCallCount.call(this);
         push.call(this.thisValues, thisValue);
         push.call(this.args, args);
-        push.call(this.callIds, callId++);
+        push.call(this.callIds, currentCallId);
+        if (matchings) {
```
mroderick: If `matchingFakes` always returns arrays, then you can simplify this code (Sinon.JS codebase uses ES5.1, so the array extras are allowed).

```javascript
matchings.forEach(function (matching) {
    incrementCallCount.call(matchings);
    push.call(matchings.thisValues, thisValue);
    push.call(matchings.args, args);
    push.call(matchings.callIds, currentCallId);
});
```

lib/sinon/spy.js
```diff
@@ -202,6 +195,13 @@ var spyApi = {
 
         push.call(this.exceptions, exception);
         push.call(this.returnValues, returnValue);
+        if (matchings) {
```
mroderick: As above, assume arrays from `matchingFakes` and prefer ES5.1 array extras over `for` loops.

lib/sinon/spy.js
```diff
@@ -210,6 +210,11 @@ var spyApi = {
             throw err;
         } catch (e) {/* empty */}
         push.call(this.errorsWithCallStack, err);
+        if (matchings) {
```
mroderick: As above, assume arrays from `matchingFakes` and prefer ES5.1 array extras over `for` loops.

lib/sinon/spy.js
```diff
@@ -271,10 +276,11 @@ var spyApi = {
         var args = slice.call(arguments);
 
         if (this.fakes) {
-            var match = matchingFake(this.fakes, args, true);
+            var matchings = this.matchingFakes(args, true);
+            var matching = matchings && matchings.pop();
```
mroderick: If `matchingFakes` always returns arrays, you don't need to test for the result first, and can just use `matchings.pop()` directly.

lib/sinon/stub.js
```diff
@@ -65,7 +67,11 @@ var uuid = 0;
 var proto = {
     create: function create(stubLength) {
         var functionStub = function () {
-            return getCurrentBehavior(functionStub).invoke(this, arguments);
+            var args = slice.call(arguments);
+            var matchings = functionStub.matchingFakes(args);
+
+            var fnStub = (matchings && matchings.pop()) || functionStub;
```
mroderick: This would also benefit if `matchingFakes` always returns an Array

lib/sinon/stub.js
```diff
@@ -70,7 +70,15 @@ var proto = {
             var args = slice.call(arguments);
             var matchings = functionStub.matchingFakes(args);
 
-            var fnStub = (matchings && matchings.pop()) || functionStub;
+            var fnStub;
+            if (matchings) {
```
mroderick: `matchingFakes` should always return arrays, so this test is not necessary

takasmiley:
Hi @mroderick,
Thank you very much for reviewing my work.
I post my opinion as below.

If `withArgs` function can understand whether each argument is matcher or not completely,
I think that `withArgsMatch` may not be necessary.
Because `withArgs` function can understand whether loosely match or exactly match.
And, existing test code is affected and it seems to don't working using `withArgs(include sinon-matcher)`.

Currently, `spy.withArgs` and `stub.withArgs` are same function.
I think that it should be not **`withArgs` and `withArgsMatch`** but **`spy.withArgs` and `stub.withArgs`**.
It seems to behavior of `spy.withArgs` and `stub.withArgs` is different.

The biggest problem I think is, how priority fakes, and the best matched fake from some fakes, and return it.
In `spy.invoke`, it should update all matched fakes(callCount ...) that is returning  matchingFakes function.
But, `spy.withArgs`, `stub.withArgs`, `stub.invoke` must select one or no fake from returning matchingFakes function.

For Example,

```
this.fakes = [
  /* 0 */ fake(1, 1),
  /* 1 */ fake(sinon.match.any, 1),
  /* 2 */ fake(sinon.match.number, 1)
]
```
In spy or stub object is above status, What returning fake is by following action?

1. `spy.withArgs(1, 1)` is called, Every fake is matched, How priority? I expect 0.
2. `spy.withArgs(1, 2)` is called, 1and2 is matched, How priority? I expect 2 rather than 1,
3. `stub.withArgs(1, 1)` is called, Every fake is matched, How priority? I expect 0.
4. `stub.withArgs(2, 1)` is called, 1and2 is matched, What fake does select?, I think no fake selected, and create a new fake(2, 1).
   If it return this.fakes[2], and its behavior updated, stub.invoke(3, 1) is affected and unexpected.
5. `stub.invoke(1, 1)` is called, Every fake is matched, How priority? I expect 0.
6. `stub.invoke(2, 1)` is called, 1and2 is matched, What fake does select?, I expect 2.
   This is different way from `stub.withArgs(2, 1)`.

It seems to be good that matchingFakes function always return array of fake matched.
Four functions which make use of matchingFakes, do each behavior, each selecting mechanism.

In addition, like above question 2, if sinon.match.number prior to sinon.match.any, priority must be defined in simon-matcher.

first of all, I should separate `spy.withArgs` and `stub.withArgs`.

takasmiley:
I'm sorry, perhaps, I might have misunderstood.
After I had send PR, I found three methods, `spy.calledWith`, `spy.calledWithExactly`, `spy.calledWithMatch`.
And, spy object has some series of methods like `***With`, `***WithExactly`, `***WithMatch`.
And, I see `calledWith`, `calledWithMatch`, `calledWithExactly` in call.js.

I see it is different from `withArgs` and new `withArgsMatch`.
Adding a new method `withArgsMatch` looks good to me.

fatso83:
Hmm ... this seems to have stalled. I am not totally sure I understand where this is going. Are you waiting for some feedback, @takasmiley? Wondering about the next step to take to get this merged.

takasmiley:
I use sinonjs in usage like case 1, and I found this problem and fixed it.
The already pushed code solves the problem of case 1 to case 3.
Please code review.

In case4 and case5 of [#1274](https://github.com/sinonjs/sinon/issues/1274), I think that the essence of the problem is different from case1.
I would like to discuss with another issue or PR.

test/issues/issues-test.js
```diff
@@ -200,4 +200,61 @@ describe("issues", function () {
             assert(firstFake !== secondFake);
         });
     });
+
+    describe("#1274 - spy.withArgs(args...).callCount is incorrect", function () {
```
fearphage: Although you're fixing a bug, I believe these tests should go in their respective test files (spy, stub, etc.)

test/issues/issues-test.js
```diff
@@ -200,4 +200,61 @@ describe("issues", function () {
...
+            obj.f(1);
+            obj.f(1, 1);
+            obj.f(1, 2);
+
+            assert.equals(spy.callCount, 4);
+            assert.equals(spy.withArgs(1).callCount, 3);
+            assert.equals(spy.withArgs(1, 1).callCount, 1);
+            assert.equals(spy.withArgs(1, 2).callCount, 1);
+        });
+        it("case2: should count accurately", function () {
```
fearphage: Are case 1 and 2 identical? Please make the descriptions more meaningful.

lib/sinon/spy.js
```diff
@@ -308,6 +307,12 @@ var spyApi = {
         return fake;
     },
 
+    matchingFakes: function (args, strict) {
```
fearphage: Since this is publicly accessible now, it should be tested thoroughly.

fatso83:
@takasmiley As @fearphage noted, this is good to merge, except for some _very_ minor fixes to the tests concerning some naming and location. We would love to have your work integrated into `master`, so could you please update the code with the last fixes?

fatso83:
Thanks for such a quick turnaround! This seems to address Phreds issues. @fearphage?

fatso83:
@mroderick do you feel there is something that needs to be addressed before merging this? I never use the matchers, so I don't feel at home in the discussion on the API behavior - even after reading through it all.

mroderick:
This looks good! Thank you!

mroderick:
@takasmiley would you mind rebasing the branch to get rid of the merge commit at the end?

fatso83:
Thank you for putting up with us and our prolonged process, @takasmiley :blush:

mroderick:
This has been released as `sinon@2.3.6` 🎆

fatso83:
@takasmiley, it looks like this might have introduced a regression. Do you have any chance to look into the findings in #1476? We are not totally sure what is happening, but maybe you do .

PR #1318, opened 2017-03-08
Define window.location when running in jsdom

joaomilho:
When using jsdom (think Jest), window is defined but `window.location` is not. So we need to return a default object in that case.

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -21,7 +21,7 @@ function responseArray(handler) {
     return response;
 }
 
-var wloc = typeof window !== "undefined" && window.location !== "undefined" ? window.location : { "host": "localhost", "protocol": "http"};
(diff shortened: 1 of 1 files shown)


mroderick:
Thank you for your pull request :)

Would you mind creating a failing test case that will be fixed with the changes made?

fatso83:
Is a test case possible here? Thought location was an undeletable property.

mroderick:
> Is a test case possible here? Thought location was an undeletable property.

There's always a way: https://github.com/mroderick/wrapple

mroderick:
This PR needs a rebase

fearphage:
Looking at the merge conflict, it seems this issue may have already been resolved with a fix for Reactive Native in #1363. @joaomilho Can you confirm?

fearphage:
As best I can tell, this looks resolved. 

@joaomilho If you disagree, feel free to reopen this ticket.

PR #1333, opened 2017-03-15
Fix issue #1332 : little bug in spy.printf "%*" formatter.

### LINKED ISSUE #1332: Little bug in spy.printf "%*" formatter
First of all, many thanks for all your hard work on sinon ;)

* Sinon version : v2.0.0
* Environment   : all ? found in node v4
* Example URL   : n/a
* Other libraries you are using: sinon-chai, chai

In the spy.printf(format, args...) function,
when using the "%*" formatter, and passing a list of strings as arguments,
I'd expect all the string to be unquoted.

**What actually happens**
the first string is unquoted,
all the others ones are quoted.

**How to reproduce**
npm install sinon@2.0.0
$ node
> const sinon = require("sinon")
> const spy = sinon.spy()
> spy.printf("%*", "a","b","c")
a, "b", "c"

**The Bug**

In lib/sinon/util/core/format.js:
```
var formatter = formatio.configure({
    quoteStrings: false,
    limitChildrenCount: 250
});

module.exports = function format() {
    return formatter.ascii.apply(formatter, arguments);
};
```
you're calling formatio.ascii with all the arguments of your sinonFormat function.

In lib/sinon/spy-formatters.js
```
module.exports = {
    ....
    "*": function (spyInstance, args) {
        return args.map(sinonFormat).join(", ");
    }
};
```
You're calling sinonFormat directly with map.
- sinonFormat is called with (value, index, collection)
- formatio.ascii is thus called with (value, index, collection)
- formatio.ascii expect (value, processedFlad, indentValue)

the index masquerade as the processedFlag, resulting in the first string being correctly unquoted (index 0 == false) and all the others being quoted (index>0 == true)

**Context**

I tried to make the sinon-chai module work with sinon@2.0.0.
I ran the tests for sinon-chai, and the only error is in the format of the expectation messages.
I traked it down to this little bug in sinon.

eauc:
Fix  #1332 

When use with a list of strings, only the first one is correctly non-quoted :
spy.printf("%*", "a", "b", "c") -> a, "b", "c"

The formatio.ascii function was called with all the arguments provided by map (value, index, collection).
Just added a wrapper in the map() call to forward only the first argument (value).

2. `npm install`
3. node
> const sinon = require("./lib/sinon.js")
> const spy = sinon.spy()
> spy.printf("%*", "a", "b", "c")
'a, b, c'

### DIFF
--- a/lib/sinon/spy-formatters.js
+++ b/lib/sinon/spy-formatters.js
@@ -95,6 +95,6 @@ module.exports = {
     },
 
     "*": function (spyInstance, args) {
-        return args.map(sinonFormat).join(", ");
+        return args.map(function (arg) { return sinonFormat(arg); }).join(", ");
     }
 };
--- a/test/call-test.js
+++ b/test/call-test.js
@@ -1282,6 +1282,16 @@ describe("sinonSpy.call", function () {
 
             assert.equals(spy.printf("%λ"), "%λ");
         });
+
+        it("*", function () {
+            var spy = sinonSpy();
+
+            assert.equals(
+                spy.printf("%*", 1.4567, "a", true, {}, [], undefined, null),
+                "1.4567, a, true, {  }, [], undefined, null"
+            );
+            assert.equals(spy.printf("%*", "a", "b", "c"), "a, b, c");
+        });
     });
 
     it("captures a stack trace", function () {


fatso83:
ha, that was quick :-)

eauc:
#1333

fatso83:
Could you please add a test case that shows the regression? It should fail without your fix and pass with the fix.

eauc:
I tried to follow the contribution guidelines for once so it took a little time ;)

I'll try to make the test yes.

eauc:
Test added.
I tested with a mix of arguments type, and in fact formatio throws an exception with array or objects...

thanks again for your work ;-)

PR #1336, opened 2017-03-16
Add support for bower

paladox:
Downstream in gerrit we use sinonjs for tests, but we use this https://github.com/blittle/sinon.js/ which is an old 1.17 version. I want to update to the newer 1.17.7 release to gain the fixes.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,5 @@
+{
+  "name": "sinonjs",
+  "version": "1.17.7",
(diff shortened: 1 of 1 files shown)


mroderick:
Can some of the community solutions in https://github.com/sinonjs/sinon/pull/253 be used to solve your problem?

paladox:
Yep. Seems it's called sinon on bower and not sinonjs.

PR #1337, opened 2017-03-17
[feature] adds spy.calledImmediatelyBefore and spy.calledImmediatelyAfter

timcosta:
Sometimes you need to be sure that you are calling `a().c()` and not `a().b().c()`. When mocking a large library like `knexjs`, being able to ensure the exact order of calls is incredibly helpful.

When mocking a large library that chains calls, you may end up writing one large mock in the `beforeEach` of the test library as to not duplicate code within your tests. Unfortunately there is no way to ensure that chained calls do not sneak into calls chains currently, as the `calledBefore` and `calledAfter` implementations simply use `>` or `<`. In order to use a large mock and reduce duplicitous code in test suites when dealing with libraries that chain calls, `calledImmediatelyBefore` and `calledImmediatelyAfter` allow you to ensure that no additional modifiers end up between your chained calls.

#### Solution  - optional
This solution works just like `calledBefore` and `calledAfter`, but uses strict equality to ensure that the calls to the spies did in fact happen sequentially.

2. `npm install`
3. `npm test` (includes 100% test coverage)

### DIFF
--- a/test/spy-test.js
+++ b/test/spy-test.js
@@ -1650,6 +1650,114 @@ describe("spy", function () {
         });
     });
 
+    describe(".calledImmediatelyAfter", function () {
+        beforeEach(function () {
+            this.spy1 = createSpy();
+            this.spy2 = createSpy();
+            this.spy3 = createSpy();
+        });
+
+        it("is function", function () {
+            assert.isFunction(this.spy1.calledImmediatelyAfter);
+        });
+
+        it("returns true if first call to A was immediately after first to B", function () {
+            this.spy2();
+            this.spy1();
+
+            assert(this.spy1.calledImmediatelyAfter(this.spy2));
+        });
+
+        it("compares calls directly", function () {
+            this.spy2();
+            this.spy1();
+
+            assert(this.spy1.getCall(0).calledImmediatelyAfter(this.spy2.getCall(0)));
+        });
+
+        it("returns false if not called", function () {
+            this.spy2();
+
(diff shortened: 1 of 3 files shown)


test/spy-test.js
```diff
@@ -1650,6 +1650,114 @@ describe("spy", function () {
...
+            this.spy1 = createSpy();
+            this.spy2 = createSpy();
+            this.spy3 = createSpy();
+        });
+
+        it("is function", function () {
+            assert.isFunction(this.spy1.calledImmediatelyAfter);
+        });
+
+        it("returns true if first call to A was immediately after first to B", function () {
```
mroderick: This is minor, but it would make the tests easier to read, if you use `A` in both description and implementation. Otherwise readers would have to understand that `A` is `this.spy1`

test/spy-test.js
```diff
@@ -1650,6 +1650,114 @@ describe("spy", function () {
...
+            this.spy1 = createSpy();
+            this.spy2 = createSpy();
+            this.spy3 = createSpy();
+        });
+
+        it("is function", function () {
+            assert.isFunction(this.spy1.calledImmediatelyBefore);
+        });
+
+        it("returns true if first call to A was immediately after first to B", function () {
```
mroderick: Same here, `A` is `this.test1`

timcosta:
Thanks for reviewing @mroderick - updated the names in my tests and the tests i stole the template from.

PR #1339, opened 2017-03-17
Redesign the template

afnizarnur:
Redesign the documentation based on #1304


fatso83:
Seems we need to exclude the docs folder from listing. Ref the failing build https://travis-ci.org/sinonjs/sinon/jobs/212364765

Would you mind, @afnizarnur?

afnizarnur:
Already done :) @fatso83

fatso83:
Took a while before I could `bundle exec jekyll serve`, but finally did. This looks _tons better_, @afnizarnur! Thanks a lot for your valuable work. The legibility is improved a lot and the general design is smoother and friendlier to the eye. I will hold off merging until another one on the core team signs this one off. What do you think, @sinonjs/sinon-core - should we starte the new week with this design?

mroderick:
This is looking pretty good!

I do think that the content area should be wide enough to accommodate 80 chars of code in the examples, without horizontal scrolling.

2017-03-19 at 19 23

I think in the case like this, visitors are most likely to be using desktop browsers with desktop screens.

What do you think?

docs/_includes/footer.html
```diff
@@ -1,24 +1,9 @@
-<footer class="site-footer">
-    <div class="wrapper">
-        <div class="footer-col-wrapper">
-            <div class="footer-col footer-col-1">
-                <ul class="contact-list">
-                    <li>{{ site.title }}</li>
-                    <li><a href="http://groups.google.com/group/sinonjs">http://groups.google.com/group/sinonjs</a></li>
```
fatso83: Are we sure we want to get rid of this? I think we should have at least the same pointers as we already have in the README on Github. If not in the same place, then somewhere.

docs/_layouts/default.html
```diff
@@ -1,20 +1,15 @@
...
-      <div class="wrapper">
-        {{ content }}
-      </div>
+<html lang="en">
+{% include head.html %}
+<body>
+  {% include header.html %}
+
+  <div class="content">
+    <div class="container container-2">
```
fatso83: This might be a bit of a nitpick, but `container-2` does not tell me much in the way of "what separates this from the normal container". Might choose a slightly more telling name?

docs/_releases/v2.0.0.md
```diff
@@ -1,10 +1,10 @@
 ---
 layout: page
-title: API documentation
+title: Sinon.JS - API documentation
```
fatso83: Good, this is a lot better for SEO.

fatso83:
I also find that we should be able to see a width of 80 chars in code snippets _on desktop_. Scrolling is fine if on mobile.

docs/_includes/footer.html
mantoni: The link is on the homepage at the very bottom now. However, I think having the Google group to the footer on all pages (additionally) is a good idea.

mantoni:
Looks great! I agree that the font size of the code snippets could be a little smaller to fit 80 characters. I'd vote to merge and deploy this with the one font-size change – or maybe even as is – and iterate on it from there. It's already a big step forward.

Thank you @afnizarnur for the nice work! 👏 👍

fatso83:
Yeah, I agree on the "merge and iterate until perfect" approach. Merge at will :-)

mantoni:
Aaaaaaand the new design has shipped! http://sinonjs.org

However, the GitHub asset is missing, which was not the case when I checked locally. Any idea why @afnizarnur ?

mantoni:
No worries, got it. GitHub hosting is apparently case sensitive while the locally served version is not.

docs/_includes/footer.html
afnizarnur: Yeah, I forget to add the google group 😅 . Will proceed it now, and add it to the footer.

docs/_layouts/default.html
afnizarnur: I agree with you. Basically, it's just different container size.

docs/_releases/v2.0.0.md
afnizarnur: Yeah, may I change all of the title pages to the better title? I found many titles in documentation is not using good SEO

afnizarnur:
Wow cool thanks! Glad to help Sinon.JS 😃  🙌 🎉🎉🎉🎉  @mantoni

PR #1343, opened 2017-03-20
Add external links to sinon.js related articles

jhartikainen:
#### Purpose

Add links to useful Sinon.js related articles (as discussed on #1304)

#### Solution

I added links into a datafile in `_data/external-howtos.yml`, and updated the howto page to include a section with "articles elsewhere on the web".

I noticed the other articles were being sorted - I tried sorting the external ones, but it was causing an error with Jekyll and I couldn't really figure out how to fix it.

I'm the author of the articles linked, and I've received a lot of positive feedback on them. I think it would be good to include more (I've had a lot of people ask me about async tests in particular), but I'm not familiar with any good articles beyond these.

### DIFF
--- a/docs/_data/external_howtos.yml
+++ b/docs/_data/external_howtos.yml
@@ -1,0 +1,15 @@
+- title: How to test an Ajax request using Sinon's fake XMLHttpRequest
+  url: https://codeutopia.net/blog/2015/03/21/unit-testing-ajax-requests-with-mocha/
+
+- title: How to stub or mock complex objects, such as DOM objects
+  url: https://codeutopia.net/blog/2016/05/23/sinon-js-quick-tip-how-to-stubmock-complex-objects-such-as-dom-objects/
+
+- title: Using Sinon.js with Promises
+  url: https://www.sitepoint.com/promises-in-javascript-unit-tests-the-definitive-guide/
+
+- title: Best practices for spies, stubs and mocks
+  url: https://semaphoreci.com/community/tutorials/best-practices-for-spies-stubs-and-mocks-in-sinon-js
+
+- title: Using Sinon.js to help test Mongoose models
+  url: https://codeutopia.net/blog/2016/06/10/mongoose-models-and-unit-tests-the-definitive-guide/
+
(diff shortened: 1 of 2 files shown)


mantoni:
LGTM. @sinonjs/owners ?

fatso83:
Good articles is always welcome. Still ... I would actually prefer to have them included in the docs. The reason is quite simply that external links rot. We could still attribute the original author with "originally published on name-of-blog-with-link" and an author field. That would keep valuable docs in one repo. Still, I don't mind linking to externals. It is just that a lot of the links on the previous Sinon page went sour in time ... :-/

fatso83:
BTW, my comment above was not "I refuse these additions", but more of a general thought I would like to shed some light on. It's up for discussion :-)

mantoni:
I agree that adding a copy of the articles into the Sinon homepage would be nicer. Some of the articles only touch on Sinon and are mainly concerned with other topics. If the relevant parts can be added here, it is more relevant content. Of cause the source should be credited with something like "Cross posted from xyz". What do you think @jhartikainen ?

jhartikainen:
The ones which are on my site (codeutopia.net) could probably be adapted. The two others maybe not. fwiw at least in terms of link-rot, my content from 10 years ago is still up, and I don't intend to go anywhere :)

cjohansen:
Just for the record, the link rot being discussed was mostly caused by me, and I didn't intend for that to happen either :) My server crashed, there was data loss, and I couldn't be bothered to re-configure the horrible CMS that used to power it. I've been meaning to bring it back up in another form, but it requires a little work that I somehow haven't found time to put in for about two years now...

fatso83:
OK, then I suggest we just merge this and kindle ask for permission to import the most relevant content as time allows :-)

@cjohansen : your content is available through the Wayback machine. I have been meaning to re-import them into the new docs.

mroderick:
I agree that having articles in our repository would be preferable. That can be a longer term goal. I see this PR as providing benefit today, so let's merge it and keep discussing.

mantoni:
Thanks everyone.

PR #1345, opened 2017-03-20
Add google group button & edit all page title

afnizarnur:
Add google group button & edit all page title – make it better SEO


mantoni:
Nice and consistent! 👍 Thank you.

PR #1355, opened 2017-03-22
Return fake to support chaining

swarthy:
Support chaining and more comfortable mock object creation (my case)

Without this pr there is no clean way to create object with stubs
```javascript
const mock = {
  method1: sinon.stub().resolves(123),
  method2: sinon.stub().rejects(new Error()),
  // etc
}
```
you need first create stubs and only then set their behavior
```javascript
const mock = {
  method1: sinon.stub(),
  method2: sinon.stub(),
  // etc
}
mock.method1.resolves(123)
mock.method2.rejects(new Error())
// etc
```

#### Solution  - optional
Return ```fake``` from behavior method

2. `npm install`
3. 
```javascript
const stubMethod = sinon.stub().returns(123) // stubMethod is undefined
```

### DIFF
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -13,6 +13,8 @@ function throwsException(fake, error, message) {
     } else {
         fake.exception = error;
     }
+
+    return fake;
 }
 
 module.exports = {
@@ -30,6 +32,8 @@ module.exports = {
         fake.callbackContext = undefined;
         fake.callArgProp = undefined;
         fake.callbackAsync = false;
+
+        return fake;
     },
 
     callsArgOn: function callsArgOn(fake, pos, context) {
@@ -42,6 +46,8 @@ module.exports = {
         fake.callbackContext = context;
         fake.callArgProp = undefined;
         fake.callbackAsync = false;
+
+        return fake;
     },
 
     callsArgWith: function callsArgWith(fake, pos) {
@@ -54,6 +60,8 @@ module.exports = {
         fake.callbackContext = undefined;
         fake.callArgProp = undefined;
         fake.callbackAsync = false;
+
+        return fake;
     },
 
     callsArgOnWith: function callsArgWith(fake, pos, context) {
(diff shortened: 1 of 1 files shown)


mroderick:
Thank you for your pull request. Your suggestion looks like it could be beneficial to everyone, especially those that use mocks.

If you want this PR merged, you should add some tests as well.

This PR will change the API (going from `undefined` to a fake as return value), so we will cut a new major version once this is merged.

harmon:
This would be AMAZING! I was just in need of this. Super helpful for "returns()" chaining.

mantoni:
@mroderick Is this really a major? Adding a return value where none was present before is adding a feature, isn't it? I don't see how this could break existing code.

lucasfcosta:
@mantoni I agree with @mroderick.

All functions in JavaScript return something and `undefined` is a valid return value.

IMO it doesn't matter whether or not someone is relying on it returning undefined, if we're changing return values this should be a new major version, even though people using it are doing something "wrong" or unexpected.

This is just my own opinion, I don't think it would be a great problem to release this bumping the minor version number only. However, if we want to comply strictly with SemVer rules I think @mroderick is correct.

mantoni:
@lucasfcosta SemVer states "Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API". My understanding is that compatibility refers to the "contract" of an API. If something is neither explicitly implemented nor documented, then there there is no "contract" to rely on. This is where I see the freedom to introduce new SemVer minor features. This includes return values.

For example, if we would document a specific function with "Always returns false", then it would be a breaking (SemVer major) change to return anything else. Functions that return implicitly undefined and have no documentation about the return value can start returning something at any time, even in a patch if that would be required internally.

swarthy:
I think I can close PR, because chaining already [supported](https://github.com/sinonjs/sinon/blob/master/lib/sinon/behavior.js#L182-L197) (and [tested](https://github.com/sinonjs/sinon/blob/master/test/stub-test.js#L202))

PR #1356, opened 2017-03-22
forEach() requires 'this' argument

hrimhari:
Otherwise, this is evaluated to undefined, which causes cryptic errors like:

`Cannot read property 'method' of undefined`

instead of:

`writeFileSync received wrong arguments(...)`

2. `npm install`
3. <your-steps-here>

### DIFF
--- a/lib/sinon/mock-expectation.js
+++ b/lib/sinon/mock-expectation.js
@@ -182,7 +182,7 @@ var mockExpectation = {
                 mockExpectation.fail(this.method + " received wrong arguments " + format(args) +
                     ", expected " + format(expectedArguments));
             }
-        });
+        }, this);
     },
 
     allowsCall: function allowsCall(thisValue, args) {
--- a/test/mock-test.js
+++ b/test/mock-test.js
@@ -475,6 +475,17 @@ describe("sinonMock", function () {
         });
 
         describe(".withArgs", function () {
+            var expectedException = function (name) {
+                return {
+                    test: function (actual) {
+                        return actual.name === name;
+                    },
+                    toString: function () {
+                        return name;
+                    }
+                };
+            };
+
             it("returns expectation for chaining", function () {
                 assert.same(this.expectation.withArgs(1), this.expectation);
             });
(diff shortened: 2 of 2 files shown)


ktsn:
I'm facing this problem when using `withArgs` too, hope this get merged.

mantoni:
Please add a test to verify.

ktsn:
I'm not sure about the assertion library (`referee`) so I may be wrong, but the `assert.exception` seems ignores any exceptions.

I confirmed that in the test case [`sinonMock .expectation .withArgs throws when called with wrong args` in `mock-test.js`](https://github.com/sinonjs/sinon/blob/master/test/mock-test.js#L507) by removing `assert.exception` that wraps `expectation(2, 2, 3);`
It throws `Cannot read property 'method' of undefined` that is not seems like the expected exception.

mantoni:
@ktsn `assert.exceptions` does not re-throw the exception, but verifies the exception type. In this case, the `name` of the exception is expected to be `"ExpectationError"`. The message you see is the `message` of the exception, which is not asserted by this test case.

ktsn:
@mantoni Yes, the second argument specifies the name of exception however the actual exception name is `"TypeError"` but there is no assertion error... Does that mean the existing test case is something wrong or `assert.exception` has any bug, doesn't it?

mantoni:
@cjohansen You know about the details in referee better than me. Can you have a look?

hrimhari:
You already have tests, but they all seem to misuse referee.assert.exception() by passing a string as matcher. My understanding of how referee.assert.exception() works is that if you pass a string as second parameter it will take it as a message, not a matcher. See https://github.com/busterjs/referee/blob/master/lib/referee.js#L551
You'd have to review all your test cases regarding exceptions.

hrimhari:
One of the tests that should be failing: https://github.com/hrimhari/sinon/blob/patch-1/test/mock-test.js#L513

hrimhari:
Any news?

hrimhari:
bump

mantoni:
Seems fine. Sorry to keep you waiting @hrimhari.

Issue #1362, opened 2017-03-23
React Native iOS: undefined is not an object (evaluating 'wloc.protocol') 


greena13:
* Sinon version : 2.1.0
* Environment   : React Native iOS
* Other libraries you are using:
React Native v0.40.0
React 15.4.1

**What did you expect to happen?**

Sinon to initialise without an error.

**What actually happens**

When running against in the React Native iOS build, the following error occurs:

    undefined is not an object (evaluating 'wloc.protocol') 

**How to reproduce**

     import sinon from 'sinon'; // in a react native project, running on iOS

PR #1363, opened 2017-03-23
Add support for React Native window location format Fixes sinonjs/sin…

greena13:
#### Purpose (TL;DR)

Fix issue #1362 and allows sinon to be used with React Native builds (where the location object is not stored directly on window). 

Preserves existing behaviour for all other environments, including fallback object.

#### How to verify
2. `npm install`
3. import sinon into a React Native project and test on Android and iOS that it does not raise the error

    undefined is not an object (evaluating 'wloc.protocol')

### DIFF
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -21,31 +21,30 @@ function responseArray(handler) {
     return response;
 }
 
-var wloc = getWindowLocation();
-
-var rCurrLoc = new RegExp("^" + wloc.protocol + "//" + wloc.host);
-
-function getWindowLocation() {
-    if ( typeof window !== "undefined") {
-        if (typeof window.location !== 'undefined') {
-            // Browsers place location on window
-            return window.location;
-        } else if((typeof window.window !== 'undefined') && (typeof window.window.location !== 'undefined')) {
-            // React Native on Android places location on window.window
-            return window.window.location;
-        } else {
-            return getDefaultWindowLocation();
-        }
-    } else {
-        // Fallback
-        return getDefaultWindowLocation();
-    }
-}
-
 function getDefaultWindowLocation() {
   return { "host": "localhost", "protocol": "http"}
 }
 
+function getWindowLocation() {
+  if ( typeof window !== "undefined") {
+    if (typeof window.location !== 'undefined') {
+      // Browsers place location on window
+      return window.location;
+    } else if((typeof window.window !== 'undefined') && (typeof window.window.location !== 'undefined')) {
+      // React Native on Android places location on window.window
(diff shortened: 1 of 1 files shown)


lib/sinon/util/fake_server.js
```diff
@@ -21,7 +21,29 @@ function responseArray(handler) {
...
+function getDefaultWindowLocation() {
+    return { "host": "localhost", "protocol": "http" };
+}
+
+function getWindowLocation() {
+    if ( typeof window !== "undefined") {
+        if (typeof window.location !== "undefined") {
+            // Browsers place location on window
+            return window.location;
+        } else if ((typeof window.window !== "undefined") && (typeof window.window.location !== "undefined")) {
```
fatso83: wow, that is proper weird, but if this works ... then fine. 

`window.window.location`? jeez 😸

lib/sinon/util/fake_server.js
```diff
@@ -21,7 +21,29 @@ function responseArray(handler) {
...
-var wloc = typeof window !== "undefined" ? window.location : { "host": "localhost", "protocol": "http"};
+function getDefaultWindowLocation() {
+    return { "host": "localhost", "protocol": "http" };
+}
+
+function getWindowLocation() {
+    if ( typeof window !== "undefined") {
+        if (typeof window.location !== "undefined") {
+            // Browsers place location on window
+            return window.location;
```
mroderick: When you use return inside `if` statements, you can avoid using `else`.

See http://rmurphey.com/blog/2012/12/10/js-conditionals

Also, you can unfold this code a bit, so there are fewer levels of indentation == easier to understand.

```javascript
function getWindowLocation() {
    if (typeof window === "undefined") {
        // Fallback
        return getDefaultWindowLocation();
    }

    if (typeof window.location !== "undefined") {
        // Browsers place location on window
        return window.location;
    }

    if ((typeof window.window !== "undefined") && (typeof window.window.location !== "undefined")) {
        // React Native on Android places location on window.window
        return window.window.location;
    }

    return getDefaultWindowLocation();

}
```

greena13: Thanks for your suggestion. I agree, it can be simplified in this way.

lib/sinon/util/fake_server.js
greena13: Yes, for reasons unknown to me, this does indeed seem to be where `location` is stored, in the versions of React Native that I tested against.

greena13:
Patch suggestions applied. :)

fatso83:
Thanks for your contributions!

mroderick:
`sinon@2.2.0` is now on `npm`

Issue #1354, opened 2017-03-22
Enhancement: Add ability to override global promise library


blacksun1:
* Sinon version : 2.1.0
* Other libraries you are using: Bluebird

Hi,

I want to be able to override the default promise library with an external library such as Bluebird or Q.

We don't want to override the Global Promise library (which is the native Promise library) but instead import/require it when needed.

Would you be interested in me providing a Pull Request for this or would I have to create a fork.

Thanks in advance,

mantoni:
Can you please elaborate why you think your use case needs a change in Sinon?

mroderick:
How do you imagine using it?

Like [`sinon-as-promised`](https://github.com/bendrucker/sinon-as-promised/blob/master/examples/bluebird/test.js) does it?

Kynde:
Use case is that our code uses Bluebird additions like .tap() and .finally(). Stubbing with the native Promises does not really work.

We used to use sinon-as-promised and that allowed using Bluebird.  Now I have horrible kludge in test suites that add such missing functions to the native Promise prototype and I would really like to get rid of that.

blacksun1:
Yes, that is our issue as well.

At the moment, our code looks like this

```js
sinon.stub(someObject, 'someMethod').callsFake(() => return bb.Promise.resolve({}));
```

I would want to work like this though.

```js
const bb = require('bluebird');
const sinon = required('sinon');

sinon.promiseLibrary = bb.Promise
sinon.stub(someObject, 'someMethod').resolves({});
sinon.stub(someObject, 'someOtherMethod').resolves({});
```

or

```js
const bb = require('bluebird');
const sinon = required('sinon');

sinon.stub(someObject, 'someMethod').resolves({}); // Would default to global promise
sinon.stub(someObject, 'someMethod').resolves({}, bb.Promise); // World use included promise
```

I think it would be important to be able to select the specific library you want to resolve with as sometimes we work with various libraries that return various promise types.

fatso83:
We could add a `configure` call like `sinon-test`? https://github.com/sinonjs/sinon-test#usage

blacksun1:
Or

```js
sinon({ promiseLibrary: bb.Promise }).stub();
```

fatso83:
Yeah, as long as we configure an instance of Sinon, and not set defaults globally I don't mind either way. We fought long and hard to get rid of globals before the release of version 2, so let's not introduce more 😄

blacksun1:
[...]

PR #1364, opened 2017-03-24
Added `usingPromise` method to stub.

blacksun1:
#### In progress

Things to still be done:

* [X] Get agreement on implementation
* [X] Add `.usingPromise` to higher level objects such as sandbox
* [x] Add documentation

#### Purpose 

The `usingPromise` method allows the setting of the promise library
that will be used by the resolve and reject method. If it is not set then it will use the default promise implementation by default.

This will eventually fix https://github.com/sinonjs/sinon/issues/1354 

#### Background

To allow the setting of a custom promise library when creating a stub that resolves or rejects. Sometimes you need to get back a custom promise library to have access to it's extended methods and so on, such as `.tap`, `.done` etc.

#### Solution

Add a `.usingPromise` method that sets the promise library to use.

Example using `sinon.stub()`:

```javascript
var assert = require("assert");
var bluebird = require("bluebird");
var sinon = require("sinon");

var myObject = {};
var myStub = sinon.stub()
    .usingPromise(bluebird.Promise)
    .resolves(myObject);

myStub()
    // Tap should now be available!!!
    .tap(function(actual) {
        assert.strictEqual(actual, myObject);
    })
    .catch(function(err) {
        console.err("Error", err);
    });
```

Example using `sinon.sandbox()`:

```javascript
var assert = require("assert");
var bluebird = require("bluebird");
var sinon = require("sinon");

var myObject = {
  myMethod: function() { return; }
};

var sandbox = sinon.sandbox.create()

sandbox.stub(myObject);

myObject.myMethod
    .usingPromise(bluebird.Promise)
    .resolves("baz");

myObject.myMethod()
    // Tap should now be available!!!
    .tap(function(actual) {
        assert.strictEqual(actual, "baz");
    });
```

#### How to verify

2. `npm install`
3. `npm test`

### DIFF
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -322,7 +322,7 @@ describe("stub", function () {
             var stub = createStub.create();
 
             assert(stub.usingPromise);
-            assert.isTrue(typeof stub.usingPromise === "function");
+            assert.isFunction(stub.usingPromise);
         });
 
         it("should return the current stub", function () {
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -317,6 +317,59 @@ describe("stub", function () {
         });
     });
 
+    describe(".usingPromise", function () {
+        it("should exist and be a function", function () {
+            var stub = createStub.create();
+
+            assert(stub.usingPromise);
+            assert.isTrue(typeof stub.usingPromise === "function");
+        });
+
+        it("should return the current stub", function () {
+            var stub = createStub.create();
+
+            assert.same(stub.usingPromise(Promise), stub);
+        });
+
+        it("should set the promise used by resolve", function () {
+            var stub = createStub.create();
+            var promise = {
+                resolve: createStub.create().callsFake(function (value) {
+                    return Promise.resolve(value);
+                })
+            };
+            var object = {};
+
+            stub.usingPromise(promise).resolves(object);
+
+            return stub().then(function (actual) {
+                assert.same(actual, object, "Same object resolved");
+                assert.isTrue(promise.resolve.calledOnce, "Custom promise resolve called once");
+                assert.isTrue(promise.resolve.calledWith(object), "Custom promise resolve called once with expected");
+            });
+        });
+
+        it("should set the promise used by reject", function () {
+            var stub = createStub.create();
+            var promise = {
+                reject: createStub.create().callsFake(function (err) {
+                    return Promise.reject(err);
+                })
+            };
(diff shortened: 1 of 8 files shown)


mantoni:
Nice work. I'm not so sure about the name `setPromise`, because you're not setting a promise instance but rather the type of promise to use. Alternatives that I can think of are `usesPromise` or `usingPromise` or `returnsPromise`. I'm not using promises myself, so I'm not the best to make a decision here.

blacksun1:
I would prefer `usePromise` or `usingPromise`. `usesPromise` sounds like something in the past.

fatso83:
I'd go for `usingPromise` (or something like that - no hard feelings on this matter).

blacksun1:
@fatso83, @mantoni

blacksun1:
If I have agreement then I will continue with the rest of the implementation this week when I have some "free time" - meaning I steal some time from somewhere else. ;)

mantoni:
@blacksun1 Sure, `usingPromise` sounds good. We all steal our time elsewhere 🍻

blacksun1:
Any idea where I should put the documentation for the feature?

mantoni:
Please put documentation into the `docs` folder (https://github.com/sinonjs/sinon/tree/master/docs). The README and the CONTRIBUTING files in there explain how to work with it.

blacksun1:
Docs added. @mantoni @fatso83

test/stub-test.js
```diff
@@ -317,6 +317,59 @@ describe("stub", function () {
         });
     });
 
+    describe(".usingPromise", function () {
+        it("should exist and be a function", function () {
+            var stub = createStub.create();
+
+            assert(stub.usingPromise);
+            assert.isTrue(typeof stub.usingPromise === "function");
```
fearphage: `assert.isFunction`

jrnail23:
I just wanted to say thanks for doing this, @blacksun1.  This should fill the gap left by obsolescing `sinon-as-promised` which left us without a way to plug `bluebird` in, and thus preventing us from upgrading to sinon v2.x 😄

test/stub-test.js
blacksun1: Done

fatso83:
Never mind my issue comment - found out I could review and merge this using even an old-skool cell phone :-) Thanks!

mroderick:
`sinon@2.2.0` is now on `npm`

PR #1365, opened 2017-03-25
Fix Reference to deprecated API

MouseZero:
Fix Reference to deprecated API

I tried to follow the docs and got the following warning message.
"sinon.stub(obj, 'meth', fn) is deprecated and will be removed from the public API in a future version of sinon. Use stub(obj, 'meth').callsFake(fn)."

1. Its pure textual you can just look at it.


fatso83:
Been wanting to fix this, thanks!

MouseZero:
No Problem.

mroderick:
Thanks!

PR #1371, opened 2017-03-29
Don't reset behavior when calling resetHistory()

### LINKED ISSUE #1361: `resetHistory` resets stubs behavior configured after `withArgs`
> We understand you have a problem and are in a hurry, but please provide us with some info to make it much more likely for your issue to be understood, worked on and resolved quickly.

* Sinon version : 2.1.0
* Environment : node 7.7.2
* Example URL : -
* Other libraries you are using: mocha 3.2.0, sinon-chai 2.9.0

Calling `resetHistory` should only reset the counters etc... not behaviors of a stub set after `withArgs`.

**What actually happens**

Calling `resetHistory` resets the return values set after `withArgs`

**How to reproduce**

```
// Setting behaviors
const test_stub = sinon.stub().returns(0);
test_stub.withArgs('test').returns(10);

// Testing everything's alright
test_stub(); // returns 0
test_stub('test'); // returns 10

// Breaking
test_stub.resetHistory();
test_stub(); // returns 0
test_stub('test'); // returns 0, shouldn't it be 10?
```

jroitgrund:
Make `stub.resetHistory()` keep the behavior defined using `stub.withArgs(...).returns`.

Fixes #1361

ATM `stub.resetHistory()` delegates to `spy.reset()` but `spy.reset()` resets all
its fakes by calling `reset()` on them. If the fakes were created from a
stub, we want to call `resetHistory()` on them instead to preserve their behavior.

Added a failing test so:

2. `npm install`
3. `npm test`

### DIFF
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -1991,6 +1991,15 @@ describe("stub", function () {
             assert(stub.calledOnce);
             assert.equals(stub.getCall(0).args[0], 2);
         });
+
+        it("doesn't reset behavior defined using withArgs", function () {
+            var stub = createStub();
+            stub.withArgs("test").returns(10);
+
+            stub.resetHistory();
+
+            assert.equals(stub("test"), 10);
+        });
     });
 
     describe(".resetBehavior", function () {
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -2000,6 +2000,15 @@ describe("stub", function () {
 
             assert.equals(stub("test"), 10);
         });
+
+        it("doesn't reset behavior", function () {
+            var stub = createStub();
+            stub.returns(10);
+
+            stub.resetHistory();
+
+            assert.equals(stub("test"), 10);
+        });
     });
 
     describe(".resetBehavior", function () {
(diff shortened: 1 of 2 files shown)


lucasfcosta:
LGTM, but maybe we could consider calling the `spy.reset` method `spy.resetHistory` in order to make it consistent with what it does and allow us to avoid this kind of check.

Flarna:
I thought about the same as it would maybe solve also #1372. Unfortunately I found no time yet to try this by myself...

test/stub-test.js
```diff
@@ -1991,6 +1991,15 @@ describe("stub", function () {
             assert(stub.calledOnce);
             assert.equals(stub.getCall(0).args[0], 2);
         });
+
+        it("doesn't reset behavior defined using withArgs", function () {
```
fearphage: Please add a test for default behaviors (without `withArgs`).

jroitgrund:
@fearphage done. anything else blocking this?

fearphage:
@jroitgrund It looks clear to merge on my end. 

@sinonjs/sinon-core Would you like to do the honors?

fatso83:
@fearphage You do have merge-rights, or have I screwed that up somehow? Because I think we totally trust you to not merge things willy-nilly :-)

fearphage:
> You do have merge-rights, or have I screwed that up somehow?

I do, but I like to get a few sets of eyes on things. :smile:

PR #1319, opened 2017-03-08
add throwsArg(index) to stubs

seppevs:
Fixes issue https://github.com/sinonjs/sinon/issues/1270 by adding a `throwsArg(index)` method to stubs

I have added unit tests & docs.

PR #1374, opened 2017-04-06
Exclude test and node_modules directories from coverage

fearphage:
istanbul is currently tracking coverage on the `test` folder. This is causes issues for PR's I think (see #1319). This addresses that by excluding the test and node_modules folders from coverage. 

Bonus: It should speed up tests slightly as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "ci-test": "npm run lint && ./scripts/ci-test.sh",
     "test-node": "mocha --recursive -R dot test/",
     "test-headless": "mochify --recursive -R dot --grep WebWorker --invert test/",
-    "test-coverage": "mochify --recursive -R dot --grep WebWorker --invert --plugin [ mochify-istanbul --report text --report lcovonly --dir ./coverage ] test/",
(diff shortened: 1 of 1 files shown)


mantoni:
Maybe we should only exclude `node_modules`. Checking coverage on tests may be beneficial because we have function calls and dynamically generated tests in there. If we get a drop in coverage from the tests, it's something worth looking into. What do you think @fearphage ?

fearphage:
> Checking coverage on tests may be beneficial because we have function calls and dynamically generated tests in there

It seems like there should be some other resource/method to track that honestly. For instance, there are tests that intentionally only run in certain environments and under certain conditions. It also makes it more difficult to track the thing I (we?) actually care about most which is "is the code being tested?".

mantoni:
Yea, I guess you're right. Since this affects the metrics on each pull request, I'd like to ask if any of the other @sinonjs/owners have any objections?

fatso83:
no objections.

mantoni:
Sorry, didn't pay attention. Thanks @fearphage 👍

PR #1379, opened 2017-04-11
Fix highlighting in matchers.md of v1.17.6 and v1.17.7.

adityavohra7:
#### Purpose (TL;DR)
The `matchers.md` files of v1.17.6 and v1.17.7 are missing some `s.

#### Solution
Add the missing `s so the files have the right highlighting.

#### How to verify
View `matchers.md`s at `f2a3f267`.


fatso83:
thanks!

PR #1380, opened 2017-04-14
🚀 Hello TypeScript

jonnyreeves:
Adds support for *gradually* migrating the sinon codebase over to TypeScript including the migration of a single utility and its test (`times-in-words`)

Well seeing as it only took us the best part of 18 months to migrate to commonJS; I thought it would be worth a shot a migrating to TypeScript ;)  In all seriousness, I've found TypeScript to have provided a massive benefit to the projects that I've been working on making code written by other developers easier to reason about - in addition, using TypeScript allows us to make use of ES6+ features whilst still targeting ES5 browsers (heck, we can even target ES3!).  The other benefit is that once all source files are TypeScript (`.ts`, as opposed to `.js`), we can autogenerate the TypeScript definition files (`.d.ts`) which provides typehints to users of sinon - something will with be greatly appreciated by some in the community.

Obviously migrating the codebase to TypeScript is a pretty massive shift (and, ultimately to some degree, boils down to a personal preference) - hence why I've done an absolute MVP here in this case: however I'm keen to explore what the 'guts' of sinon will look like typed.

Finally, for those new to the TypeScript eco-system the ES5 code generated by the TypeScript compiler is very clean (and can easily be mistaken for hand-written ES5 once formatted) - so a migration back to plan javascript in the future would be reasonably trivial (ie: minimal sunk cost).

#### Solution  - optional
- Add typescript and tslint; use a sensible `tsconfig.json` which allows interop with the existing javascript source
- Move all sinon source code in `src/` and transpile it into the `lib` folder at build time.
- Run `tsc` before each test suite - this is incredibly inefficient, however compilation speed is very fast at this stage and I didn't want to mess around too much for the MVP: this can easily be optimised later if we wish to invest in this approach.

2. `npm install`
3. npm test


fearphage:
Can you generate the typescript bindings from javascript? I personally enjoy writing javascript myself. I'm not sure about everyone else. If we were planning to write it in not-js, I would suggest something with more benefits than typescript honestly (elm, clojurescript, etc).

mantoni:
Adding TypeScript would require a compilation step to run Sinon in Node. I'm for retaining source compatibility with our target environments, at least for the language constructs. We're using browserify only for bundling, not for transpiling. We already had the same discussion regarding more modern language features.

jonnyreeves:
@mantoni the way I've configured the build results in es5 being emitted into the `lib` folder which is published to npm (so no compilation step is required by consumers of the package)

mantoni:
Ok, let me be more precise: There is a compilation step between the code as it is written and the code as it is executed. You want to add an abstraction layer to the code.

cjohansen:
Like @mantoni said, we already had this discussion. Porting the code base to TypeScript would alienate a bunch of users (myself included) and is not an option.

Compile-to-js languages are fine, but in my opinion a poor choice for a library offered to JavaScript programmers.

As a last note, if you're into using another language I'm pretty sure you could write a better library than Sinon that would leverage the language you're using and it's unique valuable features.

fearphage:
Here is an external-only approach to TypeScript:

https://github.com/sinonjs/sinon-test/issues/56#issuecomment-292738787

PR #1388, opened 2017-05-02
Remove warning about v1 as source with bundlers

mroderick:
This is obsolete by now, and just adds noise and might scare people off because of an issue that has been addressed.


mantoni:
It's also not 100% correct. I've been using Sinon 1.x with Browserify successfully at multiple large scale projects without issues. Let's remove it.

PR #1393, opened 2017-05-04
Fix markdown in `Matchers` docs

genetique-techno:
#### Purpose (TL;DR) - mandatory
Fix some bad markdown which was annoying me slightly

#### Background (Problem in detail)  - optional
Missing a few backticks, and some others were double-quotes instead

#### Solution  - optional
Add the missing backticks and fix the errant double-quotes!

#### How to verify - mandatory
Give these docs an ocular patdown.

PR #1395, opened 2017-05-04
Fix markdown in matchers docs (for reals this time)

genetique-techno:
#1393 was changing the wrong file.  This is now affecting `/release-source/release/matchers.md`


mantoni:
🎉 Thank you!

Issue #1398, opened 2017-05-07
callOrder missbehaves


mir3z:
* Sinon version : 2.1.0
* Environment   : node
* Example URL   :
* Other libraries you are using:

**What did you expect to happen?**
I expect test to fail

**What actually happens**
Test passes

**How to reproduce**

```
        const s1 = spy();
        const s2 = spy();

        s1();
        s2();
        s1();

        assert.callOrder(s2, s1, s2);
```

lucasfcosta:
Hi @mir3z, thanks for your issue!

You seem to be right, but the real problem with this issue is semantics.

I've reimplemented the `called-in-order` core function to take into account both if a spy has been called **after the previous one and before the next one**. Now your example will not pass because `s2` has not been called after `s1`.

Before the changes I've got here, [`called-in-order` only checked if a passed spy was called before the next one](https://github.com/sinonjs/sinon/blob/ad781323d610cf4c515db427f7f111a4886c56c0/lib/sinon/util/core/called-in-order.js#L9).

However, the `calledAfter` function I am using in the `called-in-order` core function has a problem.

[The `calledBefore` checks if the first call of the current spy (`this`) happened before the last call of the passed spy](https://github.com/sinonjs/sinon/blob/ad781323d610cf4c515db427f7f111a4886c56c0/lib/sinon/spy.js#L259). This means that if the passed spy has been called after the first time we invoked the current spy, the current spy will be considered to have been called before (this function will return true).

However, the `calledAfter` function has different semantics. [Instead of checking whether or not the current spy (`this`) has been called after any calls to the passed spy, it checks if the current spy has been called more recently than the passed spy](https://github.com/sinonjs/sinon/blob/ad781323d610cf4c515db427f7f111a4886c56c0/lib/sinon/spy.js#L267) (it means the current spy should have been the last one to be called instead of just being called after any calls to the passed spy).

Given this I think we must refactor the `calledAfter` function to check if the current spy has been called anytime after the first passed spy invocation.

If we want to keep the old behavior for `calledAfter` I suggest we rename it to `calledLast` or something like that.

Thus, [this test will start breaking](https://github.com/sinonjs/sinon/blob/ad781323d610cf4c515db427f7f111a4886c56c0/test/spy-test.js#L1644) because if we take into account the exact meaning of "spyA being called after spyB" it is true indeed, because even though `spyA` was not the last one to be called, it was still called after `spyB`. So we either gotta remove it or start using `calledLast` for this test.

[...]

PR #1406, opened 2017-05-17
Called in order considers callCount

lucasfcosta:
## Purpose (TL;DR)

This makes `calledAfter` symetric with `calledBefore`, as I explained in #1398.

Semantically speaking, `calledAfter` does not do what it says it does.

When you say you expect `something` to have been `calledAfter` `anotherThing`, you don't care if `something` has been called last or not, all you care about is if `something` has been called any times **after** `anotherThing`.

This is the way `calledBefore` works. It is not concerned about whether or not the current spy was the first to be called, it just checks if the current spy has been called any times before the passed spy.

## Solution

I just check if the last call to the current spy (`this`) has happened after the first call to the passed spy (`spyFn`) by comparing their `callIds`

[I also fixed a test which had the incorrect spec for this](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:simmetric-callAfter#diff-8b6a8ef53c91611b3781f26049e5d972R1644).

## How to verify

2. `npm install`
3. `npm test` -> This will run all tests including [the modified one with the correct spec](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:simmetric-callAfter#diff-8b6a8ef53c91611b3781f26049e5d972R1644)

**EDIT: Building is breaking on PhantomJS, I gotta see why. Let me check this...**

### DIFF
--- a/lib/sinon/util/core/called-in-order.js
+++ b/lib/sinon/util/core/called-in-order.js
@@ -1,15 +1,34 @@
 "use strict";
 
+var every = Array.prototype.every;
+
 module.exports = function calledInOrder(spies) {
+    var callMap = {};
+
+    function hasCallsLeft(spy) {
+        if (callMap[spy.id] === undefined) {
+            callMap[spy.id] = 0;
+        }
+
+        return callMap[spy.id] < spy.callCount;
+    }
+
     if (arguments.length > 1) {
         spies = arguments;
     }
 
-    for (var i = 1, l = spies.length; i < l; i++) {
-        if (!spies[i - 1].calledBefore(spies[i]) || !spies[i].called) {
-            return false;
-        }
-    }
+    return every.call(spies, function checkAdjacentCalls(spy, i) {
+        var calledBeforeNext = true;
 
-    return true;
+        if (i !== spies.length - 1) {
+            calledBeforeNext = spy.calledBefore(spies[i + 1]);
+        }
+
+        if (hasCallsLeft(spy) && calledBeforeNext) {
+            callMap[spy.id] += 1;
+            return true;
+        }
+
+        return false;
+    });
 };
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -200,4 +200,19 @@ describe("issues", function () {
             assert(firstFake !== secondFake);
         });
     });
+
+    describe("#1398", function () {
+        it("Call order takes into account both calledBefore and callCount", function () {
+            var s1 = sinon.spy();
+            var s2 = sinon.spy();
+
+            s1();
+            s2();
+            s1();
+
+            assert.exception(function () {
+                sinon.assert.callOrder(s2, s1, s2);
+            });
+        });
+    });
 });


lucasfcosta:
Hmm, I spent an hour playing with mochify and phantomjs related configs but I couldn't get it to work.

I even tried to run a single test and change the `callInOrder` function to always return true in order to check for any better output but I couldn't get it to work in any way.

I'll dig deeper into this as soon as I've got more time to, but if any of you can remember any detail that might help me fix this, please let me know.

fatso83:
Hmm... are you able to make it pass locally? I always found mochify (really it's phantom) to be a bit of a black box with regards to diagnostics.

lucasfcosta:
Nope, I should've tested that before opening that PR, it was my fault to submit a breaking PR, but since I cannot track down the reason of this I'm now thinking it was a good idea do open this PR in order to get feedback.

A funny thing is that if I enable the `--consolify` flag, tests pass.

mantoni:
If you run `--consolify` it generates a web page to run in a browser. It always exits with code 0. Trying to run the tests in a real browser might shed some light on this though. Also, you can try to run in real browsers with `--wd`.

lucasfcosta:
@mantoni Just tried to play around with mochify in the last hour and couldn't get it to run tests on any real browsers :(

I always end up getting this:

```
> mochify --recursive -R dot --grep WebWorker --invert test/ --wd                                                                                         │  6             callMap[spy.id] = 0;
                                                                                                                                                          │  5         }
POST /wd/hub/session                                                                                                                                      │  4
                                                                                                                                                          │  3         return callMap[spy.id] < spy.callCount;
Unexcected request error: connect ECONNREFUSED 127.0.0.1:4444                                                                                             │  2     }
    Error: connect ECONNREFUSED 127.0.0.1:4444                                                                                                            │  1
        at Object.exports._errnoException (util.js:1022:11)                                                                                               │14      if (arguments.length > 1) {
        at exports._exceptionWithHostPort (util.js:1045:20)                                                                                               │  1         spies = Array.prototype.slice.call(arguments);
        at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)                                                                                   │  2     }
                                                                                                                                                          │  3
Error: Unexcected request error: connect ECONNREFUSED 127.0.0.1:4444
```

Any ideas on how to solve this and investigate further?

mantoni:
Ok, so I checked out your branch and ran the tests locally. I found your issue: It's the arrow function in the test case. You have to use `function () { ...` or phantomjs (and Safari and IE 11 on SauseLabs) will fail with really odd messages. It needs investigation on how we can make PhantomJS / Phantomic / Mochify spit out a more helpful message in cases where the script cannot be parsed.

If you want your local setup working, check the following:

- The `--consolify` option works like this: `mochify --recursive -R dot --grep WebWorker --invert test/ --consolify test.html`, then open `test.html`.
- The `--wd` option connects to a Selenium webdriver server. In the Sinon.JS project, it's configured to connect to SauceLabs, but for you it seems to try a local connection. There are [instructions in the Mochify readme](https://github.com/mantoni/mochify.js/#selenium-webdriver-setup) on how to setup a local webdriver instance.

mantoni:
FYI: I'm working on a fix for Phantomic to catch parser errors. It's quite a hack, but I should have something soon.

mantoni:
Finally! This was a bug I was trying to hunt down for quite some time. Your issue nudged me in the right direction 🙌 

If you update phantomic to `v1.5.2` (see https://github.com/mantoni/phantomic/commit/47e7cdd248c4644561fbf6e9dc874ba7e4148487), running the tests on your branch will yield

```
# phantomjs:
SyntaxError: Unexpected token ')'
      at test/issues/issues-test.js:205
Error: Exit 1
```

🐛 🎉

fatso83:
@mantoni that is great news! debugging/finding parsing errors was the main pain point of using mochify/phantomic so far. in truth a big improvement!

mantoni:
I restarted the Travis job, but it still shows the old behavior. I think the dependencies are cached. Does anybody know how to clear `node_modules` for Travis builds?

test/issues/issues-test.js
```diff
@@ -200,4 +200,19 @@ describe("issues", function () {
...
+        it("Call order takes into account both calledBefore and calledAfter", () => {
+            var s1 = sinon.spy();
+            var s2 = sinon.spy();
+
+            s1();
+            s2();
+            s1();
+
+            assert.exception(function () {
+                sinon.assert.callOrder(s2, s1, s2);
```
fearphage: This looks broken. That is not the order they were called in.

lucasfcosta: Yup, In this test I'm expecting an error because that is not the order those spies are called in.

Basically I'm asserting this feature fails when it should.

lucasfcosta:
@fearphage thanks for your review 😄 

Since this feature has already been released I'm not sure changing its name would be a good idea (this is my personal opinion). Because it adds effort for people that area already using it and I also think that `calledInOrder` is a good name.

It denotes exactly what we want to assert on and also does not include the word `immediately`, which is also great because then people know there might be interleaving calls between the passed args.

What does everyone think? Let's gather some more opinions.

test/issues/issues-test.js
fearphage: My mistake for misreading.

lucasfcosta: No problem, buddy
I think it's ready for a merge now.

lib/sinon/util/core/called-in-order.js
```diff
@@ -1,15 +1,34 @@
...
+        if (callMap[spy.id] === undefined) {
+            callMap[spy.id] = 0;
+        }
+
+        return callMap[spy.id] < spy.callCount;
+    }
+
     if (arguments.length > 1) {
-        spies = arguments;
+        spies = Array.prototype.slice.call(arguments);
```
fearphage: MINOR: The only reason to use `slice` here would be if you plan to used `spies.every(...)` later. However you are using `every.call(...)` so you can revert this line.

fearphage:
@lucasfcosta One small thing, but I think it's good to go after that.

lib/sinon/util/core/called-in-order.js
lucasfcosta: @fearphage well noticed! I just fixed it.

Thanks for your careful review, this was indeed something that should be fixed 👍

PR #1407, opened 2017-05-17
Make calledAfter symetric with calledBefore

lucasfcosta:
## Purpose (TL;DR)

This makes `calledAfter` symetric with `calledBefore`, as I explained in #1398.

Semantically speaking, `calledAfter` does not do what it says it does.

When you say you expect `something` to have been `calledAfter` `anotherThing`, you don't care if `something` has been called last or not, all you care about is if `something` has been called any times **after** `anotherThing`.

This is the way `calledBefore` works. It is not concerned about whether or not the current spy was the first to be called, it just checks if the current spy has been called any times before the passed spy.

## Solution

I just check if the last call to the current spy (`this`) has happened after the first call to the passed spy (`spyFn`) by comparing their `callIds`

[I also fixed a test which had the incorrect spec for this](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:simmetric-callAfter#diff-8b6a8ef53c91611b3781f26049e5d972R1644).

## How to verify

2. `npm install`
3. `npm test` -> This will run all tests including [the modified one with the correct spec](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:simmetric-callAfter#diff-8b6a8ef53c91611b3781f26049e5d972R1644)

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -264,7 +264,7 @@ var spyApi = {
             return false;
         }
 
-        return this.callIds[this.callCount - 1] > spyFn.callIds[spyFn.callCount - 1];
+        return this.callIds[this.callCount - 1] > spyFn.callIds[0];
     },
 
     calledImmediatelyBefore: function calledImmediatelyBefore(spyFn) {
--- a/test/spy-test.js
+++ b/test/spy-test.js
@@ -1641,12 +1641,12 @@ describe("spy", function () {
             assert.isFalse(this.spyA.calledAfter(this.spyB));
         });
 
-        it("returns false if other called last", function () {
+        it("returns true if called anytime after other", function () {
             this.spyB();
             this.spyA();
             this.spyB();
 
-            assert.isFalse(this.spyA.calledAfter(this.spyB));
+            assert.isTrue(this.spyA.calledAfter(this.spyB));
         });
     });
 


mroderick:
This looks good! I've applied the `PATCH` label, as I feel this is fixing something that is broken, not changing a design decision. Does anyone feel differently about this?

mantoni:
Patch should be fine. I think this is not going to make any passing tests fail.

fatso83:
I think I put MAJOR on the accompanying issue as this could break users tests, which AFAIK is the definition of a breaking change (?). But this is almost a semantic discussion. If you are fine with patch, then I'm fine with patch. This is a _very_ small part of Sinon anyhow, and not one that is used _that_ frequently :-)

lucasfcosta:
@fatso83 I'd use `PATCH` too.

Even though this might break user's tests, it will break them because they were relying on incorrect behavior and since we're just correcting an error it's a fix.

Imagine if this was a more explicit error, such as:

```js
sum(2 + 2) // returns 5

// Test:
it('should sum two and two', () => {
  assert.equals(sum(2, 2), 5)
})
```

In that case, fixing `sum` would break user's test, but that's because they were relying on incorrect behavior.

mroderick:
This has become `sinon@2.3.1` on npm

PR #1415, opened 2017-05-21
Style improvements

mroderick:
This PR adds two style improvements to the site:

* increase vertical space above `<h4>` elements, which makes the documentation pages easier to read
* reduce the font size for fenced code blocks slightly, this causes the scrollbar to disappear with current code samples (on my machine, with my settings)

Visitors of sinonjs.org are most likely developers using a desktop class browser with a reasonably sized desktop display. We should optimise for reading on those displays, and not accept horisontal scrollbars when we don't need to.

#### Before

2017-05-21 at 22 59

2017-05-21 at 13 06

#### After

2017-05-21 at 22 58

2017-05-21 at 13 07

### DIFF
--- a/docs/_sass/_base.scss
+++ b/docs/_sass/_base.scss
@@ -64,6 +64,13 @@ a {
     margin-bottom: 20px;
     color: #281D18;
   }
+
+  h4 {
+    font-size: 18px;
+    margin-top: 32px;
+    color: #281D18;
+  }
+
   p {
     color: #4B382F;
   }
--- a/docs/_sass/_base.scss
+++ b/docs/_sass/_base.scss
@@ -228,3 +228,7 @@ a {
         max-width: 737px;
     }
 }
+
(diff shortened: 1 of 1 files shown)


mantoni:
While this makes the code examples wider, it decreases readability because we end up with potentially very long text lines. I thought the idea was to slightly decrease the font size in the examples to get rid of the scroll bar?

mantoni:
The `h4` styling is a good idea. I like that.

mroderick:
> I thought the idea was to slightly decrease the font size in the examples to get rid of the scroll bar?

I wasn't aware of that idea. I've implemented it and pushed new commits, and have updated the PR description with new screenshots

mantoni:
Cool, thanks 👌 We had this idea in the original issue for the new website, I think. And then I forgot about it.

mroderick:
Thanks

Issue #781, opened 2015-07-14
Stub of getters & setters not working with sandbox


vitalets:
When I stub getter using `sinon.stub()` everything works fine.
But when I use `sandbox.stub()` it does not work. 
Example:

``` js
var o = {
    get foo() { return 'foo' },
    get bar() { return 'bar' }
};

// via sinon
sinon.stub(o, 'foo', { get: function () { return 'foo stubbed' }});
console.log(o.foo); // foo stubbed

// via sandbox
var sandbox = sinon.sandbox.create();
sandbox.stub(o, 'bar', { get: function () { return 'bar stubbed' } });
console.log(o.bar); // Uncaught TypeError: Cannot set property bar of #<Object> which has only a getter
```

jsfiddle: http://jsfiddle.net/exzc5m3e/

Relates to: https://github.com/cjohansen/Sinon.JS/pull/692

Is it possible to support getters & setters in sandboxes?

Thanks!

mroderick:
Thank you for reporting this.

> Is it possible to support getters & setters in sandboxes?

That's a good question. We'll need more investigation to determine if that's possible.

mikedaly:
I recently added sinon to a project, to augment Jasmine, specifically because it supported mocking properties. I would use this feature as soon as it was available.

nunofgs:
I just hit this issue as well.

kuraga:
The same issue

amrish7:
Just ran into this issue

fatso83:
As the label says: help is much appreciated. If someone deeply craves this feature do know that we are open for pull requests ;-)

mmoss:
@fatso83 I'll give adding some units tests and fixing this a go... fingers crossed I can get it working :smile:

mantoni:
@mmoss Make sure to pull again. [The unit tests have been converted to Mocha](https://github.com/sinonjs/sinon/pull/1004).

mroderick:
This is implemented with #1416

PR #1416, opened 2017-05-21
Refactor sandbox and allow it to stub getters and setters

### LINKED ISSUE #1401: Stub get does not work with sandbox stub
`stub().get()` used to stub property getters, works well with `sinon.stub`, but I get _get is not a function_ with `sandbox.stub`.

Also, slightly related, `.get` does not exist on the typescript definition _@types/sinon_

lucasfcosta:
## Purpose (TL;DR)

This aims to fix #1401 and #781 by allowing users to stub `getters` and `setters` while using sandboxes.

In order to do this I had to refactor `sandbox.stub` and add new `value` default behavior for the sake of consistency.

*Another side-effect of this PR is adding the `value` default behavior to stubs, allowing users to stub the `value` property on property descriptors.*

As discussed in the two issues linked above, the sandbox API did not allow our users to set `getters` and `setters` for non-function properties they stubbed, because [that meant we were going to call `stubNonFunctionProperty`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-c66fb904faa27068eb89ea905c3e4d68L89) and [return a raw object with a restore method](https://github.com/sinonjs/sinon/blob/bf2bfbd929b3a21c81d07cf831e8f06b5e2f2a44/lib/sinon/stub-non-function-property.js#L15). Due to this fact, the returned object would not allow us to call the `get` and `set` default behaviors.

Another problem I explained at #781 was that by having `sandbox.stub` to accept a third argument, it was not possible to differentiate this argument from a property descriptor when we wanted to stub non-function-properties, because we would never know whether the user wanted that value to replace that property or if the user wanted that value to be the new property descriptor.

GIven the way `sandbox.stub` was working, it presented a bit of duplication and error prone code, because some of the same checks we have in `sinon.stub` were being made differently in `sandbox.stub`. This could cause unwanted `getter` triggers when checking property values and therefore inconsistent behavior compared to `sinon.stub`.

## Solution

This is how I solved these problems:

1. Allowing **getters** and **setters** (consistent sandbox API)
    I had to create [a separate utility file called `sandbox-stub.js`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-2362249c6d5cd8ba07d5ee7c94402a56) which does the job of handling calls to the old version of the `sandbox.stub` API. [The function in this file is called whenever a user passes a third argument to `sandbox.stub`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-c66fb904faa27068eb89ea905c3e4d68L84). [Otherwise, we **always** call the `sinon.stub` function](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-c66fb904faa27068eb89ea905c3e4d68R87) in order to make the behavior of `sandbox.stub` as consistent as we can.
    This allows us to have all the default stub behaviors available on the stub created by `sandbox.stub`.
    Then all we have to do is [add all the stubbed methods (which are more than one if we're stubbing an entire object) to the current collection](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-c66fb904faa27068eb89ea905c3e4d68R90).
2. Allow stubbing non-function values while still allowing `get` and `set` behaviors
    Since there is no way to differentiate a descriptor from a desired property value to replace another when we allow three arguments on the `sandbox.stub` API, I added [a new default behavior called `value`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-f46d81a3e31504be425e201260dab281R193). This makes the API consistent with stubbing other values on property descriptors, such as `get` and `set`.
    In order to restore a `stubbed` non-function property a user can just call the [`restore` method on the created fake](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-f46d81a3e31504be425e201260dab281R202).

This PR also has:
* [Tests for the new `value` behavior](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-888b5036773bbf553212fad0a7977303R2528)
* [Tests for `sandbox.stub` being able to stub `getters` and `setters`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-536d6d65592b132d984d4f646cb0448eR552)
* [Updated docs](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-fc7a9e11c0ba5324f14837e385a19d6b)
* [A minor improvement in which I add a name to a previously anonymous function on `stubNonFunctionProperty`](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-0df051cb1e1d912470223f7f38234105L16)
* Fixes to avoid tests for the old `sandbox.stub` API to log deprecation warnings to the console. Like [this](https://github.com/sinonjs/sinon/compare/master...lucasfcosta:refactor-sandbox-stub?expand=1#diff-cce87ff1f83bd217096d5f98bb2ae6f7R130), for example.

2. `npm install`
3. `npm test` - This runs both the old tests (proving nothing breaks) and the new ones (proving the new features work properly)

**As always, any feedbacks will be great, specially given it's a long PR.**

Thanks for your time!

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -106,7 +106,7 @@ Works exactly like `sinon.spy`, only also adds the returned spy to the internal 
 
 Works almost exactly like `sinon.stub`, only also adds the returned stub to the internal collection of fakes for easy restoring through `sandbox.restore()`.
 
-The sandbox `stub` method can also be used to stub any kind of property. This is useful if you need to override an object's property for the duration of a test, and have it restored when the test completes
+The sandbox `stub` method can also be used to stub any kind of property. This is useful if you need to override an object's property for the duration of a test, and have it restored when the test completes.
 
 #### `sandbox.mock();`
 
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -520,3 +520,31 @@ myObj.prop = 'baz';
 
 myObj.example; // 'baz'
 ```
+
+#### `stub.value(newVal)`
+
+Defines a new value for this stub.
+
+```javascript
+var myObj = {
+    example: 'oldValue',
+};
+
+sinon.stub(myObj, 'example').value('newValue');
+
+myObj.example; // 'newValue'
+```
+
+You can restore values by calling the `restore` method:
+
+```javascript
+var myObj = {
+    example: 'oldValue',
+};
+
+var stub = sinon.stub(myObj, 'example').value('newValue');
+stub.restore()
+
+myObj.example; // 'oldValue'
+```
+
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -3,9 +3,8 @@
 var sinonSpy = require("./spy");
 var sinonStub = require("./stub");
 var sinonMock = require("./mock");
-var throwOnFalsyObject = require("./throw-on-falsy-object");
+var sandboxStub = require("./sandbox-stub");
 var collectOwnMethods = require("./collect-own-methods");
-var stubNonFunctionProperty = require("./stub-non-function-property");
 
 var push = [].push;
 
@@ -81,13 +80,12 @@ var collection = {
     },
 
     stub: function stub(object, property/*, value*/) {
-        throwOnFalsyObject.apply(null, arguments);
+        if (arguments.length > 2) {
+            return sandboxStub.apply(this, arguments);
+        }
 
+        var stubbed = sinonStub.apply(null, arguments);
         var isStubbingEntireObject = typeof property === "undefined" && typeof object === "object";
-        var isStubbingNonFunctionProperty = property && typeof object[property] !== "function";
-        var stubbed = isStubbingNonFunctionProperty ?
-                        stubNonFunctionProperty.apply(null, arguments) :
-                        sinonStub.apply(null, arguments);
 
         if (isStubbingEntireObject) {
             var ownMethods = collectOwnMethods(stubbed);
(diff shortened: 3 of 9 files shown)


mroderick:
Excellent work @lucasfcosta!

I don't have any comments for the code.

From reading the PR, I would say that this should be a new `MINOR` version. Does anyone disagree with that?

lucasfcosta:
Thanks @mroderick! You and all the other maintainers have been doing an excellent work as well, it's great to collaborate with such good people.

I'd say this is a `MINOR` version too.

Since I didn't need to change any test cases in order to make it pass and the API contract is still the same (even though we've got a new deprecation warning) I'd say you're correct.

Btw, I always like to think about SemVer as [`breaking.feature.fix`](https://medium.com/javascript-scene/software-versions-are-broken-3d2dc0da0783) and this falls into the `feature` category.

mroderick:
I've published `sinon@2.3.0` which includes this

mantoni:
@lucasfcosta I'm floored by your work here. I usually don't work with getters a lot, but just now I had a case where this made my test trivial:

```js
const stream = new PassThrough();
sandbox.stub(process, 'stdin').get(() => stream); // This failed without your change
```

I wanted to say thank you ❤️

lucasfcosta:
@mantoni I'm happy I was able to help.

I love contributing to Sinon. All of you are always so kind, active and do great reviews. Actually I'm the one that should be thanking you for making the contribution experience so pleasurable and for taking care of this awesome codebase. ❤️ 

Whenever you need anything I'll be more than happy to help.

PR #1422, opened 2017-05-25
Make object "has" matcher recursive

mareq:
#### Purpose (TL;DR)
Enable matching properties of embedded objects using `sinon.match.has` matcher.

```
"test should assert fuzzy": function () {
    var book = {
        pages: 42,
        author: {
            name: "cj"
            surname: "no" 
        }
    };
    var spy = sinon.spy();
    spy(book);

    // currently, this will NOT match (entire object has to be passed as value)
    sinon.assert.calledWith(spy, sinon.match.has("author", { name: "cj" } ));

    // this pull request enable this to match
    sinon.assert.calledWith(spy, sinon.match.has("author.name", "cj" ));
}
```

#### Solution 
`createPropertyMatcher` has been modified such that:
* existence of `"dot.separated.embedded.property"` is checked using recursive helper function
* `deepEqual` is run only on the value of the requested property, not on the whole top-level object

2. `npm install`
3. `npm test` (test case for the new functionality has been added)

### DIFF
--- a/docs/release-source/release/matchers.md
+++ b/docs/release-source/release/matchers.md
@@ -12,6 +12,10 @@ corresponding `sinon.assert` functions as well as `spy.withArgs`. Matchers allow
     var book = {
         pages: 42,
         author: "cjno"
+        id: {
+          isbn10: "0596517742",
+          isbn13: "978-0596517748"
+        }
     };
     var spy = sinon.spy();
 
@@ -19,6 +23,7 @@ corresponding `sinon.assert` functions as well as `spy.withArgs`. Matchers allow
 
     sinon.assert.calledWith(spy, sinon.match({ author: "cjno" }));
     sinon.assert.calledWith(spy, sinon.match.has("pages", 42));
+    sinon.assert.calledWith(spy, sinon.match.has("id", sinon.match.has("isbn13", "978-0596517748")));
 }
 ```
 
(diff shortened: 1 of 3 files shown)


mantoni:
I can see how this simplifies the notation when testing details in large objects. However, there is already a way to do the same thing, and I'm not sure if @sinonjs/sinon-core wants to introduce an alternative.

```js
sinon.assert.calledWith(spy, sinon.match.has("author", sinon.match.has("name", "cj")));
```

Also, consider using `calledWithMatch`:

```js
sinon.assert.calledWithMatch(spy, {
  author: sinon.match.has("name", "cj")
});
```

mareq:
@mantoni: Thanks for hint, I did not know about the functionality already existed (actually tried that composition way first, but because of my mistake it did not work - I like the composition way more in fact).

One of the reasons (apart of me trying to match stuff on serialised string instead of real object) why I did not know the feature is already present is that I did not find unit test nor documentation for it, so I have added both. Also reverted my original changes.

mantoni:
@mareq Nice, thank you! You've added the documentation specifically for the v17 release. Can you add the documentation changes also to `/docs/release-source/release` so that it's part of the current docs and in every future release?

mareq:
@mantoni Done: amended my last commit to move it from v17 to `/docs/release-source/release`

mantoni:
Thank you!

PR #1424, opened 2017-05-25
Fix #1372: make sandbox.resetHistory also reset spies

### LINKED ISSUE #1372: sandbox.resetHistory() doesn't reset history of spies
* Sinon version : 2.1.0
* Environment   : NodeJs 4.8.1
* Other libraries you are using: mocha

`sandbox.resetHistory()` should reset history of spies.

**What actually happens**
`sandbox.resetHistory()` is not reseting history of spies.

**How to reproduce**
```
it("reset spy", function(): void {
  const sb = Sinon.sandbox.create();
  const spy = sb.spy();
  spy();
  Assert.equal(spy.callCount, 1);
  spy();
  Assert.equal(spy.callCount, 2);
  sb.resetHistory();
  spy();
  Assert.equal(spy.callCount, 3);  // should fail but doesn't
  Assert.equal(spy.callCount, 1);  // should not fail but fails
});
```

mroderick:
This PR fixes #1372, by improving the implementation of `sandbox.resetHistory()` method

#### To verify
1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -46,7 +46,13 @@ var collection = {
     },
 
     resetHistory: function resetHistory() {
-        each(this, "resetHistory");
+        getFakes(this).forEach(function (fake) {
+            var method = fake.resetHistory || fake.reset;
+
+            if (method) {
+                method.call(fake);
+            }
+        });
     },
 
     verifyAndRestore: function verifyAndRestore() {
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -187,6 +187,23 @@ describe("issues", function () {
         });
     });
 
+    describe("#1372 - sandbox.resetHistory", function () {
+        it("should reset spies", function () {
+            var spy = this.sandbox.spy();
+
+            spy();
+            assert.equals(spy.callCount, 1);
+
+            spy();
+            assert.equals(spy.callCount, 2);
+
+            this.sandbox.resetHistory();
+
+            spy();
+            assert.equals(spy.callCount, 1);  // should not fail but fails
+        });
+    });
+
     describe("#1398", function () {
         it("Call order takes into account both calledBefore and callCount", function () {
             var s1 = sinon.spy();
--- a/test/collection-test.js
+++ b/test/collection-test.js
@@ -369,20 +369,28 @@ describe("collection", function () {
         beforeEach(function () {
             this.collection = Object.create(sinonCollection);
             this.collection.fakes = [{
+                // this fake has a resetHistory method
                 resetHistory: sinonSpy()
             }, {
+                // this fake has a resetHistory method
                 resetHistory: sinonSpy()
+            }, {
+                // this fake pretends to be a spy, which does not have resetHistory method
+                // but has a reset method
+                reset: sinonSpy()
             }];
         });
 
-        it("calls resetHistory on all fakes", function () {
+        it("resets the history on all fakes", function () {
             var fake0 = this.collection.fakes[0];
             var fake1 = this.collection.fakes[1];
+            var fake2 = this.collection.fakes[2];
 
             this.collection.resetHistory();
 
             assert(fake0.resetHistory.called);
             assert(fake1.resetHistory.called);
+            assert(fake2.reset.called);
         });
     });
 


lib/sinon/collection.js
```diff
@@ -46,7 +46,16 @@ var collection = {
...
 
     resetHistory: function resetHistory() {
-        each(this, "resetHistory");
+        getFakes(this).forEach(function (fake) {
+            if (typeof fake.resetHistory === "function") {
+                fake.resetHistory();
+                return;
+            }
+
+            if (typeof fake.reset === "function" && /^spy\#/.test(fake.id)) {
```
fearphage: What else has no `resetHistory` but does have a `reset` method besides a spy?

lib/sinon/collection.js
```diff
@@ -46,7 +46,16 @@ var collection = {
     },
 
     resetHistory: function resetHistory() {
-        each(this, "resetHistory");
+        getFakes(this).forEach(function (fake) {
+            if (typeof fake.resetHistory === "function") {
+                fake.resetHistory();
+                return;
```
fearphage: Use `else-if` instead.

test/issues/issues-test.js
```diff
@@ -180,4 +180,21 @@ describe("issues", function () {
...
+        it("should reset spies", function () {
+            var spy = this.sandbox.spy();
+
+            spy();
+            assert.equals(spy.callCount, 1);
+
+            spy();
+            assert.equals(spy.callCount, 2);
+
+            this.sandbox.resetHistory();
```
fearphage: I'm assuming there are already related tests for stubs. It wouldn't hurt to have a test with both since you're potentially affecting both.

lib/sinon/collection.js
mroderick: None that I know of. Do you think I should just test for a `reset` method, and call it?

fearphage: I believe so. I believe I've already seen code like that in another pr (if resetHistory else-if reset).

lib/sinon/collection.js
mroderick: No thanks, [there is no else if](http://rmurphey.com/blog/2012/12/10/js-conditionals)

lib/sinon/collection.js
mroderick: Fixed. I've pushed a new commit that just tests for the presence of `reset` before calling it

test/issues/issues-test.js
mroderick: I have expanded the test for `collection.resetHistory` to have a fake that looks like a spy in it

lib/sinon/collection.js
fearphage: I definitely agree with point number 2. I'm even working on a PR related to that. But point number one is a little misguided. 

1. Nothing is special about `else-if` in javascript. The logic is the same in every language that I can think of. You can always rewrite an else-if as a 3 statements instead of 2. It's a core tenant of Logic.
2. Splashing random returns around seems less explicit and it's obscuring the intent. Is the goal of this code to `return undefined`?
3. The most important difference to me is consistency.  We use else-if throughout the code. In my opinion, it reads better and is easier to reason about.

Another problem I foresee is future work. So let's say there was something I wanted to do no matter which fork of the `if` branch the code went down.. I'd have to "fix" your code first before I could accomplish that. In an `if-else` case, the change would be purely additive.

How would you feel adding a (pointless) `return` statement to the other if for consistency's sake?

mroderick: OK, I'll take the bait.

If follow your original suggestion, while writing JavaScript without omitting syntax, we'll end up with this:

```javascript
getFakes(this).forEach(function (fake) {
    if (typeof fake.resetHistory === "function") {
        fake.resetHistory();
    } else {
        if (typeof fake.reset === "function") {
            fake.reset();
        }
    }
});
```

In my opinion, that requires a lot more effort to understand.

Using [`return`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/return) as a control flow statement helps me keep indentation low, which I think we all agree makes code easier to read. 

Regarding consistency: I don't think we're that consistent in this codebase. We've been using `return` as a control flow statement across this codebase for awhile, try this search:

```shell
$ ag --ignore=docs/ -Q "return;"
```

I am all for consistency across a codebase ... until we find a better way of doing things, then we should slowly refactor to adopt the better practice. I believe that using `return` to reduce indentation is a good practice.

> How would you feel adding a (pointless) return statement to the other if for consistency's sake?

ESLint understands how to keep [consistent return](http://eslint.org/docs/rules/consistent-return), which [is configured in `eslint-config-sinon`](https://github.com/sinonjs/eslint-config-sinon/blob/master/index.js#L118).

I think it is known to any (non-novice) reader, that the default return value of a function `undefined`.

\</bike-shed\>

fearphage: > In my opinion, that requires a lot more effort to understand.

I would get the above code fixed in code review. It should be an `else-if`. I agree that the way it is written is more complex than an `else-if` although logically congruent.

> Using return as a control flow statement helps me keep indentation low, which I think we all agree makes code easier to read.

> I believe that using return to reduce indentation is a good practice.

We agree there, but an `else-if` is the same amount of indentation as a plain `if`: 

```js
if (foo) {
  // one indentation
  return;
}

if (bar) {
  // one indentation
}

// VS

if (foo) {
  // one indentation
} else if {
  // one indentation
}
```

You're not saving anything whatsoever in this case. You're actually typing a few extra characters if you're concerned about keystroke counts at all.

> try this search

We use empty returns all the time. That is not the/a problem. You're not getting the full story though.

```shell
ag --ignore=docs/ -Q "return;" -A 5 -B 3
```

I only found one instance of `if-return-if` usage. PS. I had been putting off installing/using `ag` until just now.

#### RE: consistency

Apparently every fake has `resetHistory` or `reset` so we should remove the second if and make it an `if-else` [like in spy](https://github.com/sinonjs/sinon/blob/36c6cdcec83dbeae7931c1949c1b1e79bf59e77b/lib/sinon/spy.js#L134-L142):

```js
        if (this.fakes) {
            this.fakes.forEach(function (fake) {
                if (fake.resetHistory) {
                    fake.resetHistory();
                } else {
                    fake.reset();
                }
            });
        }
```

Then this whole conversation goes away. :slightly_smiling_face:

mroderick: Or just fake `fake.reset()` directly

lib/sinon/collection.js
```diff
@@ -46,7 +46,13 @@ var collection = {
     },
 
     resetHistory: function resetHistory() {
-        each(this, "resetHistory");
+        getFakes(this).forEach(function (fake) {
+            var method = fake.resetHistory || fake.reset;
+
+            if (method) {
```
fearphage: Is this check necessary? What is it catching?

mroderick: Situations where neither method exists

mroderick: Which is a bit synthetic right now .... I could also just call `fake.reset`, since that is what stub.resetHistory does ... but that just seems *wrong*

mantoni: If you're going for the check, shouldn't it throw if there is no way to reset the thing this is called on?

mroderick: > If you're going for the check, shouldn't it throw if there is no way to reset the thing this is called on?

I didn't want to create something that was vastly different from what `each` does in the same file.

https://github.com/sinonjs/sinon/blob/master/lib/sinon/collection.js#L19-L28

mantoni: Ah yes, it filtered non-function values before iterating. 👍

fearphage: Are they fakes that have neither `resetHistory` nor `reset`? It seems like we could simplify this a bit and be in the same place:

```js
(fake.resetHistory || fake.reset).call(fake);
```

Flarna: If you stub a complete object the fake has no `resetHistory()` nor `reset()`:
```
function Test() {}
Test.prototype.foo = function () {}
Test.prototype.bar = function () {}

let sandbox = Sinon.sandbox.create();
sandbox.stub(Test.prototype);
sandbox.resetHistory();
```

Flarna:
I tested the change in my local environment and it solves the issue #1372 reported by me :thumbsup:

mroderick:
> What do you think about renaming/aliasing spy.reset to spy.resetHistory and deprecating reset? Not for this ticket, but just putting the idea out there.

I think that would be a good improvement. Will you add a ticket for it and describe how it would work? Perhaps we can even attract a new contributor with it ;)

mantoni:
I don't think there is anything else to add here. Like @mroderick suggested, we could think about refactoring the whole thing.

mroderick:
I have published this change to npm as `v2.3.4`

PR #1425, opened 2017-05-25
Allow some ES6 globals

mroderick:
This is in preparation of updating `eslint-config-sinon` to set ES5 syntax, which will remove the globals.

See https://github.com/sinonjs/eslint-config-sinon/pull/1

This PR should be merged first!

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,14 @@
     "rimraf": "^2.5.3"
   },
   "eslintConfig": {
-    "extends": "eslint-config-sinon"
+    "extends": "eslint-config-sinon",
+    "globals": {
+      "ArrayBuffer": false,
+      "Map": false,
+      "Promise": false,
+      "Set": false,
+      "Symbol": false
+    }
   },
   "files": [
     "lib",


lucasfcosta:
This LGTM!

Btw, I'd like to ask you about [the `lint` task in your `package.json` file](https://github.com/sinonjs/sinon/blob/master/package.json#L24).

I've had to change it to `eslint lib test` while developing because otherwise it would take other files (build files, mainly) into account and then the lint would be unuseful both due to the amount of output it produced and due to the time it takes to run.

Am I missing anything? Shouldn't we update that to `eslint lib test` to make sure it will only check necessary files?

mroderick:
> I've had to change it to eslint lib test while developing because otherwise it would take other files (build files, mainly) into account and then the lint would be unuseful both due to the amount of output it produced and due to the time it takes to run.
> 
> Am I missing anything? Shouldn't we update that to eslint lib test to make sure it will only check necessary files?

I [fixed that earlier tonight](https://github.com/sinonjs/sinon/commit/e933add8c9df41253e4be341fd6e6b5079a5c95d)

package.json
```diff
@@ -56,7 +56,14 @@
     "rimraf": "^2.5.3"
   },
   "eslintConfig": {
-    "extends": "eslint-config-sinon"
+    "extends": "eslint-config-sinon",
+    "globals": {
+      "ArrayBuffer": false,
+      "Map": false,
```
fearphage: What about WeakMap? Not blocking.

mroderick: I only added the globals that are currently being used in the source. We can always add more

PR #1426, opened 2017-05-25
Make better use of eslint-plugin-mocha

mroderick:
This PR enables more rules from [`eslint-plugin-mocha`](https://github.com/lo1tuma/eslint-plugin-mocha)

As you can see from the error report below, there are a number of fixes to be one.

If there's traction for adopting these rules, then I'll fix the outstanding errors before we merge this PR.

```text
/Users/mroderick/code/sinon/test/call-test.js
    26:9   error  Unexpected use of Mocha `beforeEach` hook outside of a test suite  mocha/no-top-level-hooks
    66:9   error  Unexpected use of Mocha `beforeEach` hook outside of a test suite  mocha/no-top-level-hooks
    66:9   error  Unexpected use of duplicate Mocha `beforeEach` hook                mocha/no-sibling-hooks
  1249:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1250:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1251:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1252:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1253:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1254:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1255:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1256:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1257:17  error  Unexpected test nested within another test                         mocha/no-nested-tests
  1258:17  error  Unexpected test nested within another test                         mocha/no-nested-tests

/Users/mroderick/code/sinon/test/collection-test.js
  105:17  error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  216:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  229:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  323:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  344:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  365:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case

/Users/mroderick/code/sinon/test/issues/issues-test.js
  57:12  error  Unexpected skipped mocha test  mocha/no-skipped-tests

/Users/mroderick/code/sinon/test/sandbox-test.js
  152:9  error  Test title is used multiple times in the same test suite  mocha/no-identical-title

/Users/mroderick/code/sinon/test/spy-test.js
   11:9  error  Unexpected use of Mocha `beforeEach` hook outside of a test suite  mocha/no-top-level-hooks
  100:9  error  Unexpected use of Mocha `beforeEach` hook outside of a test suite  mocha/no-top-level-hooks
  100:9  error  Unexpected use of duplicate Mocha `beforeEach` hook                mocha/no-sibling-hooks
  104:9  error  Test title is used multiple times in the same test suite           mocha/no-identical-title
  108:9  error  Test title is used multiple times in the same test suite           mocha/no-identical-title
  122:9  error  Test title is used multiple times in the same test suite           mocha/no-identical-title
  130:9  error  Test title is used multiple times in the same test suite           mocha/no-identical-title
  145:9  error  Test title is used multiple times in the same test suite           mocha/no-identical-title
  155:9  error  Unexpected use of duplicate Mocha `beforeEach` hook                mocha/no-sibling-hooks
  155:9  error  Unexpected use of Mocha `beforeEach` hook outside of a test suite  mocha/no-top-level-hooks
  193:9  error  Test title is used multiple times in the same test suite           mocha/no-identical-title

/Users/mroderick/code/sinon/test/stub-test.js
   845:12  error  Unexpected skipped mocha test                                     mocha/no-skipped-tests
   858:12  error  Unexpected skipped mocha test                                     mocha/no-skipped-tests
  1668:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  1683:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  1704:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  1727:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  1762:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  1801:9   error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case

/Users/mroderick/code/sinon/test/util/core/format-test.js
  12:8  error  Unexpected skipped mocha test  mocha/no-skipped-tests

/Users/mroderick/code/sinon/test/util/core/wrap-method-test.js
  169:13  error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  180:13  error  Unexpected use of Mocha `afterEach` hook for a single test case   mocha/no-hooks-for-single-case

/Users/mroderick/code/sinon/test/util/fake-server-test.js
  918:13  error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  925:13  error  Unexpected use of Mocha `afterEach` hook for a single test case   mocha/no-hooks-for-single-case

/Users/mroderick/code/sinon/test/util/fake-timers-test.js
  156:9  error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  532:9  error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case

/Users/mroderick/code/sinon/test/util/fake-xml-http-request-test.js
   165:13  error  Unexpected use of Mocha `beforeEach` hook for a single test case  mocha/no-hooks-for-single-case
  1840:16  error  Unexpected skipped mocha test                                     mocha/no-skipped-tests
  1855:16  error  Unexpected skipped mocha test                                     mocha/no-skipped-tests

✖ 50 problems (50 errors, 0 warnings)
```

### DIFF
--- a/test/.eslintrc.js
+++ b/test/.eslintrc.js
@@ -1,4 +1,4 @@
-{
+module.exports = {
     "env": {
         "mocha": true
     },
--- a/test/.eslintrc.js
+++ b/test/.eslintrc.js
@@ -1,3 +1,7 @@
+const ERROR = "error";
+const WARN = "warn";
+const OFF = "off";
+
 module.exports = {
     "env": {
         "mocha": true
@@ -6,8 +10,20 @@ module.exports = {
         "mocha"
     ],
     "rules": {
-        "mocha/no-exclusive-tests": 2,
-        "max-nested-callbacks": 0,
-        "no-restricted-syntax": [2, "TryStatement"]
+        "max-nested-callbacks": OFF,
+        "no-restricted-syntax": [ERROR, "TryStatement"],
+
+        // Mocha Plugin - https://github.com/lo1tuma/eslint-plugin-mocha
+        "mocha/handle-done-callback": ERROR,
+        "mocha/no-exclusive-tests": ERROR,
+        "mocha/no-global-tests": ERROR,
+        "mocha/no-hooks-for-single-case": OFF,
+        "mocha/no-identical-title": ERROR,
+        "mocha/no-mocha-arrows": ERROR,
+        "mocha/no-nested-tests": ERROR,
+        "mocha/no-return-and-callback": ERROR,
+        "mocha/no-sibling-hooks": ERROR,
+        "mocha/no-skipped-tests": WARN,
+        "mocha/no-top-level-hooks": ERROR
     }
 }
--- a/test/call-test.js
+++ b/test/call-test.js
@@ -23,6 +23,7 @@ function spyCallCallSetup() {
 
 function spyCallCalledTests(method) {
     return function () {
+        // eslint-disable-next-line mocha/no-top-level-hooks
         beforeEach(spyCallSetUp);
 
         it("returns true if all args match", function () {
@@ -63,6 +64,7 @@ function spyCallCalledTests(method) {
 
 function spyCallNotCalledTests(method) {
     return function () {
+        // eslint-disable-next-line mocha/no-top-level-hooks, mocha/no-sibling-hooks
         beforeEach(spyCallSetUp);
 
         it("returns false if all args match", function () {
@@ -1251,22 +1253,22 @@ describe("sinonSpy.call", function () {
 
         describe("calls", function () {
             it("oneLine", function () {
-                function test(arg, expected) {
+                function verify(arg, expected) {
                     var spy = sinonSpy();
                     spy(arg);
                     assert.equals(spy.printf("%C").replace(/ at.*/g, ""), "\n    " + expected);
                 }
 
-                test(true, "spy(true)");
-                test(false, "spy(false)");
-                test(undefined, "spy(undefined)");
-                test(1, "spy(1)");
-                test(0, "spy(0)");
-                test(-1, "spy(-1)");
-                test(-1.1, "spy(-1.1)");
-                test(Infinity, "spy(Infinity)");
-                test(["a"], "spy([\"a\"])");
-                test({ a: "a" }, "spy({ a: \"a\" })");
+                verify(true, "spy(true)");
+                verify(false, "spy(false)");
+                verify(undefined, "spy(undefined)");
+                verify(1, "spy(1)");
+                verify(0, "spy(0)");
+                verify(-1, "spy(-1)");
+                verify(-1.1, "spy(-1.1)");
+                verify(Infinity, "spy(Infinity)");
+                verify(["a"], "spy([\"a\"])");
+                verify({ a: "a" }, "spy({ a: \"a\" })");
             });
 
             it("multiline", function () {
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -55,6 +55,7 @@ describe("issues", function () {
     });
 
     describe("#624", function () {
+        // eslint-disable-next-line mocha/no-skipped-tests
         it.skip("useFakeTimers should be idempotent", function () {
             // Issue #624 shows that useFakeTimers is not idempotent when it comes to
             // using Date.now
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -165,6 +165,7 @@ describe("sinonSandbox", function () {
                 });
         });
 
+        // eslint-disable-next-line mocha/no-identical-title
         it("must set all stubs created from sandbox with mockPromise", function () {
 
             var resolveValue = {};
--- a/test/spy-test.js
+++ b/test/spy-test.js
@@ -8,6 +8,7 @@ var refute = referee.refute;
 
 function spyCalledTests(method) {
     return function () {
+        // eslint-disable-next-line mocha/no-top-level-hooks
         beforeEach(function () {
             this.spy = createSpy.create();
         });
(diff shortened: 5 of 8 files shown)


test/.eslintrc.js
```diff
@@ -0,0 +1,27 @@
+const ERROR = "error";
```
fearphage: Why the switch to `.js`?

lucasfcosta:
My personal opinion on these rules:

* `mocha/no-top-level-hooks` - Not sure this is good, sometimes it may be useful to have this kind of global hooks defined. Even though this just makes it mandatory to explicitly state you are aware of what you are doing, I'm not sure it's worth it. However, I don't have an **strong** opinion on this matter.
* `mocha/no-nested-tests` - Strongly agree. Tests should only be nested inside "`describe`" blocks. This makes them more readable and organized.
* `mocha/no-hooks-for-single-case` - Disagree. On one side, `beforeEach` exists to avoid duplication, if there would be no duplication without it, we can just make things simpler. However, sometimes these hooks also allow us to organize the code in a more semantic way and make it easier to deal with promises and asynchronicity, because then we can encapsulate those conditions inside these hooks and call `done` inside them or just return a `Promise` instead of having to chain stuff.
* `mocha/no-skipped-tests` - Strongly agree too. Skipped tests are just like commented code. They should've been deleted if they are not being used and if we need them sometime in the future we can just use Git and all of its magic.
* `mocha/no-identical-title` - Strongly agree. This makes it harder to use grep features and differentiate test output in case we're using dot reporter or something like that.
* `mocha/no-sibling-hooks` - Strongly agree. No need for adding unnecessary code if we can just fit it all inside a single hook most of the time. Use case

Even though I know we can still use `// eslint-disable-line` to overcome the problems I've disagreed with when they happen, I think that this feature should be used only when strictly necessary and by creating linting rules for them we're discouraging our contributors of doing so. Having said that, I only want to explain that the rules I've disagreed with are not necessarily a bad thing most of times and I think it's not worth "enforcing" them automatically.

Please let me know if you disagree with anything I've said and we can talk about these rules in depth.

fearphage:
* [x] `mocha/no-hooks-for-single-case` - agree with @lucasfcosta. Always make tests as small as possible and abstract away the setup. It's for one test today, but it might make it easier in the future for others to add additional tests.
* [ ] `mocha/no-skipped-tests` - Tests are skipped for certain environments and/or browser configurations. I'm working on a PR for this to skip them better (using `this.skip()` in mocha instead of wrapping `if` blocks around them). I don't believe skipping tests are always bad. However there are some tests that are completely empty, because the body is commented out. That's a no-no. I also think there is value in having tests with comments. "This test will fail until #xyz is fixed"
* [x] `mocha/no-sibling-hooks` - prevents confusion/makes code easy to reason about.
* [x] `no-exclusive-tests` - These should never be.
* [x] `mocha/no-identical-title` - prevents confusion
* [x] `mocha/no-nested-tests` - I didn't even realize that this was possible.
* [x] `mocha/no-top-level-hooks` - Should be enabled until we need them.

What I really want (that [this plugin doesn't support](https://github.com/lo1tuma/eslint-plugin-mocha/issues/90)) is to disallow tests without assertions.

fatso83:
no strong feelings. i'll follow the lead.

mroderick:
> What I really want (that this plugin doesn't support) is to disallow tests without assertions.

That is a very nice feature of Buster.JS, that I've been missing myself. I've been wondering how I can add this to Mocha, perhaps as a plugin.

test/.eslintrc.js
fatso83: possibly just editor support (auto-complete). no biggie.

mroderick: It also allows us to use comments

fearphage: While we're making changes, I prefer using yaml for eslint personally. It's easier to read and you don't have to fuss with quotation marks and balancing braces. It also supports comments.

fatso83:
I think I am board with Phred in that skipping tests could be valuable from time to time. All others are fine. Also, using either JS or YAML is fine, but this is not a file we will be editing a lot once created, so the minor improvements you get from YAML seems negligible.

Can't we just rebase this and get it in? The errors found would be nice to get fixed anyhow - skipped tests are not :smile_cat:

test/.eslintrc.js
mroderick: I tried to get it working with a YAML file, and eventually ran out of patience. If you feel like contributing a PR that converts it, I'd be happy to merge it

mroderick:
I've rebased and fixed up some files, so we can merge this

PR #1428, opened 2017-05-26
Fix failing reset() on sandbox with unused fake server

fatso83:
`sandbox.reset()` fails when the sandbox has been configured to use a fake server and the server has not been utilized. This is the standard config for `sinon-test` and was revealed to crash in sinonjs/sinon-test#57

```javascript
    var sandbox = sinonSandbox.create({useFakeServer: true});
    sandbox.reset();
```

will crash immediately today due to the fields not being initalized.

#### Solution  - optional
Initializes the fields at creation time.

1. Check out this branch
2. `npm install`
3. `git reset --hard HEAD~1 && npm test #pre-fix the tests fail`
4. `git merge && npm test #after fix the tests pass`

### DIFF
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -56,6 +56,22 @@ describe("sinonSandbox", function () {
         assert.same(sandbox.assert, sinonAssert);
     });
 
+    it("can be reset without failing when pre-configured to use a fake server", function () {
+        var sandbox = sinonSandbox.create({useFakeServer: true});
+        refute.exception(function () {
+            sandbox.reset();
+        });
+    });
+
+    it("can be reset without failing when configured to use a fake server", function () {
+        var sandbox = sinonSandbox.create();
+        sandbox.useFakeServer();
+        refute.exception(function () {
+            sandbox.reset();
+        });
+    });
+
+
     describe(".useFakeTimers", function () {
         beforeEach(function () {
             this.sandbox = Object.create(sinonSandbox);
--- a/lib/sinon/util/fake_server.js
+++ b/lib/sinon/util/fake_server.js
@@ -100,6 +100,9 @@ var fakeServer = {
         this.xhr = fakeXhr.useFakeXMLHttpRequest();
         server.requests = [];
         server.requestCount = 0;
+        server.queue = [];
+        server.responses = [];
+
 
         this.xhr.onCreate = function (xhrObj) {
             xhrObj.unsafeHeadersEnabled = function () {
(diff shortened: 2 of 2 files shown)


test/sandbox-test.js
```diff
@@ -56,6 +56,18 @@ describe("sinonSandbox", function () {
         assert.same(sandbox.assert, sinonAssert);
     });
 
+    it("can be reset without failing when pre-configured to use a fake server", function () {
+        var sandbox = sinonSandbox.create({useFakeServer: true});
+        sandbox.reset();
```
fearphage: There are no assertions in either test. How did this fail before your changes?

fatso83: ah, good catch. I'll add some kind of refutation along with `throws` (don't remember the exact api).

failing test
```
 1) sinonSandbox can be reset without failing when configured to use a fake server:
     TypeError: Cannot set property 'length' of undefined
      at Object.resetBehavior (lib/sinon/util/fake_server.js:279:51)
      at Object.reset (lib/sinon/util/fake_server.js:274:14)
      at lib/sinon/collection.js:26:21
      at Array.forEach (native)
      at each (lib/sinon/collection.js:25:19)
      at Object.reset (lib/sinon/collection.js:41:9)
      at Context.<anonymous> (test/sandbox-test.js:68:17)

```

fatso83: OK, pushed a fix. Output before fix:

```

  1306 passing (846ms)
  4 pending
  2 failing

  1) sinonSandbox can be reset without failing when pre-configured to use a fake server:
     AssertionError: [refute.exception] Expected not to throw but threw TypeError (Cannot set property 'length' of undefined)
      at referee.fail (node_modules/referee/lib/referee.js:193:25)
      at Object.fail (node_modules/referee/lib/referee.js:90:21)
      at assertion (node_modules/referee/lib/referee.js:98:21)
      at Function.referee.(anonymous function).(anonymous function) [as exception] (node_modules/referee/lib/referee.js:118:23)
      at Context.<anonymous> (test/sandbox-test.js:61:16)

  2) sinonSandbox can be reset without failing when configured to use a fake server:
     AssertionError: [refute.exception] Expected not to throw but threw TypeError (Cannot set property 'length' of undefined)
      at referee.fail (node_modules/referee/lib/referee.js:193:25)
      at Object.fail (node_modules/referee/lib/referee.js:90:21)
      at assertion (node_modules/referee/lib/referee.js:98:21)
      at Function.referee.(anonymous function).(anonymous function) [as exception] (node_modules/referee/lib/referee.js:118:23)
      at Context.<anonymous> (test/sandbox-test.js:69:16)


```

mroderick:
This has become part of `sinon@2.3.2`, which is on npm

PR #1435, opened 2017-05-28
Add test for #950, show that it has been fixed

### LINKED ISSUE #950: First execution of a spy as a method renames that spy
I find this quite bizarre. Is this intended behavior? It really conflicts with clear test output. Executing a spy as a method of an object _renames_ that spy to the method name, but only once:

``` js
function bob () {}
var spy = sinon.spy(bob);
console.log(spy); // function bob

var obj = { methodName: spy };
console.log(spy); // function bob

spy();
console.log(spy); // function bob

obj.methodName.call(null);
console.log(spy); // function bob

obj.methodName();
console.log(spy); // function methodName

obj.otherProp = spy;
obj.otherProp();
console.log(spy); // function methodName
```

mroderick:
This PR adds a test for the situation described in #950, showing that it has been fixed.

Fixes #950

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -122,6 +122,33 @@ describe("issues", function () {
         });
     });
 
+    describe("#950 - first execution of a spy as a method renames that spy", function () {
+        it("should not rename spies", function () {
+            var expectedName = "proxy";
+
+            function bob() {}
+            var spy = sinon.spy(bob);
+
+            assert.equals(spy.name, expectedName);
+
+            var obj = { methodName: spy };
+            assert.equals(spy.name, expectedName);
+
+            spy();
+            assert.equals(spy.name, expectedName);
+
+            obj.methodName.call(null);
+            assert.equals(spy.name, expectedName);
+
+            obj.methodName();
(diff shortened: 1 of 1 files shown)


mantoni:
👍

mantoni:
So why on earth did the build pass on this branch and then failed on master in this very test case?

```
 1) issues #950 - first execution of a spy as a method renames that spy should not rename spies:
     AssertionError: [assert.equals] undefined expected to be equal to proxy
      at referee.fail (node_modules/referee/lib/referee.js:197)
      at ctx.fail (node_modules/referee/lib/referee.js:90)
      at assertion (node_modules/referee/lib/referee.js:98)
      at referee[type][name] (node_modules/referee/lib/referee.js:118)
      at Anonymous function (test/issues/issues-test.js:132)
```

mantoni:
Only in IE 11 ™️  🙄

mroderick:
Grrrr

mantoni:
Ah, I remember now that we explicitly disabled sauce-labs browser testing on pull requests for security reasons. The browser tests only run on master, unfortunately.

PR #1436, opened 2017-05-29
Feature detect function name property in issue 950

mantoni:

Fix the sauce labs build on master

IE 11 does not support the name property on functions.

#### Solution  - optional

The test case for issue #950 is only executed if the name property is supported.

2. `npm install`
3. `npm run test-cloud`


mantoni:
Check the non-whitespace diff: https://github.com/sinonjs/sinon/pull/1436/files?w=1

mroderick:
👍

Issue #617, opened 2014-12-02
Behavior of sandbox creation inconsistent with documentation.


divmain:
The provided [documentation](https://github.com/cjohansen/sinon-web/blob/master/docs/index.html#L1482-L1502) seems to indicate that a call to `sinon.sandbox.create` will create a sandbox with a default configuration (namely, enabled fake timers and fake server).

However, this is inconsistent with the actual behavior.  See [example fiddle](http://jsfiddle.net/8k0zv6tp/).  Instead, an unconfigured instance is returned with `sinon.sandbox` in the prototype chain ([link](https://github.com/cjohansen/Sinon.JS/blob/master/lib/sinon/sandbox.js#L110-L112)).

It doesn't appear as if there is any default configuration behavior for sandboxes, so this section of the documentation is misleading at best.  Either the documentation or the behavior should be updated to be consistent with the other.  I'm happy to submit a PR for either of these cases, if you can point me in the preferred direction.

tjunghans:
I can confirm this with sinon@1.7.2. Not only is the documentation misleading as described by @divmain, but even if the default configuration as shown in the [sandbox documentation](http://sinonjs.org/docs/#sandbox) is used (see snippet below), only the `clock` property is available.

```js
sinon.defaultConfig = {
    // ...
    injectInto: null,
    properties: ["spy", "stub", "mock", "clock", "server", "requests"],
    useFakeTimers: true,
    useFakeServer: true
}
```

I'm also happy to help (/cc @mantoni)

mroderick:
The [current documentation shows how to create a sandbox without a config](https://github.com/sinonjs/sinon/blob/454e0fdb178dccb514f2ccf6b306fe6c44ec6b1d/docs/release-source/release/sandbox.md#var-sandbox--sinonsandboxcreate) and clearly states that `sinon.sandbox.create(config)` is an integration feature, that users are likely not to need.

I consider this solved, and am closing this. Feel free to create a PR for the new documentation if you have improvements for this area of the documentation.

ryan-roemer:
I think perhaps a further exploration of the issue may help.

### Incorrect docs and/or behavior

The documentation says:

> #### `var sandbox = sinon.sandbox.create(config);`
> 
> The `sinon.sandbox.create(config)` method is mostly an integration feature, and as an end-user of Sinon.JS you will probably not need it.
> 
[...]

PR #1443, opened 2017-06-01
Docs: Fix incorrect sandbox configuration documentation.

ryan-roemer:
#### Purpose (TL;DR)

The documentation for `sinon.sandbox` is incorrect and this PR fixes it. The bug was originally identified in #617.

@divmain identified that the `sinon.sandbox` documentation is wrong in #617 and in a fiddle http://jsfiddle.net/8k0zv6tp/

Here's a concise example:

The documentation says that calling:

```js
var sandbox = sinon.sandbox.create({});
```

essentially means:

```js
var sandbox = sinon.sandbox.create({
    // ...
    injectInto: null,
    properties: ["spy", "stub", "mock", "clock", "server", "requests"],
    useFakeTimers: true,
    useFakeServer: true
});
```

This is wrong. In the shell we have:

```js
> var sandbox = sinon.sandbox.create({})
undefined
> sandbox.clock
undefined
> sandbox.server
undefined
```

What really happens is equivalent to:

```js
var sandbox = sinon.sandbox.create({
    // ...
    injectInto: null,
    properties: ["spy", "stub", "mock"],
    useFakeTimers: false,
    useFakeServer: false
});
```

The upstream #617 issue was closed with https://github.com/sinonjs/sinon/issues/617#issuecomment-304498837 comment of:

> The current documentation shows how to create a sandbox without a config and clearly states that sinon.sandbox.create(config) is an integration feature, that users are likely not to need.
>
> I consider this solved, and am closing this. Feel free to create a PR for the new documentation if you have improvements for this area of the documentation.

As part of the public API, some of our biggest clients use `sandbox` exclusively for large test suites for better fake management and cleanup, we have a vested interest in making sure the documentation and behavior of sandbox is correct.

So, after the issue was closed I asked "is this a bug? or a documentation bug?" and I haven't received a response. So, I took a hint from the closing comment and making the simplifying assumption it's a documentation bug and here's the requested PR!

Enjoy!

#### How to verify

The documentation behavior being wrong can be seen with:

```
$ npm install sinon
$ node
> var sandbox = sinon.sandbox.create({})
undefined

// These **should** be defined.
> sandbox.clock
undefined
> sandbox.server
undefined

// Now, let's do what the PR improved docs say.
> sandbox = sinon.sandbox.create(sinon.defaultConfig)
undefined

// These **should** be defined.
> sandbox.clock
{ now: 0,
  hrNow: 0,
  timeouts: {},
  Date: 
   { [Function: ClockDate]
// ... STUFF ...
> sandbox.server
{ logError: [Function: logError],
  requests: [],
  requestCount: 0,
  queue: [],
  responses: [] }
```

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -41,16 +41,32 @@ describe('myAPI.hello method', function () {
 
 #### `var sandbox = sinon.sandbox.create();`
 
-Creates a sandbox object
+Creates a sandbox object with spies, stubs, and mocks.
 
 
 #### `var sandbox = sinon.sandbox.create(config);`
 
-The `sinon.sandbox.create(config)` method is mostly an integration feature, and as an end-user of Sinon.JS you will probably not need it.
+The `sinon.sandbox.create(config)` method is often an integration feature, and can be used for scenarios including a global object to coordinate all fakes through.
 
-Creates a pre-configured sandbox object. The configuration can instruct the sandbox to include fake timers, fake server, and how to interact with these.
+Sandboxes are partially configured by default such that calling:
 
-The default configuration looks like:
+```javascript
+var sandbox = sinon.sandbox.create({});
+```
+
+will merge in extra defaults analogous to:
+
+```javascript
+var sandbox = sinon.sandbox.create({
+    // ...
+    injectInto: null,
+    properties: ["spy", "stub", "mock"],
+    useFakeTimers: false,
+    useFakeServer: false
+});
+```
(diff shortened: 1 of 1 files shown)


mroderick:
I think this is a great contribution, thank you!

Let's get another set of eyes on it, before we merge it

mantoni:
I agree. Very good contribution. We should get rid of the `defaultConfig` thingy, in my opinion. The code already has a comment saying that it's exposed for backward compatibility.

Issue #1445, opened 2017-06-02
Sandbox stubbing static function property throws error


raulmatei:
* Sinon version : 2.3.2
* Environment: Node
* Other libraries you are using: Mocha

**What did you expect to happen?**
Should be able to stub a function static property. It worked with v2.2.0

**What actually happens**
Throws an error:
`TypeError: Attempted to wrap object property content as function`

**How to reproduce**

```js
// code
const fn = () => false;

fn.content = {
  a: 'A',
  b: 'B',
};
```

```js
// test
before(() => {
  sandbox = sinon.sandbox.create();
  
  sandbox.stub(fn, 'content').value({
    a: '__A__',
    b: '__B__',
  });
});

after(() => {
  sandbox.restore();
});
```

fatso83:
Seems like we have a regression. Could it be related to the new getter/setter stubbing functionality, @lucasfcosta?

raulmatei:
It looks to me that this was introduced with this commit https://github.com/sinonjs/sinon/commit/278c2ce66cb88e81e3752c672ed630369f15fc34#diff-048fb1194de4b43f7f7e45758f900749R21, changing that check to:

```js
var isStubbingNonFuncProperty = (typeof object === "object" || typeof object === "function")
                                    && typeof property !== "undefined"
                                    && (typeof actualDescriptor === "undefined"
                                    || typeof actualDescriptor.value !== "function")
                                    && typeof descriptor === "undefined";
```

seems to make it work.

Thanks,
Raul

lucasfcosta:
@fatso83, yup! Good catch @raulmatei.

I missed a check to see if the passed object itself was a function.

That should indeed solve the problem.

raulmatei:
@lucasfcosta will come with a PR later this week. Hope it won't be to late for others.

Raul.

lucasfcosta:
Hi folks. I can't close issues but I believe this one is done.

PR #1450, opened 2017-06-06
Fix 1445: make stubbing of static function properties possible

raulmatei:
Fix issue #1445 related to stubbing function static properties by adding an extra check for the `object` type to be a function


mroderick:
I think this looks like a new `MINOR` version. What say you @sinonjs/core?

lucasfcosta:
@mroderick I think this looks more like a `PATCH`, because this should have been possible before. AFAIK this is what we expected from the sandbox API since the beginning of its design.

According to the docs:

> The sandbox stub method **can also be used to stub any kind of property**. This is useful if you need to override an object’s property for the duration of a test, and have it restored when the test completes.

mroderick:
Cheers @lucasfcosta!

I've published this change as `v2.3.3`

PR #1454, opened 2017-06-11
Fix “it’s” when used in the possessive form

keirog:


mantoni:
Very nice! Thanks a lot 👍

Issue #1453, opened 2017-06-11
Proposal: extract fake xhr and fake server to separate module


mroderick:
I would like to propose that we excise fake xhr and fake server from `sinon` and publish them as a separate module.

Down the line, I would like to remove them from core sinon, and just have them live their own life. 

## Strategy

* Create new repository for `xhr` and `server`, publish them under a new name `nise` (偽 - adjective "fake" in Japanese ... "fake server", "fake xhr", etc).
* Import `nise` back into `sinon`, keeping exactly the same api, adding deprecation warnings
* Publish new `MAJOR` version, that removes old deprecations, and uses `nise` for fake xhr and server
* Some time in the future, remove `nise` from `sinon` in a new major version

For the first part, I have created a temporary repository under my name: https://github.com/mroderick/nise and have started importing things.

I am trying to leave the source and tests fairly unchanged for now. If we want to make significant improvements to this module, let's do that when it is extracted completely.

#### Open questions and tasks

- [ ] Shall we remove **all** previously deprecated things with this new major version?

mroderick:
Ping: @cjohansen @fatso83 @fearphage @jonnyreeves @mantoni @lucasfcosta

mantoni:
Absolutely in favor of extracting them. Your plan to deprecate and then remove makes sense. I'm wondering though whether we should deprecate in a major and remove in a following major, since most people will receive implicit upgrades in their builds through ^ dependencies.

fatso83:
We talked about this before, and all agreed back then, so yeah: still in favor.

@mantoni The caret won't upgrade major versions, will it? Thought it would do patches and minor versions, but not major. That would make very little sense.

Don't see the big point in deprecating now that we eliminated most hurdles for doing major releases more frequently, but if it helps in any way: sure, why not.

mantoni:
@fatso83 Yes, `^` only updates minor and patch. My question is: Do we add deprecation warnings in a minor release, or in a major? Obviously, we're going to remove the APIs in a(nother) major.

fearphage:
Like the concept. Not a fan of the tests and source living next to each other ([mass hysteria](https://youtu.be/JmzuRXLzqKk)), but that's something we can work on later.

> remove fake-xhr-and-server from sinon

Can we work on a better name? `sinon-networking`? `sinon-net`? Just throwing things out there.

> we should deprecate in a major and remove in a following major

Agreed.

[...]

PR #1461, opened 2017-06-17
Import nise

mroderick:
This PR is part of #1453, it replaces fake xhr and fake server with [`nise`](https://github.com/mroderick/nise)

#### To try it out

1. Check out this branch
1. `npm install`
1. `npm test`


mroderick:
Added the `MAJOR` label because:

* 944f6cf removes previously deprecated properties
* We might break something

It's better that people only get this release through conscious upgrade, and not part of regular install (damn you loose versioning!!!!)

fearphage:
> * 944f6cf removes previously deprecated properties

FYI these changes should end up int he release notes.

mroderick:
I have pushed new commits that imports the new nise module from npm.

There are still some deprecations we can remove:

* [`sandbox.stub(obj, 'meth', val)`](https://github.com/sinonjs/sinon/blob/2395f35b01c74eab50a27f38ad903c3d497dd931/lib/sinon/sandbox-stub.js#L13-L18)
* [`sinon.stub(obj, 'meth', fn)`](https://github.com/sinonjs/sinon/blob/2395f35b01c74eab50a27f38ad903c3d497dd931/lib/sinon/stub-descriptor.js#L12-L17)
* [`sinon/util/core`](https://github.com/sinonjs/sinon/blob/2395f35b01c74eab50a27f38ad903c3d497dd931/lib/sinon.js#L7-L26)

Shall we remove them as part of this `MAJOR` release, or save them for another day?

fatso83:
Remove, remove, remove!

fatso83:
I think we can merge this and remove any remaining deprecations in another PR. Anyone opposed?

mroderick:
> Anyone opposed?

Me! I am already working on removing the last deprecations, so I'd rather just keep this branch open for a bit longer, than either have `master` unshippable again or having to maintain another legacy branch.

mroderick:
OK, I think I have mopped up all the previously deprecated stuff.

If we're going to put out a new `MAJOR` release based on this PR, then I think it deserves a thorough review.

mroderick:
I am attending a family event today, so won't be able to contribute anything until tonight at the earliest and perhaps not even until sometime next week.

mroderick:
Ping @sinonjs/core do you have any further comments for this PR?

fatso83:
Njet. All good.

lib/sinon/stub-non-function-property.js
```diff
@@ -3,15 +3,13 @@
...
 
-function stubNonFunctionProperty(object, property, value) {
+function stubNonFunctionProperty(object, property) {
     var original = object[property];
 
     if (!hasOwnProperty.call(object, property)) {
         throw new TypeError("Cannot stub non-existent own property " + valueToString(property));
     }
 
-    object[property] = value;
```
fearphage: Looks like most of the property tests were removed. I see negative tests, but no positive tests.

lib/sinon/util/core/index.js
```diff
@@ -2,7 +2,6 @@
 
 module.exports = {
     calledInOrder: require("./called-in-order"),
-    configureLogError: require("./log_error"),
```
fearphage: If this was accessible, we should document that it was removed even if it was previously deprecated.

lib/sinon.js
```diff
@@ -49,20 +41,13 @@ exports.useFakeTimers = fakeTimers.useFakeTimers;
 exports.clock = fakeTimers.clock;
 exports.timers = fakeTimers.timers;
 
-var event = require("./sinon/util/event");
-exports.Event = deprecated.wrap(event.Event, deprecated.defaultMsg("Event"));
```
fearphage: Documentation for all of these changes.

lib/sinon/stub-descriptor.js
```diff
@@ -1,41 +0,0 @@
-"use strict";
-
-var deprecated = require("./util/core/deprecated");
-var spy = require("./spy");
-var wrapMethod = require("./util/core/wrap-method");
-
-// This is deprecated and will be removed in a future version of sinon.
-// We will only consider pull requests that fix serious bugs in the implementation
-function stubDescriptor(object, property, descriptor) {
```
fearphage: Documentation.

test/sandbox-test.js
```diff
@@ -597,6 +597,7 @@ describe("sinonSandbox", function () {
 
         it("allows stubbing setters", function () {
             var object = {
+                foo: undefined,
```
fearphage: Did this need to be added?

lib/sinon/stub-non-function-property.js
```diff
@@ -3,15 +3,13 @@
 var valueToString = require("./util/core/value-to-string");
 var hasOwnProperty = Object.prototype.hasOwnProperty;
 
-function stubNonFunctionProperty(object, property, value) {
+function stubNonFunctionProperty(object, property) {
```
mroderick: > This should throw like the other stub method with too many arguments.

It is internal only, and is called inside `sandbox.stub`, which does throw for too many arguments.

mroderick: > The replacement for this usage is not documented

The practice of using `sandbox.stub` for [stubbing out properties has been documented for awhile](http://sinonjs.org/releases/v2.3.8/sandbox/#sandboxstub), admittedly, not very well.

lib/sinon/stub-descriptor.js
mroderick: This was an internal only file.

That `sinon.stub` only accepts two arguments has been [documented in this very commit](https://github.com/mroderick/sinon/blob/fdea27c9b77ea30e97bd0233853cabb7a77e0b7c/docs/release-source/release/stubs.md#var-stub--sinonstubobject-method-func)

test/sandbox-test.js
mroderick: It is necessary, otherwise we run into the [input validation of collection.stub, that checks for own properties on stubbing targets](https://github.com/sinonjs/sinon/blob/5eca462a5f769c7de1622886be66d76a67f245a1/lib/sinon/collection.js#L93-L95).

`TypeError: Cannot stub non-existent own property foo`

lib/sinon/stub-non-function-property.js
mroderick: Stubbing properties is the responsibility of `collection.js`, it has [several tests covering this](https://github.com/sinonjs/sinon/blob/5eca462a5f769c7de1622886be66d76a67f245a1/test/collection-test.js#L131-L182).

lib/sinon/collection.js
```diff
@@ -85,9 +85,13 @@ var collection = {
         return this.add(sinonSpy.apply(sinonSpy, arguments));
     },
 
-    stub: function stub(object, property/*, value*/) {
+    stub: function stub(object, property) {
         if (arguments.length > 2) {
-            return sandboxStub.apply(this, arguments);
+            throw new TypeError("stub(obj, 'meth', fn) has been removed, see documentation");
```
mroderick: Indeed, well spotted

lib/sinon/stub-non-function-property.js
mroderick: I have pushed another commit to improve documentation for stubbing out non-function properties

mroderick:
@fearphage thank you for your thorough review.

I think I have addressed all your comments.

If you're satisfied, then I'll fold the review commit into the appropriate commit and merge it.

test/stub-test.js
```diff
@@ -1027,11 +996,7 @@ describe("stub", function () {
                     fail("should not call getter");
                 }
             };
-            var stub = createStub(obj, "prop", {get: function () {
-                return 43;
-            }});
-
-            assert.equals(obj.prop, 43);
+            var stub = createStub(obj, "prop");
```
fearphage: This test has no assertions now. Add an assertion or remove the test.

mroderick: That was a mistake! I've fixed this in the offending commit and have rebased my branch

mroderick:
I think everything has been addressed. Since #1496 is also a `MAJOR`, I would like to merge that one first and then this one, and then just cut one new `MAJOR` release, instead of two.

PR #1469, opened 2017-06-22
Run tests on Travis in node 8

mroderick:
This PR adds node 8 as a test target on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: node_js
 node_js:
-  # LTS release per February 2016
-  - "4"
-  - "6"
+  # https://github.com/nodejs/LTS
+  - "4" # ends April 2018
+  - "6" # ends April 2019
+  - "8" # ends December 2019
 
 sudo: false
 


mroderick:
I can't quite figure out why the builds are not working in node 8. It works fine on my machine

fatso83:
Weird. "Error: Cannot find module 'lolex' from '/home/travis/build/sinonjs/sinon/lib/sinon/util'". Hoping it's a temporary fluke in the NPM install. Restarting the build.

fearphage:
Not a fix, but can one of you click buttons here? I've used CircleCI before and I've always prefered it over Travis. It's faster (anecdotely) both in processing and not having to wait for builds and the UI is a bit more polished.

https://circleci.com/add-projects/gh/sinonjs

mroderick:
I could

lucasfcosta:
I know this is related to #1479 so I'd like to give my 2 cents on this:

First of all I'd like to say I'm not against this change. For me I'd be happy with whichever works, however, in order to avoid unnecessary work maybe we could try fixing problems with TravisCI.

I know CircleCI works great but I think TravisCI does too. We've had a few problems with SauceLabs on ChaiJS too and our builds were flaky for a few weeks, but if I remember correctly it was due to the number of concurrent builds we were running, so if we can get SauceLabs configs that work fine maybe we could keep the CI process on Travis.

mantoni:
I tried to run the build again with cleared caches (I'm sure you already tried yourself). When I run it locally on Node 8 it works fine, including SauceLabs tests (except there is a Safari 8 issue with one of the tests).

Since the exception is thrown in the `build.js` file and given this works fine locally for anybody who's doing releases, we could change the `prepublish` script to `prepublishOnly`, as npm suggests. This would simply _not_ run the build step when installing. After all, `prepublish` is deprecated anyway.

fatso83:
I restarted the build. It now builds. Although an error in the prepublish step can be seen, the process exits with exit code 0.

mroderick:
It's not difficult to remove it again, if we have to

PR #1473, opened 2017-06-27
Simplified eslint config (js -> yaml)

fearphage:
It's easier to read and edit since there's not much extra cruft (quotes, braces, commas, etc).


test/.eslintrc
```diff
@@ -0,0 +1,22 @@
+env:
+  mocha: true
+
+plugins:
+  - mocha
```
mroderick: Why does this get a `-`, but the ones in `rules` do not?

(I was cursing at YML last night)

fearphage: `plugins` is an array. `rules` is an object.

mroderick: Cheers!

PR #1477, opened 2017-06-30
Add missing documentation for spy.getCalls()

mroderick:
[`spy.getCalls()`](https://github.com/sinonjs/sinon/blob/833ff2c25f2a6af96b782422c1f1bef068d63bb2/lib/sinon/spy.js#L242-L251) has been part of the codebase since at least 8d69a482, which was
committed on (2013-09-08 15:18:55 +0100)

### DIFF
--- a/docs/_releases/v1.17.6/spies.md
+++ b/docs/_releases/v1.17.6/spies.md
@@ -351,6 +351,11 @@ assertEquals("/stuffs", spyCall.args[0]);
 ```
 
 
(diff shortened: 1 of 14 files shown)


grassator:
@mroderick Thanks for this, the method is very handy to find one of the calls with [Array.prototype.find](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

PR #1478, opened 2017-06-30
Fix #1476: stub#firstCall seems broken on sinon v2.3.6

### LINKED ISSUE #1476: stub#firstCall seems broken on sinon v2.3.6
Hello,

First of all, thanks for the work done :)

Today, one of our unit tests has failed after the update of sinon. (v2.3.5 -> v2.3.6)
We think the latest release introduced a regression.

* Sinon version : 2.3.6
* Environment   : Debian Jessie, Node v6.11.0

I expect both stub.getCall(0) and stub.firstCall to return the same thing, even with a stub created with withArgs.

**What actually happens**
When I create a stub responding to particular arguments with stub.withArgs, this stub can't be inspected with firstCall.
stub.getCall(0) works fine but stub.firstCall returns null. In the previous release (v2.3.5), calling stub.getCall(0) and getting stub.firstCall produced the same result. 
The same issue can be observed with secondCall and thirdCall as well.

**How to reproduce**
```javascript
const stub = sinon.stub();

stub.withArgs('broken').returns(42);
stub('broken');

stub.withArgs('broken').callCount; // 1
stub.withArgs('broken').calledOnce; // true
stub.withArgs('broken').returnValues[0]; // 42
stub.withArgs('broken').getCall(0) // the first call
stub.withArgs('broken').firstCall; // null
```

takasmiley:
Fix issue #1476 
The cause of this bug is that `createCallProperties(matching)` is never called in spy.js.

### DIFF
--- a/test/spy-test.js
+++ b/test/spy-test.js
@@ -327,7 +327,7 @@ describe("spy", function () {
     });
 
     it("should work with combination of withArgs arguments and order of calling withArgs", function () {
-        var assertSpy = function (spy) {
+        function assertSpy(spy) {
             // assert callCount
             assert.equals(spy.callCount, 4);
             assert.equals(spy.withArgs(1).callCount, 3);
@@ -335,15 +335,15 @@ describe("spy", function () {
             assert.equals(spy.withArgs(1, 2).callCount, 1);
 
             // assert call
-            assert.equals(spy.getCall(0).args[0], undefined);
+            refute.defined(spy.getCall(0).args[0]);
             assert.equals(spy.getCall(1).args[0], 1);
-            assert.equals(spy.getCall(1).args[1], undefined);
+            refute.defined(spy.getCall(1).args[1]);
             assert.equals(spy.getCall(2).args[0], 1);
             assert.equals(spy.getCall(2).args[1], 1);
-            assert.equals(spy.getCall(2).args[2], undefined);
+            refute.defined(spy.getCall(2).args[2]);
             assert.equals(spy.getCall(3).args[0], 1);
             assert.equals(spy.getCall(3).args[1], 2);
-            assert.equals(spy.getCall(3).args[2], undefined);
+            refute.defined(spy.getCall(3).args[2]);
             ["args", "callCount", "callId"].forEach(function (propName) {
                 assert.equals(spy.withArgs(1).getCall(0)[propName],
                               spy.getCall(1)[propName]);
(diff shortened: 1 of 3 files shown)


fatso83:
That was quick :-)

test/stub-test.js
```diff
@@ -115,6 +115,15 @@ describe("stub", function () {
         assert.equals(stub(2), 0);
     });
 
+    it("should work with combination of withArgs arguments", function () {
+        var stub = createStub();
+
+        stub.withArgs(1).returns(42);
+        stub(1);
+
+        assert.isObject(stub.withArgs(1).firstCall);
```
fearphage: I would assert it's not `null` here. While being an object has the same effect, is not null is a bit more explicit.

test/spy-test.js
```diff
@@ -326,7 +326,62 @@ describe("spy", function () {
         });
     });
 
-    it("counts with combination of withArgs arguments and order of calling withArgs", function () {
+    it("should work with combination of withArgs arguments and order of calling withArgs", function () {
+        var assertSpy = function (spy) {
```
fearphage: Just make it a function expression instead. `function assertSpy()...`

test/spy-test.js
```diff
@@ -326,7 +326,62 @@ describe("spy", function () {
...
+            assert.equals(spy.getCall(3).args[1], 2);
+            assert.equals(spy.getCall(3).args[2], undefined);
+            ["args", "callCount", "callId"].forEach(function (propName) {
+                assert.equals(spy.withArgs(1).getCall(0)[propName],
+                              spy.getCall(1)[propName]);
+                assert.equals(spy.withArgs(1).getCall(1)[propName],
+                              spy.getCall(2)[propName]);
+                assert.equals(spy.withArgs(1).getCall(2)[propName],
+                              spy.getCall(3)[propName]);
+                assert.equals(spy.withArgs(1).getCall(3),
```
fearphage: Use `assert.isNull` throughout.

test/spy-test.js
```diff
@@ -326,7 +326,62 @@ describe("spy", function () {
...
+    it("should work with combination of withArgs arguments and order of calling withArgs", function () {
+        var assertSpy = function (spy) {
+            // assert callCount
+            assert.equals(spy.callCount, 4);
+            assert.equals(spy.withArgs(1).callCount, 3);
+            assert.equals(spy.withArgs(1, 1).callCount, 1);
+            assert.equals(spy.withArgs(1, 2).callCount, 1);
+
+            // assert call
+            assert.equals(spy.getCall(0).args[0], undefined);
```
fearphage: Use `refute.defined` assertion throughout.

jvanoostveen:
Hmm it seems with this patch (in 2.3.7) withArgs().lastCall() is broken.
`TypeError: Cannot read property 'lastCall' of undefined`

fatso83:
99 bugs of Sinon on the wall, 99 bugs of Sinon.
Take one down and pass it around, 101 bugs of Sinon on the wall

:sob: 

Seriously, we need some more tests to catch these things. These things are fragile.

@jvanoostveen, could you create an issue with a small reproducible testcase?

jvanoostveen:
It isn't all bad, I've been very happy with Sinon overall. :)
Sorry about the double post, I wanted to add it to the issue instead of the PR.

After updating from 2.3.5 to 2.3.7 we suddenly have some failing tests. (2.3.6 was also failing, can't remember if they where the same tests.)
I'll try if I make an isolated case with the same error (and post is a new issue, linking to this one/conversation).

Issue #1405, opened 2017-05-17
Discussion: Abandon PhantomJS


mroderick:
Chrome 59 will have headless support

* https://developers.google.com/web/updates/2017/04/headless-chrome
* https://www.chromestatus.com/features/5678767817097216

The single maintainer of PhantomJS is stepping down as a consequence of this: https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE

We should probably start thinking about what we want to replace PhantomJS with.

* headless chrome
* https://github.com/tmpvar/jsdom
* your suggestion here

lucasfcosta:
IMO headless chrome is the most adequate substitution.

When using headless browsers to make tests we do not want to test whether or not our code is using the latest APIs, we just want to make sure it works on a browser environment so **I** can't see why we should use other headless platforms since Chrome is what most users have, has great support, great and active team and works extremely fine.

Anyway, I'd love to hear what other people have to say and make sure I'm not missing any detail involved in this discussion.

fatso83:
Chrome, for reasons mentioned above.

mantoni:
I agree with @lucasfcosta.

When I read about headless Chrome, I instantly wanted to add support for it into Mochify. My approach to replace it would be to create a new project with an API and functionality similar to [phantomic](https://github.com/mantoni/phantomic). It would then be trivial to add it to Mochify.

Also, I don't want to be a bottleneck in the development there. If anyone here wants to help, please ping me and I'll add collaborators. I might also group the related projects in a new org. Suggestions are very welcome.

mantoni:
I managed to put an initial, somewhat working Chromium based implementation into Mochify. You can find a somewhat working version here: https://github.com/mantoni/mochify.js/tree/chromium

Tests pass, it works with Sinon and some other projects I tested with. It's quite a bit faster running the tests, but startup time seems not much better than PhantomJS. I'm pretty underwhelmed with the performance gain. On a large test suite like Sinon, it only save around 200 milliseconds.

I went with [Puppeteer](https://github.com/GoogleChrome/puppeteer) which provides everything needed.

There is one downside to this: If an asynchronous error is thrown, it's handled correctly by Mocha, but there is no stack trace and some error details are lost. This works fine with PhantomJS. I'm not sure where this originates. I did a couple of experiments but was unable to get a stack trace other than the internal one from Puppeteer.

Outstanding work:

- Specify external binary (see skipped test case).
[...]

PR #1479, opened 2017-07-03
CircleCI Integration

fearphage:
I'm working on the CircleCI integration to potentially replace TravisCI. So far, the problem I'm having is I don't understand why the `webworker` test (`test-webworker`) is failing on Node 6 and 8. FYI it's not run on Node 4. I mostly copied the logic of the travis config into circle.

Notes: I'm using their new fancy 2.0 configuration which is new to me.

@sinonjs/sinon-core Any ideas? I can't reproduce locally and I don't see anything out of the ordinary. Is there a way to get any more feedback from the [failure](https://circleci.com/gh/fearphage/Sinon.JS/28)?


fatso83:
sorry, no idea. this change smells a bit like the YAML vs JSON vs JS issue for me. Personally, I don't really care as long as it works, as it's just too little time saved to care, but remind me again why you are doing this; aren't these things pretty much the same?

fearphage:
> this change smells a bit like the YAML vs JSON vs JS issue for me

It has nothing to do with the config file itself to my knowledge.

> TravisCI has always worked well for me and it has great integration with many services. Have we been having any problems with it lately? Sometimes it could be easier to get in touch with their support and see if these problems can be solved so that we don't have to change infrastructure.

I just looked into, because of the problems with running tests in Node 8 on travis. I prefer CircleCI and that's what I've used professionally. I have yet to see an integration that Travis has over Circle. I'm just putting it out there as an option. I also like that it reports each node version individually instead of as a set. Thus you don't have to click through to see where the failures are.

> This StackOverflow answer might help you.

Check out the actual build. It's running phantomic with that option: 

```shell
browserify --no-commondir --full-paths -p [ mocaccino -R spec --color ] test/webworker/webworker-support-assessment.js | phantomic --web-security false
```

https://circleci.com/gh/fearphage/Sinon.JS/28

fatso83:
I didn't mean it was a config issue; I meant it was a superficial change that is functionally equivalent. As long as it works and doesn't make things harder it's all fine to me 😁

> because of the problems with running tests in Node 8 on travis.

Which ones are that? I don't know of any. Superficially googled and [came up with this](https://github.com/nodejs/node-gyp/issues/1211), which seems to be a solved problem.

fatso83:
Sorry, I am an idiot. Just saw the list of PRs ... #1469, of course.

fatso83:
Two thoughts:
- could it be that it picks up a phantomjs version on the local path, instead of the one we bundle? (Out on a limb with this one, but could explain the "works on my machine" issue)
- try bumping the phantomic version to 1.5.2. We are on 1.4.0, but it only got [official support for phantom 2.0 with version 1.5.2](https://github.com/mantoni/phantomic/blob/master/CHANGES.md). I might not read the docs right, though ...

fatso83:
[This issue](https://github.com/amir20/phantomjs-node/issues/345) is what made me think of it, btw. Might not be related, but could be.

mroderick:
We should probably make it a priortiy to replace PhantomJS with headless chrome (#1405). We are likely to do this soon anyway, and it is likely to obviate many of these problems.

fatso83:
While researching #1498, I understood what the failing synchronous network request was. It simply happens because ./scripts/build.js (the `prepublish` step) has not been called before `npm test` is run. Therefore there is no `../../pkg/sinon.js` to load in the webworker file. The `importScripts()` call in WebWorkers is synchronous, as they need to load the code before anything else is run.

I'll try to get this fixed as part of my #1498 fixes.

fatso83:
This should work as soon as #1504 lands.

fearphage:
> This should work as soon as #1504 lands.

Thanks for looking into this!

fatso83:
No idea why you are getting `EPEERINVALID`
```
npm ERR! peerinvalid The package eslint@3.19.0 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer eslint-config-sinon@1.0.3 wants eslint@^4.2.0
npm ERR! peerinvalid Peer eslint-plugin-mocha@4.11.0 wants eslint@^2.0.0 || ^3.0.0 || ^4.0.0
```

fatso83:
Hi, this has been building for quite a while. Any reason not to merge it? For some reason, it is already being used in other PRs here, even though it hasn't been merged to `master` :smile:

fatso83:
Squashed and merged. Thanks, Phred!

PR #1462, opened 2017-06-17
Fix: check configurable on a prop before creating (fixes #1456)

gyandeeps:
#### Purpose (TL;DR) - mandatory

Fixes #1456 by making sure propert is made `configurable` based on the object props.

@lucasfcosta  have created a fix with my beast understanding of your proposal. Feedback is welcome. Thanks

Issue #1456, opened 2017-06-12
Attempting to configurable attribute of unconfigurable property.


gyandeeps:
### System info

**Node:** 8
**babel-polyfill:** 6.23.0
**core-js:** ^2.4.0
**phantomjs-prebuilt:** 2.1.14
**Sinon**: 2.3.1 (no issue) - 2.3.2 (has issue)

**What did you expect to happen?**

* It runs fine on chrome
* It throws error on phantomjs
* No issues with sinon 2.3.1. Issue after that.

**What actually happens**

I use `karma` run unit test. `webpack` for bundling and `babel` for code translation and polyfill. Running my unit test using `phantomjs` and i get the following error:

```sh
PhantomJS 2.1.1 (Windows 7 0.0.0) Positioning detectFlipPositionY Special cases Should return original if element exceeds limit on top and bottom FAILED
TypeError: Attempting to configurable attribute of unconfigurable property. in C:/Users/<dir/dir>/node_modules/babel-polyfill/dist/polyfill.js (line 4908)
```

**How to reproduce**

```js
describe("Positioning", () => {
    const sandbox = sinon.sandbox.create();

    const mockWindowWithAttr = (width = 100, height = 100) => {
        sandbox.stub(window, "innerWidth").value(width);
        sandbox.stub(window, "innerHeight").value(height);
    };

    beforeEach(() => {
        mockWindowWithAttr();
    });

    afterEach(() => {
        sandbox.restore();
    });

    it("test", () => {
        assert.isTrue(true);
    });
});
```

### UPDATE

Regression: https://github.com/sinonjs/sinon/pull/1419

* If I remove `configurable: true` then it works fine.

fatso83:
Where do you mean you set the configurable option? It's not in the code. Thanks for finding the regression btw

gyandeeps:
#1419 PR sets a `configurable: true` at  many places when using `Object.defineProperty`. Thats where If i remove that then it works fine..

fatso83:
👌. I guess we could wrap those defineProperty calls in a try/catch. Unless there is some way of knowing if they are configurable before attempting?

lucasfcosta:
@fatso83 Yup, there's an easy way to do it.

[...]

PR #1481, opened 2017-07-04
Avoid running test for #1456 on Safari

fatso83:
#### Purpose
This avoids [breaking the build](https://travis-ci.org/sinonjs/sinon/jobs/249565441) by avoiding running the breaking test on Safari. See background in #1462.

Safari throws an error when trying to configure properties on `window` that are not configurable. AFAIK this is not a problem Sinon can solve: if the prop cannot be stubbed, then we shouldn't suppress the error or avoid trying to stub it.

1. Check out this branch
2. `npm install`
3. Configure Sauce Labs:
```bash
export SAUCE_USERNAME=sinon-sauce-name
export SAUCE_ACCESS_KEY=abcd3-12345-...
```
4. `npm run test-cloud`

BTW, the code is not so nice ATM. I was just so annoyed by the errors from Travis, so just needed to put a fix out there ... Not totally sure where to put the runtime detection of how this breaks and I am not so sure about how I skip the test. Maybe a simple `return` would suffice?

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -7,7 +7,6 @@ var configureLogError = require("../../lib/sinon/util/core/log_error.js");
 var assert = referee.assert;
 var refute = referee.refute;
 
-
 describe("issues", function () {
     beforeEach(function () {
         this.sandbox = sinon.sandbox.create();
@@ -210,10 +209,10 @@ describe("issues", function () {
             var argsB = match(suffixB);
 
             var firstFake = readFile
-              .withArgs(argsA);
+                .withArgs(argsA);
 
             var secondFake = readFile
-              .withArgs(argsB);
+                .withArgs(argsB);
 
             assert(firstFake !== secondFake);
         });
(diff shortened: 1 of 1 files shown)


test/issues/issues-test.js
```diff
@@ -252,7 +251,20 @@ describe("issues", function () {
...
+        try {
+            Object.defineProperty(window, "innerHeight", { value: 10, configureable: true, writeable: true });
+            Object.defineProperty(window, "innerHeight", preDescriptor);
+        } catch (err) {
+            throwsOnUnconfigurableProperty = true;
+        }
+        /* eslint-enable no-restricted-syntax */
+
         describe("#1456", function () {
+            if (throwsOnUnconfigurableProperty) { return; }
```
fearphage: Any reason not to do this all inline instead of setting a single-use variable?

Also if you do the check in the `beforeEach`, you can call `this.skip();` to skip all the related tests.

fatso83: nope, it was a quick and dirty thing, and I wasn't aware of the this.skip thing.

mroderick: @fatso83 will you make a slow and clean version, or shall we merge this one?

fatso83: ah, sorry, I totally forgot. Was offline and on holiday for a week.

fatso83:
I think the issues have been addressed, @fearphage.

mroderick:
👍

Issue #1485, opened 2017-07-13
Documentation: useFakeTimers options


mroderick:
> the documentation for useFakeTimers says that you can list functions to fake but doesn't explain what that would mean if you do vs don't

Improvements should be made to https://github.com/sinonjs/sinon/blob/master/docs/release-source/release/fake-timers.md, and all the versions for 2.x.

elad-nach:
@mroderick I can take this up. I'd also want to hook up the useFakeTimers implementation to the `shouldAdvanceTime` feature I submitted in lolex. I could take of the docs in the process.

mroderick:
That would be lovely, thanks 🙏

PR #1488, opened 2017-07-15
Clarify behavior of useFakeTimers

VynceMontgomery:
I no longer use sinon, so I don't know if it still has the apparent bugs that cause me to look for this documentation, but here's the clarification I wanted at the time.

address #1485 

useFakeTimers() documentation is still unclear bout why one would ever want to fake some but not all functions, but now at least it's clear that you don't have to list every function to fake them all. 

2. `npm install`
3. <your-steps-here>

Read documentation. See if it is now clear and correct.


mroderick:
Thank you for the contribution 💯

PR #1489, opened 2017-07-15
Allow anonymous mock functions to be named

Rowno:
#### Purpose
Makes debugging anonymous mock functions a lot easier by allowing an optional method name to be specified. Since there's currently nothing in the `sandbox.verify()` error message/stacktrace that points to which mock is broken.

#### How to verify
1. Check out this branch.
2. Write a test with `sinon.mock().once().verify();`.
3. Write a test with `sinon.mock('functionName').once().verify();`.


mroderick:
Thank you 💯 

This has become `v2.4.0`

PR #1490, opened 2017-07-18
Fix #1487: incorrect withArgs().returnValue

### LINKED ISSUE #1487: Incorrect returnValue when using withArgs() and firstCall or lastCall
* Sinon version : 2.3.8 (likely since 2.3.6)
* Environment   : Windows 10, Node 8.1.4
* Other libraries you are using: sinon-chai

Related to: https://github.com/sinonjs/sinon/pull/1478 and https://github.com/sinonjs/sinon/issues/1476

It should not matter how the stub is queried, i.e., getCall(0), firstCall or lastCall should all return the same value if there is only one call using withArgs().
In version 2.3.5, this test succeeds, but when updating to any version above it (tested it with 2.3.8), the test will fail.

**What actually happens**
The test fails when using `withArgs()` together with `firstCall.returnValue` or `lastCall.returnValue`. But somehow, when using `getCalls(0).returnValue` or `returnValues[0]`, it does work.

**How to reproduce**
Given the following test case (sorry, it's in TypeScript).
The inject stub will return a different fake component function based on the arguments that are passed. The return value of the inject stub is then used for example to check the arguments passed etc.
```ts
import { SinonStub } from 'sinon';
import * as chai from 'chai';
import * as sinon from 'sinon';
import * as sinonChai from 'sinon-chai';

chai.use(sinonChai);
const { expect } = chai;

let testComponentA = (config: object) => ({ render: () => 'test a' });
let testComponentB = (config: object) => ({ render: () => 'test b' });

describe('Sinon firstCall, lastCall test', () => {
  let inject: SinonStub;

  beforeEach(() => {
    let fakeComponent = (variant: string) => {
      return sinon.stub().returns({
        render: sinon.stub().returns(`fake component ${variant}`)
      });
    };
    inject = sinon.stub().throws('Nothing set');
    inject.withArgs(testComponentA).returns(fakeComponent('a'));
    inject.withArgs(testComponentB).returns(fakeComponent('b'));
  });

  it('returnValues', () => {
    let config = { option: 'a' };
    let component = inject(testComponentA)(config);

    expect(inject).to.have.been.calledWith(testComponentA);
    expect(inject).to.not.have.been.calledWith(testComponentB);

    expect(component).to.respondTo('render');
    expect(component.render()).to.equal('fake component a');

    expect(inject.withArgs(testComponentA).returnValues[0]).to.have.been.calledWith(config); // Ok
    expect(inject.withArgs(testComponentA).getCall(0).returnValue).to.have.been.calledWith(config); // Ok

    expect(inject.withArgs(testComponentA).firstCall.returnValue).to.have.been.calledWith(config); // Fails
    expect(inject.withArgs(testComponentA).lastCall.returnValue).to.have.been.calledWith(config); // Fails
    // TypeError: undefined is not a spy or a call to a spy!
    // at assertCanWorkWith (node_modules\sinon-chai\lib\sinon-chai.js:53:19)
    // at Proxy.<anonymous> (node_modules\sinon-chai\lib\sinon-chai.js:98:13)
    // at Proxy.methodWrapper (node_modules\chai\lib\chai\utils\addMethod.js:57:25)
    // at Context.it (sinon-test.ts:39:79)
  });
});
```

takasmiley:
Fix issue #1487
The cause of this bug is that createCallProperties(matching) is never called to make returnValue and exception.

### DIFF
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -127,7 +127,7 @@ describe("stub", function () {
     describe("should work firstCall and lastCall", function () {
         var testComponentA = function () { return { render: function () { return "test a"; } }; };
         var testComponentB = function () { return { render: function () { return "test b"; } }; };
-        var inject;
+        var stub;
 
         beforeEach(function () {
             var fakeComponent = function (variant) {
@@ -135,26 +135,26 @@ describe("stub", function () {
                     render: createStub().returns("fake component " + variant)
                 });
             };
-            inject = createStub().throws("Nothing set");
-            inject.withArgs(testComponentA).returns(fakeComponent("a"));
-            inject.withArgs(testComponentB).returns(fakeComponent("b"));
+            stub = createStub().throws("Nothing set");
+            stub.withArgs(testComponentA).returns(fakeComponent("a"));
+            stub.withArgs(testComponentB).returns(fakeComponent("b"));
         });
 
         it("returnValues", function () {
             var config = { option: "a" };
-            var component = inject(testComponentA)(config);
+            var component = stub(testComponentA)(config);
 
-            assert.isTrue(inject.calledWith(testComponentA));
-            assert.isFalse(inject.calledWith(testComponentB));
+            assert.isTrue(stub.calledWith(testComponentA));
+            assert.isFalse(stub.calledWith(testComponentB));
 
             assert.isFunction(component.render);
             assert.equals(component.render(), "fake component a");
 
-            assert.isTrue(inject.withArgs(testComponentA).returnValues[0].calledWith(config));
-            assert.isTrue(inject.withArgs(testComponentA).getCall(0).returnValue.calledWith(config));
+            assert.isTrue(stub.withArgs(testComponentA).returnValues[0].calledWith(config));
+            assert.isTrue(stub.withArgs(testComponentA).getCall(0).returnValue.calledWith(config));
 
-            assert.isTrue(inject.withArgs(testComponentA).firstCall.returnValue.calledWith(config));
-            assert.isTrue(inject.withArgs(testComponentA).lastCall.returnValue.calledWith(config));
+            assert.isTrue(stub.withArgs(testComponentA).firstCall.returnValue.calledWith(config));
+            assert.isTrue(stub.withArgs(testComponentA).lastCall.returnValue.calledWith(config));
         });
     });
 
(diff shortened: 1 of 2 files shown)


test/stub-test.js
```diff
@@ -124,6 +124,39 @@ describe("stub", function () {
...
+            assert.isTrue(this.stub.calledWith(testComponentA));
+            assert.isFalse(this.stub.calledWith(testComponentB));
+
+            assert.isFunction(component.render);
+            assert.equals(component.render(), "fake component a");
+
+            assert.isTrue(this.stub.withArgs(testComponentA).returnValues[0].calledWith(config));
+            assert.isTrue(this.stub.withArgs(testComponentA).getCall(0).returnValue.calledWith(config));
+
+            assert.isTrue(this.stub.withArgs(testComponentA).firstCall.returnValue.calledWith(config));
```
fatso83: You could do the same for this call.

test/stub-test.js
```diff
@@ -124,6 +124,39 @@ describe("stub", function () {
...
+            var component = this.stub(testComponentA)(config);
+
+            assert.isTrue(this.stub.calledWith(testComponentA));
+            assert.isFalse(this.stub.calledWith(testComponentB));
+
+            assert.isFunction(component.render);
+            assert.equals(component.render(), "fake component a");
+
+            assert.isTrue(this.stub.withArgs(testComponentA).returnValues[0].calledWith(config));
+            assert.isTrue(this.stub.withArgs(testComponentA).getCall(0).returnValue.calledWith(config));
```
fatso83: Do note that having `calledWith` is an uneeded complexity that has little with the core of the original issue. Just assert equality between the returnValue.

test/stub-test.js
```diff
@@ -124,6 +124,39 @@ describe("stub", function () {
...
+
+            assert.isTrue(this.stub.withArgs(testComponentA).firstCall.returnValue.calledWith(config));
+            assert.isTrue(this.stub.withArgs(testComponentA).lastCall.returnValue.calledWith(config));
```
fatso83: This could have been a separate, second test to avoid confusion, where you might call the stub twice, just to be sure that `getCall(1).returnValue` and `lastCall.returnValue` returns the same.

brian-mann:
Really looking forward to having this land!

fatso83:
@brian-mann, if you check out the branch and fix the remaining issues I mentioned above you can get it out whenever you want. Just push it out and open another PR, leaving the original commit to recognize the original work. I'll see to it that it gets merged pronto.

brian-mann:
Okay @chrisbreiding is gonna hack on this.

fatso83:
This just got merged as part of #1500, so closing. Thanks a lot for the work, @takasmiley! We appreciate it greatly.

fatso83:
This will be out as soon as we find out what is causing the problems in #1498.

PR #1491, opened 2017-07-20
Update fake-xhr-and-server.md

escapedcat:
__before__:  
`assert(server.requests.length > 0)` 
__now__:  
`assert(this.server.requests.length > 0)`

Fix documentary to be consistent. 
I'm still not too familiar with sinon but looking at the code above this line I think it should be `this.server` as well. Sorry if I'm wrong here.


fatso83:
Thanks!

Issue #431, opened 2014-03-11
XHR.timeout is not implemented


kumar303:
In this example,

```
var xhr = new XMLHttpRequest();
xhr.timeout = 1;
xhr.open(...);
xhr.send(...);
```

the timeout property does not seem to be implemented by Sinon's fake XHR. https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#Properties

I was expecting it to trigger the `timeout` event. It would be nice if this were implemented or at least documented as such.

I'm using Sinon from node @1.9.0

marian-r:
No updates on this?

cjohansen:
PR welcome!

mroderick:
It's been more than 500 days since last activity on this. I am closing it. If anyone wants to create a PR that would be great

PR #1495, opened 2017-07-25
Issue 431

Munksey:
Attempt to resolve issue #431. This issue was closed, but has a `help-wanted` label on it. I have a bunch of uncommented tests in a few codebases that try to check that timeout is working as expected, but I can't simulate them passing because there is no implementation here. It seems that despite the issue being inactive, it's still unresolved. I'd like to submit this PR as a possible resolution.

I added a method to XMLHttpRequest called `triggerTimeout`. This can be used to simulate a timeout event. 

At the moment, I've changed `send` to call `setInterval` *ONLY* if there is a fake timer available. I've added some comments to the code, but any tick caused by the user should trigger the `setInterval` callback. According to the spec, timeout can be changed while the request is being fetched. So, we can't use `setTimeout` unless we add a `getter` and `setter` method for `timeout` and create more variables that are intended to be private.

I made sure that the request was not complete before triggering the callback. This should prevent fakeServerWithClocks odd (jQuery legacy) behavior from causing unnecessary triggering after calling `respond`.

I've added a number of tests and even refactored a few things. None of the existing tests break and I think I've covered a lot of the possible edge cases for timeout.

Some outstanding thoughts on my end:

1) I chose to only call `setInterval` with a fake timer because it would seem bad if the library actually ran timeouts in a test environment with the `FakeXMLHttpRequest`. It seems like you don't want your requests to actually timeout in a test environment. I did not consult anyone in the issues thread, though. Perhaps I should do this?

2) I haven't written any tests for when `supportsTimeout` is false. It wouldn't be too difficult to do, though. I'm also not sure if I should mock the test so it runs as though supportsTimeout is true and supportsTimeout is false. I noticed a lot of conditional tests, so I followed the pattern. Let me know if you want more.

3) PhantomJS does not seem to support timeouts in their XMLHttpRequest implementation. Since a lot of people run tests in that browser, I almost wonder if it makes sense to assume that `supportsTimeout` works as expected (or allow a way to override it). Relevant issue: https://github.com/ariya/phantomjs/issues/12837.

4) Let me know if you want me to annotate the PR with github comments.

5) Let me know if you want me to squash commits.

### DIFF
--- a/lib/sinon/util/fake_xml_http_request.js
+++ b/lib/sinon/util/fake_xml_http_request.js
@@ -485,6 +485,32 @@ extend(FakeXMLHttpRequest.prototype, sinonEvent.EventTarget, {
             this.onSend(this);
         }
 
+        // Only listen if setInterval is a stubbed clock.
+        if (typeof setInterval.clock === "object" && typeof Date.clock === "object") {
+            var initiatedTime = Date.now();
+            var self = this;
+
+            // Listen to any possible tick by fake timers and check to see if timeout has
+            // been exceeded. It's important to note that timeout can be changed while a request
+            // is in flight, so we must check anytime the end user forces a clock tick to make
+            // sure timeout hasn't changed.
+            // https://xhr.spec.whatwg.org/#dfnReturnLink-2
+            var clearIntervalId = setInterval(function () {
+                // Check if the readyState has been reset or is done. If this is the case, there
+                // should be no timeout. This will also prevent aborted requests and
+                // fakeServerWithClock from triggering unnecessary responses.
+                if (self.readyState === FakeXMLHttpRequest.UNSENT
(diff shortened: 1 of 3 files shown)


fatso83:
Thank you for doing this! As you can see from the last comments in the discussion of #431, the issue was closed due to inactivity, not due to being implemented. Still, might I suggest you send the changes to [the nise project](https://github.com/sinonjs/nise), as #1461 is essentially removing all traces of XHR out of Sinon.

As to your questions, I don't have much to contribute, other than that we will probably start using Chrome headless as soon as someone sets aside time for [Mochify](https://github.com/mantoni/mochify.js) to start using it.

Munksey:
Thank you for bringing this to my attention! I will close this PR and open an issue on nise, to see if the behavior is desired.

PR #1496, opened 2017-07-25
Issue #1485: Update lolex version and update fakeTimers documentation

### LINKED ISSUE #1485: Documentation: useFakeTimers options

elad-nach:
This PR is to take care of updating to the latest lolex version, updating the fakeTimers method signatures and related tests.
I've also started updating the documentation but wanted to consult beforehand.

The question that arises is how to expose the lolex configuration (if at all).
Right now the implementation is straightforward - just sends a copy of the `config` object that was sent to `useFakeTimers()`. the copy was made because otherwise lolex alters the config object with the default values, resulting in tests breaking due to altered values showing up in the spies used in the tests.

However I believe that we should just allow users to send a lolex config object down the pipe and not just whitelist certain features. 

This is up to you however, just let me know what's your take on this - if we should whitelist certain features of lolex or we should just forward the object to the install method.

In the case of useFakeTimers I think that we should still support the possibility to `useFakeTimers(<Date|Epoch>)` and just break the tests that use other features from lolex e.g., `config.toFake` with a proper exception.

I will also edit the documentation accordingly.

fixes #1485 
fixes expressjs/express#2447

### DIFF
--- a/docs/release-source/release/fake-timers.md
+++ b/docs/release-source/release/fake-timers.md
@@ -37,40 +37,101 @@ set of features (Sinon uses it under the hood) and was previously extracted from
 }
 ```
 
-
 ## Fake timers API
 
 
 #### `var clock = sinon.useFakeTimers();`
 
-Causes Sinon to replace the global `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`, `setImmediate`, `clearImmediate` and `Date` with a custom implementation which is bound to the returned `clock` object.
+Causes Sinon to replace the global `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`, `setImmediate`, `clearImmediate`, `process.hrtime`, `performance.now`(when available) and `Date` with a custom implementation which is bound to the returned `clock` object.
 
-Starts the clock at the UNIX epoch (timestamp of 0).
+Starts the clock at the UNIX epoch (timestamp of `0`).
 
 
 #### `var clock = sinon.useFakeTimers(now);`
 
-As above, but rather than starting the clock with a timestamp of 0, start at the provided timestamp.
+As above, but rather than starting the clock with a timestamp of 0, start at the provided timestamp `now`.
 
 *Since `sinon@2.0.0`*
 
 You can also pass in a Date object, and its `getTime()` will be used for the starting timestamp.
 
-#### `var clock = sinon.useFakeTimers([now, ]prop1, prop2, ...);`
+#### `var clock = sinon.useFakeTimers(config);`
 
-Sets the clock start timestamp and names functions to fake. If the first argument is not numeric, it sets the clock to 0 and treats all arguments as names of functions to fake.
+As above, but allows further configuration options, some of which are:
 
-Possible functions are `setTimeout`, `clearTimeout`, `setInterval`, `clearInterval`, `setImmediate`, `clearImmediate` and `Date`.  Any functions not listed continue to use the original version of the function, including the actual time for the timestamp. This can have surprising results and should be done with care.
+- `config.now` - *Number/Date* - installs lolex with the specified unix epoch (default: 0)
+- `config.toFake` - *String[ ]* - an array with explicit function names to fake. By default lolex will automatically fake all methods *except* `process.nextTick`. You could, however, still fake `nextTick` by providing it explicitly
+- `config.shouldAdvanceTime` - *Boolean* - tells lolex to increment mocked time automatically based on the real system time shift (default: false)
 
-Note that if no functions are listed, the default behavior is to replace all eligible functions.
+Please visit the `lolex.install` [documentation](https://github.com/sinonjs/lolex#var-clock--lolexinstallconfig) for the full feature set.
+
+**Important note:** when faking `nextTick`, normal calls to `process.nextTick()` would not execute automatically as they would during normal event-loop phases. You would have to call either `clock.next()`, `clock.tick()`, `clock.runAll()` or `clock.runToLast()` (see example below). Please refer to the [lolex](https://github.com/sinonjs/lolex) documentation for more information.
+
+#### Examples:
+
+Installs fake timers at January 1st 2017 and fakes `setTimeout` and `process.nextTick` only:
+
+```javascript
+var clock = sinon.useFakeTimers({
+                now: 1483228800000, 
+                toFake: ["setTimeout", "nextTick"] 
+            }); 
+
+var called = false;
+
+process.nextTick(function () {
+    called = true;
+});
+
+clock.runAll(); //forces nextTick calls to flush synchronously
+assert(called); //true
+```
+
+Install at the same date, advancing the fake time automatically (default is every `20ms`), causing timers to be fired automatically without the need to `tick()` the clock:
+
+```js
+var clock = sinon.useFakeTimers({
+                now: 1483228800000, 
+                shouldAdvanceTime: true 
+            });
+
+setImmediate(function () {
+    console.log('tick'); //will print after 20ms
+});
+
+setTimeout(function () {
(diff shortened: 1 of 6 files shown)


lib/sinon/util/fake_timers.js
```diff
@@ -2,17 +2,34 @@
...
 var llx = require("lolex");
 
-exports.useFakeTimers = function () {
-    var now;
-    var methods = Array.prototype.slice.call(arguments);
+/**
+ *
+ * @param config {number|Date|Object} the unix epoch value to install with (default 0) or
+ */
+exports.useFakeTimers = function (args) {
```
mroderick: To me, this method is a bit difficult to read, because

* `for` loop for creating a new `Object` with the own properties from `args`. This can be done with [`Object.create`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) in a single statement
* nested conditionals make everything more difficult to read
* `typeof null` returns `"object"`, which would cause errors if you call the method with `null`.

I have made an attempt at addressing those issues, I think this is functionally equivalent

```javascript
exports.useFakeTimers = function (args) {
    if (typeof args === "string" || Array.isArray(args) && args.length > 1) {
        throw new TypeError("useFakeTimers expected 1 argument." +
            " see https://github.com/sinonjs/sinon");
    }

    var config;

    if (!Boolean(args)) {
        config = {
            now: 0
        };        
    }
    
    if (!config && typeof args === "number" || args instanceof Date) {
        config = {
            now: args
        };
    }

    if (!config && typeof args === "object") {
        config = Object.create(null, args);
    }

    if (!config) {
        throw new TypeError("useFakeTimers called with " + String(args) +
            " but expected epoch or config object. see https://github.com/sinonjs/sinon");
    }

    var clock = llx.install(config);
    clock.restore = clock.uninstall;
    return clock;
};
```

Admittedly, it has more lines, but I think it reads a lot easier with the reduced indentation (and no use of `else`).

What do you think?

elad-nach: I actually like `if ... else if ... else` because when you read it you know that once a certain condition has been met the execution breaks out of the block, not allowing for other blocks in the condition to evaluate and execute.

In the example you wrote, on first sight the second and third conditions might look also deceiving since: 
```javascript
> args = new Date()
2017-07-26T08:54:10.426Z
> args instanceof Date
true
> typeof args === "object"
true
```


The `Object.create()` doesn't fly and doesn't deep copy the array. sticking with the `for` loop for now.

On REPL do:
```javascript
let a= {a: ['a','b','c']}
undefined
> a
{ a: [ 'a', 'b', 'c' ] }
> let b = Object.create(null,aa)
undefined
> b
{}
```

I'm suggesting a shorter version of the `useFakeTimers` function, cleaner and shorter.
```javascript
exports.useFakeTimers = function (args) {
    var config = {};

    if (typeof args === "undefined" || args === null) {
        config.now = 0;
    } else if (typeof args === "number" || args instanceof Date) {
        config.now = args;
    } else if (args !== null && typeof args === "object") {
        var keys = Object.keys(args);
        for (var i = 0; i < keys.length; i++) {
            if (args.hasOwnProperty(keys[i])) {
                config[keys[i]] = args[keys[i]];
            }
        }
    } else {
        throw new TypeError("useFakeTimers expected epoch or config object." +
            " see https://github.com/sinonjs/sinon");
    }

    var clock = llx.install(config);
    clock.restore = clock.uninstall;
    return clock;
};
```

What do you think?

mroderick: 👍

mroderick:
Would you mind doing a rebase to remove the extra commits? (like there should not be any already existing or merge commits in this branch)

lib/sinon/util/fake_timers.js
elad-nach: So in regards to the config object - will update the documentation.

Next up - the sinonjs website - if I'd like to keep the [table format from lolex](https://github.com/sinonjs/lolex) - would this work with the current page margins of the documentation section on [sinon](http://sinonjs.org/releases/v2.4.0/fake-timers/)? 
It looks about a hundred pixels narrower. Or maybe I should find some other format to display.

Thoughts?

mroderick:
Almost there!

Here's how to get rid of the first commit:

```shell
# assuming sinonjs/sinon remote is named upstream
git fetch upstream
git rebase upstream/master
```

Then set the commit message of the commit to something that reads better when you can only see the first line

```shell
git commit --amend
```

And then replace the text with

```text
Upgrade lolex to 2.1.2

* updated useFakeTimers function signature
* updated documentation
```

elad-nach:
Done. but please don't merge this just yet. I would like to make a few more changes

elad-nach:
@mroderick I think I'm done more or less. could you have a look?

docs/release-source/release/fake-timers.md
```diff
@@ -43,34 +43,95 @@ set of features (Sinon uses it under the hood) and was previously extracted from
...
+
+Install at the same date, advancing the fake time automatically (default is every 20ms), causing timers to be fired automatically without the need to `tick()` the clock:
+
+```js
+var clock = sinon.useFakeTimers({
+                now: 1483228800000, 
+                shouldAdvanceTime: true 
+            });
+
+setImmediate(() => {
```
mroderick: Minor: since we're going for **wide** adoption, targeting ES5.1, I would prefer it if the examples were also ES5.1.

lib/sinon/util/fake_timers.js
```diff
@@ -2,17 +2,30 @@
...
+    } else if (args !== null && typeof args === "object" && arguments.length === 1) {
+        var keys = Object.keys(args);
+        for (var i = 0; i < keys.length; i++) {
+            if (args.hasOwnProperty(keys[i])) {
+                config[keys[i]] = args[keys[i]];
+            }
+        }
     } else {
-        now = methods.shift();
+        throw new TypeError("useFakeTimers expected epoch or config object." +
```
mroderick: The error message doesn't need to be two lines, and "see" should start with capitalised "S".

test/util/fake-timers-test.js
```diff
@@ -1066,11 +1066,79 @@ describe("fakeTimers.clock", function () {
...
+            this.clock.setImmediate(function () {
+                this.clock.restore();
+                done();
+            }.bind(this));
+        });
+
+        it("throws on old useFakeTimers signatures", function () {
+            assert.exception(function () {
+                fakeTimers.useFakeTimers("setImmediate");
+            }, { name: "TypeError" });
```
mroderick: [`assert.exception`](http://docs.busterjs.org/en/latest/modules/referee/#exception) also supports matching on the message, which would make for sturdier tests

lib/sinon/util/fake_timers.js
mroderick: The sinonjs website uses GitHub Flavoured Markdown, so anything you can do on GitHub, you can do in the docs :)

elad-nach:
Done. I'm not sure how the table will land in the sinon pages after the release-source will be concatenated. is there a css that covers it already?

mroderick:
I don't know. You can try running the site locally. There are directions in one of the README files

elad-nach:
So... as suspected the table doesn't show up properly.
Did small modifications - looks good now.

fatso83:
Merged after all issues have been resolved.

mroderick:
Nice work @elad-nach! Thank you 💯

elad-nach:
@mroderick my pleasure 🙏

Issue #992, opened 2016-02-24
Sinon mock expectations do not remember multiple withArgs() or withExactArgs() calls.


TonyApuzzo:
SInon 1.17.3 doesn't seem to work with mocks as expected.  I would think that I can treat mocks like stubs, but the with*Args() methods seem to only remember the last invocation, it doesn't create an internal map of argument to return value like you get with stubs.

``` javascript
describe('Sinon Verification', function () {
  it('Should work with mocks', function () {
    var u = { inc : function(x){} };
    var m = sinon.mock(u);
    var e = m.expects("inc");
    e.withExactArgs(0).returns(1);
    e.withExactArgs(1).returns(2);
    e.withExactArgs(2).returns(3);

    u.inc(2).must.eql(3); // OK
    u.inc(1).must.eql(2); // Fails expectation
    u.inc(0).must.eql(1); // Fails expectation
  });
});
```

roderickhsiao:
having the same issue

yaronyg:
Sounds like #984

hrimhari:
@yaronyg not exactly. There were discussions about multiple withArgs() being chainable or not like #984 is trying to do. But the problem reported here is that mocks just can't have multiple withArgs(), regardless of their declaration being in the same chain or not.

mroderick:
It is correct that mocks only keep the last value of arguments to `withArgs`.

This has been the case since 84dbb22ae72a94380b846ee58a7d2b0238c5ea0d, which was committed on Apr 18, 2010.

```shell
git log -S "this.expectedArguments = slice.call(arguments);"
```

The same is true for `withExactArgs`, which calls `withArgs`.

I have created https://github.com/sinonjs/sinon/issues/1432 to document this issue

hrimhari:
@mroderick how would you test a loop over different objects if your mock can only accept the same parameters?

mroderick:
> @mroderick how would you test a loop over different objects if your mock can only accept the same parameters?

With `sinon.mock`, you would have to create a new one for each member in your collection. Using `sinon.stub`, you can inspect each call.

mroderick:
Issue #1193 also has a good explanation of this deficiency.

mroderick:
Fixed by #1432

PR #1501, opened 2017-07-28
Fix #1432: add details around expectations.withArgs behavior to docs

### LINKED ISSUE #1432: Explain that mock.withArgs only uses arguments from last call
The documentation should be updated to show that only the arguments from the last call are being used.

https://github.com/sinonjs/sinon/blob/master/lib/sinon/mock-expectation.js#L227-L236

This will help us close issues like #992

dcolucci:
#### Purpose
Fix issue #1432 - adds details to documentation of `expectations` methods `withArgs` and `withExactArgs` to prevent confusion, such as #992

#### How to verify
Probably not necessary to run the docs application - please review the copy and confirm whether it is clear, informative, etc, and whether I've touched the correct / necessary files.

### DIFF
--- a/docs/release-source/release/mocks.md
+++ b/docs/release-source/release/mocks.md
@@ -146,10 +146,14 @@ Expect the method to be called exactly `number` times.
 
 Expect the method to be called with the provided arguments and possibly others.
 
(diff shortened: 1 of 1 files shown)


dcolucci:
Thanks @fatso83, my mistake - updated

mroderick:
Thank you!

Issue #1498, opened 2017-07-27
master build failing


mroderick:
We have merged #1461, which resulted in an interesting failure where Chrome would fail the tests. I tried to fix that with `proxyquire-universal`.

Now I have run into a problem with the [webworker test failing](https://travis-ci.org/sinonjs/sinon/jobs/258052323).

The setup to get the web worker test going is not exactly my strongest suit.

Ping @mantoni @fatso83: do you think you could take a look at this and see if you can figure out what the problem might be?

fatso83:
I can have a look at it. Is the proxyquire change checked in somewhere?

mroderick:
> Is the proxyquire change checked in somewhere?

It's in `master`, because only `master` is built using SauceLabs. Perhaps we should try turning it on for PRs as well?

https://github.com/sinonjs/sinon/blob/master/.travis.yml#L30

fatso83:
AFAI remember, we couldn't turn it on for pull requests due to dangers in exposing the sauce lab credentials. As anyone can push a pull request where they can `console.log(process.env.SAUCE_LABS_SECRET)` we could risk loosing the key, which we have assured Sauce Labs we would not.

mroderick:
Ok. Then let's just stick to master

fatso83:
OK, I have played with this a while - without fixing it, and thought I could report back some findings to you, @mroderick.

As for the WebWorker error, this is basically about the new build not creating a `sinon` object on the global scope. You can see this from the webworker script that is being loaded (`test/webworker/webworker-script.js`):

```javascript
    importScripts("../../pkg/sinon.js");

    var stub = sinon.stub().returns("worker response");
```

This will fail if `sinon` is undefined. To see that this has nothing to do with WebWorkers, I created this CLI test:

```
npm i && ./build.js \
&& echo 'console.log("sinon:",window.sinon)' >> append.js \
&& cat pkg/sinon.js append.js \
      | $(npm bin)/phantomic --phantomjs node_modules/phantomjs-prebuilt/bin/phantomjs 
```

For `40f1599` (before the proxyquire changes) it outputs:
```
sinon: [object Object]
sinon: [object Object]
sinon: [object Object]
sinon: [object Object]
```

No idea why it happens four times, but it shows that `sinon` is defined.

After the `proxyquire` changes it outputs:

```
sinon: undefined
sinon: undefined
sinon: undefined
sinon: undefined
sinon: undefined
```

[...]

PR #1503, opened 2017-07-28
Add method to set custom formatter

chrisbreiding:

Allows overriding the default formatter (`formatio`) with a custom one.

In [Cypress](https://www.cypress.io/), we prefer to use our own formatting for printing out values instead of what `formatio` provides. With sinon 1.x, we overrode `sinon.format` to achieve this. 

Although `sinon.format` is still exposed in 2.x, overriding it doesn't work anymore because the modules within sinon use `util/core/format` directly. 

Overriding `sinon.format` was always a bit of hack anyway. This provides a better way to achieve setting a custom formatter.

#### Solution  - optional

This solution adds `sinon.setFormatter`, which is a method that takes a formatter function. The formatter function's signature is the same as `formatio.ascii`. It's called with a value and should return a formatted version of that value.

1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon.js
+++ b/lib/sinon.js
@@ -54,3 +54,6 @@ var behavior = require("./sinon/behavior");
 exports.addBehavior = function (name, fn) {
     behavior.addBehavior(exports.stub, name, fn);
 };
+
+var format = require("./sinon/util/core/format");
+exports.setFormatter = format.setFormatter;
--- a/lib/sinon/util/core/format.js
+++ b/lib/sinon/util/core/format.js
@@ -7,6 +7,22 @@ var formatter = formatio.configure({
     limitChildrenCount: 250
 });
 
-module.exports = function format() {
+var customFormatter;
+
+function format() {
+    if (customFormatter) {
+        return customFormatter.apply(null, arguments);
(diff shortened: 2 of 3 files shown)


brian-mann:
I know we opened the PR but...

Shouldn't `setFormatter` be exposed globally on `sinon`.

As it stands this would require the consumer to understand how `formatter` is currently stored / structured within `sinon`.

I'd prefer just something like

```js
sinon.setFormatter(fn)
```

chrisbreiding:
@brian-mann It is exposed on the `sinon` global 😅 https://github.com/cypress-io/sinon/blob/fb1a01feb5d5e90b894699db46685e4139958fcd/lib/sinon.js#L59

brian-mann:
Oh, derp.

mroderick:
I like it!

fatso83:
This is all good to go, but I would prefer not merging anything in until we have fixed #1498, so that we avoid introducing new regressions without knowing.

fatso83:
All builds are now passing, so gladly merging this. Thanks!

Issue #1502, opened 2017-07-28
Documentation: migration guide from 2.x to 3.0


elad-nach:
If I understand correctly the next release will be a major release.
Therefore we should add a new migration guide section with the relevant information.

I can take this upon myself, but would need some further information on what has to be included.

fatso83:
See the commit id for the last release, and then look through the list of merged pull requests to find changes. I think that is what we basically do :)

elad-nach:
I was thinking that :) but actually to be explicit I just wanted to know from @mroderick if he has any specific comments about the `nise` import and if it breaks anything in particular that I should note for (there's 5000 deletions on that PR 😂).

elad-nach:
I was also thinking - maybe we could include the extracted modules as git submodules on sinon, then if we would refactor the different sections in the submodule documentation we could actually have sinon to generate the sections on the documentation from the submodules.

It would just mean we would need to refactor the different parts of the documentation on the submodules then concat them somehow with an npm job.

Thoughts?

elad-nach:
Another addition:
I'm just going through the migration guide from 1.x to 2.0, and there's a list of helpers that should be removed at 3.0 but weren't. The marked items were indeed removed but on a first look the rest weren't:
* `sinon.calledInOrder`
* `sinon.create`
* `sinon.deepEqual`
* `sinon.format`
* `sinon.functionName`
* `sinon.functionToString`
* `sinon.getConfig`
* `sinon.getPropertyDescriptor`
* `sinon.objectKeys`
* `sinon.orderByFirstCall`
* `sinon.restore`
* `sinon.timesInWorlds`
* `sinon.valueToString`
* `sinon.walk`
* `sinon.wrapMethod`
- [x] `sinon.Event`
- [x] `sinon.CustomEvent`
- [x] `sinon.EventTarget`
- [x] `sinon.ProgressEvent`
* `sinon.typeOf`
* `sinon.extend`

Proceeding with removal of the rest if there's no objection.

fatso83:
@elad-nach  Are you sure there are references all over the codebase, or just in `_docs`? A quick test for `calledInOrder`, `create`, and `extend` showed no occurences:

`ag sinon.walk lib/ test/`

elad-nach:
in `assert.js`:
```
var calledInOrder = require("./util/core/called-in-order");
var orderByFirstCall = require("./util/core/order-by-first-call");
var timesInWords = require("./util/core/times-in-words");
[...]

PR #1504, opened 2017-08-01
Fix build and test bugs in #1498 and #1479

fatso83:
#### Purpose 
Should fix the errors we are seeing in #1498, introduced by a proxyquire config bug, as well as a case of a missing file causing trouble for the builds in #1479.

See the discussion in #1498 and #1479.

1. Check out this branch
2. `npm install`
3. Set the sauce lab credentials (or run [selenium-standalone](https://www.npmjs.com/package/selenium-standalone) locally to test with local browsers)
4. `npm test`
5. `npm run test-cloud`

Those steps cover #1498. 

To test that #1479 failed before, but not after, do this for on `master` and on this branch:
1. `rm -r pkg`
2. `npm test`


fatso83:
The commit messages should also have quite a bit of info on the details.

fatso83:
P.S. One thing I am not totally not sure about, is where to put the `build` step. Maybe you have some input on that? 

Right now `build` is called from `test`, `prepublishOnly` and the Travis script for Node v6. I was pondering about stuffing it in the `test-webworker` script, but that line was already far too long.

test/call-test.js
```diff
@@ -852,7 +853,10 @@ describe("sinonSpy.call", function () {
         });
 
         // https://github.com/sinonjs/sinon/issues/1066
+        /* eslint-disable consistent-return */
         it("does not throw when the call stack is empty", function (done) {
+            if (!supportPromise) { return this.skip(); }
```
fearphage: You don't need to return here. To my knowledge, `this.skip()` throws a special error (`SkipError`?) that mocha is expecting so execution should stop immediately. Then you can remove the eslint toggle.

MINOR: You can inline the conditional since it's only used/checked once.

test/sandbox-test.js
```diff
@@ -167,6 +169,7 @@ describe("sinonSandbox", function () {
 
         // eslint-disable-next-line mocha/no-identical-title
```
fearphage: Why do they have identical titles? That's weird. I realize it isn't your change though.

package.json
```diff
@@ -16,14 +16,15 @@
   "scripts": {
     "test-node": "mocha --recursive -R dot test/",
     "test-dev": "npm run test-node -- --watch -R min",
-    "test-headless": "mochify --recursive -R dot --grep WebWorker --invert test/",
+    "test-headless": "mochify --recursive -R dot --grep WebWorker --invert --plugin [proxyquire-universal] test/",
     "test-coverage": "mochify --recursive -R dot --grep WebWorker --invert --plugin [ mochify-istanbul --exclude '**/+(test|node_modules)/**/*' --report text --report lcovonly --dir ./coverage ] test/",
     "test-cloud": "npm run test-headless -- --wd",
     "test-webworker": "browserify --no-commondir --full-paths -p [ mocaccino -R spec --color ] test/webworker/webworker-support-assessment.js | phantomic --web-security false",
-    "test": "npm run test-node && npm run test-headless && npm run test-webworker",
+    "test": "run-p test-node test-headless && run-s build test-webworker",
```
fearphage: Is the output corrupted when running in parallel?

package.json
```diff
@@ -16,14 +16,15 @@
...
     "lint": "eslint .",
     "precommit": "lint-staged",
-    "prepublish": "rimraf pkg && ./build.js",
+    "build": "rimraf pkg && ./build.js",
```
fearphage: Use `prebuild` for the cleanup.

package.json
```diff
@@ -16,14 +16,15 @@
   "scripts": {
     "test-node": "mocha --recursive -R dot test/",
     "test-dev": "npm run test-node -- --watch -R min",
-    "test-headless": "mochify --recursive -R dot --grep WebWorker --invert test/",
+    "test-headless": "mochify --recursive -R dot --grep WebWorker --invert --plugin [proxyquire-universal] test/",
     "test-coverage": "mochify --recursive -R dot --grep WebWorker --invert --plugin [ mochify-istanbul --exclude '**/+(test|node_modules)/**/*' --report text --report lcovonly --dir ./coverage ] test/",
     "test-cloud": "npm run test-headless -- --wd",
     "test-webworker": "browserify --no-commondir --full-paths -p [ mocaccino -R spec --color ] test/webworker/webworker-support-assessment.js | phantomic --web-security false",
```
fearphage: Instead of changing the externals, why don't you make `build` a requirement for `test-webworker`?

`pretest-webworker: 'npm run build',`

This change would theoretically allow you to revert every instance where you used `run-s` (`travis.yml` and `package.json`).

package.json
```diff
@@ -16,14 +16,15 @@
   "scripts": {
     "test-node": "mocha --recursive -R dot test/",
     "test-dev": "npm run test-node -- --watch -R min",
-    "test-headless": "mochify --recursive -R dot --grep WebWorker --invert test/",
+    "test-headless": "mochify --recursive -R dot --grep WebWorker --invert --plugin [proxyquire-universal] test/",
```
fearphage: Spacing is off (different) from all the others below.

package.json
fatso83: Wasn't aware that adding `pre` was a general pattern outside of the standard lifecycle methods (publish, install, ...). You learn something every day :smile: Good improvement!

Having it as a dependency solves the grievances I had otherwise.

test/sandbox-test.js
fatso83: I saw that too - I couldn't tell immediately, and decided not to tackle it now.

package.json
fatso83: Could absolutely have been an issue, but the startup time of the phantom stuff actually makes it sequential by chance. You are probably right that it doesn't make sense to save 1.5 seconds in risk of having garbled output. I usually parellize tasks when having interactive/watched builds, where I don't really care all that much.

mroderick:
This looks great! I don't have anything to add other than what was already pointed out by @fearphage

fatso83:
Thanks for the review, @fearphage. Good tips on the lifecycle events. Regarding the merits of `npm-run-all` I just usually end up using it to improve readability / avoid long one-liners.

Typically:
`run-s test-node test-headless build test-webworker`
vs
`npm run test-node && npm run  test-headless  && npm run build  && npm run test-webworker`

The parallel stuff is more of a side-note. I can remove it, no problem, but I think it has some (minor) merit, just because we can't work much with readability when being restricted to oneliners.

fearphage:
I'm not against it. Just making sure it's worthwhile. If it simplifies things, then I'm all for it. 

Questions:

* How does it handle failure? 
* Can you tell what fails? For instance with `npm-s A B C D`, can you tell C failed?
* Do you still get the exit code as expected?

fatso83:
@fearphage This is an example run where to scripts simply do `exit 0` and `exit 100`:
```
$ $(npm bin)/run-s a b

> sinon@2.4.1 a /home/developer/dev/sinon
> exit 0


> sinon@2.4.1 b /home/developer/dev/sinon
> exit 100

npm ERR! code ELIFECYCLE
npm ERR! errno 100
npm ERR! sinon@2.4.1 b: `exit 100`
npm ERR! Exit status 100
npm ERR! 
npm ERR! Failed at the sinon@2.4.1 b script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/developer/.npm/_logs/2017-08-01T16_06_43_505Z-debug.log
ERROR: "b" exited with 100.

```
Doing `$ echo $?` return 1.

So yes, you can tell what failed from the output (`Failed at the sinon@2.4.1 b script`) and what the exit status was, but as a whole the exit status is simply 1 from the tool if any of the tasks fail.

fearphage:
> It will fail on the first task to fail - unless configured to do otherwise

Sounds legit to me. :+1:

fatso83:
As I have addressed all the issues, I'll merge this.

mroderick:
Well done!

fatso83:
Btw, this should pave the way for releasing 3.0; not sure if it hinges on #1502

PR #1511, opened 2017-08-03
Add support for passing a function to stub.throws(...)

### LINKED ISSUE #1507: Stack trace for throws(new Error(...)) starts at constructor invocation
When I pass an error object to `stub.throws(...)` I noticed that it shows the stack trace of the constructor in mocha and Chrome, not sure yet about Firefox. It would be nice to have the stack trace start from the point that the exception is thrown so I could identify what pieces of code interacted with the stubbed function. I was going to file a PR to attach a new stack trace to the exception but I noticed that there's a comment warning against interacting with the stack trace [here](https://github.com/sinonjs/sinon/blob/master/lib/sinon/spy.js#L210). Can someone give me guidance on whether replacing the stack trace is desirable, or if there's a workaround from the end user's perspective I should use instead? Thanks

noisecapella:
Fixes #1507 
This PR adds functionality to allow `stub.throws(...)` to accept a function as an argument. This function is invoked at the point that the exception is thrown which allows the end user to have a more descriptive stack trace for the error. This also delays instantiating the `Error` until it is thrown (except when the user has already instantiated it) to create a more detailed stack trace.

At least in node, the stack trace for an `Error` is created when the object is instantiated rather than when it's thrown ([from here](https://nodejs.org/api/errors.html#errors_class_error)). The stack trace is sometimes useful in figuring out what pieces of code are involved with an error. Currently errors used in `throws(...)` are instantiated by the caller or soon after in the setup of the stub which results in a stack trace showing where the setup occurred rather than where the problem occurred.

#### Solution  - optional
In cases where the exception is not passed into `throws(...)`, a function which will instantiate the exception is used internally instead and assigned to a new field `exceptionCreator`. When the exception is thrown `exceptionCreator` is instantiated and the result is assigned to `exception` to allow inspection via `toString` or other means.

To see the difference in stack trace see the test I included in the linked issue.

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -271,6 +271,11 @@ Causes the stub to throw an exception of the provided type.
 Causes the stub to throw the provided exception object.
 
 
+#### `stub.throws(function() { return new Error(); });`
+
+Causes the stub to throw the exception returned by the function.
+
+
 #### `stub.rejects();`
 
 Causes the stub to return a Promise which rejects with an exception (`Error`).
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -127,6 +127,10 @@ var proto = {
 
         if (this.exception) {
             throw this.exception;
+        } else if (this.exceptionCreator) {
+            this.exception = this.exceptionCreator();
+            this.exceptionCreator = undefined;
+            throw this.exception;
         } else if (typeof this.returnArgAt === "number") {
             return args[this.returnArgAt];
         } else if (this.returnThis) {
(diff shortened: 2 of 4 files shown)


fatso83:
Looks good, but need to test it to get a feel for it :-)

mantoni:
Nice work! I'm missing test cases verifying that the error object is now created lazily for all three cases.

noisecapella:
I added some tests, are those assertions what you were thinking of?

noisecapella:
I see, I'll make the change

mantoni:
Great tests! Thank you very much 👍

fatso83:
This introduced regression #1526. Mind taking a look at it, @noisecapella?

noisecapella:
Yes I'll take a look

Issue #1514, opened 2017-08-04
Rename factory functions


mantoni:
As a follow up on #1000, what remains are the factory function renamings:

- `sinon.sandbox.create()` -> `sinon.createSandbox()`
- `sinon.fakeServer.create()` -> `sinon.createFakeServer()`

Do we want to introduce the new ones and deprecate the old ones in 4.0, and then remove the old ones in 5.0? I think this procedure worked well for the 1.x -> 3.x API transition.

fatso83:
I like this strategy. Works, so why not continue on that way.

mroderick:
Since we're not changing any existing behaviour, only expanding on the api (and not involving separate modules), I think we can get away with creating the new factory methods and adding deprecations to the old ones in a `MINOR` release. Then we can remove the old factory methods in the next `MAJOR` release.

mantoni:
@mroderick Also works for me. It's just that upgrading to `^3.0.0` might have caused some deprecation warnings which the mindful developer might have fixed, and then a minor update causes more deprecation warnings. I'm personally fine with your suggestion, but it might be annoying for some people. We can add the new APIs in a minor anyway. The question is just about when to introduce the deprecation warning.

mroderick:
Fair point @mantoni

mroderick:
Let's add the new API and update the documentation first

PR #1515, opened 2017-08-06
Add factory functions for sandbox and fake server

mantoni:
As discussed in #1514, this adds the following methods as aliases:

- `createSandbox` = `sandbox.create`
- `createFakeServer` = `fakeServer.create`
- `createFakeServerWithClock` = `fakeServerWithClock.create`

Also adjusted the docs to show the new functions. Is it okay to suggest the new functions only? Or do we want to keep API docs for the "old" functions?

Did I miss anything else?

### DIFF
--- a/test/sinon-test.js
+++ b/test/sinon-test.js
@@ -10,34 +10,61 @@ if (!hasPromise) {
 
 describe("sinon module", function () {
     var sinon,
+        fakeSandbox,
         fakeNise;
 
     beforeEach(function () {
 
         fakeNise = {
-            fakeServer: "47c86a4c-6b48-4748-bb8c-d853f999720c",
-            fakeServerWithClock: "e69974f8-4568-48d1-a5e9-2b511a59c14b",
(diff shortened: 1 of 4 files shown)


test/sinon-test.js
```diff
@@ -10,34 +10,61 @@ if (!hasPromise) {
...
                 useFakeXMLHttpRequest: "ba8bd609-c921-4a62-a1b9-49336bd426a4"
             }
         };
+        fakeSandbox = {
+            create: "dc61d622-407f-46eb-af24-7a83bb30b8bf"
+        };
         sinon = proxyquire("../lib/sinon", {
-            nise: fakeNise
+            nise: fakeNise,
+            "./sinon/sandbox": fakeSandbox
```
mantoni: I have to go through proxyquire to get the correct reference to the sandbox instance. I tried using a reference to the sandbox module directly, which worked on node but it failed in browsers.

Issue #1512, opened 2017-08-03
sandbox.stub fails to stub methods on the prototype


RaceProUK:
* Sinon version: 2.4.1 and 3.0.0
* Environment: Node 8 (also happens in earlier versions of Node)

**What did you expect to happen?**
I expected that calling `sandbox.stub(object, 'property')`, where `property` is a function on the prototype, would result in the function being successfully stubbed with a no-op stub.

**What actually happens**
We receive the exception `TypeError: Cannot stub non-existent own property <property>`

**How to reproduce**
(Test case added August 7 by @fatso83)
```
const sinon = require('sinon')

class TestObject {
    someFunction() {
        console.error('sinon should stub me!');
    }
}

const sandbox = sinon.sandbox.create();

const instance = new TestObject();
const stub = sandbox.stub(instance, 'someFunction'); // This fails on Sinon 3.0, it worked on Sinon 2.x
```

**Diagnosis**
It appears the issue is in the check on like 89 of collections.js, specifically the call `Object.prototype.hasOwnProperty.call(object, property)`. It appears that `hasOwnProperty()` doesn't enumerate ES6 functions and properties.

**Proposed fix**
Instead of calling the above, use the test `Object.getOwnPropertyNames(Object.getPrototypeOf(object)).indexOf(property) > -1`. Our testing indicates that this method _does_ enumerate ES6 functions and properties in addition to ES5 functions and properties.

fatso83:
Thank you for your bug report, but I have no idea what "an es6 function" is. Do you mean an "arrow function" - a function whose context is the environment in which it was defined, or are you thinking about the proposed standard for class methods in esnext that are pre-bound to the instance? Or something else entirely? Not getting wiser by the prose, I am afraid.

A reproducible snippet of code speaks more than three paragraphs of text :-)

RaceProUK:
My apologies 😊

By ES6 function, I mean defining a member function with the ES6 class syntax, as we do here:
https://github.com/SockDrawer/SockBot/blob/master/providers/nodebb/index.js#L157-L160

One of the tests we're seeing with this issue is here:
https://github.com/SockDrawer/SockBot/blob/master/test/providers/nodebb/indexTest.js#L265

As the first link shows, the function does exist, but it's not enumerated in the `hasOwnProperty()` check, which is causing these tests to fail:
https://travis-ci.org/SockDrawer/SockBot/jobs/260642010#L969

[...]

PR #1516, opened 2017-08-07
Fix for 1512: sandbox.stub() fails on prototype props

fatso83:
This PR restores the functionality in sandboxes that made it possible to stub methods that were higher on the prototype chain. 

#### Background 
The bug introduced in 74263dd happened as the check for a property only checked on the instance, not on the properties further up the prototype chain. That meant it was no longer possible to stub out methods on instances of a "class". See issue #1512 for background info.

#### Solution 
Simply checks for existence of the (enumerable) property.

1. Check out this branch 
2. `npm install`
3. `npm test` (will run the testcase) 
4. `git checkout HEAD~1 && npm test` should fail

### DIFF
--- a/lib/sinon/collection.js
+++ b/lib/sinon/collection.js
@@ -86,7 +86,8 @@ var collection = {
     },
 
     stub: function stub(object, property) {
-        if (object && typeof property !== "undefined" && !Object.prototype.hasOwnProperty.call(object, property)) {
+        if (object && typeof property !== "undefined"
+            && !(property in object)) {
             throw new TypeError("Cannot stub non-existent own property " + valueToString(property));
         }
 
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -320,4 +320,25 @@ describe("issues", function () {
             assert.equals(this.stub.withArgs("arg").lastCall.returnValue, "return value");
         });
     });
+
+    describe("#1512", function () {
+        var sandbox;
+
+        beforeEach(function () {
+            sandbox = sinon.sandbox.create();
+        });
+
+        afterEach(function () {
+            sandbox.restore();
+        });
+
+        it("can stub methods on the prototype", function () {
+            var proto = { someFunction: function () {} };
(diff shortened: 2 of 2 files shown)


fatso83:
The test only checks for enumerable properties though ... I can change it to use `Object.hasOwnProperty`, but it will need to recurse up the prototype chain to the top. Do we have such a utility?

lucasfcosta:
Hi @fatso83, excellent work, as always.

We do not have that exact utility, but we have [`get-property-descriptor`](https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/core/get-property-descriptor.js), which goes up the prototype chain until it finds a property descriptor, otherwise it returns `undefined`.

I think we could use that and check if the `descriptor` returned is defined or not.

fatso83:
Ha, I was wrong: the original code was 100% correct, and I just learned how to read the EcmaScript specification :smile: 

Some exploratory tests using `prop in object` on non-enumerable objects made me confused, as it showed that it actually could tell if a non-enumerable property was present. I thought that it was the complement of the `for ... in`, but then I 
1. found [the definition](https://www.ecma-international.org/ecma-262/5.1/#sec-11.8.7) of how the _in_ operator works, which lead me to
2. [`[[HasOperator]]`](https://www.ecma-international.org/ecma-262/5.1/#sec-8.12.6), which again led me to 
3. [`[[GetProperty]]`](https://www.ecma-international.org/ecma-262/5.1/#sec-8.12.2), which essentially is exactly what our internal util does - a recursive lookup on the prototype to find the descriptor!

So, basically I'll just merge this as it is, as it does exactly what was intended (somewhat by chance) :see_no_evil:

mroderick:
👍

lucasfcosta:
@fatso83 awesome!

Btw, if you're interested into doing some more research into the MetaObject Protocol I highly recommend you [this link](http://www.2ality.com/2014/12/es6-proxies.html). Which explains them and relates them to how `Proxies` work.

Issue #1368, opened 2017-03-26
Support for Stub#resolvesThis()


kanehara:
* Sinon version : 2.1.0
* Environment   :  Node 7.7.4

**What did you expect to happen?**
Stubs have #returnsThis() and #resolves() functionality but it is not possible to use either to return the Stub's context as a _Promise_.  It would be nice to have a #resolvesThis() function in Stubs.  For example, Mongoose's Model#save() method returns the saved Model object on success.  Given the following function:

```javascript
function save(userInfo) {
  const user = new User({
    ...userInfo
  });
  return user.save().catch(err => console.log(err));  
}
```
It would be nice to test that the saved `user` object actually contains all the info in `userInfo` via:

```javascript
const stub = sinon.stub(User, "save");
stub.resolvesThis();
const savedUser = await save(mockUserInfo);
assert(savedUser.id === mockUserInfo.id);
```

#returnsThis returns the context but not as a promise and #resolves does not allow for returning the Stub context.

PR #1517, opened 2017-08-07
fixes issue #1368 by adding stub#resolvesThis

HugoMuller:
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -155,6 +155,7 @@ module.exports = {
     resolves: function resolves(fake, value) {
         fake.returnValue = value;
         fake.resolve = true;
+        fake.resolveThis = false;
         fake.reject = false;
         fake.returnValueDefined = true;
         fake.exception = undefined;
@@ -173,12 +174,19 @@ module.exports = {
         }
         fake.returnValue = reason;
         fake.resolve = false;
+        fake.resolveThis = false;
         fake.reject = true;
         fake.returnValueDefined = true;
         fake.exception = undefined;
         fake.fakeFn = undefined;
 
         return fake;
+    },
+
+    resolvesThis: function resolvesThis(fake) {
+        fake.resolveThis = true;
+        fake.resolve = false;
+        fake.reject = false;
     },
 
     callThrough: function callThrough(fake) {
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -187,6 +187,9 @@ module.exports = {
         fake.resolveThis = true;
         fake.resolve = false;
         fake.reject = false;
+        fake.returnValueDefined = false;
+        fake.exception = undefined;
+        fake.fakeFn = undefined;
     },
 
     callThrough: function callThrough(fake) {
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -184,8 +184,9 @@ module.exports = {
     },
 
     resolvesThis: function resolvesThis(fake) {
-        fake.resolveThis = true;
+        fake.returnValue = undefined;
         fake.resolve = false;
+        fake.resolveThis = true;
         fake.reject = false;
         fake.returnValueDefined = false;
         fake.exception = undefined;
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -292,7 +292,7 @@ describe("stub", function () {
             }
         });
 
-        it("returns a promise resolves with this", function () {
+        it("returns a promise resolved with this", function () {
             var instance = {};
             instance.stub = createStub.create();
             instance.stub.resolvesThis();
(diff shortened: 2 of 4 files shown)


lib/sinon/default-behaviors.js
```diff
@@ -181,6 +183,12 @@ module.exports = {
         return fake;
     },
 
+    resolvesThis: function resolvesThis(fake) {
+        fake.resolveThis = true;
+        fake.resolve = false;
+        fake.reject = false;
```
lucasfcosta: IMO we should keep this consistent with other behaviors, such as `rejects`, which also sets many other flags, as you can see [here](https://github.com/sinonjs/sinon/pull/1517/files#diff-f46d81a3e31504be425e201260dab281R165). Even though this is enough to make it work, I think that by keeping the whole API consistent when it comes to flagging fakes is extremely important.

lucasfcosta:
Hi @HugoMuller, thanks for this awesome Pull Request!

I really liked your tests and your clean code. It works, indeed and it is an extremely important feature IMO. I just had a comment regarding the flags being set when the `resolvesThis` behavior is used.

Please let me know if you disagree with anything or if you have any further doubts.

I'm looking forward to being able to merging this as soon as everyone approves it.

Thanks again for your work 😄

lib/sinon/default-behaviors.js
fatso83: I think you created the wrong link? It just points back to your comment.

test/stub-test.js
```diff
@@ -285,6 +285,65 @@ describe("stub", function () {
...
+
+            return instance.stub().then(function (actual) {
+                assert.same(actual, instance);
+            });
+        });
+
+        var strictMode = (function () {
+            return this;
+        }()) === undefined;
+        if (strictMode) {
```
mroderick: I am not sure a conditional for strict mode is necessary:

* Sinon's source is meant to be ES5.1, strict mode and expects to run in engines that supports that.
* The test file has `use strict;` at the top.

If for some reason, someone decides to run the tests in non-strict mode or in engines that doesn't support that, then I think the tests **should** be allowed to fail.

test/stub-test.js
```diff
@@ -285,6 +285,65 @@ describe("stub", function () {
...
+                stub.resolvesThis();
+
+                // Due to strict mode, would be `global` otherwise
+                return stub().then(function (actual) {
+                    assert.same(actual, undefined);
+                });
+            });
+        }
+
+        it("stub respects call", function () {
```
mroderick: Can this expectation be more specific, to make it easier to understand what "respects call" means?

test/stub-test.js
```diff
@@ -285,6 +285,65 @@ describe("stub", function () {
...
+            var stub = createStub.create();
+            stub.resolvesThis();
+            var object = {};
+
+            return stub.call(object).then(function (actual) {
+                assert.same(actual, object);
+            });
+        });
+
+        it("stub respects apply", function () {
```
mroderick: Can this expectation be more specific, to make it easier to understand what "respects apply" means?

test/stub-test.js
```diff
@@ -2171,6 +2230,16 @@ describe("stub", function () {
             refute.defined(instance.stub());
         });
 
+        it("cleans 'resolvesThis' behavior", function () {
```
mroderick: It is generally better to be very specific when writing expectations, especially in codebases with distributed contributors. Generic expectations often lead to readers trying to derive the intent from the implementation of the test. If there's a mistake in the implementation, then the reader is likely to end up with an incorrect understanding of then intended behaviour.

I would suggest rewriting the expectation to something like this:

```js
it("cleans 'resolvesThis' behavior using stub.resetBehavior", function () {
it("calls stub's resetBehavior method to clean behavior", function () {
```

test/stub-test.js
HugoMuller: Actually, the tests I wrote are very similar to those written for returnsThis.
So the same logic applies here. But if this one does not make sense, then I can remove it.
Same remark for the unclear expectations.

test/stub-test.js
HugoMuller: I'll fix these expectations. Thanks for the suggestions.

lib/sinon/default-behaviors.js
HugoMuller: I guess I can set `returnValueDefined` and `exception` flags to false to keep consistency. Do you think some other flags need to be turned off?

mroderick:
👍

lib/sinon/default-behaviors.js
lucasfcosta: @fatso83 I think that link only works inside the files tab. When I click it inside of it the link works.

Basically I think the following flags should be set:

* `fake.returnValue`
* `fake.returnValueDefined`
* `fake.exception`

HugoMuller: Okay now resolves, rejects and resolvesThis all act on the same flags.
I set returnValue to undefined as the stub resolves with itself.
Is it okay by you?

PR #1519, opened 2017-08-07
Update docs for sandbox config

fatso83:
#### Purpose (TL;DR)
The [migration docs for 3.0](http://sinonjs.org/releases/v3.0.0/migrating-to-3.0/) mentions that the format for the config object to `sinon.useFakeTimers()` has changed, but this is also used by the sandboxes. This PR updates the migration docs with a new section, updates the sandbox docs with the omission, and also fixes a couple of relative links that pointed to 404s.

I needed to change both the source and the released docs, as it can be regarded as a bug that this fix is not already present.

1. Build the docs per the instructions in `docs/CONTRIBUTING.MD`.
2. See the changes ([1](http://localhost:4000/releases/v3.0.0/sandbox) and [2](http://localhost:4000/releases/v3.0.0/migrating-to-3.0/))
3. Check the links


mantoni:
Hmmm, this pull request reverted my changes from #1515. Was that on purpose?

fatso83:
Damn! No, that was NOT intentional. Not sure how that happened.

fatso83:
Fixed by 2a04615

Issue #1521, opened 2017-08-09
Cannot stub `Array.prototype.filter`


pomek:
* Sinon version : 3.1.0
* Environment: Node.js and Chrome 60.0.3112.78
* Example URL: https://gist.github.com/pomek/1f135ec4e1bbf872254a66fb9170da7b
* Other libraries you are using: Mocha

**What did you expect to happen?**
The tests should pass.

**What actually happens**

<details>
    <summary>RangeError: Maximum call stack size exceeded</summary>

```
    at defineProperty (native)
    at Array.proxy (node_modules/sinon/lib/sinon/spy.js:69:79)
    at Function.matchingFakes (node_modules/sinon/lib/sinon/spy.js:343:35)
    at Function.invoke (node_modules/sinon/lib/sinon/spy.js:166:30)
    at Array.proxy (node_modules/sinon/lib/sinon/spy.js:69:54)
    at Function.matchingFakes (node_modules/sinon/lib/sinon/spy.js:343:35)
    at Function.invoke (node_modules/sinon/lib/sinon/spy.js:166:30)
    /* ... */   
    at Array.proxy (node_modules/sinon/lib/sinon/spy.js:69:54)
    at Function.matchingFakes (node_modules/sinon/lib/sinon/spy.js:343:35)
    at Function.invoke (node_modules/sinon/lib/sinon/spy.js:166:30)
    at Array.proxy (node_modules/sinon/lib/sinon/spy.js:69:54)
    at Context.it (packages/ckeditor5-dev-env/tests/asda.js:25:19)
```
</details>

**How to reproduce**

Just run the tests. It worked in Sinon `1.x` and does not work in `2.x` and `3.x`.

mroderick:
Thank you for the detailed error report 🏆 

Using your test case I have been able to determine that this defect was introduced in e72fee495d121a4fa3ad9a623a5a5a078412eab7, which was part of https://github.com/sinonjs/sinon/pull/1313 which became `v2.3.6`.

Until this is fixed, I guess you can use `v2.3.5`.

fatso83:
[...]

PR #1523, opened 2017-08-10
Cache references to Array.prototype.filter

fatso83:
#### Purpose
Fixes issue #1521 by caching references to the `filter` method so that it can be stubbed. Also adds a hint to the pull request template on things that should be in place.

1. Check out this branch
2. `npm install`
3. `npm test`


fatso83:
I added no extra tests for the second commit. I simply couldn't find out how to add relevant tests, so I gave up after close to an hour of messing around with `call.js` and `call-test.js`, but at least all existing tests pass.

mroderick:
This has become part of [`v3.2.0`](https://github.com/sinonjs/sinon/releases/tag/v3.2.0)

PR #1527, opened 2017-08-11
⚠️ Experiment: use mocha-chrome to replace PhantomJS

mroderick:
⚠️ Experiment ⚠️ 

This PR is an experiment to see if we can use [`mocha-chrome`](https://github.com/shellscape/mocha-chrome) to get closer to a solution to #1405 (abandoning PhantomJS).

This first commit took just a few minutes to set up. I am not heavily invested in the solution, and am happy to scrap it entirely or turn it into something else.

Admittedly, this first commit doesn't work very well, in that I don't think it actually uses all the tests, and one test fails.

But, it shows a lot of promise!

#### To try it out locally

1. Check out this branch
1. `npm install`
1. `npm run test:chrome`

#### Open questsions

- [ ] Is it acceptable to just load everything as a bundle, or shall we try to load files individually to get better stack traces?
- [ ] Would it be worthwhile to also make a small development server, so that people can even run tests in regular Chrome and easily use Chrome's built-in debugging tools?

Ideas, suggestions and feedback welcome!


test/test.html
```diff
@@ -0,0 +1,17 @@
+<!doctype>
```
fearphage: `<!doctype html>`, no?

mroderick: Indeed :

fatso83:
I am not totally sold on this: why need to reinvent Karma? It already gives us the possibility of running in any browser (passing a `--headless` option if we want to), has a development server, etc. 

The optimal solution IMHO would be to update `mochify` to support Chrome headless to be able to reuse all that we have.

PR #1530, opened 2017-08-14
Fix regression for issue #1526 regarding onFirstCall().throws()

### LINKED ISSUE #1526: .throws() in conjuction with .onFirstCall() doesn't work
Hi,

After upgrade to Sinon 3.2.0, I've spotted that some of my tests are failing.
As a result of a quick investigation, I've found that .throws() is not throwing when it's used together with .onFirstCall(), onCall() etc.

I've checked this issue for .withArgs() and it seems to work correctly.

* Sinon version : 3.2.0
* Environment   : nodejs
* Example URL   :
* Other libraries you are using:

It should throw an Error only on the first call or no of the call specified in the .onCall() method.
**What actually happens**
It doesn't throw any error.
**How to reproduce**

```javascript

  const stub = sinon.stub();
  stub
    .onFirstCall()
    .throws();

  try {
    stub();
    console.log('WRONG');
  } catch(e) {
    console.log('OK');
  }
```

noisecapella:
Fix #1526 by adding `exceptionCreator` to the list of fields checked in `isPresent`

The PR #1511 broke this behavior because it added `exceptionCreator` as a field which stored a function which would create the exception.

#### Solution  - optional
The check was added in `isPresent` so that the behavior is detected as being set.

If you run the new test on master it should fail, and it should pass on this branch.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -115,6 +115,7 @@ var proto = {
     isPresent: function isPresent() {
         return (typeof this.callArgAt === "number" ||
                 this.exception ||
+                this.exceptionCreator ||
                 typeof this.returnArgAt === "number" ||
                 this.returnThis ||
                 this.resolveThis ||
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -137,6 +137,19 @@ describe("stub", function () {
                 stub();
             });
         });
+
+        it("throws only on the first call", function () {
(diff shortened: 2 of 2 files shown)


fatso83:
Thanks! Manually verified too.

mroderick:
Thank you :)

mroderick:

PR #1532, opened 2017-08-16
Fix for `sinon.createFakeServer()` exception

### LINKED ISSUE #1531: sinon.createFakeServer() throws `TypeError: server.configure is not a function`
Sinon version: 3.2.0
Node version: 8.2.0

**To reproduce:**
```js
require("sinon").createFakeServer();
```

**Expected:**  
Fake server is created successfully.

**Actual:**
Error thrown `TypeError: server.configure is not a function`

**Analysis:**
When the function `createFakeServer` is [defined](https://github.com/sinonjs/sinon/blob/1b81ec67c30046f5117c9299faa24a66035d7fd5/lib/sinon.js#L32), the context of `nise.fakeServer` is not perserved.  
Instead, the context becomes the `sinon` object, which, as the error says, does not have a `configure` function.

**Suggested fix:**
`exports.createFakeServer = nise.fakeServer.create.bind(nise.fakeServer);`

**Temporary solution:**
Use `sinon.fakeServer.create()` instead.

EyalAr:
Fixes #1531 by preserving the context of `nise.fakeServer.create` and `nise.fakeServerWithClock.create`

### DIFF
--- a/test/sinon-test.js
+++ b/test/sinon-test.js
@@ -17,10 +17,14 @@ describe("sinon module", function () {
 
         fakeNise = {
             fakeServer: {
-                create: "47c86a4c-6b48-4748-bb8c-d853f999720c"
+                create: function () {
+                    return "47c86a4c-6b48-4748-bb8c-d853f999720c";
+                }
             },
             fakeServerWithClock: {
-                create: "e69974f8-4568-48d1-a5e9-2b511a59c14b"
+                create: function () {
+                    "e69974f8-4568-48d1-a5e9-2b511a59c14b";
+                }
             },
             fakeXhr: {
                 xhr: "958e8996-0cc3-4136-8a0e-6a120f5311bc",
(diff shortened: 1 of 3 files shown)


test/sinon-test.js
```diff
@@ -17,10 +17,14 @@ describe("sinon module", function () {
...
             fakeServer: {
-                create: "47c86a4c-6b48-4748-bb8c-d853f999720c"
+                create: function () {
+                    return "47c86a4c-6b48-4748-bb8c-d853f999720c";
+                }
             },
             fakeServerWithClock: {
-                create: "e69974f8-4568-48d1-a5e9-2b511a59c14b"
+                create: function () {
+                    "e69974f8-4568-48d1-a5e9-2b511a59c14b";
```
fatso83: This function will only return `undefined`. Add `return`.

fatso83:
Thank you very much for this PR. This is good work, but there was just a minor issue that needs some fixing before merging this.

test/sinon-test.js
EyalAr: Good catch. Fixed.

mroderick:
Thank you 👍

mroderick:

PR #1533, opened 2017-08-16
resolvesThis should override previous throws

noisecapella:
Fixed missing reset of `exceptionCreator` for `resolvesThis` which exists in other similar functions. Where `exception` is set to `undefined` `exceptionCreator` should also be set to `undefined.

In the branch, remove the change to `default-behaviors.js` and run the tests. One of the two new tests should fail, showing that the previous behavior is inconsistent regarding how `throws` affects `resolveThis`.

#### Checklist for author

- [ ] `npm run lint` passes


mroderick:
Thank you 👍

mroderick:

PR #1543, opened 2017-08-25
Add support to instanceOf for [Symbol.hasInstance]

alex94puchades:
#### Background
Though it is easy to replicate this functionality (eg: ```sinon.match(x => s instanceof foo)```), the fact that this is not supported out-of-the-box can produce confusion, as the behaviour of instanceOf matchers and the instanceof operator differs regarding [Symbol.hasInstance].

#### Solution
This PR addresses this issue by lifting the limitation to allow not only functions, but any object with [Symbol.hasInstance] method, as defined [by the standard](https://tc39.github.io/ecma262/#sec-instanceofoperator), if supported by the interpreter.

#### How to verify
1. Check out this branch
2. `npm install`
3. `node` (make sure your node version supports ES6 symbols)
4. `var sinon = require("./")`
5. `var matcher = sinon.match.instanceOf({ [Symbol.hasInstance]: () => true })`
6. `matcher.test("foo")` // returns true

#### Checklist for author
- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -185,7 +185,7 @@ match.typeOf = function (type) {
 };
 
 match.instanceOf = function (type) {
-    if (typeof Symbol === "undefined") {
+    if (typeof Symbol === "undefined" || typeof Symbol.hasInstance === "undefined") {
         assertType(type, "function", "type");
     } else {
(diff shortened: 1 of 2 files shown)


fatso83:
Thank you!

Issue #1535, opened 2017-08-17
Assert lacks `calledWithNew`


tsclaus:
* Sinon version : 2.3.8 but newest documentation still lacks it
* Environment   : N/A
* Example URL   : N/A
* Other libraries you are using: N/A

For the most part, the Assert methods mirror the Spy methods. However, Spy supports `calledWithNew` while Assert does not. Not a huge inconvenience but it forces test files to use something besides the assert pattern. `assert.calledWithNew(spy)` makes the most sense for how it would be implemented.

fatso83:
Thanks for the bug report. Would you mind having a go at implementing it? I think that would be the quickest win, if you want to have it within a reasonable timeframe.

mroderick:
A pull request for this would be very welcome.

To update the documentation, you should add to the files in `docs/release-source`, which are the ones that will be used when we cut the new release.

rgroothuijsen:
Just tested this with the latest version, and assert.calledWithNew appears to work as expected. Documentation still needs to be fixed, though.

mroderick:
Fixed by #1544

PR #1544, opened 2017-08-26
Documentation update for assert.calledWithNew

rgroothuijsen:
#### Purpose (TL;DR)

Documentation update for #1535 that adds missing assert.calledWithNew.

It was reported that there was no assert.calledWithNew, while Assert did feature most other assertions that Spy had.

#### Solution 

Upon investigation, it turned out that Assert did in fact have this method, but it had been left out in the API documentation.

#### How to verify
1. Check out this branch
2. `npm install`
3. `node`
4. `var sinonAssert = require("./lib/sinon/assert")`
5. `var sinonSpy = require("./lib/sinon/spy")`
6. `var testSpy = sinonSpy()`
7. `new testSpy()`
7. `sinonAssert.calledWithNew(testSpy)` // this should not throw an exception

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/assertions.md
+++ b/docs/release-source/release/assertions.md
@@ -127,6 +127,11 @@ Passes if `spy` was always called with matching arguments.
 This behaves the same way as `sinon.assert.alwaysCalledWith(spy, sinon.match(arg1), sinon.match(arg2), ...)`.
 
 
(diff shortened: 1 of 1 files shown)


fatso83:
Thanks!

PR #1550, opened 2017-09-06
Fix double backticks in inline code examples

kuba-orlik:

In some places in the file the inline code examples are terminated with double backtics, which makes the second backtick visible in rendered output

1. Compile the markdown
2. See that the additional backtics are not visible


fatso83:
thanks :)

Issue #1552, opened 2017-09-07
sinon.stub(object, "missingMethod") silently fails


tsclaus:
* Sinon version : 2.3.8
* Environment : Node.js 6.9.5
* Example URL : N/A
* Other libraries you are using: Karma

**What did you expect to happen?**

I expected sinon to either throw an error immediately or stub the method despite it not existing yet (I think the first one is better to make errors due to typos more obvious).

**What actually happens**

Essentially nothing. `object.missingMethod` is undefined after the call but the call doesn't error either.

**How to reproduce**

```
'use strict';

const sinon = require('sinon')

describe('stub for missing method', function() {
  beforeEach(function() {
    this.test = {};

    sinon.stub(this.test, 'foo');
  });

  it('should exist if it didn\'t error in the beforeEach', function() {
    this.test.foo();

    sinon.assert.calledOnce(this.test.foo);
  });
});
```

The test will break on `this.test.foo()` since it is still undefined. My sinon version is outdated, but looking at `function stub(object, property)`, it seems this would also be an issue in the newest version. EDIT: Looking closer, I don't think is a problem in 3.X. If you don't fix old versions and can't reproduce this in 3.X, this issue is irrelevant.

If this is indeed an issue, I will try to address it along with the other issue I have open this weekend.

fatso83:
Please verify if this is a problem using the latest sinon version. Just do `npm install sinon@latest` and run those tests again. This is related to #1512 and #1537

tsclaus:
The test still fails with 3.2.1, specifically with:
```
stub for missing method should exist if it didn't error in the beforeEach:
     TypeError: this.test.foo is not a function
      at Context.<anonymous> (test.spec.js:13:15)
```

fatso83:
Thanks for the report! This is verified to be a duplicate of #1508. You can track progress there.

fatso83:
Fixed.

Issue #1537, opened 2017-08-18
Sandbox throws error 'cannot stub non-existent own property'


ZebraFlesh:
I just tried upgrading from 2.4.1 to 3.2.1 and encountered the following issue. This code works in 2.4.1:

```
        const spy = sandbox.spy();
        sandbox.stub(window, 'google').value({
            maps: {
                LatLng: x => x,
                Map: spy
            }
        });
```

But in 3.2.1 it throws an exception: `TypeError: Cannot stub non-existent own property google`

It's not mentioned in the [migration guide](http://sinonjs.org/releases/v3.0.0/migrating-to-3.0/) so it appears to be a regression.

fatso83:
Related to #1512.

fatso83:
If the property doesn't exist you don't need to add it to the sandbox. Simply overwrite it. But yeah, if it worked before, and we haven't explicitly said it should change, then it's a regression. 

Not sure what we should do here. Update the docs to say stubbing non-existing values doesn't make sense and is unsupported, or make it possible?

ZebraFlesh:
> If the property doesn't exist you don't need to add it to the sandbox. Simply overwrite it.

The nice thing about adding the property to the sandbox is that sinon then helps me keep my global test environment clean between each test via `sandbox.restore()`. It is an extremely useful feature, especially when dealing with 3rd party libraries like Google Maps where I don't control the API. It would be great if it could be made to work in the 3.x line.

Also I just noticed that I committed the sin of not providing a complete example. My sandbox is being created with 2.4.1 format:
```
let sandbox;

before(() => { sandbox = sinon.sandbox.create(); })
afterEach(() => { sandbox.restore(); })
```

Not sure if that's important; apologies for not providing it sooner.

mroderick:
I think that in the scenarios like the one @ZebraFlesh describes, I would prefer to have the text fixture be more explicit.

```js
// not so explicit, doesn't work with sinon@3.x
beforeEach(function() {
    const spy = sandbox.spy();
    sandbox.stub(window, 'google').value({
        maps: {
            LatLng: x => x,
            Map: spy
        }
    }); 
});
```

```js
// more explicit, works with sinon@2, sinon@3
function setGoogleMapsFixture(sandbox) {
    window.google = {
        maps: {
            LatLng: x => x,
            Map: sandbox.spy()
        }
    };
}

function removeGoogleMapsFixture() {
    delete window.google;
}

beforeEach(function() {
[...]

Issue #1508, opened 2017-08-02
Sinon no longer throws when stubbing non-existent properties


schreifels:
* Sinon version : 2.3.7
* Environment   : OSX, Node 6 (Jest)
* Other libraries you are using: Mocha

**What did you expect to happen?**

When calling `sinon.stub(obj, 'nonexistentProperty');`, I would expect it to throw an error, e.g. "Attempted to wrap undefined property".

**What actually happens**

It does not throw (nor does it create a stub on the object at that property).

**How to reproduce**

Interestingly, if you use the deprecated 3rd argument to `sinon.stub`, it does still work, as shown in this example:

```
'use strict';

const assert=require('assert');
// const sinon = require('./lib/sinon'); // if checking out the sinon project
const sinon = require('sinon');

describe('sinon', () => {
  context('with the standard syntax', () => {
    it('throws when stubbing a non-existent property', () => {
      const obj = {};

      // FAIL: expected [Function] to throw an error
      assert.throws(() => {
        sinon.stub(obj, 'nonexistentProperty');
      });
    });
  });

  context('with the deprecated syntax', () => {
    it('throws when stubbing a non-existent property', () => {
      const obj = {};

      // PASS
      assert.throws(() => {
        sinon.stub(obj, 'nonexistentProperty', () => {});
      }, 'Attempted to wrap undefined property');
    });
  });
});
```

(I would have expected both tests to pass.)

fatso83:
Thanks to your test I was able to automate the search for this using `git bisect`. This behavior has been consistent throughout the entire 2.x series and was introduced by PR #1297, and the change is mentioned in [the list of changes](https://github.com/sinonjs/sinon/pull/1297#issuecomment-285877735).

From the discussion, I am not totally sure how it was decided we should  go from throwing on non-existent properties to support stubbing them, but maybe @lucasfcosta remembers more? Anyway, this change was intentional and can't be reversed now (if so, use 1.x). The only thing I feel is an issue, is that this is a breaking change (as you show), that is [not covered in the migration guide from 1.x to 2.x](http://sinonjs.org/releases/v3.1.0/migrating-to-2.0/). That should be remedied.

lucasfcosta:
Hi @schreifels, thanks for your issue!

As @fatso83 commented, this has been an intentional change in order to allow stubbing values, if I'm not mistaken.

[...]

PR #1557, opened 2017-09-12
Remove support for stubbing undefined props

fatso83:
#### Purpose
This removes support for stubbing undefined properties, aligning the behavior with how sandbox stubbing works. The [reasoning behind not allowing](https://github.com/sinonjs/sinon/issues/1537#issuecomment-323948482) this is 

> stubbing non-existing own properties is likely to lead to mistakes in tests, where the test passes because the author mistyped the name of the existing property they were aiming to stub

#### Background
This was a longer discussion that can be found in the
issues #1508, #1552 and #1537 - and internally amongst the core team.

Support for stubbing props using get()/set() was added in #1237
and also worked for undefined properties.

This type of behavior (stubbing undefined props) has been
explicitly disabled for sandboxes, so in order for Sinon to
behave in a consistent way, we should do the same here.

This change will bring normal stubs back to how it has
used to be historically, and will make sandbox stubs
and normal sinon stubs behave the same.

It might break things for people relying on the behavior
that has been present since Sinon 2.0, but it should
make things more reliable going forward.

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -9,6 +9,7 @@ var getPropertyDescriptor = require("./util/core/get-property-descriptor");
 var wrapMethod = require("./util/core/wrap-method");
 var stubEntireObject = require("./stub-entire-object");
 var throwOnFalsyObject = require("./throw-on-falsy-object");
+var valueToString = require("./util/core/value-to-string");
 
 var slice = Array.prototype.slice;
 
@@ -18,6 +19,10 @@ function stub(object, property) {
     }
 
     throwOnFalsyObject.apply(null, arguments);
+
+    if (object && typeof property !== "undefined" && !(property in object)) {
+        throw new TypeError("Cannot stub non-existent own property " + valueToString(property));
+    }
 
     var actualDescriptor = getPropertyDescriptor(object, property);
     var isStubbingEntireObject = typeof property === "undefined" && typeof object === "object";
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -5,7 +5,6 @@ var createStub = require("../lib/sinon/stub");
 var createStubInstance = require("../lib/sinon/stub").createStubInstance;
 var createSpy = require("../lib/sinon/spy");
 var sinonMatch = require("../lib/sinon/match");
-var getPropertyDescriptor = require("../lib/sinon/util/core/get-property-descriptor");
 var deprecated = require("../lib/sinon/util/core/deprecated");
 var assert = referee.assert;
 var refute = referee.refute;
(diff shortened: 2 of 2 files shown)


fatso83:
The alternative to removing this functionality is the inverse of this:
- remove the restriction introduced for sandboxes in 3.1 (#1537)
- fix the stub that is being produced (the reason for #1552 is that the stub is only produced at the point we call `set(...)` - this could be done earlier)

mroderick:
You can count on me to pretty much always err on the side of stricter use of languages and apis, and this is another one of those cases.

I vote 👍

mroderick:
@sinonjs/sinon-core any objections to merging this?

fatso83:
@mroderick [Only the objections raised in this other issue](https://github.com/sinonjs/sinon/issues/1389#issuecomment-330173705). Thought it might be a nuisance to have two major releases in short succession, but it's not a big thing, so you might as well. There's not like there is a PR out there for the factory function refactor.

mroderick:
> @mroderick Only the objections raised in this other issue. Thought it might be a nuisance to have two major releases in short succession, but it's not a big thing, so you might as well. There's not like there is a PR out there for the factory function refactor.

I don't see having several MAJOR releases as a big deal, as long as their changelogs are small at least :)

Glogo:
This may be unrelated to this issue, but for those that have the same problem as me with version 4 now.
Instead of doing this:
```js
sinon.stub(obj, 'yourMethod', () => Promise.resolve('whatever'))
```

you do:
```js
sinon.stub(obj, 'yourMethod')
  .returns(() => Promise.resolve('whatever'));
```

fatso83:
@Glogo, actually we have had [promise support](http://sinonjs.org/releases/v4.5.0/stubs/#stubresolvesvalue) for several versions, so it is even shorter:

```javascript
sinon.stub(obj, 'yourMethod').resolves('whatever');
```

paulmelnikow:
> stubbing non-existing own properties is likely to lead to mistakes in tests, where the test passes because the author mistyped the name of the existing property they were aiming to stub

Maybe; or maybe it's deliberate: https://github.com/sinonjs/sinon/issues/1537#issuecomment-433647887

cphoover:
Why was support for this removed?

As others have said adding

window.someKey = 'someValue';
// ... testCode here
delete window.someKey

is problematic if you have multiple differing values between tests, or it may be defined as something else in the future. With sandbox.restore we know we cleanup after ourselves so if that value **is** defined in the future we shouldn't run into issues. I think it makes sense to allow stubbing undefined properties maybe add another method if you are worried about this.

Issue #1556, opened 2017-09-11
FeatureRequest: nested property matcher


servel333:
Request to add this function signature: `sinon.match.hasNested(keypath, expected)`

where `keypath` functions like [Lodash.get](https://lodash.com/docs/#get) supporting `"outer.inner"` or `"outer[inner]"`.

Turning this syntax into this syntax

    sinon.match.has('outer', sinon.match.has('middle', sinon.match.has('inner', sinon.match.string)));
    sinon.match.hasNested('outer.middle.inner', sinon.match.string);

Personally, I would prefer `has` supported nested keypaths, but that would probably break backwards compatibility. The implementation is simple if the addition of the `lodash.get` library is acceptable.

```
const get = require("lodash.get");

match.hasNested = createPropertyMatcher(function (actual, property) {
    return get(actual, property) !== undefined;
}, "hasNested");

```

fatso83:
`lodash.get` is about 6k (2k gzipped), but we are not required to load in microseconds, so that should be fine. If you can supply a PR, that would be welcome.

mroderick:
This was fixed with #1560

PR #1560, opened 2017-09-13
Adds sinon.match.hasNested

servel333:
PR for #1556, adds `sinon.match.hasNested` so `sinon.match.has` chains can be avoided.

    sinon.match.has('outer', sinon.match.has('middle', sinon.match.has('inner', sinon.match.string)));
    sinon.match.hasNested('outer.middle.inner', sinon.match.string);

Nested array's are supported too.

    sinon.match.hasNested('outer[0].inner', sinon.match.string);

- Adds dependency `lodash.get@^4.4.2`.
- Adds tests for `hasNested`
- Adds documentation for `hasNested`

#### Solution  - optional
I was unable to use the existing `createPropertyMatcher` in the marcher code because it expects the property to not be nested when comparing the expectation.

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes
  - What?

### DIFF
--- a/docs/release-source/release/matchers.md
+++ b/docs/release-source/release/matchers.md
@@ -220,6 +220,22 @@ The property might be inherited via the prototype chain. If the optional expecta
 Same as `sinon.match.has` but the property must be defined by the value itself. Inherited properties are ignored.
 
 
+#### `sinon.match.hasNested(propertyPath[, expectation])`
+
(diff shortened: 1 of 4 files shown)


fatso83:
Thank you for this! As to what caching standard library functions means, see the link to #1523 for an elaboration :smile:

fatso83:
Released as Sinon 3.3.0

servel333:
> Thank you for this! As to what caching standard library functions means, see the link to #1523 for an elaboration

I see the link in my own PR description now... I can't believe I missed that.

PR #1565, opened 2017-09-18
Patch docs

fatso83:
This patches the docs for the 3.x releases to include info on the breaking change in sandboxes (see #1537), and also adds a migration guide for 4.0


mroderick:
Are we waiting for anything before releasing 4.0?

fatso83:
Don't think so

PR #1580, opened 2017-10-04
Dependency hygeine

mroderick:
This PR does a few things: 

* introduce an npm task to verify that all `dependencies` are used by code
* move `native-promise-only` and `sam-sam` to `devDependencies`
* remove unused dependencies: `text-encoding`, `path-to-regexp`
* upgrade `mocha` to latest

The check that verifies all dependencies in used will prevent us from shipping with unused dependencies again.


fatso83:
Perf!

Issue #1584, opened 2017-10-09
Fakeserver should support triggering progress events


tarjei:
Hi, AFAIK there is no way to get the fake server to trigger progress events during an upload.

It would be very useful this was possible in a simple way. Maybe:
```
    server = sinon.fakeServer.create()
    server.responWith((r) => {...})
// this fires for all requests - not ideal but very accessable and would solve 80% of all usecases
    server.updateProgress({lengthComputable: true, loaded: 100, total: 200 })
```

* Sinon version :  4.0
* Other libraries you are using: Mocha, Karma

**What did you expect to happen?

I would expect the progress event to fire for all requests.

fatso83:
Are you sure about this? We have support for `ProgressEvent`, and a quick grep of `nise` shows:

```
2375:        it("progress event is triggered with xhr.uploadProgress({loaded, 20, total, 100})", function (done) {
2376:            this.xhr.upload.addEventListener("progress", function (e) {
```

fatso83:
OK, I see, you basically want a utility wrapper that aggregates the events. Could you close this and open a feature request on the `sinonjs/nise` project? Or a PR ;-)

tarjei:
Will do. I'll also try to come up with a PR for the docs on using server.requests to trigger progress events :)

PR #1585, opened 2017-10-09
Document how to trigger progress events

tarjei:
Also, add in a small bit of info on the difference between download and upload progress events.

Document triggering progress events #1584 

I spent some time figuring this out - I hope someone else doesn't...


fatso83:
Looks great, thanks!

Issue #1559, opened 2017-09-13
Idea for future milestone(s)


mroderick:
### Background

`sinon.stub` / `sandbox.stub` has become the kitchen sink of 
configurable behaviour with issues that are often difficult to find and fix without regressions.

I think that the root cause of the difficulties is that it `stub` has far too many responsibilities.

Further, `stub` also has problematic usage caused by the fact that behaviour is set *after* it has been created, and can be redefined many times over.

```js
var myStub;

beforeEach(function(){
    myStub = sinon.stub().resolves('apple pie :)');
});

// several hundred lines of tests later
myStub = sinon.stub().rejects('no more pie :(');

// several hundred lines of tests later
// what behaviour does myStub currently have? Can you tell without 
// reading the entire file?
// can you safely change the behaviour without affecting tests further 
// down in the file?
```

And then there are the more confusing scenarios

```js
var myStub = sinon.stub()
                .withArgs(42)
                .onThirdCall()
                .resolves('apple pie')
                .rejects('no more pie')
```

What does that even do?

Instead of continuing to add more responsibilities to `stub`, I propose that instead we introduce new members to `sinon`, which are much *narrower* in scope.

The most important I can think of would be an immutable stand in for a function.

We can then later figure out what we're going to do about properties (as a separate, new, single responsibility member).

### `sinon.fake`

A `fake` (the return value of calling `sinon.fake`) is a **pure** and **immutable** `Function`. It does one thing, and one thing only. It has the same behaviour **on each and every** call. Unlike `stub`, its behaviour cannot be redefined. If you need different behaviour, make a new `fake`.

#### Single responsibility

A fake can have **one** of these responsibilities

* resolve a `Promise` to a value
* reject a `Promise` to an `Error`
* return a value
* throw an `Error`
* yield value(s) to a callback

If you want/need side effects, and you still want the spy interface, then either use the real function, use a `stub` or make a custom function

```javascript
sinon.replace(myObject, myMethod, sandbox.spy(function(args) {
    someFunctionWithSideEffects(args);
});
```

#### Throws errors generously

Will be generous with throwing errors when user tries to create/use them in unsupported ways.

```js
// will throw TypeError when `config` argument has more than one property
const fake = sinon.fake({
    resolves: true, 
    returns: true
});
```

#### Uses the spy API

Except for `.withArgs`, as that violates immutability

#### Usage ideas

```js
// will return a Promise that resolves to 'apple pie'
[...]

PR #1586, opened 2017-10-11
New API: default sandbox and fake

mroderick:
This PR is a solution for the proposal in #1559 

This PR should not be merged into `master` until we have given it a good trial in real world projects.

Contributions (code, fixes, tests, documentation) are very welcome. Please make pull requests into this branch.

I'll keep the branch up to date by pulling from `master` at intervals. Once we're happy with the branch, then I'll rebase against master and get rid of the merge commits.

#### Tasks

- [x] Make the main `sinon` object a default `sandbox`
- [x] Create `sandbox.replace` method
- [x] Document `sandbox.replace`
- [x] Create `sandbox.replaceGetter` method
- [x] Document `sandbox.replaceGetter`
- [x] Create `sandbox.replaceSetter` method
- [x] Document `sandbox.replaceSetter`
- [x] Create `sinon.fake`
- [x] Clean unwanted methods from `fake` [`.withArguments`, etc]
- [x] Document `sinon.fake`
- [ ] Put release candidates on npm as `sinon@next` to get feedback

#### To try it out in your project directly from source

1. Clone the repo
1. `git checkout default-sandbox-and-fake`
1. `npm link`
1. In your project: `npm link sinon`
1. Run your tests

Once we start publishing release candidates on npm, then we'll add instructions for testing in your projects here.


lib/sinon.js
```diff
@@ -1,42 +1,64 @@
...
 };
 
-var format = require("./sinon/util/core/format");
-exports.setFormatter = format.setFormatter;
+var legacySandboxAPI = {
+    sandbox: {
+        create: deprecated.wrap(
+            createSandbox,
+            // eslint-disable-next-line max-len
+            "`sandbox.create()` is deprecated. Use default sandbox at `sinon.sandbox` or create new sandboxes with `sinon.createSandbox()` or `new sinon.Sandbox()`"
```
fatso83: >  create new sandboxes with `sinon.createSandbox()` or `new sinon.Sandbox()
Why have two ways of creating sandboxes? We just made a new factory function.

mroderick: I'll remove the second way and tidy up the documentation. Thanks

mroderick:
> Should we expose the Sandbox prop if we don't mean to mean to make it public?

In the current version of the branch (5883852), that property is not exposed.

```js
> var sinon = require('./lib/sinon');
undefined

> typeof sinon.Sandbox
'undefined'
```

mroderick:
Ping @sinonjs/sinon-core

This is getting close to being reviewable, however two of the tests are failing in Chromium.

Any ideas why that might be?

mantoni:
@mroderick Nice work! I had a look at the failing tests. It's `proxyquire` changing the instance, so `actual instanceof Sandbox` also returns `false`. What you can do to fix the tests is adding `sinon = require("../lib/sinon");` or use proxyquire only where it's actually needed.

docs/release-source/release/stubs.md
```diff
@@ -178,6 +176,10 @@ This is equivalent to calling both `stub.resetBehavior()` and `stub.resetHistory
 
 *Updated in `sinon@2.0.0`*
 
+*Since `sinon@5.0.0`*
+
+You can reset all stubs using `sinon.reset()`
```
fearphage: Clarify what this resets. I assume it's behavior and history here, right?

docs/release-source/release/stubs.md
```diff
@@ -213,7 +221,11 @@ stub.resetHistory();
 stub.called // false
 ```
 
-*Since `sinon@2.0.0`*
+
+*Since `sinon@5.0.0`*
+
+You can reset behaviour of all stubs using `sinon.resetHistory()`
```
fearphage: Typo: `resetHistory` doesn't reset behavior.

mroderick: > Also do we consistently use the Old English (British?) spelling of behavior in the docs? The actual method name uses the American English spelling.

My natural language uses (mostly) British English spelling. Sometimes when writing code and documentation, that spelling slips through.

(I wonder if there's a spelling linter where you can set the language preference)

mroderick:
Tests are passing now (Travis was having a bit of a day). If you don't have any further comments for this PR, then I'll rewrite the history to read better.

Do we need more documentation of `fake`? Are you happy with me publishing it as `sinon@next`?

test/fake-test.js
```diff
@@ -0,0 +1,193 @@
...
+    it("should return a Sinon proxy", function () {
+        var actual = func(argument);
+
+        assert.isProxy(actual);
+    });
+}
+
+function noop() {}
+
+var hasFunctionNameSupport = typeof noop.name !== "undefined";
```
fearphage: `noop.name === 'noop'` is a bit more straight forward.

mroderick: Good point, thanks :)

docs/release-source/release/fakes.md
```diff
@@ -0,0 +1,116 @@
...
+
+fake();
+
+console.log(fake.callCount);
+// 1
+```
+
+### Fakes with behaviour
+
+Fakes can be created with behaviour, which cannot be changed once the fake has been created.
```
lucasfcosta: IMO it would be good to point out that it is possible to do [a `restore` as mentioned here](https://github.com/sinonjs/sinon/pull/1586/files#diff-8e4eec829308104e4aa45dca14337830R116) and that would be enough for the user to get the original property back into the object.

As a user I think that would be something I'd like to know how to do as soon as I read that:

> [a fake] cannot be changed once the fake has been created

docs/release-source/release/fakes.md
```diff
@@ -0,0 +1,116 @@
...
+```js
+var fake = sinon.fake.returns('42');
+
+sinon.replace(console, 'log', fake);
+
+console.log('apple pie');
+// 42
+```
+
+When you want to restore the replaced properties, `sinon.restore` method.
```
lucasfcosta: Even though this is really simple it might be worth having an example added.

Also, I'd add another header for this to make it as visible as possible given the importance of this method.

docs/release-source/release/sandbox.md
```diff
@@ -26,23 +24,26 @@ describe('myAPI.hello method', function () {
...
 });
 ```
 
 ## Sandbox API
 
-#### `var sandbox = sinon.createSandbox();`
+#### Default sandbox
 
-Creates a sandbox object with spies, stubs, and mocks.
+Since `sinon@5.0.0`, the `sinon` object is a default sandbox. Unless you have a very advanced setup or need a special configuration, you probably want to just use that one.
```
lucasfcosta: I think I'd also add a line here explaining why this change happened and how this makes testing easier, especially mentioning how useful it is to clear/restore all methods at once without having to create lots of sandboxes on each test.

docs/release-source/release/sandbox.md
```diff
@@ -127,19 +128,86 @@ A convenience reference for [`sinon.assert`](./assertions)
 
 *Since `sinon@2.0.0`*
 
+#### `sandbox.replace(object, property, replacement);`
+
+Replaces `property` on `object` with `replacement` argument.
+
+`replacement` can be any value, including `spies`, `stubs` and `fakes`.
+
+This method only works on non-accessor properties, for replacing accessors, use `sandbox.replaceGetter()` and `sandbox.replaceSetter()`.
```
lucasfcosta: This is great idea btw.

Since we've been dealing with `getters/setters` we've had a lot of trouble making the APIs generic enough to deal with them, but I guess that encapsulating that logic into separate methods is a lot better and allows us to avoid having lots of special conditions and a complex control-flow.

docs/release-source/release/fakes.md
mroderick: `replace` and `restore` belong together. Fakes are created separately from replacing properties. 

Fakes are just one possible argument to `replace`, you could use something else if you wanted to (like `mock`, `spy` or `stub` ... or even `Date`).

I'd like to keep the documentation focused, so not mention topics that are not necessary for understanding the current topic. With that in mind, I think that mentioning `replace` in the documentation about creating fakes with behaviour, would reduce the focus and make the text a bit confusing.

docs/release-source/release/fakes.md
mroderick: Good point, I've pushed new commits

docs/release-source/release/sandbox.md
mroderick: At first I agreed with you, and wrote some documentation.

But then I thought about it some more. The motivation is only really interesting when you're transitioning from earlier versions to this MAJOR version. Once you're over that hump, stating that in the documentation just adds noise.

Instead, I've put details into the migration guide for v5, and have pushed a new commit with that.

mroderick:
I've tried using this branch with an internal project with 1793 tests, most of them using Sinon. I am happy to report, that there were no problems.

It would be great if someone else could take it for a spin, before we publish it as `sinon@next`.

docs/guides/migrating-to-5.0.md
```diff
@@ -0,0 +1,47 @@
...
+### Before `sinon@5.0.0`
+
+In earlier versions, you would manually have to create sandboxes and keep references to them in order to restore them.
+
+```js
+const sandbox = sinon.createSandbox();
+
+describe('myFunction', function() {
+    afterEach(function () {
+        this.sandbox.restore();
```
mantoni: ~~this.~~

mroderick: You saw nothing! Nothing!

fatso83:
I also 
1. checked out this branch
2. ran `npm link` in the sinon repo
3. ran `npm link sinon` and `npm test` in a project with 200 tests - many sinon

All green :smile: 

Word of warning: I write really simple tests and I don't have access to projects from my previous employers that had much more elaborate setups. But at least initial impressions are good! And the fact that internal sinon tests pass is very promising.

mroderick:
Thanks @fatso83. I'll try to find some time in near future to put a `sinon@next` on npm, then we can ask the bigger audience to try it out.

docs/release-source/release/fakes.md
lucasfcosta: I totally understand, but then maybe we could rephrase this to make it more specific.

The first time I read this I had the strong impression that you would not be able to restore any property replaced with a fake.

The wording is a bit difficult though, what do y'all think? 🤔

mroderick:
This is now on npm as `sinon@next`

docs/release-source/release/fakes.md
mantoni: @lucasfcosta I think the documentation already points out that `replace` is a sandbox method that _may_ use a fake as an argument. The standbox can then be restored. There is nothing to restore in the fake itself. I agree with @mroderick here that talking about `restore` here would make the documentation more confusing.

mroderick:
@sinonjs/core what do you think, shall we get this ready to merge?

fatso83:
Seems fine. This means we introduce fakes, but still keep the old spy and stub APIs?

mroderick:
> This means we introduce fakes, but still keep the old spy and stub APIs?

I think that's best. I think if we deprecate spy/stub without first proving that fakes are a better choice, we risk making a mistake that would be difficult to reverse and cause a lot of extra work for consumers.

mantoni:
Hi all! I finally found some time to play with the new toys. I'm trying to write tests ONLY using the new fakes and sandbox features. First of all, I think this is great stuff. It feels almost like an all new framework.

I was wondering why `yields` is something completely different now. I have to pass a callback to the fake that will then be invoked – first, the implementation is currently missing a wrapped function, so it will attempt to invoke the callback immediately. I don't think that was the intention, right? But shouldn't it really behave like how `yields` works for stubs and invoke the last function argument once the fake is invoked?

mantoni:
Another thing I just noticed: I can't do `fake.yield()`, while `fake.firstCall.yield()` works.

mantoni:
I’d like to help with this effort and implement the suggestions / fix the issues I filed, if we agree on them. To make reviews simpler, I would suggest we create a “next” branch, merge this PR into “next” and then make more, smaller follow-up PRs against “next” until we’re all happy. How does that sound?

 @mroderick @fatso83 It would be great if one of you could make this branch mergable against latest master since you know better what happened. I’m mostly watching from the sideline lately.

mroderick:
> I’d like to help with this effort and implement the suggestions / fix the issues I filed, if we agree on them. To make reviews simpler, I would suggest we create a “next” branch, merge this PR into “next” and then make more, smaller follow-up PRs against “next” until we’re all happy. How does that sound?

That's great news! I'll work on making that branch mergeable :)

mroderick:
This has been replaced by #1700

PR #1589, opened 2017-10-14
Slim down npm package

avaly:

Slim down NPM package in size

Sinon is downloaded from NPM around 3.8 millions times per month.

This change will save around `3.1 MB` of uncompressed files and `0.9 MB` of NPM package tarball (`3.42 TB` of NPM traffic per month).

#### Solution  - optional

The build step should only create one set of packaged files.

The NPM CDN URL is still functional, since it supports package versioning: https://cdn.jsdelivr.net/npm/sinon@4.0.1/pkg/sinon.js

1. Check out this branch
2. `npm install`
3. `npm run build`
5. `npm pack`
4. `npm run postversion`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/build.js
+++ b/build.js
@@ -30,7 +30,6 @@ function makeBundle(name, config) {
 
             var script = preamble + buffer.toString();
             fs.writeFile("pkg/" + name + ".js", script);
-            fs.writeFile("pkg/" + name + "-" + pkg.version + ".js", script);
         });
 }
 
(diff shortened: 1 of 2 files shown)


fatso83:
@mroderick Do you see any problems with this?

mroderick:
I haven’t had time to take a closer look due to vacation. I’ll look into it soon

fatso83:
I don't see any immediate problems with it, as I would assume very few people are depending on the actual file format of the versioned files. And if they do, they probably refer to older versions, which means pruning the extra files going forwards should be fine. A :+1: from me.

mroderick:
I took a closer look, this looks great!

Thank you for the PR!

PR #1604, opened 2017-11-01
Remove package-lock.json

mroderick:
It turns out that `package-lock.json` doesn't quite do what we thought
it did.

From https://docs.npmjs.com/files/package-lock.json

> One key detail about package-lock.json is that it cannot be published,
> and it will be ignored if found in any place other than the toplevel
> package. It shares a format with npm-shrinkwrap.json, which is
> essentially the same file, but allows publication. This is not
> recommended unless deploying a CLI tool or otherwise using the
> publication process for producing production packages.

As a package provider, we are not getting any benefits from having a `package-lock.json` at all, since it won't ensure that package consumers
will use the same dependencies.

If we want to lock down dependencies when Sinon is installed, we should be using [`npm-shrinkwrap.json`](https://docs.npmjs.com/files/shrinkwrap.json).


fatso83:
Hmm ... I am a bit at a loss at what to do going forwards, as `npm-shrinkwrap.json` is [discouraged](https://docs.npmjs.com/files/shrinkwrap.json) for libraries as well:
> The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies. It's strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

Doesn't `package-lock.json` make our life as developers of the project a bit simpler? Or maybe it's good to get bitten every now and then by differing dependencies  ... not sure.

mroderick:
In my opinion, npm's loose versioning of dependencies is a **huge** design flaw. Even if we apply exact version numbers for **our** dependencies, the recursive resolution of their dependencies is most likely still going to be using loose versioning.

It's such a time sink having to deal with this mess:

* life as an application developer is NOT easier
* life as a package maintainer is NOT easier

I wish we only had strict versions 😞 

Anyway, I still don't really see what `package-lock.json` offers us as library authors (even if dependency resolution is faster).

fatso83:
No strong feelings about this. It's just nice that `package-lock.json` guarantees that if you got it working, then I can get started developing without any hassles.

mantoni:
> Anyway, I still don't really see what `package-lock.json` offers us as library authors

What it offers is a more reproducible dev environment. With `package-lock.json` in the repository, everybody can easily install the same dependency tree (with the exact versions, recursively).

> I wish we only had strict versions 😞

I disagree that strict versions would make anything better. We'd constantly find ourselves chasing maintainers to publish a new version of their package because of some patch being applied to one of their dependencies. I wouldn't want to live in that world.

Regarding this particular PR, I don't really have strong feelings either. I just wanted to highlight that there is actually a potential advantage in having the `package-lock.json`.

mroderick:
OK, let's leave it be.

Issue #1607, opened 2017-11-03
4.1.0 with supportsColor breaks testing on node < 4 


Yeti-or:
I am complaining because sinon supports old node.js, it's written in package.json  https://github.com/sinonjs/sinon/blob/master/package.json#L95

Okay, I understand that it is not mainstream, but I have a package that should be working on old env.
And everything was okay until release of 4.1.0.

But than `supports-color` was added: https://github.com/sinonjs/sinon/pull/1583/files#diff-dab05c918e781b4a65cc1fdbcdc6a7aaR3

Some options we could do:

* drop support of old node.js, because as I see in `.travis.yml` you are not testing on it anyway
* remove `supports-color` from dependencies 
and put necessary code from `supports-color` to `lib/sinon/color.js`
* try to convince `supports-color`, to switch from `const` && `let`, to `var` but actually I am not believing in this option =/

## Sinon version : 4.1.0

**What did you expect to happen?**
tests on node 0.12.0 works

**What actually happens**
no they not

mantoni:
The "engine" field in the `package.json` file is outdated and should be removed or corrected. The Sinon.js documentation page has a compatibility section wich states that Node 4 is the minimum requirement: http://sinonjs.org/releases/v4.1.0/

If you need Sinon on Node 0.12 please use Sinon 1.x. We require Node 4 since Sinon 2.

PR #1608, opened 2017-11-03
Remove "engines" from package.json

mroderick:
As discussed in #1607, this is outdated and should be removed


mroderick:
This has been published as `sinon@4.1.1`

PR #1609, opened 2017-11-03
Prefer node 8 for single run tests in Travis CI

mroderick:
`node@8` will be faster than `node@6` at running all these tests, and will offer better compatibility as our dependencies update and stop supporting older LTS versions


fatso83:
`Fail! Modules in package.json not used in code: coveralls`

mroderick:
@fatso83 I fixed it 👍

Issue #1610, opened 2017-11-04
Remove deprecated methods from the documentation


micnic:
`sinon.restore()` is not available as a public API, though it is documented in `utilities` section.

P.S. `sinon.format()` and `sinon.log()` also seems to not be available.

fatso83:
Those methods [were deprecated](deprecation-of-internal-helpers), starting with Sinon 2. Thank you for this reminder :-) We need to remove them. Would you like to [do the honours](https://github.com/sinonjs/sinon/blob/master/docs/CONTRIBUTING.md), @micnic?

micnic:
I just started using `sinon` few days ago in a project and had the need of using a functionality like `sinon.restore()`. So I could contribute to remove the `sinon.format()` and `sinon.log()` from the docs, but I'm interested in keeping / adding back `sinon.restore()`, what was the reason behind removing this method? the link that you provided gives 404.

Just to describe the case that I had, it's a class with 5 static methods, one of which calls every other static method, I created 4 stubs for these methods for testing the one that calls them and then I had to restore the methods to use them in other tests with the original behavior. I discovered that `sinon.restore()` is not available and had to write the `.restore()` method for each of the stubs, it's not much work, but it would be nicer to have such method for restoring the original behavior. For `sinon.format()` and `sinon.log()` I agree that they do not belong to the `sinon` public API.

fatso83:
Sorry about the dead link. Basically you need to check out [sandboxes](http://sinonjs.org/releases/v4.1.1/sandbox/). They achieve what you want. Or, for auto-correct, employ [Sinon-test](https://www.npmjs.com/package/sinon-test). 

Basically, the utils were removed due to either encouraging global state or exposing internal utils, neither of which is good.

micnic:
@fatso83 thanks for pointing out `sandboxes`, did not get my hands on them yet, will try this approach. Will come back soon with a PR for removing the deprecated methods.

PR #1613, opened 2017-11-07
Remove deprecated methods from documentation

micnic:
#### Purpose
Remove from docs `sinon.restore()`, `sinon.format()` and `sinon.log()`
These methods were long time deprecated and removed from code, but were still available in the utilities docs

related issue #1610


fatso83:
Thank you!

mroderick:
@fatso83 you were a bit too quick to merge this, we still need to remove them from the documentation of all the old releases where they were removed. Shall I do this?

fatso83:
:flushed: Ah, that's right ... If you can bother, then yes please.

PR #1615, opened 2017-11-07
Update documentatation for older releases

mroderick:
This PR is a followup to #1610, #1613, where we removed deprecated methods from documentation ... but only for `docs/release-source`.

This PR removes the documentation from the documentation from all the releases.

As alway, read commits separately to keep a firm grasp on sanity.

### DIFF
--- a/docs/_releases/v2.0.0/utils.md
+++ b/docs/_releases/v2.0.0/utils.md
@@ -39,14 +39,3 @@ The given constructor function is not invoked. See also the [stub API](../stubs)
 Formats an object for pretty printing in error messages using [formatio](https://github.com/busterjs/formatio). Feel free to
 override this method with your own implementation if you prefer different
 visualization of e.g. objects. The method should return a string.
-
-#### `sinon.log(string);`
-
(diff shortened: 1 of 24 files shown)


fatso83:
This will not fail if merging #1616 first, @mroderick.

fatso83:
Why keep the docs for `sinon.restore(object);` (and others in the 2.x docs) if we deprecated the method? It's no longer officially supported.

mroderick:
> Why keep the docs for sinon.restore(object); (and others in the 2.x docs) if we deprecated the method? It's no longer officially supported.

The were removed in 3.x, not 2.x

mroderick:
Read the commit messages, Luke!

fatso83:
I did :smiley_cat:  I just meant that even if they were _present_, it was just to ease transition for previous users, so in my mind they were already gone by the time we wrapped them with a deprecation helper in the pre-2.0 branch.

mroderick:
Thanks.

Let's discuss a strategy for how long to keep documentation around when we deprecate things, and ***how** to keep it around.

PR #1618, opened 2017-11-09
`stub.callsArgAsync(index);` is already listed in line #466

malev:
1. Open docs/release-source/release/stubs.md and verify that `stub.callsArgAsync(index);` appears only once.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -463,8 +463,6 @@ Like `callArg`, but with arguments.
 
 Same as their corresponding non-Async counterparts, but with callback being deferred (executed not immediately but after short timeout and in another "thread")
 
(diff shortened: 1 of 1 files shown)


mroderick:
Good find! There is definitely something weird going on in this section of the documentation.

The following sentence is also duplicated:

> Same as their corresponding non-Async counterparts, but with callback being deferred (executed not immediately but after short timeout and in another "thread")

I think we need to do more for this section, than just delete the duplication. Perhaps a level 3 header about async variations using the sentence above.

Please note that it is not enough to only change `docs/release-source` when documenting existing functionality, you will also need to go through the documentation in `docs/_releases` and update that.

Would you be up for making those changes?

PR #1619, opened 2017-11-09
Move yieldsAsync* documentation to the first yieldsAsync*

malev:

> Move the description for `yieldsAsync*` methods to the first `yieldsAsync*` because it makes more sense to be there.

> If you visit  http://sinonjs.org/releases/v4.1.2/stubs/, at first glance it looks like the `yieldsAsync*` methods lack on documentation.
> The generic explanation for all the `yieldsAsync*` is under `yieldsToOnAsync` which is the last method on the list
> Makes more sense to have the generic documentation under the first `yieldsAsync*` of the list.

1. Check out this branch
2. `npm install`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -473,13 +473,13 @@ Same as their corresponding non-Async counterparts, but with callback being defe
 
 #### `stub.yieldsAsync([arg1, arg2, ...]);`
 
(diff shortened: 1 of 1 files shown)


fatso83:
To me, this looks fine, and I don't immediately see the reason to retroactively fix this in the previous versions, since it's not a bug.

mroderick:
Thank you :)

Issue #1377, opened 2017-04-10
Upgrading to Sinon 2.1.0 resulted in "HTMLElement is not defined" error


GentlemanHal:
I've actually solved the issue for myself but wanted to raise this in case others had the same issue. 

Sorry if this is just creating noise for the team. Is there be a better way to document these kind of things in the future?

* Sinon version: 2.1.0
* Other libraries you are using: mocha, chai, sinon-chai, jsdom, react

**What did you expect to happen?**

After reading the migration guide and the docs I expected everything to carry on working as I wasn't using any removed functionality.

**What actually happens**

Several tests failed with the error `HTMLElement is not defined`.

```javascript
ReferenceError: HTMLElement is not defined
    at typeDetect (node_modules/sinon/node_modules/type-detect/index.js:159:24)
    at typeOf (node_modules/sinon/lib/sinon/util/core/typeOf.js:6:12)
    at match (node_modules/sinon/lib/sinon/match.js:103:16)
    at node_modules/sinon/lib/sinon/call.js:50:43
    at Array.reduce (native)
    at Object.calledWithMatch (node_modules/sinon/lib/sinon/call.js:47:36)
    at Function.spyApi.(anonymous function) [as calledWithMatch] (node_modules/sinon/lib/sinon/spy.js:372:47)
    at Assertion.<anonymous> (node_modules/sinon-chai/lib/sinon-chai.js:95:48)
    at Assertion.ctx.(anonymous function) [as calledWithMatch] (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
```

**My solution**

I need a DOM to test some of our React components so the `type-detect` module was entering the `isDom` block and then failing. I was able to fix this by adding the following to the script that creates the virtual DOM for my tests:

```javascript
global.HTMLElement = () => {
}
```

The [commit with the upgrade](https://github.com/build-canaries/nevergreen/commit/ca1a98988faf48ad4be964db558ecb228a97bb39).

Note: This wasn't an issue in 1.x and I haven't done any investigation as to what changed in 2.x to result in this issue.

ripter:
I am also seeing this issue since upgrading to 2.1.0. I also see it on 2.0.0.

Downgrading to 1.17.7 fixes the issue.

jyboudreau:
Also seeing the same thing, but we're running our tests in Node using Ava.

damenturnbull:
I also encountered a similar error as @GentlemanHal 

```
[...]

PR #1621, opened 2017-11-11
Upgrade type-detect dependency to 4.0.5

mroderick:
This includes a fix for #1377


fatso83:
No point in pushing a new sinon release, though. This patch will automatically be installed by npm due to lack versioning.

mroderick:
Yep

PR #1626, opened 2017-12-06
[feature] Spy passes through calling with `new`

### LINKED ISSUE #1265: Support constructor functions and ES6 classes in spies and stubs
Currently spies and stubs don't call original or stub function with `new` by design - even if they were called with `new`. The suggestion is to add this behaviour to them.

To my knowledge, some of the cases could be easily treated with `createStubInstance`, while in other cases it will be difficult. Since `createStubInstance` creates an instance, a wrapper function should be created manually, and if there should be static methods, they should be created manually too. If test case presumes that wrapper function should be called with or without `new`, this should be coded manually too.

I believe that doing this in `sinon.spy` and `sinon.stub` would result in more consistent behaviour.

* Sinon version : 2.0.0-pre.5
* Environment   : Node.js

 I would expect that something like this will happen inside a spy/stub:

<a name="#expectation">Code</a>
```js
var newTarget;

try {
  newTarget = eval('new.target');
} catch (err) {
  if (err instanceOf SyntaxError)
    newTarget = this && (typeof this === 'object' || typeof this === 'function')
      && this.constructor;
}

if (newTarget) {
  return new originalFn.bind(this, ...args)
} else {
  return originalFn(...args);
}
```

**What actually happens**

Original spied/stubbed classes and constructor function aren't called with `new`. This includes ES6 classes that were transpiled with Babel and have a safeguard against being called directly. This results in error:

> Class constructor MockedBar cannot be invoked without 'new'

**How to reproduce**

```js
const foo = { Bar: class Bar {
  static factory() { return new Bar }
} };

sinon.stub(foo, 'Bar', class MockedBar {
  static factory() { return new MockedBar }
} )

foo.Bar.factory()
new foo.Bar()
```

ProLoser:
#### Purpose (TL;DR)
This changes the spy behavior so that if the spy is `calledWithNew()` then
the original function will also get called with `new`.

This was throwing an error if you tried spying on an ES6 class object.

Fixes #1265

#### How to verify
```
let x = {
  y: class {}
};

spy(x, 'y');

// This would normally throw during callThrough()
new x.y();

assert(x.y.calledWithNew());
```

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -196,7 +196,7 @@ var spyApi = {
                 // Call through with `new`
                 // eslint-disable-next-line new-parens
                 returnValue = new (Function.prototype.bind.apply(this.func || func, [thisValue].concat(args)));
-                
+
                 if (typeof returnValue !== "object") {
                     returnValue = thisValue;
                 }
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -193,10 +193,11 @@ var spyApi = {
             var thisCall = this.getCall(this.callCount - 1);
 
             if (thisCall.calledWithNew()) {
-              // Call through with `new`
-              returnValue = new (Function.prototype.bind.apply(this.func || func, [thisValue].concat(args)));
+                // Call through with `new`
+                // eslint-disable-next-line new-parens
+                returnValue = new (Function.prototype.bind.apply(this.func || func, [thisValue].concat(args)));
             } else {
-              returnValue = (this.func || func).apply(thisValue, args);
+                returnValue = (this.func || func).apply(thisValue, args);
             }
 
             if (thisCall.calledWithNew() && typeof returnValue !== "object") {
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -196,12 +196,12 @@ var spyApi = {
                 // Call through with `new`
                 // eslint-disable-next-line new-parens
                 returnValue = new (Function.prototype.bind.apply(this.func || func, [thisValue].concat(args)));
+                
+                if (typeof returnValue !== "object") {
+                    returnValue = thisValue;
+                }
             } else {
                 returnValue = (this.func || func).apply(thisValue, args);
-            }
-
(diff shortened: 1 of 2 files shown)


lib/sinon/spy.js
```diff
@@ -190,9 +190,15 @@ var spyApi = {
...
             var thisCall = this.getCall(this.callCount - 1);
+
+            if (thisCall.calledWithNew()) {
+              // Call through with `new`
+              returnValue = new (Function.prototype.bind.apply(this.func || func, [thisValue].concat(args)));
+            } else {
+              returnValue = (this.func || func).apply(thisValue, args);
+            }
+
             if (thisCall.calledWithNew() && typeof returnValue !== "object") {
```
fatso83: Perhaps nest this within the first check?

ProLoser:
@fatso83 Updated.

Sorry for doing everything on the Github website vs a streamlined PR, didn't want to take time to clone/install/test/etc when it's all built into CI/CD

fatso83:
Thank you! I had a hard time understanding what was going on, so needed to play with the code and run it in a debugger and examine it before I could accept it. Made a tiny change, but your test was sufficient to prove everything still worked. 

Since the everything done here logically belonged in one commit it was ok that the history was a bit messy, as I could squash everything, including my commit. And yes, the unit test you added was _fine_ :-)

fatso83:
[Released as 4.1.3](https://www.npmjs.com/package/sinon)

lib/sinon/spy.js
```diff
@@ -195,7 +196,7 @@ var spyApi = {
             if (thisCall.calledWithNew()) {
                 // Call through with `new`
                 // eslint-disable-next-line new-parens
```
ProLoser: This line could probably have been removed with your addition.

mroderick:
Thank you!

lib/sinon/spy.js
fatso83: ah, true. thanks. will fix.

PR #1629, opened 2017-12-08
Upgrade to Mochify 5 to run tests in Headless Chrome

### LINKED ISSUE #1405: Discussion: Abandon PhantomJS

mantoni:
This brings the number of tests that are executed up to the same number
as in Chrome and Firefox (+21). These tests where skipped in PhantomJS
due to missing features.

It also allows the WebWorker test script to be simplified removing
mocaccino, phantomic and phantomjs-prebuilt from the dependencies.

Fixes #1405 

---

Open issues:

- [x] Fix Travis build.
- [x] Fix CircleCI build. While it's green, it doesn't seem to run the browser tests at all.
- [x] Figure out why `npm run test-coverage` hits only 639 tests while it used to run 1138.
- [ ] Verify this can be used on Windows.


mantoni:
Uuuh, so the main issue with this PR is that Mochify does requires Node 6 __at install time__ because of the Puppeteer download script being written in a Node 4 incompatible way. This is annoying because we're not even running Mochify on Node 4.

Anybody a clever idea how to either exclude Mochify from the install or how to prevent the install script from being executed?

fatso83:
`error while loading shared libraries: libX11-xcb.so.1: cannot open shared object file: No such file or directory` :nauseated_face:

fatso83:
No clever thoughts on the Node 4 thing. We could just skip it? With regards to the crashing on Node 8:
Installing the[ `libx11-xcb1` library should probably fix it](https://github.com/GoogleChrome/puppeteer/issues/290#issuecomment-322838719). It would require [adding a `pre` section](https://circleci.com/docs/1.0/installing-custom-software/) to the circle ci config (`circle.yml`). Something like

```yaml
dependencies:
  pre:
    - sudo apt-get update; sudo apt-get install libx11-xcb1
```

.travis.yml
```diff
@@ -19,6 +19,8 @@ env:
 before_install:
   - npm config set strict-ssl false
   - npm install coveralls
+  # Prevent mochify -> puppeteer install script to run unnecessarily
+  - if [ "x$TRAVIS_NODE_VERSION" != "x8" ]; then npm config set ignore-scripts true; fi
```
fatso83: That was new! Cool to know.

fatso83:
So it failed again on Node 4, but I don't understand why. It says it executed this:

```
npm config set strict-ssl false
npm install
```

But the `before_install` section contains
```
npm config set strict-ssl false
npm install coveralls
```

fatso83:
Oh, I mixed up the Travis config and the Circle CI.

mantoni:
Yeah, I just experimented with the Travis build as I don't know much about CircleCI yet. I think we could do a similar hack there and maybe only run Mochify on Node 6 and disable the install script otherwise. Once is enough anyway.

mantoni:
Err, I don't really understand what't going on in the CircleCI Node 6 build now. The pre-test thingy sais `sh: =~: unknown operand` while that works fine on the other builds. Also verified the notation locally. 😠🤔

fatso83:
@mantoni It's running POSIX shell, not Bash:
```
#!/bin/sh -eo pipefail
npm config set strict-ssl false
if [[ ! "$(node --version)" =~ ^v6 ]]; then
  # Avoid downloading chromium unnecessarily
  npm config set ignore-scripts true
fi
npm install
sh: =~: unknown operand
```

fatso83:
Anyway, it works, so I think this is ready for merge! The change for shell is outside the config, and perhaps something @fearphage might know about? It's probably in the VM setup for Circle CI.

mantoni:
Well, it doesn't seem to run the browser tests on CircleCI because of that. Testing something else now.

test/webworker/webworker-script.js
```diff
@@ -4,7 +4,7 @@
 
 // Abort if we are not running in a WebWorker
 if (typeof importScripts !== "undefined") {
-    importScripts("../../pkg/sinon.js");
+    importScripts("https://localhost:8080/pkg/sinon.js");
```
fatso83: shouldn't this be just `"/pkg/sinon.js"`?

test/webworker/webworker-support-assessment.js
```diff
@@ -8,7 +8,7 @@ if (typeof Worker !== "undefined") {
 
     describe("WebWorker support", function () {
         it("should not crash", function (done) {
-            var worker = new Worker("file://" + __dirname + "/webworker-script.js");
+            var worker = new Worker("https://localhost:8080/test/webworker/webworker-script.js");
```
fatso83: ditto: could be made relative.

mantoni:
Also slightly disturbing is the drop in tests being executed in `npm run test-coverage`. It only executes 639 tests while it used to run 1138 (see https://travis-ci.org/sinonjs/sinon/jobs/313574869#L1709 vs https://travis-ci.org/sinonjs/sinon/jobs/313396213#L1939)

test/webworker/webworker-script.js
mantoni: Good point. Adjusted in b3cd2d81.

mroderick:
This is an exciting PR, watching from the sidelines (with popcorn)

mantoni:
I got the conditionals fixed for the CircleCI build. Also, it's necessary to activate npm scripts again after installing or `npm run test-node` does nothing.

Now there's a new issue with launching chroming:

```
Error: Failed to launch chrome! spawn /root/sinon/node_modules/puppeteer/.local-chromium/linux-515411/chrome-linux/chrome ENOENT
```

I tried with a ssh session and don't see what the problem is. The issues that people are seeing with missing libraries seem to be something else – maybe we're going to run into those once we can actually run the chrome binary.

The thing is, we have a nicely working Travis build and if CircleCI is giving us a hard time, I don't see the point putting more effort into it.

mantoni:
@sinonjs/sinon-core I'd rather remove the headless test-runs from CircleCI for now (or comment them out). What do you think?

.circleci/config.yml
```diff
@@ -18,12 +18,19 @@ jobs:
           name: Install dependencies
           command: |
               npm config set strict-ssl false
+              NODE_VERSION=$(node --version)
+              if [[ ${NODE_VERSION:0:3} != "v6." ]]; then
```
fearphage: MINOR: `if [[ ! "$(node --version)" =~ ^v6 ]]; then`

.circleci/config.yml
```diff
@@ -18,12 +18,19 @@ jobs:
...
               npm install
+              npm config set ignore-scripts false
       - run:
           name: Pre-Test
-          # ESLint only supports Node >=4
+          # ESLint only supports Node >=4 and Mochify requires Node >= 6
           command: |
-              if [[ ! "$(node --version)" =~ ^v4 ]]; then
+              NODE_VERSION=$(node --version)
+              if [[ ${NODE_VERSION:0:3} == "v6." ]]; then
```
fearphage: MINOR: `if [[ "$(node --version)" =~ ^v6 ]]; then`

mantoni: Yeah, that is what I tried first, but it failed saying that `=~` was an invalid operator.

fearphage: Oh, fair enough. I didn't look at the history.

mantoni:
Since nobody seems to have a strong opinion, I removed the headless test script invocations on CircleCI for now. The tests run successfully on Travis which has to be good enough. If we want to move entirely to CircleCI, the Headless Chromium issue is a blocker.

I also realized that we've been generating the coverage reports on each node version in the Travis build. I've put an `if` around the invocation to only run it on Node 8, like the other headless test runs.

With these changes the build is green. Merge?

fearphage:
We can look into the circleci tests at some point, but I think it should be in a separate PR as long as Travis is running them.

mroderick:
I think we should merge this

mantoni:
Thanks everybody! 🚀

Issue #1446, opened 2017-06-02
Deprecate spy.reset() in favor of the more explicit spy.resetHistory()


fearphage:
Spy currently has a `reset` method which functions as `resetHistory`. It would make the intention and functionality of the method more clear if it were named appropriately.

fearphage:
The idea for this ticket started in #1424.

GCHQDeveloper500:
Picked this one up as it was marked as "Help Wanted" - hope that's okay.

mroderick:
This was done with #1630

PR #1630, opened 2017-12-08
Alias spy.reset to spy.resetHistory

GCHQDeveloper500:
As per #1446 - deprecates and aliases `spy.reset` in favour of `spy.resetHistory`

See ticket. 

#### Solution  - optional
* Change existing `spy.reset` test to instead use `spy.resetHistory`
* Introduce `spy.reset` test to make sure alias works
* Implement changes (just a function name change and an alias)
* Updated docs to reflect changes

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/spies.md
+++ b/docs/release-source/release/spies.md
@@ -395,11 +395,6 @@ If the call did not explicitly return a value, the value at the call's location 
 Resets the state of a spy.
 
 
-#### `spy.reset()`
-
-Alias for `spy.reset();`
-
-
 #### `spy.restore()`
 
 Replaces the spy with the original method. Only available if the spy replaced an existing method.
--- a/docs/release-source/release/spies.md
+++ b/docs/release-source/release/spies.md
@@ -390,9 +390,14 @@ Array of return values, `spy.returnValues[0]` is the return value of the first c
 If the call did not explicitly return a value, the value at the call's location in `.returnValues` will be `undefined`.
 
 
-#### `spy.reset()`
+#### `spy.resetHistory()`
 
 Resets the state of a spy.
+
+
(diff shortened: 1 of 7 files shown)


docs/release-source/release/spies.md
```diff
@@ -390,11 +390,16 @@ Array of return values, `spy.returnValues[0]` is the return value of the first c
 If the call did not explicitly return a value, the value at the call's location in `.returnValues` will be `undefined`.
 
 
-#### `spy.reset()`
+#### `spy.resetHistory()`
 
 Resets the state of a spy.
 
 
+#### `spy.reset()`
```
fatso83: Remove this. If it's deprecated we don't need to spread knowledge of its existence. Only old code-bases should use this.

lib/sinon/spy.js
```diff
@@ -413,6 +413,8 @@ function delegateToCalls(method, matchAny, actual, notCalled) {
     };
 }
 
+spyApi.reset = spyApi.resetHistory;
```
fatso83: This needs to be be wrapped in our deprecation helper. See `lib/sinon/util/core/deprecated.js`

GCHQDeveloper500:
Thanks for the feedback @fatso83 I'll take a look after the weekend.

GCHQDeveloper500:
Okay @fatso83 this is ready for re-review - do you think we need a test for the deprecated `reset` method? Doesn't appear to have been something that people have done in the past.

fatso83:
This was fine, no worries. Thanks a lot!

mroderick:
This has been released with `sinon@4.1.4`

PR #1636, opened 2017-12-19
Fix headless chrome in CircleCI

fearphage:
This re-enables headless chrome testing in CircleCI. It has one [huge hack](https://github.com/sinonjs/sinon/pull/1636/files#diff-1d37e48f9ceff6d8030570cd36286a61R44) in it right now (manually editing a dependency). I think perhaps we should wait on merging this for a cleaner solution. I filed an issue with mochify (mantoni/mochify.js#162) (cc @mantoni).

This reverts small amounts of #1629.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,22 +27,14 @@ jobs:
               else
                 npm install
               fi
-
-              # install latest chrome unstable version
-              if node --version | grep -q '^v6'; then
-                wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
-                echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list
-                apt-get update
-                apt-get install -y --no-install-recommends google-chrome-unstable
-              fi
       - run:
           name: Pre-Test
           # ESLint only supports Node >=4
           command: |
               if node --version | grep -q '^v6'; then
                 npm run lint;
-                npm run test-headless -- --chrome $(which google-chrome-unstable) --allow-chrome-as-root;
-                npm run test-webworker -- --chrome $(which google-chrome-unstable) --allow-chrome-as-root;
+                npm run test-headless -- --allow-chrome-as-root;
+                npm run test-webworker -- --allow-chrome-as-root;
               fi
       - run:
           name: Test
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -18,17 +18,35 @@ jobs:
...
           # ESLint only supports Node >=4
           command: |
-              NODE_VERSION=$(node --version)
-              if [[ ${NODE_VERSION:0:3} == "v6." ]]; then
+              if node --version | grep -q '^v6'; then
                 npm run lint;
+                # HACK: modify mochify to pass args to chrome to allow it to run as root
+                #   --no-sandbox --disable-setuid-sandbox
+                # See https://github.com/mantoni/mochify.js/issues/162
+                sed -i "s#'--allow-insecure-localhost'#'--allow-insecure-localhost', '--no-sandbox', '--disable-setuid-sandbox'#" node_modules/mochify/lib/chromium.js && echo "modified mochify/lib/chromium.js";
```
fearphage: This is gross but a necessary evil for now. This is the [result without the hack](https://circleci.com/gh/fearphage/Sinon.JS/73).

mantoni: We'll get this mixed in Mochify.

fearphage:
@mantoni merged the fix to mochify (mantoni/mochify.js#163) and I tested it and everything is working. Just need an npm version of mochify to point to wrap this puppy up. (Thanks for the quick merge.)

mantoni:
@fearphage Already there. Just published `v5.1.0`.

fearphage:
You're quick!

mantoni:
Out of curiosity, why did you go with a separate download of chromium instead of relying on the puppeteer install script?

fearphage:
> why did you go with a separate download of chromium instead of relying on the puppeteer install script?

I read it was necessary (when researching #1629). I didn't even try the default installed version.

mantoni:
Hm, I think it now used the cache instead of doing a new npm install and therefore doesn't find Chrome.

fearphage:
I'm done fiddling with this one. Review/merge away.

> I think it now used the cache instead of doing a new npm install and therefore doesn't find Chrome.

Did a fresh npm install and that [build failed](https://circleci.com/gh/fearphage/Sinon.JS/107).

mantoni:
@fearphage Thank for trying. Hopefully this gets resolved with a future release of puppeteer.

Issue #1640, opened 2017-12-28
Spying on objects with Symbol method names breaks assertion error messages


fongandrew:
* Sinon version : 4.1.3
* Environment : Ubuntu

Code for formatting assertion error messages does not properly account for non-string method names (i.e. symbols). This may be a regression of https://github.com/sinonjs/sinon/issues/1002.

**Reproduction Steps**

Running this bit of code:
```js
const sinon = require('sinon');
const sym = Symbol('my symbol');
const obj = { [sym]: (a, b) => a + b };
const spy = sinon.spy(obj, sym)
sinon.assert.calledWith(spy, 1, 2);
```

Results in this error:
```
TypeError: Cannot convert a Symbol value to a string
    at String.replace (<anonymous>)
    at Function.printf (/path/to/node_modules/sinon/lib/sinon/spy.js:375:31)
    at Object.assert.(anonymous function) [as calledWith] (/path/to/node_modules/sinon/lib/sinon/assert.js:80:68)
```

Instead of the expected `AssertError: expected Symbol(my symbol) to be called with arguments`.

Possible fix would be to explicitly type-cast symbol to string at https://github.com/sinonjs/sinon/blob/5e131b3e0a88eb703db551d4cbbb8b45f8c25d91/lib/sinon/spy.js#L379 (i.e. `return String(formatter.call(null, spyInstance, args))`)

mroderick:
Thank you for this very thorough bug report!

Would you be up for creating a pull request to fix it?

fongandrew:
Created!

mroderick:
Thank you for your contribution 👍

PR #1642, opened 2017-12-30
Assertion error messages handle symbolic method names

fongandrew:
Resolves https://github.com/sinonjs/sinon/issues/1640 (assertion failure errors with "Cannot convert a Symbol value to a string")

Assertion error messages may reference a method's name (which may now be a symbol, not a string). However, symbols cannot be implicitly converted to strings.

#### Solution
Explicitly cast symbols to strings with `String`. The value returned should match http://www.ecma-international.org/ecma-262/6.0/#sec-symboldescriptivestring.

1. Check out this branch
2. `npm install`
3. Run the following in node:

```
const sinon = require('./lib/sinon');
const sym = Symbol('my symbol');
const obj = { [sym]: (a, b) => a + b };
const spy = sinon.spy(obj, sym)
sinon.assert.calledWith(spy, 1, 2);
```

The error message should say something about expecting `Symbol(my symbol)` to be called, not an error about converting the symbol to a string.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/test/assert-test.js
+++ b/test/assert-test.js
@@ -1649,4 +1649,42 @@ describe("assert", function () {
                           "    actual = { foo: 1 }");
         });
     });
+
+    if (typeof Symbol === "function") {
+        describe("with symbol method names", function () {
+            var obj = {};
+
+            function setupSymbol(symbol) {
+                obj[symbol] = function () {};
+                sinonSpy(obj, symbol);
+            }
+
+            function createExceptionMessage(method, arg) {
+                try { // eslint-disable-line no-restricted-syntax
+                    sinonAssert[method](arg);
+                } catch (e) {
+                    return e.message;
+                }
+            }
+
+            it("should use the symbol's description in exception messages", function () {
+                var symbol = Symbol("Something Symbolic");
+                setupSymbol(symbol);
+
+                assert.equals(createExceptionMessage("called", obj[symbol]),
+                    "expected Symbol(Something Symbolic) to have been " +
+                    "called at least once but was never called");
+            });
+
+            it("should indicate that an assertion failure with a symbol method name " +
+               "occured in exception messages, even if the symbol has no description", function () {
+                var symbol = Symbol();
+                setupSymbol(symbol);
+
+                assert.equals(createExceptionMessage("called", obj[symbol]),
+                    "expected Symbol() to have been " +
+                    "called at least once but was never called");
+            });
+        });
+    }
 });
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -376,7 +376,7 @@ var spyApi = {
             formatter = spyApi.formatters[specifyer];
 
             if (typeof formatter === "function") {
-                return formatter.call(null, spyInstance, args);
+                return String(formatter.call(null, spyInstance, args));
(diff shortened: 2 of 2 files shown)


test/assert-test.js
```diff
@@ -1649,4 +1649,45 @@ describe("assert", function () {
                           "    actual = { foo: 1 }");
         });
     });
+
+    if (typeof Symbol === "function") {
+        describe("with symbol method names", function () {
+            function setupSymbol(symbol) {
```
fatso83: Could you please simplify the logic by making it more functional and more explicit? It seems needlessly complex right now :smiley:

- change the setup method
- remove the `bind`  call in line 1662
- simplify `this.message` function (see below) and rename it
- beforeEach => before
- improve test descriptions

test/assert-test.js
```diff
@@ -1649,4 +1649,45 @@ describe("assert", function () {
...
+                this.obj = {};
+                this.obj[symbol] = function () {};
+                sinonSpy(this.obj, symbol);
+            }
+
+            beforeEach(function () {
+                this.setupSymbol = setupSymbol.bind(this);
```
fatso83: remove

test/assert-test.js
```diff
@@ -1649,4 +1649,45 @@ describe("assert", function () {
...
+
+                /*eslint consistent-return: "off"*/
+                this.message = function (method) {
```
fatso83: Simplify into
```javascript 
function createExceptionMessage(method, arg) {
    try { sinonAssert[method](arg); }
    catch (e) { return e.message; }
}
```

All the uses have at most one argument so we can drop the slicing.

test/assert-test.js
```diff
@@ -1649,4 +1649,45 @@ describe("assert", function () {
...
+                    try { // eslint-disable-line no-restricted-syntax
+                        sinonAssert[method].apply(sinonAssert, [].slice.call(arguments, 1));
+                    } catch (e) {
+                        return e.message;
+                    }
+                };
+            });
+
+            it("assert.called exception message with symbol with description", function () {
```
fatso83: The sentence is not valid English. Make it read out aload in a natural way, for instance `it("should create an exception message for symbols that describe the symbol"`

test/assert-test.js
```diff
@@ -1649,4 +1649,45 @@ describe("assert", function () {
...
+                var symbol = Symbol("Something Symbolic");
+                this.setupSymbol(symbol);
```
fatso83: `setupSymbol(symbol);`

test/assert-test.js
```diff
@@ -1649,4 +1649,45 @@ describe("assert", function () {
...
+
+                assert.equals(this.message("called", this.obj[symbol]),
```
fatso83: `assert.equals(createExceptionMessage("called", obj[symbol]), ...`

fongandrew:
Cleaned up the tests. Please take another look. Thanks!

fatso83:
Very nice, thanks!

mroderick:
@fatso83 shall we cut a new release?

fatso83:
yes, please :)

mroderick:
This has been released with `sinon@4.1.4`

PR #1653, opened 2018-01-13
Upgrade dependencies

mroderick:
This PR upgrades outdated dependencies.

I've looked at the changelogs and made necessary changes, I don't think this is a risky upgrade as only `supports-color` is used by end users, and is covered with updated tests.


mantoni:
👍

PR #1658, opened 2018-01-16
Use "stale" as label for stale issues

mroderick:
This PR changes the label to use for stale issues from "wontfix" to "stale"


mroderick:
I'll update the other repositories directly

PR #1660, opened 2018-01-16
Fix #176 by ensuring the most specific fake is used in .withArgs if several of them match

### LINKED ISSUE #176: Chaining sinon.stub() with multiple .withArgs only stubs the last declaration
Chaining sinon.stub() with multiple .withArgs only stubs the last declaration

var stubbedFunction = sinon.stub().withArgs(argument1).returns(value1).withArgs(argument2).returns(value2);

then only the last declaration (.withArgs(argument2).returns(value2)) is actually stubbed

davidgf:
As discussed in  #176, `.withArgs` is implemented so that it picks the last matching fake. However, if the last time `.withArgs` was called passing `sinonMatch.any`, it will always pick that fake, even if there's any other more accurate match. For example:

```javascript
const s = sinon.stub();
s.withArgs(1).returns(1)
 .withArgs(2).returns(2)
 .withArgs(sinon.match.any).returns('my-default-value')

s(1) // my-default-value
s(2) // my-default-value
s(5) // my-default-value
```

This commit avoids this and returns the appropriate fake.

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -10,6 +10,7 @@ var wrapMethod = require("./util/core/wrap-method");
 var stubEntireObject = require("./stub-entire-object");
 var throwOnFalsyObject = require("./throw-on-falsy-object");
 var valueToString = require("./util/core/value-to-string");
+var sinonMatch = require("./match");
 
 var slice = Array.prototype.slice;
 
@@ -82,6 +83,10 @@ function getCurrentBehavior(stubInstance) {
     var currentBehavior = stubInstance.behaviors[stubInstance.callCount - 1];
     return currentBehavior && currentBehavior.isPresent() ? currentBehavior : getDefaultBehavior(stubInstance);
 }
+
+function fakeUsesSinonAnyMatcher(fake) {
+    return fake.matchingArguments.length === 1 && fake.matchingArguments[0] === sinonMatch.any;
+}
 /*eslint-enable no-use-before-define*/
 
 var uuid = 0;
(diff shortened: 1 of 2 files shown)


mantoni:
Why would we treat `any` different from other matchers? In your example, shouldn‘t a `number` matcher also have lower priority than an actual number? You‘d end up implementing a scoring system for argument matching. I don‘t think we want to go there.

As commented on the original issue you can work around this by putting the `any` matcher first.

An alternative solution could be a custom matcher or to configure the spy with a custom implementation by passing in a function.

davidgf:
@mantoni fair enough, I see the point in the `number` matcher example, so I absolutely understand why you wouldn't want this. Anyway, it seems pretty odd that the stub can return the `sinonMatch.any` value even though there's an exact match.

cincodenada:
I landed here and just wanted to say thanks for this discussion, and also for leaving it as-is! This issue (and #176) gave me a better solution to the problem of overriding any `withArgs()` later on, so I can do something like this:
```
let responseStub
beforeEach(function() {
  responseStub = Sinon.stub()
    .resolves({data: 'the usual thing'})
    .withArgs(match({specific: 'match'}).resolves({data: 'some other usual thing'})
})
it('should behave differently if responseStub returns unexpected things', function() {
  responseStub.withArgs(match.any).resolves({error: 'this is a totally different response'})
  const response = responseStub({specific: 'match'})
  should.exist(response.error)
  should.not.exist(response.data)
  should.equal('this is a totally different response', response.error)
})
```

Being able to override a stub's behavior entirely is very useful for testing error cases where the thing being stubbed is in some condition where it doesn't pay attention to the args being passed in. So there's a use case where this behavior is not only less complicated, but also useful!

Issue #1624, opened 2017-11-16
sinon.match.each ?


cyrilchapon:
It would be very great to have a matcher that exactly act like `sinon.match.has` but on every enumerated value in an array / object

Use case :
```
          ...
          const res = httpMocks.createResponse();
          res.json = sinon.spy(res.json);

          return stuffController.listStuff(req, res, next)
          .then(function() {
            assert.calledWith(res.json, sinon.match.array);
            assert.calledWith(res.json, sinon.match.has('length', expectedStuff.length));
            assert.calledWith(res.json, sinon.match(function(result) {
              return _.every(result, function(stuff) {
                return stuff instanceof Stuff;
              });
            }, 'Every return object is an stuff'));
          });
```

That could be turned to
```
          ...
          const res = httpMocks.createResponse();
          res.json = sinon.spy(res.json);

          return stuffController.listStuff(req, res, next)
          .then(function() {
            assert.calledWith(res.json, sinon.match.array);
            assert.calledWith(res.json, sinon.match.has('length', expectedStuff.length));
            assert.calledWith(res.json, sinon.match.each(result, sinon.match.instanceOf(Stuff)));
          });
```

cyrilchapon:
It could be named `sinon.match.each`, or lodash-like `sinon.match.every`

cyrilchapon:
In addition, one could find useful to check that at least one enumerated value matches.
We could have `sinon.match.some` for this

mroderick:
I think that would be a great addition to matchers.

My preference leans more towards using names that are more like [`Array.prototype.every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) and [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some), than using lodash conventions. Keeping in mind just how inconsistent Sinon's API currently is, I don't think that preference matters that much.

mroderick:
This has been implemented by #1661 and shipped as `sinon@4.2.0`

PR #1661, opened 2018-01-16
Add match.every and match.some (#1624)

nivsherf:
Add match.every and match.some as requested in issue #1624.

Steps to verify:

1. npm install
2. npm test

### DIFF
--- a/test/match-test.js
+++ b/test/match-test.js
@@ -688,6 +688,79 @@ describe("sinonMatch", function () {
         }
     });
 
+    describe(".some", function () {
+        it("throws if given argument is not a matcher", function () {
+            assert.exception(function () {
+                sinonMatch.some({});
+            }, "TypeError");
+            assert.exception(function () {
+                sinonMatch.some(123);
+            }, "TypeError");
+            assert.exception(function () {
+                sinonMatch.some("123");
+            }, "TypeError");
+        });
+
+        it("returns matcher", function () {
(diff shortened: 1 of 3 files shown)


test/match-test.js
```diff
@@ -615,6 +615,79 @@ describe("sinonMatch", function () {
...
+        it("wraps the given matcher message with an \"every()\"", function () {
+            var every = sinonMatch.every(sinonMatch.number);
+
+            assert.equals(every.toString(), "every(typeOf(\"number\"))");
+        });
+
+        it("fails to match anything that is not an object or an iterable", function () {
+            var every = sinonMatch.every(sinonMatch.any);
+
+            assert.isFalse(every.test(1));
```
mroderick: Instead of using `assert.isFalse`, you can use [`refute`](http://docs.busterjs.org/en/latest/modules/referee/#refute)... you just have to import it from `referee` at the top of the file.

nivsherf:
Thanks for your feedback @mroderick. Edited the tests and added documentation for the new matchers.

mroderick:
This has caused tests in IE11 for `master` to fail. I am investigating. Once we have figured out what the error is caused by, I'll push a fix and we can get this shipped.

mroderick:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Browser_compatibility

mroderick:
This has been released as `sinon@4.2.0`.

PR #1666, opened 2018-01-21
Add eslint-plugin-ie11

mroderick:
This PR adds `eslint-plugin-ie11`, which would have prevented me from merging #1661, which caused a failing build (and more work for me). Instead, the author would have seen the error and fixed it.

#### Background
In IE11 `Map` and `Set` constructors do not accept arguments, but instead of throwing an error, they fail silently, and just produce the empty state.

```js
// works in modern browsers, not in IE11
new Set([1, 2]);
```

1. Check out this branch
2. `npm install`
3. Add the code above to any linted JS file, save it
4. `npm run lint`
5. Observe the failure

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -641,6 +641,15 @@
         "parse-asn1": "5.1.0"
       }
     },
+    "browserify-zlib": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/browserify-zlib/-/browserify-zlib-0.2.0.tgz",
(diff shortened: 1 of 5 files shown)


mantoni:
For this type of issues we cannot have enough automation. Thank for digging out the plugin. I didn't know about it 👍

mroderick:
Thanks.

I also looked into eslint-plugin-compat (which I use elsewhere), but that doesn't catch this specific issue.

mroderick:
I'll update the other projects to add this

Issue #1509, opened 2017-08-03
Name differences


henit:
It might be logical reasons that I don't know about, but I wonder about some of the name differences between sinon-methods:

spy.notCalledWith
assert.neverCalledWith

Should they have the same name?

stub.callsFake
stub.callsArg
stub.callsArgWith
stub.callThrough
stub.callArg
stub.callArgWith

I often end up checking the docs where there is 'calls*' and where there is 'call*'.

mroderick:
Thank you for pointing it out!

I agree, there should be more consistency in naming.

I think an approach like this could work:

* create a `README` or wiki page, describing the naming conventions
* rename all methods that do not follow the naming conventions
* add deprecation warnings to all deprecated method names
* publish new version
* in a later `MAJOR` version, remove the deprecated methods

Do you want to take a stab at it?

henit:
That sounds like a good approach. 

I think Sinon is a very good library, and would very much like to contribute to it if possible. Unfortunately, there is no room for this at the moment, hopefully it will change soon.

mantoni:
Related discussion in #1000.

PR #1667, opened 2018-01-21
Fix inconsistent naming

mroderick:
This PR is a fix for #1509, it renames the following methods

* `stub.callArg ` => `stub.callsArg`
* `stub.callArgOnWith` => `stub.callsArgOnWith`
* `stub.callArgWith ` => `stub.callsArgWith`
* `stub.callThrough` => `stub.callsThrough`

AFAICT, this fixes the inconsistent naming of the `calls*` methods.

I've added deprecation warnings to all the old names. We can remove them in the next major version, or the one after that.

1. Carefully read the PR
1. Observe that tests pass

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -454,9 +454,9 @@ Useful if a function is called with more than one callback, and simply calling t
 }
 ```
 
-#### `stub.callArgWith(argNum, [arg1, arg2, ...])`
+#### `stub.callsArgWith(argNum, [arg1, arg2, ...])`
 
-Like `callArg`, but with arguments.
+Like `callsArg`, but with arguments.
 
 #### Asynchronous calls
 
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -435,7 +435,7 @@ Like `yield`, `yieldTo` grabs the first matching argument, finds the callback an
 ```
 
 
-#### `stub.callArg(argNum)`
+#### `stub.callsArg(argNum)`
 
 Like `yield`, but with an explicit argument number specifying which callback to call.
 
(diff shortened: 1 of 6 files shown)


mantoni:
I’m irritated. Maybe I’m confusing things now, but I though `callThrough` means “do call through reicht now” and `callsThrough` means “when this stub is called it will call through”. Like, programming the stub to do something vs doing it right away. No?

mroderick:
I am not happy about how stubs have grown and grown, hence #1586. `fake` is going to be much simpler, I am going to try and protect it from the feature creep.

This PR is about making the stub api consistent.

I *think* I understand your point, but it's very difficult to spot from the documentation. Perhaps what's needed here is a documentation improvement, and perhaps an alias, for when people still don't get it?

mantoni:
Looked a little closer at the changes now and realized you changed it on calls. Your description was talking about renaming those functions on stubs. However, both namings already exist on stubs with the "convention" that I was trying to line out above.

mroderick:
This whole API is a bit confusing

mantoni:
Yes, I agree. The documentation changes you made are not correct. E.g. `stub.callArg` exists and does something different from `stub.callsArg` which also exists.

mroderick:
😢

mantoni:
Ok, let me illustrate what I mean with two examples that both pass:

```js
var fetch = sandbox.stub(remote, 'fetch')
var spy = sinon.spy()

fetch(spy)

sinon.assert.notCalled(spy)

fetch.yield()

sinon.assert.called(spy)
```

vs

```js
var fetch = sandbox.stub(remote, 'fetch')
var spy = sinon.spy()
fetch.yields()

fetch(spy)

sinon.assert.called(spy)
```

The same convention is implemented for `callArgs` and `callsArgs` etc.

I checked the stub documentation, and I think it's correct as is. The code changes you made are effectively working with calls. IMO the calls API should be in line with the convention used on stubs.

Issue #1627, opened 2017-12-07
Sinon stub with an object and a property has low performance.


NoamPaz:
Sinon stub with an object and a property name has low performance due to new Error in the wrap-method.js.

When I create stub with sandbox.stub and pass an object and a property name, it reach this line in the wrap-method.js:
method.stackTrace = (new Error("Stack Trace for original")).stack;

https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/core/wrap-method.js#L110

This line takes about 0.2-0.3 second, therefore my tests runs for 6 minutes.
If I comment it out, it runs for 7 seconds (instead of 6 minutes).

1) I'm using the latest version of sinon.
2) I tried to use node v6 and node v8. in both the issue has occurred.
3) I'm using Linux.
4) here is an example of the stack trace this new Error generates:

> Error: Stack Trace for original
    at wrapMethod (/home/noam/t2kdev/echo/node_modules/sinon/lib/sinon/util/core/wrap-method.js:110:26)
    at stub (/home/noam/t2kdev/echo/node_modules/sinon/lib/sinon/stub.js:60:44)
    at Object.stub (/home/noam/t2kdev/echo/node_modules/sinon/lib/sinon/collection.js:102:33)
    at global.createStub (/home/noam/t2kdev/echo/.tmp/mocha-webpack/1512641354923/webpack:/node_modules/@t2k/t2k-js-scripts/test-setup.js:33:1)
    at Context.<anonymous> (/home/noam/t2kdev/echo/.tmp/mocha-webpack/1512641354923/webpack:/test/spec/js/components/App.spec.jsx:208:4)
    at callFnAsync (/home/noam/t2kdev/echo/node_modules/mocha/lib/runnable.js:371:21)
    at Hook.Runnable.run (/home/noam/t2kdev/echo/node_modules/mocha/lib/runnable.js:318:7)
    at next (/home/noam/t2kdev/echo/node_modules/mocha/lib/runner.js:309:10)
    at Immediate.<anonymous> (/home/noam/t2kdev/echo/node_modules/mocha/lib/runner.js:339:5)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
    at processImmediate [as _immediateCallback] (timers.js:617:5)

fatso83:
[...]

PR #1671, opened 2018-01-23
Fix to #1627

NoamPaz:
Fix issue #1627 by creating an error object and use the property 'stack' only when needed, instead of read stack property in the creation time of the error object.

The slowness seems to come from accessing the .stack property on the error, as v8 seems to go through considerable trouble in order to convert it from its internal representation to the string version it uses to report errors.

#### Solution  
the fix for this is to change the line to method.stackTraceError = new Error("Stack Trace for original"), and only access the .stack property if an error needs to be reported.

####Test
passed: 'npm run test' and 'npm run lint'

### DIFF
--- a/lib/sinon/util/core/wrap-method.js
+++ b/lib/sinon/util/core/wrap-method.js
@@ -43,8 +43,8 @@ module.exports = function wrapMethod(object, property, method) {
         }
 
         if (error) {
-            if (wrappedMethod && wrappedMethod.stackTrace) {
-                error.stack += "\n--------------\n" + wrappedMethod.stackTrace;
+            if (wrappedMethod && wrappedMethod.stackTraceError) {
+                error.stack += "\n--------------\n" + wrappedMethod.stackTraceError.stack;
             }
             throw error;
         }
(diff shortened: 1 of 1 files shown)


mroderick:
Thank you for taking the time to investigate this issue and making a pull request 🚀 

Your example script in the gist doesn't exercise actual Sinon code. I wanted to make sure that there would be an improvement for an example that exercises Sinon.

So, I created a little benchmark script using [Benchmark.js](https://github.com/bestiejs/benchmark.js).

```js
"use strict";

var Benchmark = require("benchmark");
var suite = new Benchmark.Suite();
var sinon = require("./lib/sinon");
var sandbox = sinon.createSandbox();

// add tests
suite.add("stub test", function () {
    var obj = {
        hello: function hello() {
            return "world";
        }
    };

    sandbox.stub(obj, "hello");
});
// add listeners
suite.on("cycle", function (event) {
    console.log(String(event.target));
});
suite.on("complete", function () {
    console.log("Fastest is " + this.filter("fastest").map("name"));
});
// run async
suite.run({
    "async": true
});
```

When run against current `master` 5063c3353549599dba4b1de84976e625a3f92f2d, these are the results on my ageing laptop:

```text
stub test x 4,852 ops/sec ±3.78% (59 runs sampled)
```

Run against this branch:

```text
stub test x 10,129 ops/sec ±3.17% (64 runs sampled)
```

The benchmarks show a pretty significant speedup!

Excellent pull request 👍

mroderick:
This has been published as `sinon@4.2.1`

PR #1672, opened 2018-01-23
"throttle" is actually a "debounce" in the example

ALeschinsky:


mroderick:
Thank you for the contribution 👍

PR #1688, opened 2018-02-13
Document that it is allowed to assert on dedicated spy calls.

### LINKED ISSUE #1471: Doc: Pass a SinonSpyCall to sinon.assert.calledWith()
* Sinon version : 2.3.5
* Environment   : -
* Example URL   : -
* Other libraries you are using: -

Documentation states that a `Spy` instance needs to be passed to `sinon.assert.calledWith()` (and others like `calledOn()` and `calledWithExactly`).

I found the passing a `SpyCall` works also fine and it sounds reasonable that this is correct.

Is this something which works just per luck and can go away at any time or is it intended that this works and documentation should be updated?

Flarna:

Document that some asserts can be done on a `SpyCall` not only on a `Spy`. Fixes #1471

I updated only master of the doc. Once I'm told that wording is ok and change is accepted I can update also older releases.

N/A - just doc update

#### Checklist for author

- [x] `npm run lint` passes


mroderick:
Thank you for your pull request @Flarna 

Reviewing this pull request allowed me to discover that we are missing tests for some of the assertions. [Only in one place do we verify that an assertion also works with a `spyCall` as the first argument](https://github.com/sinonjs/sinon/blob/2c2bc0dbd2e4ed6666bdfd75fc6fc9a0fd30ab44/test/assert-test.js#L732-L741).

Would you be up for expanding your pull request?

* Add missing documentation for all the assertions that supports both `spy` and `spyCall`
* Add missing tests for all the assertions that supports both `spy` and `spyCall`
* Change documentation examples to use `spyOrSpyCall` as the name of the first argument
    `sinon.assert.calledWith(spyOrSpyCall, arg1, arg2, ...);`

Flarna:
sure, will do that - hopefully soon.

> Add missing documentation for all the assertions that supports both spy and spyCall

I thought I found all of them - but looking again I think that I forgot `calledWithNew`. Do you know if there are others which should support `spy` and `spyCall`?

Flarna:
@mroderick Updated docs and added tests. I tried to act like a a roman in rome in the tests; hope it fits.

Still pending is the merge into older docs.

mroderick:
Looks good to me 👍 

Do go ahead and fold the documentation updates into the older docs.

Flarna:
@mroderick older docs are updated now

Issue #903, opened 2015-10-27
It would be useful that yield, yieldTo, etc returns the return value of the yielded function


julienw:
``` javascript
var stub = sinon.stub({
  hello: function(callback) {}
});

stub.hello(
  () => { console.log('hello !'); return 'world'; }
);

var value = stub.hello.yield();
console.log(value); // should be 'world', currently false
```

fearphage:
What's the use case? Can you show a real world example?

julienw:
We have an API that goes like this:

```javascript
MessageManager.getMessages({
  each: <function>
  done: <function>
});
```

It's an iterator that calls the `each` callback several times until there is no more items or until the `each` callback returns false.

When testing the caller, I mock getMessages, but I want to check that the `each` callback returns `false` in a specific situation.

So the test could go like this:

```
sinon.stub(MessageManager, 'getMessages');
callTestedMethod();
var result = MessageManager.getMessages.yieldTo('each');
assert.isFalse(result);
```

I don't think there is a way to do it currently, besides stubbing with a custom function.

fearphage:
Aren't you passing the object to `getMessages`? It seems like you're tightly coupling two unrelated things.

```js
var object = {each: fn};

assert.isFalse(object.each(5));
// or
assert.isFalse(fn(5));
```

julienw:
The tested code is passing the object to getMessages.

Yet I realize from your comment that I could do:
```
sinon.stub(MessageManager, 'getMessages');
callTestedMethod();
var object = callTestedMethod.lastCall.args[0];
var result = object.each();
assert.isFalse(result);
```

But this defeats the purpose of yieldTo IMO. Using `args` is not really expressive... If you tell me "ok, just use `args`", then why do we have all the useful methods in the API?

tshelburne:
@fearphage This is also useful in the case of caching - cache-manager provides the following interface:

```javascript
cache.wrap('request-key', function() {
  return fetch('some/url')
})
.then(function(res) {...})
.catch(function(err) {...})
```

In this case, being able to stub `cache.wrap` to yield *and* return the value of that yield is essential to test.

globetro:
This would also be helpful for testing asynchronous callbacks. Example:

```javascript
stub.onCallback(async () => { /* do some async stuff */});

it('executes the async stuff', async () => {
  await stub.onCallback.yield();
  // expect async callback to have happened
});
```

julienw:
[...]

PR #1689, opened 2018-02-13
Change return value of yield and callArg

stoically:
Hi, thanks for the work on sinon!

#### Purpose

Instead of returning false this change makes `yield`, `yieldOn`, `yieldTo`, `yieldToOn`, `callArg`, and `callArgOn` return the callbacks return values. This was also requested in #903.

#### Background

When doing blackbox feature tests, where a codebase itself is not exposed to the tests, but instead adds listeners to global sinon stubs (imitating the WebExtensions API in this case), then it's necessary to trigger behavior in the codebase by yielding those listeners. Currently we have to rely on `process.nextTick` to wait for resolving promises and we have no way of getting the return value, e.g.

https://github.com/mozilla/multi-account-containers/blob/61da6b5e99e3265ed8fcfb40cc46e3ae01d0c2fa/test/issues/940.test.js#L32-L38

With this change we could await the actual returned promise and also verify return values.

#### Solution

Thanks to the tidy sinon codebase this was only a matter of adding some return statements and a slight change to `delegateToCalls`, which now accepts an additional parameter that triggers to return callback values instead of true or false. The parameter is only set to true for the yield and callArg functions. Also added tests and a sentence to the docs.

#### How to verify
1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

#### Note

Like suggested in #903 did I test this against #1586 and it's working fine.

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -409,6 +409,8 @@ Invoke callbacks passed to the `stub` with the given arguments.
 
 If the stub was never called with a function argument, `yield` throws an error.
 
+Returns an Array with all callbacks return values in the order they were called, if no error is thrown.
+
 Also aliased as `invokeCallback`.
 
 
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -98,20 +98,20 @@ var callProto = {
     },
 
     callArg: function (pos) {
-        this.args[pos]();
+        return this.args[pos]();
     },
 
     callArgOn: function (pos, thisValue) {
-        this.args[pos].apply(thisValue);
+        return this.args[pos].apply(thisValue);
     },
 
     callArgWith: function (pos) {
-        this.callArgOnWith.apply(this, [pos, null].concat(slice.call(arguments, 1)));
+        return this.callArgOnWith.apply(this, [pos, null].concat(slice.call(arguments, 1)));
     },
 
     callArgOnWith: function (pos, thisValue) {
         var args = slice.call(arguments, 2);
-        this.args[pos].apply(thisValue, args);
+        return this.args[pos].apply(thisValue, args);
     },
 
     throwArg: function (pos) {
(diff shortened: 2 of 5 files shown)


mroderick:
This PR looks good to me, thank you @stoically 

Just to be sure that I haven't overlooked anything major, I'd like one of the other stewards of Sinon to  give it a review before we merge it.

mantoni:
Nice work! I agree this is good to merge 👌

mroderick:
This has been published as `sinon@4.4.0`

ungrim97:
Does this also provide the functionality to `yields`, `yieldsOn`, `yieldsTo`, `yieldsToOn`, `callsArg`, `callsArgOn` and `callsArgWith`?

mantoni:
@ungrim97 Can you provide an example where this would be useful?

ungrim97:
@mantoni for exactly the same reasons as the original change was proposed. 

I currently have a function that I am testing that returns the result of a function (that I am stubbing) that receives a call back as its argument and returns the a Promise that is resolved upon some calculation:

```
module.exports = class MyClassWithDropTableFunction {
  constructor(data) {this.knex = data.knex;}

  function dropTable(tableName) {
    return this.knex.transaction(txn => {
      return Promise.resolve(txn.schema.dropTableIfExists(tableName));
    });
};
```
```
test('dropTable', async () => {
  const txn = { schema: { dropTableIfExists: sinon.stub().withArgs('testTable') } };
  const knex = { transaction: sinon.stub().yield(txn) };
  const obj = new MyClassWithDropTableFunction({knex: knex});

  await obj.dropTable('testTable');

  assert.ok(txn.schema.dropTableIfExists.calledOnce, 'table dropped');
});
```

Hopefully that makes sense.

Fundimentally my understanding is that yields et al are functionally similar to yield et al. Only that they only call the first callback available. This I would expect that they would provide a similar behaviour for similar reasons.

mantoni:
@ungrim97 Yes, I see now how this makes sense with promises. The change adressed here does not implement the return behavior for `yields` and friends.

Can I ask you to either file a new issue, or even better, make a pull request?

ungrim97:
@mantoni Happy to make a PR if I can figure out where it needs to go. Will take a look in a bit and see if I can figure it out, otherwise will raise an issue.

Issue #1694, opened 2018-02-20
Sinon internal error TypeError: this.args[pos].apply is not a function


Alexsey:
* Sinon version: **4.3.0**
* Environment : **NodeJS**
* Other libraries you are using: **proxyquire**, **mocha**

**What did you expect to happen?**
Error with a descriptive message should be thrown (probably I'm doing something wrong?)

**What actually happens**
Internal Sinon error is thrown
```
TypeError: this.args[pos].apply is not a function
    at Object.callArgOnWith (node_modules\sinon\lib\sinon\call.js:114:24)
    at Object.callArgWith (node_modules\sinon\lib\sinon\call.js:109:28)
    at Function.spyApi.(anonymous function) (node_modules\sinon\lib\sinon\spy.js:408:47)
    at Context.it (foo.test.js:13:12)
    at <anonymous>
```

**How to reproduce**
`foo.js`:
```js
const rp = require('request-promise')

module.exports = () => rp('http://www.google.com')
```
`foo.test.js`
```js
const sinon = require('sinon')
const sandbox = sinon.createSandbox()
const proxyquire = require('proxyquire')

const rpStub = sandbox.stub()

const foo = proxyquire('./foo', {
    'request-promise': rpStub
})

it('should throw from Sinon internals', async () => {
    await foo()
    rpStub.callArg(0)
})
```

fatso83:
This is a usage question. We try to keep this issue tracker for bugs. Try posting to StackOverflow where you'll find plenty of help. 

Your problem is twofold:
- you await the stub, but you forget to return a promise from the stub. See the docs for stub#resolves().
- you try to call the argument passed to the stub, which makes no sense as it is supposed to receive a string

Alexsey:
@fatso83 my issue is not a question. It's about Sinon crash. Crash is something that should not happen even if something is doing wrong. If something is doing wrong then error with descriptive message should be thrown

Actually, it's a bug and to fix it we need to add checks in places like [this](https://github.com/sinonjs/sinon/blob/master/lib/sinon/call.js#L101), [this](https://github.com/sinonjs/sinon/blob/master/lib/sinon/call.js#L105) and [this](https://github.com/sinonjs/sinon/blob/master/lib/sinon/call.js#L114) that argument is a function. I can make a PR for that if you would reopen this issue

BTW `await` of non-promise is 100% valid

fatso83:
[...]

PR #1699, opened 2018-02-24
Ensure that argument is a function befor attempting to call it

Alexsey:
Resolves issue https://github.com/sinonjs/sinon/issues/1694

BTW it looks like constructions like
```js
assert.exception(function () {
    call.callArgOnWith({}, thisObj);
}, "TypeError");
```
doesn't check that error is actually `TypeError` - you can change `TypeError` to any other string, even to empty, and the test will remain green. There are a lot of them in the tests

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -98,10 +98,12 @@ var callProto = {
     },
 
     callArg: function (pos) {
+        this.ensureArgIsAFunction(pos);
         return this.args[pos]();
     },
 
     callArgOn: function (pos, thisValue) {
+        this.ensureArgIsAFunction(pos);
         return this.args[pos].apply(thisValue);
     },
 
@@ -110,6 +112,7 @@ var callProto = {
     },
 
     callArgOnWith: function (pos, thisValue) {
+        this.ensureArgIsAFunction(pos);
         var args = slice.call(arguments, 2);
         return this.args[pos].apply(thisValue, args);
     },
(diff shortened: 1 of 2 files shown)


Alexsey:
Do not merge, please. There are some issues. Going fix them right away

Alexsey:
Now it's ok. BTW it looks like constructions like
```js
assert.exception(function () {
    call.callArgOnWith({}, thisObj);
}, "TypeError");
```
doesn't check that error is actually `TypeError` - you can change `TypeError` to any other string or remove it completly and test will remane green

mroderick:
@Alexsey you're correct about this codebase and it's incorrect use of the second argument to `assert.exception`. That should be an `Object`, where the `name` property describes the type of error.

http://docs.busterjs.org/en/latest/modules/referee/#exception

https://github.com/sinonjs/referee/blob/master/lib/referee.js#L546-L584

We have recently taken over stewardship of `referee`. The plan is to retrofit it with newer developer tools and unit tests ported to Mocha. That'll take awhile. Once that work is done, I'd be happy to make the api better.

In the meantime, I've created #1702 to track that problem.

Alexsey:
Updated

Alexsey:
Another BTW - I have noticed that all `yield` object properties are defined as `"yield"`, but according to strict mode this word is reserved only for use as variable or function name but use it as property field is ok

mroderick:
This has been published to npm as `sinon@4.4.2`

PR #1700, opened 2018-02-25
Next fakes

mroderick:
This PR replaces #1586. 

It is a solution for the proposal in #1559, and will be merged into the `next` branch. 

The only difference between the two is that I used `git rebase -i` a lot and tidied up the history, so it will read better for Future Contributor

#### Tasks

- [x] Make the main `sinon` object a default `sandbox`
- [x] Create `sandbox.replace` method
- [x] Document `sandbox.replace`
- [x] Create `sandbox.replaceGetter` method
- [x] Document `sandbox.replaceGetter`
- [x] Create `sandbox.replaceSetter` method
- [x] Document `sandbox.replaceSetter`
- [x] Create `sinon.fake`
- [x] Clean unwanted methods from `fake` [`.withArguments`, etc]
- [x] Document `sinon.fake`
- [x] Put release candidates on npm as `sinon@next` to get feedback


mroderick:
Thanks

Issue #1697, opened 2018-02-21
sinon@next: Shouldn't fake.yields() have the same semantics as stub().yields()?


mantoni:
The current implementation of `fake.yields()` is completely different from what `yields` meant for stubs. Is this intentional @mroderick? I would prefer keeping it in line with the `stub` semantics so that a migration from stubs to fakes is not becoming too confusing.

I think what `fake.yields(callback)` is currently doing should really be named `fake.calls(callback)`.

Via #1586

mroderick:
> The current implementation of fake.yields() is completely different from what yields meant for stubs. Is this intentional? I would prefer keeping it in line with the stub semantics so that a migration from stubs to fakes is not becoming too confusing.

I agree. Let's try to make the transition as easy as possible.

> I think what `fake.yields(callback)` is currently doing should really be named `fake.calls(callback)`.

That would be the same as `fake(callback)` or `sinon.spy(callback)`, wouldn't it? If that's correct, then we don't need to add it.

mantoni:
Yes, you're right. We actually don't need "calls". I just re-discovered that there is another variation of `yields`: `yieldsRight`. I remember being confused that `yields` would invoke the first function it finds in the argument list. `yieldsRight` is what would have been intuitive to me.

I would suggest to make `yields` very strict and attempt to invoke the last argument as a function. If it's not a function it should throw. If someone wants to invoke a different argument as a function, they can do so with `callsArg(index, args...)`.

mroderick:
> I would suggest to make `yields` very strict and attempt to invoke the last argument as a function. If it's not a function it should throw

That sounds like a reasonable approach to me 👍 

> If someone wants to invoke a different argument as a function, they can do so with callsArg(index, args...).

When creating a fake, these properties were removed (as they contribute to make it stateful, or are just adding noise)

* `instantiateFake`
* `callArg`
* `callArgOn`
* `callArgOnWith`
* `callArgWith`
* `invokeCallback`
* `throwArg`
* `withArgs`
* `yield`
* `yieldOn`
* `yieldTo`
* `yieldToOn`

I think `fake` is capable enough as it is, and would like to wait with adding more features, before we know we have to.

mantoni:
I'm quite sure that we'll have to add some of those functions back in. I've already experienced cases where I wanted `yield` myself.

[...]

PR #1703, opened 2018-02-25
Change yields semantics to match stub behaviour

mantoni:
As discussed in #1697 the `fake.yields` semantics should match the `stub.yields` semantics.

We also agreed that we wouldn't want to support `yieldsRight`. As a solution, the fake `yields` and `yieldsAsync` behaviour is more strict and requires the last argument to be a callback function. If it's not, an exception is thrown.

This should match most implementations and make a migration from stub to fake possible (with a caveat).

### DIFF
--- a/test/fake-test.js
+++ b/test/fake-test.js
@@ -160,14 +160,31 @@ describe("fake", function () {
     describe(".yields", function () {
         verifyProxy(fake.yields, noop, "42", "43");
 
-        it("should call the callback with the provided values", function () {
+        it("should call a callback with the provided values", function () {
             var callback = sinon.spy();
-            var myFake = fake.yields(callback, "one", "two", "three");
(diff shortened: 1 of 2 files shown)


test/fake-test.js
```diff
@@ -160,25 +160,42 @@ describe("fake", function () {
...
+
+            sinon.assert.calledOnce(callback);
+        });
+
+        it("should throw if the last argument is not a function", function () {
+            var myFake = fake.yields();
+
+            assert.exception(function () {
+                myFake(function () {}, "not a function");
+            }, /TypeError: Expected last argument to be a function/);
```
mroderick: Does [`assert.exception`](http://docs.busterjs.org/en/latest/modules/referee/#exception) work as expected with regular expressions for the matcher argument?

mantoni: Yes

Issue #1695, opened 2018-02-21
sinon@next: sinon.replace only works on own property


mantoni:
While playing with #1586, I noticed that `sinon.replace(obj, 'prop', fake)` only works if 'prop' is an _own_ property of `obj`. I can only make this work with `sinon.replace(Object.getPrototypeOf(obj), 'prop', fake)`.

`sinon.replace` should allow to replace properties up the prototype chain, like `sandbox.stub` does.

mroderick:
I tried adding a test to reproduce your scenario

```js
it("should replace property on prototype", function () {
    var Person = function (name) {
        this.name = name;
    };
    var Customer = function (name) {
        Person.call(this, name);
    };

    Customer.prototype = Object.create(Person.prototype);
    Customer.prototype.constructor = Customer;

    var customer = new Customer("Some Name");
    var replacement = "Some Other Name";

    this.sandbox.replace(customer, "name", replacement);

    assert.equals(customer.name, replacement);
});
```

However, this passes with no changes.

I don't think I understood your description correctly. Can you suggest an improvement to the test, that would make it clear what `replace` is currently not capable of?

mantoni:
@mroderick The reason your example doesn't fail is because `this.name = name` creates an own property of the object. If you'd add a method to the `Person` prototype and try to replace it, that would fail. You don't need the derived `Customer`.

Anyway, I've created #1704 to illustrate the issue with a failing test.

mroderick:
Fixed with #1705

PR #1704, opened 2018-02-25
Allow replacing non-own object properties

mantoni:
As discussed in #1695, the new `sandbox.replace` function requires the replaced property to be an own property of the object. Inherited properties can not be replaced.

For now, this PR adds a failing test case to demonstrate the issue.


mantoni:
@mroderick Why did you merge? I thought we should fix it on this branch first 😆

mroderick:
Oh, I thought I'd just make a PR to fix it ... I'll do that now ... no one will know ;)

PR #1705, opened 2018-02-25
Allow replacing non-own object properties

### LINKED ISSUE #1695: sinon@next: sinon.replace only works on own property

mroderick:
This fixes #1695

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -165,7 +165,7 @@ function Sandbox() {
     }
 
     sandbox.replace = function replace(object, property, replacement) {
-        var descriptor = Object.getOwnPropertyDescriptor(object, property);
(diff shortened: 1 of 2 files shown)


mantoni:
Why do my PRs trigger CircleCI and yours don't? That's weird.

mroderick:
> Why do my PRs trigger CircleCI and yours don't? That's weird.

I use a fork, you push branches to the sinonjs repository

mantoni:
Ah, ok. Then it makes sense.

Issue #1711, opened 2018-02-27
Simple stub fails for ES6 import


ctaylo21:
* Sinon version : 4.4.2
* Environment   : MacOSX High Sierra 10.13.3
* Example URL   :
* Other libraries you are using: Webpack, babel, mocha

**What did you expect to happen?**
Given a module like so:
```js
// mod1.js
function test() {
  return 'fail';
}

export {
  test
};
```
And a test like this:
```js
// mod1Spec.js
import { assert } from 'chai';
import sinon from 'sinon';

import * as mod1 from 'Modules/settings/mod1';

describe('test', function () {
  it('should correctly mock module method', () => {
    sinon.stub(mod1, 'test').returns('pass');
    assert.strictEqual(mod1.test(), 'pass');
  });
});
```
The test should pass.

**What actually happens**
The method is never mocked. It still returns 'fail' even with the direct stub above it. I've seen some issues around mocking ES6 modules, but they all seem to imply that using `import * as blah` will allow you to stub correctly (ES6 classes may be a different story).

Can you actually stub ES6 modules with sinon? If so, are there any known libraries that can interfere with the stubbing? I imagine babel or webpack would be valid culprits as they may bundle/transpile modules and break stubbing.

I've been tearing my code apart, so I want to confirm that it's possible to stub ES6 modules without using anything like [babel-plugin-rewire](https://github.com/speedskater/babel-plugin-rewire) or if something like that plugin is now required.

**How to reproduce**
See code above.

mroderick:
I've created [a runnable test](https://github.com/mroderick/sinon-issue-1711) from your example, using `@std/esm` instead of `babel`.

It shows very clearly that (under node), we're not allowed to modify modules that are imported using `import`. It looks like [imports are read-only views on exports](http://exploringjs.com/es6/ch_modules.html#_imports-are-read-only-views-on-exports). 

So, `sinon.stub` is not able to modify the imported module. 

I guess you will have to use some [link seam](https://www.informit.com/articles/article.aspx?p=359417&seqNum=3) with your chosen module loader to redefine what the dependencies mean.

If you're going to use a [link seam](https://www.informit.com/articles/article.aspx?p=359417&seqNum=3), you might be interested in the new [`sinon.fake` api](https://github.com/sinonjs/sinon/blob/next/docs/release-source/release/fakes.md) (install as `sinon@next`). The api is still under development, feedback is very welcome.

[...]

Issue #1623, opened 2017-11-16
Stubbing default exported functions


sebakerckhof:
It is a common use case to mock ES modules.

e.g. I have a module that imports getEventByEventId
```js
import getEventByEventId from '../queries/getEventByEventId';

export default async (
  {
    connection,
    session
  }: ResolverContextType
) => {
  if (!session || !session.userId) {
    throw new Error('User must be authenticated.');
  }

  const event = await getEventByEventId(connection, eventId);

  // ..
}
```
In my test I want to mock getEventByEventId, e.g.
```js
import test from 'ava';
import sinon from 'sinon';
import moment from 'moment';
import createReservation from '../../../src/mutators/createReservation';

test('throws an error if the event is in the past', async (t) => {
  const parameters: any = {};

  const context: any = {
    session: {
      userId: 1
    }
  };

  const stub = sinon.stub().returns({
    date: moment().format('YYYY-MM-DD'),
    time: moment(new Date().getTime() - 1000 * 60).format('HH:mm')
  });

  // How to use the stub to mock `getEventByEventId`?

  await t.throws(createReservation(context), 'Cannot create a reservation for a past event.');
});
```

This was asked before here: https://github.com/sinonjs/sinon/issues/1358 . But this got unrightly closed as a duplicate of: https://github.com/sinonjs/sinon/issues/1121 , while it is a totally different issue.
This was already mentioned in that issue, but it doesn't seem to get looked after anymore.

mroderick:
Sinon.JS does not concern itself with module systems. I am closing this.

Because of the many requests that keep coming in, and that the stewards of the project also write tests for a living, we have published [How To Use Link Seams With CommonJS](http://sinonjs.org/how-to/link-seams-commonjs/).

Please help us keep the issues list manageable and ask usage questions on the [Sinon.JS users mailinglist](https://groups.google.com/forum/#!forum/sinonjs) or [StackOverflow](https://stackoverflow.com/search?q=es6+stub+default+export). Both are good resources where the wider audience of users can offer help.

nikoremi97:
@sebakerckhof checkout this  https://github.com/sinonjs/sinon/issues/562#issuecomment-399090111

PR #1715, opened 2018-02-28
Throw meaningful error stubbing ECMAScript Module

fatso83:
#### Purpose (TL;DR)
Throw a meaningful error when encountering ES Modules in `sinon.stub()`

#### Background
Ref #1711 and #1623 for background.

This feature is simply about throwing a meaningful error, instead of one that tells the user that the property cannot be deleted. As the exports from ECMAScript modules are immutable once created we
cannot do anything about them from the view of Sinon, so it's preferable to simply tell the user that when we can.

This should not affect transpiled code (say, using Babel), as the resulting modules are no longer true ECMAScript modules (but rather some form of CommonJS module).

1. Check out this branch
2. `npm install`
3. `npm run test-es-module` (included in `npm test`)

#### Checklist for author

- [x] `npm run lint` passes
- [x]  References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).


test/es2015/module-support-assessment-test.mjs
```diff
@@ -0,0 +1,14 @@
+"use strict";
+import referee from "referee";
+import sinon from "../../lib/sinon";
+import * as aModule from "./a-module";
+const {assert} = referee;
+
+describe("Explicit lack of support for stubbing Modules", function() {
+    it("should give a proper error message for modules without default export", function() {
```
mroderick: Any export (or rather import) would be read-only, wouldn't it? This seems to suggest that read-only only applies to `default` exports.

test/es2015/module-support-assessment-test.mjs
```diff
@@ -0,0 +1,14 @@
...
+        assert.exception(function() {
+            sinon.stub(aModule, "anExport").returns(400);
```
mroderick: We will have the same problem for spies, won't we?

lib/sinon/stub.js
```diff
@@ -18,6 +18,14 @@ function stub(object, property) {
         throw new TypeError("stub(obj, 'meth', fn) has been removed, see documentation");
     }
 
+    if (
+        object &&
+        typeof Symbol !== "undefined" &&
+        object[Symbol.toStringTag] === "Module"
+    ) {
+        throw new TypeError("No support for stubbing ES Modules");
```
mantoni: This sound a bit like "Sinon does not have this feature", but really, it's impossible. Maybe "ES modules cannot be stubbed"?

lib/sinon/stub.js
```diff
@@ -18,6 +18,14 @@ function stub(object, property) {
         throw new TypeError("stub(obj, 'meth', fn) has been removed, see documentation");
     }
 
+    if (
```
mroderick: It would be great to move this into a function, so it can be re-used (upcoming `sinon@next` has a new `replace` method).

test/es2015/module-support-assessment-test.mjs
fatso83: Yes. Pushed a fix for this.

test/es2015/module-support-assessment-test.mjs
fatso83: I was thinking of adding some additional tests, but couldn't be bothered at that time of the night :-) I added some more tests now, and as you can see, it's only-halfway true. You can actually stub some forms of exports. Although you can't modify the exported properties, one can modify the objects that has been exported (check the content of the tests), so one can modify the props of `export default { a: 1, b: 2 }`, just not point `default` to another object.

test/es2015/module-support-assessment-test.mjs
```diff
@@ -2,13 +2,54 @@
...
-        /TypeError: No support for stubbing ES Modules/);
+import aModuleWithDefaultExport from "./a-module-with-default";
+import functionModule, * as functionModuleAlternative from "./a-function-module";
+const {assert, refute} = referee;
+
+function createTestSuite(action){
+    var stub;
+    var errorRegEx = /TypeError: ES Modules cannot be (stubbed|spied)/;
+
+    afterEach(function() {
```
mroderick: This seems duplicated with the same `afterEach` inside the `describe` below ... is that intentional?

test/es2015/module-support-assessment-test.mjs
```diff
@@ -2,13 +2,54 @@
...
+        stub && stub.restore && stub.restore();
     });
-});
+
+    describe("sinon." + action + "()", function(){
+        describe("Modules with objects as their default export", function() {
+    afterEach(function() {
```
mroderick: Hopefully ESLint complains about indentation here ...

fatso83: No it did not, but I have now added a `lint-staged` script to catch this.

PR #1719, opened 2018-03-03
Replace referee dependency with @sinonjs/referee

mroderick:
This PR replaces the `referee` dependency with `@sinonjs/referee`.

1. Observe that all tests pass on Travis
2. ?
3. Win!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,29 @@
         }
       }
     },
+    "@sinonjs/referee": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/referee/-/referee-2.0.0.tgz",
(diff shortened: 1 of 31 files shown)


mantoni:
As we discussed offline, `@sinonjs/referee` needs a patch to turn the dependencies into devDependencies. Otherwise they'd become direct dependencies of sinon as well. Let's do that first and update this PR.

mroderick:
> As we discussed offline, @sinonjs/referee needs a patch to turn the dependencies into devDependencies. Otherwise they'd become direct dependencies of sinon as well. Let's do that first and update this PR.

That was for `@sinonjs/referee-sinon`, which doesn't affect this project (but has already been patched).

mantoni:
Oh yes, right. I'm getting confused 😆

mroderick:
It has been some intense days, with lots of commits and PRs thrown around

PR #1724, opened 2018-03-05
Return returned value of invokant when using yields* and callsArg*

ungrim97:
#### Purpose
Ensure that the yields* and callsArg* methods of the stub cause the stub to return the returned value of the invoked callback. Will defer to all other return value handlers (returns, returnsThis, resolves, rejects ect)

#### Background
#1689 allows the yield* and callArg* methods to return the value returned by the invoked callback. It is preferable that the stub methods of yields* and callsArg* provide the same functionality to allow testing of functions that are invoked as a callback on the stub that return a promise.

This prevents the need for the test to wait for `Process.nextTick()` to check if the promise has resolved but can instead us `await` or `.then` to do so instead. Thus guaranteeing that the promise is resolved

#### Solution
Adds a new property to the behaviour of yields that is set to true for all yields* and callsArg* methods in `default-behaviour.js`.

Ensure that if the yields property is true, and no other return setting is enabled (such as returns, returnsThis ect) then the returned value of the invoked callback is returned. by the `invoke` function in `behaviour.js`

#### How to verify
1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

-  [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -92,9 +92,11 @@ function callCallback(behavior, args) {
                 func.apply(behavior.callbackContext, behavior.callbackArguments);
             });
         } else {
-            func.apply(behavior.callbackContext, behavior.callbackArguments);
+            return func.apply(behavior.callbackContext, behavior.callbackArguments);
         }
     }
+
+    return undefined;
 }
 
 var proto = {
@@ -125,7 +127,7 @@ var proto = {
     },
 
     invoke: function invoke(context, args) {
-        callCallback(this, args);
+        var returnValue = callCallback(this, args);
 
         if (this.exception) {
             throw this.exception;
@@ -156,7 +158,12 @@ var proto = {
             return (this.promiseLibrary || Promise).reject(this.returnValue);
         } else if (this.callsThrough) {
             return this.stub.wrappedMethod.apply(context, args);
+        } else if (this.returnValueDefined) {
+            return this.returnValue;
+        } else if (this.yields) {
+            return returnValue;
         }
+
         return this.returnValue;
     },
 
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -158,7 +158,7 @@ var proto = {
             return (this.promiseLibrary || Promise).reject(this.returnValue);
         } else if (this.callsThrough) {
             return this.stub.wrappedMethod.apply(context, args);
-        } else if (this.returnValueDefined) {
+        } else if (typeof this.returnValue !== "undefined") {
             return this.returnValue;
         } else if (typeof this.callArgAt === "number") {
             return returnValue;
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -160,7 +160,7 @@ var proto = {
             return this.stub.wrappedMethod.apply(context, args);
         } else if (this.returnValueDefined) {
             return this.returnValue;
-        } else if (this.yields) {
+        } else if (typeof this.callArgAt === "number") {
             return returnValue;
         }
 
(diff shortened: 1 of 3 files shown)


lib/sinon/behavior.js
```diff
@@ -156,7 +158,12 @@ var proto = {
             return (this.promiseLibrary || Promise).reject(this.returnValue);
         } else if (this.callsThrough) {
             return this.stub.wrappedMethod.apply(context, args);
+        } else if (this.returnValueDefined) {
```
mantoni: Where is `returnValueDefined` being set? I don't see it anywhere else.

lib/sinon/behavior.js
```diff
@@ -156,7 +158,12 @@ var proto = {
...
+            return this.returnValue;
+        } else if (this.yields) {
```
mantoni: Is a new property really required? I think checking for `callsArgAt` would also do, no?

test/stub-test.js
```diff
@@ -1271,34 +1306,41 @@ describe("stub", function () {
...
-            assert(spy.calledOnce);
+            var callback = createStub().returns("return value");
+            assert.same(stub.call(obj, callback), obj);
+            assert(callback.calledOnce);
+        });
+
+        it("returns the result of the yielded callback", function () {
+            var stub = createStub().yields();
+            var callback = createStub().returns("return value");
+            assert.same(stub(callback), "return value");
```
mantoni: Minor styling: Can you separate setup / testing code from assertions with a blank line? This style should be used everywhere. You did it right in the first few tests already. It makes the tests easier to read :)

mantoni:
Nice pull request with very good description and testing. Thanks!

I've added a few comments. Maybe @mroderick or @fatso83 have something to add?

lib/sinon/behavior.js
ungrim97: Its set in `default-behaviour.js` for returns, and resolves. Not sure it was the perfect answer. Original thought of `typeof this.returnValue !== "undefined"`. Open to suggestions

ungrim97:
- [x] Add spaces between setup and assertions in tests
- [x] Use callArgAt property rather than new property
- [x] Use returnValue rather than returnValueDefined

lib/sinon/behavior.js
ungrim97: It would seem that callArgAt will work for this.

lib/sinon/behavior.js
ungrim97: I have gone back and used `typeof this.returnValue !== "undefined"` as methods such as returnsThis don't set `returnValueDefined`

test/stub-test.js
```diff
@@ -1496,6 +1555,50 @@ describe("stub", function () {
                 this.stub(callback);
             });
         });
+
+        it("plays nice with throws", function () {
```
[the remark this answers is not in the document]
ungrim97: Not at all. Does "Throws takes precedent over yields return value" work here?

ungrim97:
- [ ] Rename new "plays well with X" tests to have more explicit test names

test/stub-test.js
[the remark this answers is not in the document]
ungrim97: @mroderick Done

ungrim97:
@mroderick You change should be implemented now.

mantoni:
@mroderick @fatso83 Feel free to merge this. I’m AFK for a couple of days.

mroderick:
@ungrim97 very nice! There's a minor misspelling: "precident" => "precedent" ... would you mind fixing that before we merge it?

ungrim97:
@mroderick fixed

mroderick:
This has been published as `sinon@4.4.6` on npm

PR #1730, opened 2018-03-08
stub.throws and TypeErrors

MichaelPaulukonis:
Docs and tests misleadingly suggest that string parameters will generate
an error of that type. Documentation and test changes should make
clear(er) that string params supply generic error name and/or message
parameters.

Docs and tests misleadingly suggest that string parameters will generate an error of that type

Example from stubs.md:
```
stub.throws("TypeError");

Causes the stub to throw an exception of the provided type.
```

Stub does not throw a "TypeError" exception. Instead, it throws an Error with the `name` property set to "TypeError"

#### Solution  - optional

Relevant docs and unit-tests have had the magic string "TypeError" changed to "Named Error".

1. Check out this branch
2. `npm install`
3. `npm test`

NOTE: existing issue related to `referee` requires workaround to successfully run all tests. This workaround (updating afftected file to use `@sinonjs/referree`) is not part of the PR.

#### Checklist for author

- [x] `npm run lint` passes


docs/_releases/v4.1.4/stubs.md
```diff
@@ -261,15 +261,18 @@ The Promise library can be overwritten using the `usingPromise` method.
 Causes the stub to throw an exception (`Error`).
 
 
-#### `stub.throws("TypeError");`
+#### `stub.throws("Error msg");`
 
-Causes the stub to throw an exception of the provided type.
+Causes the stub to throw an exception with the provided message.
```
fatso83: That's incorrect. The `Error#message` is not set, but the `Error#name` is. See this example:
```javascript
> stub=sinon.stub().throws("An_error")
> stub2=sinon.stub().throws(new TypeError("The message") )
> try { stub() } catch (err) { console.log(' err.name', err.name, '\n', 'err.message', err.message);   }
 err.name An_error 
 err.message 
undefined
> try { stub2() } catch (err) { console.log(' err.name', err.name, '\n', 'err.message', err.message);   }
 err.name TypeError 
 err.message The message
```

I suggest just saying it like it is: 
> ".... an exception with the `name` property set to the provided string"

fatso83: Also, the second parameter is not exposed in the docs, so an alternative to adding it could be to just mention it for this one:

```markdown
#### `stub.throws("ErrorType", "Optional error message")

Causes the stub to throw an `Exception` with the `name` property set to the type argument. The message argument is optional and will set the `message` prop of the error.
```

MichaelPaulukonis: Thanks for the feedback - I expect to update this tomorrow (Monday) evening.

MichaelPaulukonis:
I've made a new commit based on the suggestions.

Additionally, I think I should have been editing the "master" (?) copy of the documentation, so I applied my updated there, and reverted the published-doc changes.

MichaelPaulukonis:
ping

fatso83:
This was merged as 3d18f04 and released as Sinon 4.4.10. Thanks!

I updated the released docs as well, btw, as this was an improvement that could well be part of the older docs as well. For later reference, I used [this script](https://gist.github.com/fatso83/5547fcfa37c38e26dc7f6fe8d8b8eda9) to make the job go a lot quicker.

PR #1732, opened 2018-03-11
Fix inconsistent newline usage %D

### LINKED ISSUE #1717: Inconsistent newlines in %D replacement using spy.printf
* Sinon version: 4.4.2
* Environment: Ubuntu 16.04, Node v8.9.4, npm 5.7.1
* Other libraries you are using: [sinon-chai](https://github.com/domenic/sinon-chai) 2.14.0

I expected the `%D` replacement for `spy.printf` to have the same newline behavior regardless of the number of calls to the spy.

**What actually happens**
When there's only one call to a spy and I use the `%D` replacement with `spy.printf`, a newline is appended to the start of the replacement thanks to [this statement](https://github.com/sinonjs/sinon/blob/f89392c419dd3825d7af7fa0d58cc6ddeabfa3a6/lib/sinon/spy-formatters.js#L58) which is always executed.  However, when multiple calls have been made to the spy, a newline is _not_ placed at the beginning of the replacement because of [this conditional](https://github.com/sinonjs/sinon/blob/f89392c419dd3825d7af7fa0d58cc6ddeabfa3a6/lib/sinon/spy-formatters.js#L51) which only adds newlines before the "Call _n_:" label after the first iteration of the loop.

**How to reproduce**
```javascript
const Sinon = require('sinon');
const spy = Sinon.spy();
spy(1, 2, 3);
console.log(spy.printf('%D'));  // => "\n1\n2\n3"
spy(4, 5, 6);
console.log(spy.printf('%D'));  // => "Call 1:\n1\n2\n3\nCall 2:\n4\n5\n6"
```

I this an oversight in formatting or is this an intentional decision?  There don't seem to be any explicit unit tests for `%D` with the rest of the [`printf` tests](https://github.com/sinonjs/sinon/blob/f89392c419dd3825d7af7fa0d58cc6ddeabfa3a6/test/call-test.js#L1328), but altering how `%D` works causes tests like the [`assert.alwaysCalledWithMatch exception message` test](https://github.com/sinonjs/sinon/blob/f89392c419dd3825d7af7fa0d58cc6ddeabfa3a6/test/assert-test.js#L1664) to fail.

JayAndCatchFire:
- Fix inconsistent newline usage %D
- Update affected unit tests
- Add tests for spy.printf %D text replacement

See #1717 for the detailed bug report.  This PR should resolve #1717.

1. Check out this branch
2. `npm install`
3. Make sure that `%D` always returns `\n` as its first character.
```js
const Sinon = require('sinon');
const spy = Sinon.spy();
spy(1, 2, 3);
console.log(spy.printf('%D').charCodeAt(0));  // => 10
spy(4, 5, 6);
console.log(spy.printf('%D').charCodeAt(0));  // => 10
```
Or just run `npm test`, which should pass.

#### Checklist for author

- [x] `npm run lint` passes

(That second checklist item isn't really applicable to these changes.)

### DIFF
--- a/lib/sinon/spy-formatters.js
+++ b/lib/sinon/spy-formatters.js
@@ -48,10 +48,7 @@ module.exports = {
         for (var i = 0, l = spyInstance.callCount; i < l; ++i) {
             // describe multiple calls
             if (l > 1) {
-                if (i > 0) {
-                    message += "\n";
-                }
-                message += "Call " + (i + 1) + ":";
+                message += "\nCall " + (i + 1) + ":";
             }
             var calledArgs = spyInstance.getCall(i).args;
             for (var j = 0; j < calledArgs.length || j < args.length; ++j) {
--- a/test/assert-test.js
+++ b/test/assert-test.js
@@ -1463,7 +1463,7 @@ describe("assert", function () {
             this.obj.doSomething(1, 3, "not");
 
             assert.equals(this.message("calledWith", this.obj.doSomething, 1, 3, "hey").replace(/ at.*/g, ""),
-                "expected doSomething to be called with arguments " +
+                "expected doSomething to be called with arguments \n" +
                         "Call 1:\n" +
                         color.red("4") + " " + color.green("1") + " \n" +
                         "3\n" +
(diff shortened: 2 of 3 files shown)


mroderick:
Thank you for the PR @JayAndCatchFire.

I don't think you need to update package-lock.json for the changes made in the other files. Do you agree?

JayAndCatchFire:
The changes to the lock file were only a consequence of running `npm install`. But if you'd rather the lock file not be updated as part of this PR, I'll revert it. Give me a few hours and I'll be able to get in front of a computer and make that change.

JayAndCatchFire:
@mroderick The lock file has been reverted.

mroderick:
This has been published as `sinon@4.4.3`

Issue #1736, opened 2018-03-14
stub.onCall(#).yields(x).returns(y) returns undefined instead of y


nfriedly:
I just updated some code from sinon 1.x to 4.x to take advantage of the newer promise features, and tests that had been passing started failing. These tests were stubbing `http.request`, which both returns a value and calls a callback. The return `req` value changed to `undefined`, and so code that called `req.on(...)` started failing. 

* Sinon version : 4.4.5
* Environment   : Node.js 8.9.4, macOS
* Example: see below
* Other libraries you are using: just sinon + mocha in the example below

**What did you expect to happen?**
I expect the stub to return whatever I pass into `.returns()`

**What actually happens**
When `onCall()` is used, it returns `undefined`

**How to reproduce**

```js
var assert = require('assert')
var sinon = require('sinon');

describe('sinon yield + returns', function() {
    // this passes
    it('works without without onCall', () => {
        const stub = sinon.stub().yields('yield val').returns('return val');
        const cb = sinon.spy();
        const ret = stub(cb);
        assert.equal(ret, 'return val');
        assert.deepEqual(cb.args, [['yield val']]);
    });
    // this fails
    it('should return the correct value with onCall', () => {
        const stub = sinon.stub().onCall(0).yields('yield val').returns('return val');
        const cb = sinon.spy();
        const ret = stub(cb);
        assert.equal(ret, 'return val'); // throws because ret is undefined
        assert.deepEqual(cb.args, [['yield val']]);
    });
});
```

save the above example to a test.js file, install sinon, and run `npx mocha test.js`

nfriedly:
Doh, I think https://github.com/sinonjs/sinon/pull/1724 might fix this.

mroderick:
@nfriedly thank you for your contribution!

I've just published #1724 as `sinon@4.4.6` to npm. Can you confirm that it fixes the issue for you?

nfriedly:
Hum, it actually doesn't seem to from a quick test:

```sh
MLREM1712156:test nathan.friedly$ npm ls sinon
nathan.friedly@1.0.0 /Users/nathan.friedly/test
└── sinon@4.4.6 

MLREM1712156:test nathan.friedly$ npx mocha test.js


  sinon yield + returns
    ✓ works without without onCall
    1) should return the correct value with onCall


  1 passing (8ms)
  1 failing

  1) sinon yield + returns
       should return the correct value with onCall:
[...]

PR #1738, opened 2018-03-15
Fix onCall(#) with chained behaviors

nfriedly:
Restores previous Sinon behavior and fixes issue #1736 by making all chained behaviors (yields, returns, throws, etc.) added after an `.onCall(#)` apply to that call rather than only the first behavior applying, and subsequent ones reverting to the default behavior.

As mentioned in the original issue, in previous versions of Sinon, the following code added a `yeilds` and a `returns` behavior to call 0:

```js
stub.onCall(0).yields(x).returns(y);
```

However, in the current release of Sinon, that code would cause the first call to yield x and return undefined, and then all subsequent calls would return y.

The codebase I'm working on currently, uses the above pattern in numerous places for mocking `http.request` (which both returns and yields), and upgrading to the current version of Sinon caused a number of tests to start failing.

#### Solution  - optional

behaviorMethods always return `this` (the behavior object) instead of the original stub. I also changed the behavior object to be a function that calls through to the original stub.

#### WARNING

It's possible that some users inadvertently depend on the buggy behavior with code such as this:

```js
var stub = sinon.stub().onCall(0).returns('foo');
stub.returns('bar');
```
In the current release, the first call will return 'foo', and subsequent calls will return 'bar'. With the fix, it will return 'bar' and then undefined. However, I would argue that the above code is buggy and this fix merely exposes that bug, and therfore this fix shouldn't constitute a breaking change.

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -101,7 +101,7 @@ function callCallback(behavior, args) {
 
 var proto = {
     create: function create(stub) {
-        var behavior = extend({}, proto);
+        var behavior = extend(stub.bind(), proto);
         delete behavior.create;
         delete behavior.addBehavior;
         delete behavior.createBehavior;
@@ -219,7 +219,7 @@ function createBehavior(behaviorMethod) {
 function addBehavior(stub, name, fn) {
     proto[name] = function () {
         fn.apply(this, [this].concat([].slice.call(arguments)));
-        return this.stub || this;
+        return this;
     };
 
     stub[name] = createBehavior(name);
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -420,4 +420,21 @@ describe("issues", function () {
             refute.isTrue(spyProp.get.called);
         });
     });
+
+    describe("#1736 - onCall + yield + returns", function () {
+        it("should return the correct value without onCall", function () {
+            var stub = sinon.stub().yields("yield val").returns("return val");
+            var cb = sinon.spy();
+            var ret = stub(cb);
+            assert.equals(ret, "return val");
+            assert.equals(cb.args, [["yield val"]]);
+        });
+        it("should return the correct value with onCall", function () {
+            var stub = sinon.stub().onCall(0).yields("yield val").returns("return val");
+            var cb = sinon.spy();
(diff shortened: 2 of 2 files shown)


lib/sinon/behavior.js
```diff
@@ -101,7 +101,7 @@ function callCallback(behavior, args) {
 
 var proto = {
     create: function create(stub) {
-        var behavior = extend({}, proto);
+        var behavior = extend(stub.bind(), proto);
```
fatso83: what does this do? I am guessing this is an easy way of wrapping the stub in a closure so you can change the copy instead of the original. am I right?

nfriedly: yes. We could replace it with 

```js
var behavior = extend(function(){
   var args = Array.prototype.slice.call(arguments);
   return stub.apply(null, args);
}, proto);
```

But the .bind version is more concise.

nfriedly:
> how would one make a stub that returns bar in most cases, but 'foo' in the 42nd call when using your fix?

Like this:
```js
var stub = sinon.stub();
stub.returns('foo');
stub.onCall(42).returns('bar');
```

(The second two lines could be swapped without changing the behavior.)

Or, a chained version:

```
var stub = sinon.stub().returns('foo').onCall(42).returns('bar');
```

(The behavior depends on the order here.)

We could perhaps add a method to get back to the default behavior, e.g.

```
var stub = sinon.stub().onCall(42).returns('bar').otherwise().returns('foo');
```

and `otherwise()` would just be a function that returned `this.stub`. Probably needs a better name, though.

And, like I said, this change restores the behavior that was present in Sinon 1.x. I'm not exactly sure when it changed, but I'm sure we could track it down if it matters.

fatso83:
@mroderick This seemed fine and I finally understood the nuances of the implications, but I am not sure this should result in a new major or a patch! It is an API change, but as the previous change never was official/published I regard this as more of a bugfix that gives us the old behaviour back.

fatso83:
And thanks, @nfriedly, for the great work!

mroderick:
Good work team!

Issue #1737, opened 2018-03-15
Postinstall script log asking for donations?


SimonSchick:
So this is not strictly about sinon functionality therefor I am not going to use the template.

I have never seen a package actively ask for donations via a `postinstall` script, this is not exactly malicious but I don't think this is appropriate use of `postinstall` either as it also unnecessarily slows down installs...

mantoni:
It slows down your install time by something like a few milliseconds. We think that’s totally reasonable. We‘re not the only ones doing this. We also ask for donations on our website and in the readme.

billyjanitsch:
@mantoni I empathize with the motivation to introduce the postinstall message: y'all deserve to be compensated for the time you spend on the project, and I appreciate the need to find ways to ask users to contribute.

However I'm not sure the impact is as low as you think, especially when you consider scaling. A typical npm tree contains thousands of packages. What if many packages start doing this?

I timed the postinstall script at ~80ms (starting a `node` process isn't free), which is negligible on its own, but it would only take 10 packages doing this to add ~1s to every local and CI install.

More importantly, `npm install` currently produces a useful log of lifecycle scripts and a tree summary. Even if _only_ sinon's 14 (incl. transitive) dependencies included similar messages, sinon's own message would be lost in the noise -- defeating the purpose of it -- and it would be way less clear to users what actually happened in the lifecycle stage.

So I feel like this kind of solution, at best, only works briefly for the first couple of packages that do it, and then results in a worse ecosystem overall. To be clear, I'm not saying that you should work for free, and I'm happy that OSS revenue solutions are starting to pop up (`npx thanks`, Open Collective, etc.) I just don't think a postinstall script is a good medium. I understand if you disagree. ❤️

mroderick:
The postinstall script was implemented by me as an experiment to try to drive more donations to our Open Collective.

It’s inspired by (imitating) Remy’s strategy for nodemon: https://remysharp.com/2018/01/10/open-source-with-a-cap-in-hand

So far, I am still about €500 out of pocket to pay the costs of bringing the stewards together for the first time, in order to get `sinon@2` released. That was a year ago and has changed how we work on Sinon and related projects.

Once we have accelerated the community support, then I am happy to remove it again.

Other strategies to help pay for our (minimal) costs would be most welcome.

ljharb:
I truly do appreciate that you need funding, but polluting the install logs of all consumers with an advertisement really isn't a great way to generate empathy and incentivize donations.

[...]

PR #1742, opened 2018-03-21
Reduce noisy output in donations message

fatso83:
#### Purpose 
- Basically get's rid of most of the echoed `console.log` noise.
- Minor change in the wording :pray: 

Don't think this is much different from the oneliner performance wize (ref #1737).

1. `npm run postinstall`


mroderick:
This has been published as `sinon@4.4.7` ❤️

PR #1749, opened 2018-03-22
Fix #1746 by removing library

### LINKED ISSUE #1746: support-sinon.js breaking builds for older versions of node.js
The new support-sinon.js post install script breaks build, because it uses the "supports-color" module, which uses ES6 syntax, which requires a new version of node.js.

We are on an older version of node.js.

* Sinon version : 4.4.8
* Environment   :  node v0.10.22
* Example URL   :
* Other libraries you are using: karma-sinon-chai

```
node scripts/support-sinon.js


node_modules/sinon/node_modules/supports-color/index.js:2
error	21-Mar-2018 11:27:27	const os = require('os');
error	21-Mar-2018 11:27:27	^^^^^
error	21-Mar-2018 11:27:27	SyntaxError: Use of const in strict mode.
error	21-Mar-2018 11:27:27	    at Module._compile (module.js:439:25)
error	21-Mar-2018 11:27:27	    at Object.Module._extensions..js (module.js:474:10)
error	21-Mar-2018 11:27:27	    at Module.load (module.js:356:32)
error	21-Mar-2018 11:27:27	    at Function.Module._load (module.js:312:12)
error	21-Mar-2018 11:27:27	    at Module.require (module.js:364:17)
error	21-Mar-2018 11:27:27	    at require (module.js:380:17)
error	21-Mar-2018 11:27:27	    at Object.<anonymous> (node_modules/sinon/lib/sinon/color.js:3:21)
error	21-Mar-2018 11:27:27	    at Module._compile (module.js:456:26)
error	21-Mar-2018 11:27:27	    at Object.Module._extensions..js (module.js:474:10)
error	21-Mar-2018 11:27:27	    at Module.load (module.js:356:32)
```

fatso83:
Basically fixes #1746 by removing a library dependency in the postinstall to avoid
breaking installs on Node versions < 4.

*Also* (and this is up for concideration), supports a use case where users can disable the support message, in case they rely on formatting being consistent in the output for parsing. This would be used like: `NO_SUPPORT_MESSAGE='' npm run postinstall -s`. I can remove this commit, but I think this strikes an ok compromise, as you _know_ people have read the message in case they explicitly want to turn it off. Unless this just turns into a standard flag convention in other packages, but that is getting ahead of things :smile:

### DIFF
--- a/scripts/support-sinon.js
+++ b/scripts/support-sinon.js
@@ -1,12 +1,17 @@
 "use strict";
 /* eslint-disable no-console */
 
-var color = require("../lib/sinon/color");
+var green = "\u001b[32m";
+var white = "\u001b[22m\u001b[39m";
+var boldCyan = "\u001b[96m\u001b[1m";
+var reset = "\u001b[0m";
 
 var output =
-    color.green(
-        "Have some ❤️  for Sinon? You can support the project via Open Collective:"
-    ) +
-    color.white("\n > ") +
-    color.cyan(color.bold("https://opencollective.com/sinon/donate\n"));
+    green +
+    "Have some ❤️  for Sinon? You can support the project via Open Collective:" +
+    white +
+    "\n > " +
+    boldCyan +
+    "https://opencollective.com/sinon/donate\n" +
+    reset;
 console.log(output);
--- a/scripts/support-sinon.js
+++ b/scripts/support-sinon.js
@@ -14,4 +14,7 @@ var output =
     boldCyan +
     "https://opencollective.com/sinon/donate\n" +
     reset;
-console.log(output);
+
+if (!("NO_SUPPORT_MESSAGE" in process.env)) {
+    console.log(output);
+}


mroderick:
:+1:

mantoni:
Good change. 👍 Maybe rename the environment variable to something containing `SINON` (1) to make it easier to remember what the flag is for and (2) to avoid potential conflicts with other environment variables. How about `SINON_OPEN_COLLECTIVE_MESSAGE`?

mantoni:
Oh, I was unaware of the opencollective-cli. Maybe we should just use that module and leave the opt-out question to them. I'm not really willing to remove the banner because a single person is complaining loudly. It's one of the things I learned from parenting: If complaining loudly wins, it's just getting more noisy.

mantoni:
bildschirmfoto 2018-03-23 um 10 18 57

Shall I make it a PR? @mroderick @fatso83

fatso83:
Ha, I think that won't make the number of complaints go down, @mantoni :laughing: Anyway, as Morgan says: we need more backers to not fund this too much out of our own pockets, and if we can increase the numbers of backers we can either kill the banner or enable a flag that does it.

mroderick:
@mlucool The observable facts (from where I am sitting) are that you're encouraging package maintainers to create a flag that would return the text output during install of their packages to what it was before the postinstall banner. Apparently this is to save some work for people that write parsers for the install output. Postinstall-parser-writers are not the primary audience of this package (nor of Nodemon, I am sure).

I agree, "crusade" is a strong term.

I am all for productive conversation. If you have ideas that will lead to Sinon.JS (and Nodemon) getting more backers, then I would love to discuss them.

PR #1753, opened 2018-03-27
Fix #1714: Add .callback and .lastArg to fake and call

### LINKED ISSUE #1714: sinon@next: Provide a way to invoke callbacks on fakes
This is a follow up issue from the discussion on #1696.

The Sinon spy and stub APIs have these functions:

- `yield` (aliased as `invokeCallback`)
- `yieldOn`
- `yieldTo`
- `yieldToOn`
- `callArg` (aliased as `callArgWith`)
- `callArgOn` (aliased as `callArgOnWith`)

They all intend to invoke callbacks that were passed to the spy or stub as an argument.

Instead of providing various ways to invoke arguments as functions, we could expose them in a more convenient way and let users invoke the functions. This allows to make use of `.call` and `.apply` to change the function scope.

Currently, arguments can be accessed through calls, e.g. the `fake.firstCall.args` array.

I would like to propose that we expose more references to arguments in order to avoid having to add all of the above methods to the new `fake` api:

- `fake.callback`: Referring to the last argument being passed to the fake in the last call, if it's a function. Otherwise this property does not exist.
- `fake.lastArg`: Referring to the last argument being passed to the fake in the last call.
- `fake.getCall(n).callback`: Referring to the last argument being passed to this specific call, if it's a function. Otherwise this property does not exist.
- `fake.getCall(n).lastArg`: Referring to the last argument being passed to this specific call.

The "legacy" API functions would map to the new fake API as follows:

| Legacy | Fake API |
|----|----|
| `yield(a, b, c)` | `fake.callback(a, b, c)` |
| `yieldOn(scope, a, b, c)` | `fake.callback.call(scope, a, b, c)` |
| `yieldTo(property, a, b, c)` | `fake.lastArg[property](a, b, c)` |
| `yieldToOn(property, scope, a, b, c)` | `fake.lastArg[property].call(scope, a, b, c)` |
| `callArg(index, a, b, c)` | `fake.lastCall.args[index](a, b, c)` |
| `callArgOn(index, scope, a, b, c)` | `fake.lastCall.args[index].call(scope, a, b, c)` |

The difference between `fake.lastArg` and `fake.callback` is purely semantical and implies a type check for `fake.callback`. I'm suggesting this, because it would be weird to have an object under `fake.callback`. On the other hand, `fake.lastArg()` doesn't read very nice.

mroderick:
This PR is a solution for #1714 

@mantoni does this match your idea outlined in #1714? If it does, then I'll take a stab at writing some documentation for it.

### DIFF
--- a/lib/sinon/call.js
+++ b/lib/sinon/call.js
@@ -209,10 +209,16 @@ function createSpyCall(spy, thisValue, args, returnValue, exception, id, errorWi
     if (typeof id !== "number") {
         throw new TypeError("Call id is not a number");
     }
+
     var proxyCall = Object.create(callProto);
+    var lastArg = args.length > 0 && args[args.length - 1] || undefined;
+    var callback = lastArg && typeof lastArg === "function" ? lastArg : undefined;
+
     proxyCall.proxy = spy;
     proxyCall.thisValue = thisValue;
     proxyCall.args = args;
+    proxyCall.lastArg = lastArg;
+    proxyCall.callback = callback;
     proxyCall.returnValue = returnValue;
     proxyCall.exception = exception;
     proxyCall.callId = id;
--- a/lib/sinon/fake.js
+++ b/lib/sinon/fake.js
@@ -32,8 +32,25 @@ function cleanProxy(f) {
     return f;
 }
 
+var uuid = 0;
 function wrapFunc(f) {
-    return cleanProxy(spy(f));
+    var fakeInstance = function () {
+        var lastArg = arguments.length > 0 && arguments[arguments.length - 1] || undefined;
+        var callback = lastArg && typeof lastArg === "function" ? lastArg : undefined;
+
+        /* eslint-disable no-use-before-define */
+        p.lastArg = lastArg;
+        p.callback = callback;
+        /* eslint-enable no-use-before-define */
+
+        return f && f.apply(f, arguments);
+    };
+    var p = cleanProxy(spy(fakeInstance));
+
+    p.displayName = "fake";
+    p.id = "fake#" + uuid++;
+
+    return p;
 }
 
 function fake(f) {
--- a/test/call-test.js
+++ b/test/call-test.js
@@ -424,6 +424,41 @@ describe("sinonSpy.call", function () {
         });
     });
 
+    describe(".callback", function () {
+        it("it should be a reference for the callback", function () {
+            var spy = sinonSpy();
+            var callback1 = function () {};
+            var callback2 = function () {};
+
+            spy(1, 2, 3, callback1);
+            assert.equals(spy.getCall(0).callback, callback1);
+
+            spy(1, 2, 3, callback2);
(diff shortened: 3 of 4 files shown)


mroderick:
I've merged the branch into `next`. I'll write some documentation for it next. Do you want to take it for a spin in one of your projects, to see that it works as expected?

mantoni:
Sure will do, but don‘t expect anything before Friday 🤖

fatso83:
Does this mean closing #1714 should wait until the docs are in place?

mroderick:
Yes please. I have some docs in a branch, however I'd like to add them upstream, so we don't end up in rebase hell

mantoni:
I played with this in a small project (only a dozend tests) where I'm trying to use the newly introduced APIs exclusively. With this change, it works great! It was the last missing bit 🎉

The features I've been using are:

- `sinon` global sandbox with `sinon.restore()` called after each test
- `sinon.replace(...)`
- `sinon.fake()`
- `sinon.fake.returns(...)`
- `sinon.fake.yields(...)`
- `fake.callback(...)`
- `call.callback(...)`

I've been mixing those with fake timers and various `sinon.assert` functions.

I like the API. The resulting test code reads very well and I didn't find any new issues. 👍 

@mroderick If you're happy the current state, you could cut another `sinon.5.0.0-next` release and I'll refactor an open source project to use our new toys. Can't share the project I've been using as a playground (yet).

mroderick:
Thank you for the feedback. 

I am working on a couple of pull requests:

* add `lastArg` and `callback` to `spyCall` (#1759)
* Improve documentation for fakes and spyCall (not opened yet)

Once we get those in, then I think we should cut a new `next` and focus on getting documentation ready for the new major release.

mroderick:
I've pushed a new branch `next.4`, which has all changes from `master` up to `sinon@4.5.0` and all of `next`, including a commit to add documentation for `fake().lastArg` and `fake().callback()`.

Could you try it out in your playground? If it works there, then I think we can publish this as `next.4`, and then focus our efforts on finishing documentation.

What do you think @mantoni?

mroderick:
On a related note, I hate long running branches and always run into merge hell. I think we should use `next.4` branch going forwards.

mantoni:
@mroderick It's looking good. I ran the branch in my playground and on a couple of other projects that use `sinon@4.x`. The only difference was the deprecation warning reminding me to migrate `sinon.sandbox.create` 👍

mroderick:
```
$ npm install sinon@next --save-dev

> sinon@5.0.0-next.4 postinstall /private/tmp/next-install/node_modules/sinon
> node scripts/support-sinon.js

Have some ❤️ for Sinon? You can support the project via Open Collective:
 > https://opencollective.com/sinon/donate

+ sinon@5.0.0-next.4
```

mantoni:
🎉  https://github.com/javascript-studio/studio-log/commit/ca85d2d8fa03d736a2804f30c16710e17e31e2b1

Issue #1709, opened 2018-02-27
Max call stack on objects with circular reference


scripter-co:
* Sinon version : 4.4.2
* Environment   : node
* Example URL   : https://github.com/scripter-co/sinon-circular-reference
* Other libraries you are using: `mocha`

**What did you expect to happen?**

Sinon does not throw an error when it comes across a circular reference

**What actually happens**

Sinon fails with the following:

```
➜  circular-reference git:(master) ✗ npm t

> circular-reference@1.0.0 test /Users/mahdan01/workspace/circular-reference
> mocha ./tests



  1) throws a "Maximum call stack size exceeded" error in sinon deepEqual

  0 passing (28ms)
  1 failing

  1) throws a "Maximum call stack size exceeded" error in sinon deepEqual:
     RangeError: Maximum call stack size exceeded
      at isElement (node_modules/sinon/lib/sinon/util/core/deep-equal.js:28:19)
      at deepEqual (node_modules/sinon/lib/sinon/util/core/deep-equal.js:37:9)
      at deepEqual$matcher (node_modules/sinon/lib/sinon/util/core/deep-equal.js:111:16)
      at deepEqual (node_modules/sinon/lib/sinon/util/core/deep-equal.js:84:45)
      at deepEqual$matcher (node_modules/sinon/lib/sinon/util/core/deep-equal.js:111:16)
      at deepEqual (node_modules/sinon/lib/sinon/util/core/deep-equal.js:84:45)
      at deepEqual$matcher (node_modules/sinon/lib/sinon/util/core/deep-equal.js:111:16)
      at deepEqual (node_modules/sinon/lib/sinon/util/core/deep-equal.js:84:45)
      at deepEqual$matcher (node_modules/sinon/lib/sinon/util/core/deep-equal.js:111:16)
```

**How to reproduce**

https://github.com/scripter-co/sinon-circular-reference

simply:

```js
const sinon = require('sinon');

it('throws a "Maximum call stack size exceeded" error in sinon deepEqual', () => {
  const spy = sinon.spy();

  const firstObj = {};
  firstObj['aKeyName'] = firstObj;

  const secondObj = {};
  secondObj['aKeyName'] = secondObj;

  spy(firstObj);

  sinon.assert.calledWith(spy, secondObj);
});
```

**Additional information:**

[...]

PR #1764, opened 2018-04-06
[WIP] Handle cyclic objects

scripter-co:

Resolves issue #1709 by adding `@sinonjs/samsam`. Removed old dependency and updated references.

### DIFF
--- a/lib/sinon/util/core/deep-equal.js
+++ b/lib/sinon/util/core/deep-equal.js
@@ -1,54 +1,14 @@
 "use strict";
 
-var div = typeof document !== "undefined" && document.createElement("div");
+var samsam = require("@sinonjs/samsam");
 
 function isReallyNaN(val) {
     return val !== val;
 }
 
-function isDOMNode(obj) {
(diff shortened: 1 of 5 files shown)


mroderick:
Nice work!

I've added a PR to your PR with a refactoring of `deepEqual`.

---

However, the matcher still doesn't handle cyclic input.

The way I see it, the matcher part of  `deepEqual` really doesn't belong in `deepEqual`. It should be in `match.js`. Further, `deepEqual.use` should be move to `match.js`.

That still won't fix the problem.

I think what we need to do is:

* let `deepEqual` worry about strict comparisons, not loose comparisons
  * move the matcher specific code from `deepEqual` into `match.js`
  * move `deepEqual.use` into `match.js`
* update `match.js` to be able to handle circular objects

Perhaps the update to `match.js` can do what samsam does, or maybe we can extract a function for traversal of potentially circular structures and re-use it

What do you think?

mroderick:
This has been replaced with #1853

PR #1768, opened 2018-04-27
Prepare 5 release

mroderick:
This PR contains a rewritten branch of `next`, most notable changes are

* default `sandbox` on `sinon`
* introducing `fake`
* introducing `sinon.replace*` methods
* removing deprecated `spy.reset`


mantoni:
Thanks Morgan!

One thing I noticed: The documentation for fake `yields` and `yieldsAsync` does not match my understanding of the implementation. It should be something like “The fake expects the last argument to be a callback  and will invoke it with the given arguments.”

I remember that we talked about this and thought we already fixed the docs, but somehow that seems to have went under the radar.

Unfortunately I can’t make that contribution myself from where I currently  work ¯\_(ツ)_/¯

mroderick:
@mantoni I've pushed an extra commit to amend the documentation and provide a better example

lib/sinon/util/core/next-tick.js
```diff
@@ -0,0 +1,20 @@
...
+
+function nextTick(callback) {
+    setTimeout(callback, 0);
+}
+
+if (hasNextTick) {
+    module.exports = process.nextTick;
+}
+
+if (!hasNextTick && hasSetImmediate) {
```
dazld: in terms of async scheduling, `nextTick` and `setImmediate` aren't exactly the same - `nextTick` will append the call to the current spin of the loop, and `setImmediate` schedules it for the next, whenever that happens. Maybe `Promise.resolve().then(callback)` would be closer, as that will (afaik) do the same thing as `nextTick`...? It's subtle, but it might be closer to the intended semantics.

ie:

```
function ping(m){
    return function(){
        console.log(m)
    }
}

ping('start')()
setImmediate(ping('immediate'))
ping('start2')()
setTimeout(ping('timeout'),0)
ping('start3')()
process.nextTick(ping('tick'))
ping('start4')()
Promise.resolve().then(ping('promise'))
ping('start5')()


```

gives:
```
start
start2
start3
start4
start5
tick
promise
immediate
timeout
```

dazld: it'll be slightly faster when making lots of calls too.

PR #1776, opened 2018-05-02
Fix #1775: Default sandbox does not restore stubs, spies, mocks

### LINKED ISSUE #1775: Default sandbox's restore method does not restore stubs/spies
When using any sandbox method on the default sandbox, `restore()` does not restore mocks or spies properly. A minimal repro case on Sinon v5.0.2:

```js
const myApi = {
  someMethod() {
    console.log('test method!');
  }
};

sinon.stub(myApi, 'someMethod');
sinon.restore();
sinon.stub(myApi, 'someMethod');
// TypeError: Attempted to wrap someMethod which is already wrapped
```

It seems that in the [call to `extend` in `lib/sinon.js`](https://github.com/sinonjs/sinon/blob/master/lib/sinon.js#L63), the un-sandboxed `stub/spy/etc.` exports from `apiMethods` end up shadowing the sandbox methods, and so `restore()` doesn't do anything. The [Sandbox API docs](http://sinonjs.org/releases/v5.0.2/sandbox/) and source in `lib/sinon.js` seem to suggest that this is intended usage.

I only tested `stub` and `spy`, but I would expect any other shadowed methods also do not work as intended.

I ran these tests on Node v8.9.4 and Chrome 66.0.3359.117 via Karma/Chromedriver, though this issue seems unrelated to environment.

mroderick:
This PR is a fix for #1775

The bug was caused by adding stub, spy, mock explicitly to apiMethods, thus
shadowing the methods from Sanbox, which are the ones that collect the
fakes for restoring.

Removing these methods, allows the caller to use the sandbox methods,
and all is well.

### DIFF
--- a/lib/sinon.js
+++ b/lib/sinon.js
@@ -15,10 +15,7 @@ var apiMethods = {
     assert: require("./sinon/assert"),
     fake: require("./sinon/fake"),
     match: require("./sinon/match"),
-    spy: require("./sinon/spy"),
     spyCall: require("./sinon/call"),
-    stub: require("./sinon/stub"),
-    mock: require("./sinon/mock"),
 
     expectation: require("./sinon/mock-expectation"),
     createStubInstance: require("./sinon/stub").createStubInstance,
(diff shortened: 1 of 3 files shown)


mroderick:
This has been published as `sinon@5.0.3`

PR #1784, opened 2018-05-05
Remove support-sinon.js as postinstall

mroderick:
We have managed to pay back outstanding expenses, and have attracted enough backing to have an annual budget of $3k.

I think that's good enough for now 😄 

Opinions?


fatso83:
I think this has been very successful in getting the attention we needed for funding, but even though it is effective, it has served its intended purpose and I concur that we can retire it. We still keep the front-page links, of course.

mantoni:
I don’t intend to file any expenses. So this is fine with me anyway.

mroderick:
`sinon@5.0.6`

Issue #1779, opened 2018-05-03
Calling sandbox.replace not cleaning up if called more than once


mantoni:
When calling `sandbox.replace` twice for the same `object` + `property`, the original state can not be restored.

Reproduce:

```js
sinon.replace(console, "log", sinon.fake());
sinon.replace(console, "log", sinon.fake());
sinon.restore();
console.log; // Still a proxy object
```

There are two different ways to solve this:

- We throw on the second attempt to use `sinon.replace`, like `sinon.stub(obj, prop)` fails if it's already a fake
- We allow multiple calls to use `sinon.replace`, but restore to the original state properly

I'd try to come up with an implementation if we know which direction to take. I'm not really decided here.

mroderick:
Trying to replace a fake with a fake seems like something that should warrant an investigation by the programmer. I would prefer to have tests fail, rather than sinon glossing over my mistakes. Throwing get's my vote.

mroderick:
This was fixed with `sinon@5.0.5`

PR #1785, opened 2018-05-06
Refuse to replace already replaced values

mantoni:
Fix for issue #1779

When values where replaced multiple times, `restore` didn't restore the original value. This change causes `sandbox.replace` to throw if the given object / property combination was already used.

1. Check out this branch
2. `npm install`
3. `npm test`

or

```js
sinon.replace(console, "log", sinon.fake());
sinon.replace(console, "log", sinon.fake()); // now throws
```

## Versioning

While this fixes an issue where `sandbox.restore()` fails to restore values that where replaced multiple times, this might also cause existing tests to fail. However, I would argue that we're fixing a broken behavior in the new `sandbox.replace` API to bring it in line with `sandbox.stub` which also throws in this case. Therefore I'd vote for semver patch.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -161,9 +161,20 @@ function Sandbox() {
     function getFakeRestorer(object, property) {
         var descriptor = getPropertyDescriptor(object, property);
 
-        return function restorer() {
+        function restorer() {
             Object.defineProperty(object, property, descriptor);
-        };
+        }
+        restorer.object = object;
+        restorer.property = property;
+        return restorer;
+    }
+
+    function verifyNotReplaced(object, property) {
+        fakeRestorers.forEach(function (fakeRestorer) {
+            if (fakeRestorer.object === object && fakeRestorer.property === property) {
+                throw new TypeError("Attempted to replace " + property + " which is already replaced");
+            }
+        });
     }
 
     sandbox.replace = function replace(object, property, replacement) {
@@ -189,6 +200,8 @@ function Sandbox() {
             throw new TypeError("Cannot replace " + typeof object[property] + " with " + typeof replacement);
         }
 
+        verifyNotReplaced(object, property);
+
         // store a function for restoring the replaced property
         push.call(fakeRestorers, getFakeRestorer(object, property));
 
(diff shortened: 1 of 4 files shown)


lib/sinon/stub.js
```diff
@@ -104,10 +104,10 @@ var proto = {
             return getCurrentBehavior(fnStub).invoke(this, arguments);
         };
 
-        functionStub.id = "stub#" + uuid++;
         var orig = functionStub;
         functionStub = spy.create(functionStub, stubLength);
         functionStub.func = orig;
+        functionStub.id = "stub#" + uuid++;
```
mantoni: Uh, sorry. Commited this by accident. I noticed that the generated `stub#${uuid}` id property is replaced by `spy.create` with `spy#{uuid}`. This change fixes it, but it's not really part of this PR. Shall I revert or just leave it?

lib/sinon/sandbox.js
```diff
@@ -161,9 +161,21 @@ function Sandbox() {
...
             Object.defineProperty(object, property, descriptor);
-        };
+        }
+        restorer.object = object;
+        restorer.property = property;
+        return restorer;
+    }
+
+    function verifyNotReplaced(object, property) {
+        for (var i = 0; i < fakeRestorers.length; i++) {
```
mroderick: Since we're targeting ES5.1, it's fine to use the ES5 Array Extras... it is much harder to create off-by-one errors, when you don't need to maintain an index 😉 

```js
function verifyNotReplaced(object, property) {
    fakeRestorers.forEach(function(fakeRestorer) {
        if (fakeRestorer.object === object && fakeRestorer.property === property) {
            throw new TypeError("Attempted to replace " + property + " which is already replaced");
        }
    });
}
```

lib/sinon/stub.js
mroderick: Maybe just put it into it's own commit?

lib/sinon/sandbox.js
mantoni: Yeah, you're right. And even `forEach` would give us an index as the second argument, if needed.

mroderick:
This has been published as `sinon@5.0.5`

PR #1786, opened 2018-05-07
Fix stub id prefix

mantoni:

Each spy, stub and fake exposes an `id` property which should be named `spy#{n}`, `stub#{n}` and `fake#{n}` with `{n}` being incremented for each new instance. In case of a stub this property is generated but then replaced with `spy#{n}` due to a wrong function call order. This pull request fixes this and adds missing test cases (there was one for fakes, but none for spy or stub).

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -105,8 +105,8 @@ var proto = {
         };
 
         var orig = functionStub;
-        functionStub.id = "stub#" + uuid++;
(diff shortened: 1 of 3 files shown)


mroderick:
This has been published as `sinon@5.0.7`

mantoni:
Thank you 🙂

Issue #1651, opened 2018-01-13
Proposal: documentation overhaul and hostnames


mroderick:
I think it's time for a major overhaul of the documentation of the `sinon` package, to make it easier to understand.

However, I'd like **all** our projects to have good documentation. The minor projects do not have great documentation.

If we decide on this, then I think we should consider using http://www.mkdocs.org with the Read The Docs theme for our API documentation projects.

I am using it at work. I find it's very easy to use, with good conventions.

I think that if we create an example of how we would like the documentation to end up, then we can recruit contributors to help speed up the work ... and in the process also get more regular contributors into the organisation.

#### Hostnames

In line with the effort to improve documentation for our projects, I have been thinking about adding more hostnames or sites.

* https://sinonjs.org — jumpstation or hub for all the active projects, tutorials, etc.
* https://lolex.sinonjs.org — API documentation
* https://nise.sinonjs.org — API documentation
* https://sinon.sinonjs.org — API documentation
* https://sinon-test.sinonjs.org — API documentation

mantoni:
Makes a lot of sense to me. Again, nice and consistent with GitHub (and the npm namespace). 👍

mgred:
Hey @mroderick 
Is this still a relevant topic? I would like to contribute if so.
The minor projects seem to be migrated to MkDocs already. Should that be done for the major one as well?

mgred:
Alright, sounds good.

As an inspiration so far, check out:
* **[Docusaurus](https://docusaurus.io/en/)**
  * Babeljs - https://babeljs.io/docs/en/
  * Jest - https://jestjs.io/docs/en/getting-started
* **[Docsify](https://docsify.js.org/)**
 * Showcase - https://docsify.js.org/#/awesome?id=showcase
* **[Asciidoctor](https://asciidoctor.org/)

mroderick:
I've updated the original post with more thoughts. Feedback and suggestions welcome

mgred:
> Use a Docker container to make it a one minute task to start contributing

That's a nice idea. I created a [branch](https://github.com/mgred/sinon/tree/sinon/1651) that includes a `docker-compose` file to serve the site from a container using a simple `jekyll` image as a first attempt.
You can run:

```bash
npm run serve-docs
```
**Note:**  You need `docker` and `docker-compose` installed.

I also updated the Documentation to point that out.

> Only a single documentation base, not one per version

[...]

PR #1794, opened 2018-05-14
Discussion: turn examples into runnable, testable fragments of code

mroderick:
🚧 do not merge 🚧 

This PR is just an opening for a discussion of how we can turn the API examples into runnable, testable code.

It often happens that when you try to learn to use a new JS package, that the examples our outdated or have typos in them. Sinon is no exception to that.

I'd like to fix this, by making all the examples used in the API documentation for `sinon@5.x` be runnable and have a basic unit test.

The changes should go into `release-source/`, I am just using the `5.0.7` as a working copy, as `release-source/` does not get built by Jekyll.

🚧 do not merge 🚧 

#### How to verify
1. Check out this branch
1. `cd docs`
1. `bundle exec jekyll serve`
1. Navigate to http://localhost:4000/releases/v5.0.7/fakes/
1. Observe the slightly updated examples
1. `npm run test-examples`
1. Observe that the tests pass

### DIFF
--- a/docs/_includes/examples/.eslintrc.yml
+++ b/docs/_includes/examples/.eslintrc.yml
@@ -1,0 +1,23 @@
+env:
+  mocha: true
+
+plugins:
+  - mocha
+
+rules:
+  no-console: off
+  max-nested-callbacks: off
+  no-restricted-syntax: [error, 'TryStatement']
+
+  # Mocha Plugin - https://github.com/lo1tuma/eslint-plugin-mocha
+  mocha/handle-done-callback: error
+  mocha/no-exclusive-tests: error
+  mocha/no-global-tests: error
+  mocha/no-hooks-for-single-case: off
+  mocha/no-identical-title: error
+  mocha/no-mocha-arrows: error
+  mocha/no-nested-tests: error
+  mocha/no-return-and-callback: error
+  mocha/no-sibling-hooks: error
+  mocha/no-skipped-tests: warn
+  mocha/no-top-level-hooks: error
--- a/docs/_includes/examples/.eslintrc.yml
+++ b/docs/_includes/examples/.eslintrc.yml
@@ -7,7 +7,7 @@ plugins:
 rules:
   no-console: off
   max-nested-callbacks: off
-  no-restricted-syntax: [error, TryStatement]
+  no-restricted-syntax: [off, TryStatement]
 
   # Mocha Plugin - https://github.com/lo1tuma/eslint-plugin-mocha
   mocha/handle-done-callback: error
--- a/docs/_includes/examples/.eslintrc.yml
+++ b/docs/_includes/examples/.eslintrc.yml
@@ -7,7 +7,7 @@ plugins:
 rules:
   no-console: off
   max-nested-callbacks: off
-  no-restricted-syntax: [error, 'TryStatement']
+  no-restricted-syntax: [error, TryStatement]
 
   # Mocha Plugin - https://github.com/lo1tuma/eslint-plugin-mocha
   mocha/handle-done-callback: error
--- a/docs/_includes/examples/fakes/__tests.js
+++ b/docs/_includes/examples/fakes/__tests.js
@@ -1,0 +1,42 @@
+"use strict";
+
+var proxyquire = require("proxyquire").noCallThru();
+var sinon = require("../../../../lib/sinon");
+
+describe("`fake` examples", function () {
+    beforeEach(function () {
+        sinon.replace(global.console, "log", sinon.fake());
+    });
+
+    afterEach(function () {
+        sinon.restore();
+    });
+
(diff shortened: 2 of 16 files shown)


fatso83:
I have thought the same thoughts many a times, and doings this is exactly in line with how the [Pragmatic Programmer](https://pragprog.com/book/tpp/the-pragmatic-programmer) was made. They also made sure all the code compiled and built before their docbook conversion :smile: 

Good initiative!

docs/_includes/examples/.eslintrc.yml
```diff
@@ -0,0 +1,23 @@
+env:
+  mocha: true
+
+plugins:
+  - mocha
+
+rules:
+  no-console: off
+  max-nested-callbacks: off
+  no-restricted-syntax: [error, 'TryStatement']
```
fearphage: `TryStatement` doesn't need quotes.

mroderick:
I've updated this PR with runnable, testable examples for pretty much all of the documentation for `fakes`.

I've updated the "How to verify" instructions.

@fatso83 @fearphage it would be great if you could spare the time to review and verify this PR before I continue.

Once there are no more comments for this PR, then I'll update all the `5.x` releases and `release-source` to use these examples and we can merge it.

docs/_includes/examples/fakes/__tests.js
```diff
@@ -0,0 +1,185 @@
+"use strict";
+
+var proxyquire = require("proxyquire").noCallThru();
```
fatso83: I think there should be a comment here that says why we are using `proxyquire`. Immediately, it didn't make sense to me why we should be stubbing out `sinon` below, until it hit me that it's of course because we want to use the most recent version.

docs/_includes/examples/fakes/__tests.js
```diff
@@ -0,0 +1,185 @@
...
+            sinon.assert.calledOnce(global.console.log);
+            sinon.assert.calledWithExactly(global.console.log, "apple pie");
+        });
+    });
+
+    // throws.js
+    describe("throws", function () {
+        it("should throw Error with message: 'not apple pie'", function () {
+            try {
+                proxyquire("./throws", {
```
fatso83: I didn't understand why `proxyquire` was necessary at first, ref comment above. Another thing is that I think all these proxyquire calls obscures what is going on. To be honest, if we intend to replace all instances of `sinon` in these tests with a link seam, I think this perfectly validates using the [`@global`](https://github.com/thlorenz/proxyquire#globally-override-require) flag to proxyquire. Using that, the tests would be simpler - most could be one-liners.

Example (this test):

```javascript
assert.exception(require.bind(null, "./throws"), "not apple pie");
```

mroderick: Maybe I am not reading the documentation right, but I am having a hard time seeing how the use of `@global` is going to turn the tests into one liners.

Could try to create a runnable example that demonstrates what you're thinking?

docs/_includes/examples/fakes/__tests.js
```diff
@@ -0,0 +1,146 @@
...
+    beforeEach(function () {
+        sinon.replace(global.console, "log", sinon.fake());
+    });
+
+    afterEach(function () {
+        sinon.restore();
+    });
+
+    // basic.js
+    describe("basic invocation", function () {
```
fearphage: These `describe` blocks here serve no purpose. If you want to make the test names more descriptive, I would suggest you just add to the test names. Could we simplify this a bit? It would be nice to use the names in the `proxyquire` as well although that's not necessary. This could be simplified into an object:

```js
{ 'basic - test name': testFn } 
// or
{ 'test name': ['basic' /* filename to load with proxyquire? */, testFn] }

// ...
Object.keys(testObject)
  .forEach(label => {
    it(label, testFn);
  })
;
```

fearphage: Another thought, why don't we read the names of the files in the test directory and just load and run them dynamically?Then we could use the file names as the key in the `testObject` above:

```js
{  './basic': testFn }
```

docs/_includes/examples/fakes/__tests.js
fatso83: Sorry - my comment doesn't make sense post-morning. Must have been high - on sleep depravation.

docs/_includes/examples/fakes/__tests.js
mroderick: Good points. I'll try to re-work this first version to incorporate these

fearphage:

fatso83:
Would love to have this merged (against `release-source`, of course).

mroderick:
This was a good discussion, thank you all for participating. I think we've proven that the idea has legs, and so I will add it as a feature for #1651

PR #1799, opened 2018-05-20
Name removal

JayAndCatchFire:
This PR is a request to remove my name and email from the AUTHORS file.

(I'm fully aware of the irony of opening a PR to do this and leaving a ridiculous paper trail of the removal of my own personal information.)


fatso83:
not sure how to handle this. this file is built automatically from the commits. removing your name would involve rewriting the history. thoughts?

JayAndCatchFire:
Meh, in that case it's not worth the trouble.

mroderick:
While it might be considered too much effort, one way to accomplish this without adding the name to another file would be to keep a hashed list of names to remove from the AUTHORS file. This list could the be used by a cleaning script to remove lines from the AUTHORS file.

Issue #1780, opened 2018-05-03
sinon.clock not found


scinos:
**Environment**

* Sinon version : 5.0.3
* Environment: OSX with Node 8.9.0

**What did you expect to happen?**

After Sinon 5, I expect Sinon root object to behave like a Sandbox, including fake timers. Sinon should be a drop-in replacement for most use cases of sandboxes.

**What actually happens**

Extra logic is required if we were using fakeTimers with the sandbox.

**What actually happens**

**How to reproduce**

With Sinon 5, I replaced all my usages of sandboxes with just sinon (like suggested in the guide), but it doesn't work for clocks.

Before:

```
const sandbox = sinon.sandbox.create();
sandbox.useFakeTimers();
//...
sandbox.clock.restore(); //this works
```

After:
```
sinon.useFakeTimers();
sinon.clock.restore(); //sinon.clock is undefined
```

Workaround:
```
const clock = sinon.useFakeTimers();
clock.restore();
```

The work around is trivial (and actually documented in the `useFakeTimers` documentation), but `sinon.clock` should exist so `sinon` is a drop-in replacement for a sandbox.

brian-mann:
Just came across this. Totally agree.

scinos:
Apologies for the confusion, but looks like I got the error wrong. There **is** a `sinon.clock`, but it is not an instance of `useFakeTimers` (`sinon.clock !== sinon.useFakeTimers()`).

This `sinon.clock` was added ages ago (it was there in v 0.5.0), but I can't find any documentation or tests for it. It is defined in https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/fake_timers.js#L31 and re-exported in https://github.com/sinonjs/sinon/blob/master/lib/sinon.js#L28. The last time it was used internally was in v1.9.1 (https://github.com/sinonjs/sinon/blob/v1.9.1/lib/sinon/util/fake_timers.js#L372), but I don't think it has been used ever since.

So the question is... is this public API? Do we want to proceed and overwrite `sinon.clock` with the result of `sinon.useFakeTimers()`?

mantoni:
I checked the docs and couldn't find the `sinon.clock` object being documented anywhere. All objects used to be created in this way (`sinon.clock.create()`, `sinon.sandbox.create()`, ...) and I think this bit is a leftover from those days.

[...]

PR #1802, opened 2018-05-21
Fix  #1801: spied fakeTimers are not restored correctly

### LINKED ISSUE #1801: Spied fakeTimers are not restored correctly
**Describe the bug**
When a fakeTimer is spied, `sandbox.restore()`/`sinon.restore()` does not restore `setTimeout` to the original function. 

Related to: #1775, #1780

**To Reproduce**

```javascript
it("restores spied fake timers when then sanddox is restored", function () {
    var originalSetTimeout = setTimeout;

    this.sandbox.useFakeTimers();
    this.sandbox.spy(global, "setTimeout");

    this.sandbox.restore();

    assert.same(originalSetTimeout, global.setTimeout, "fakeTimers restored");
});

it("should restore spied fake timers", function () {
    var originalSetTimeout = setTimeout;

    sinon.useFakeTimers();
    sinon.spy(global, "setTimeout");

    sinon.restore();

    assert.same(originalSetTimeout, global.setTimeout, "fakeTimers restored");
});

```

**Expected behavior**
Tests above should pass.

* Library version: 5.0.7
* Environment: OSX, node 8.10.0

ifrost:

When a fakeTimer is spied, sandbox.restore()/sinon.restore() does not restore setTimeout to the original function.

See #1801

When `sandbox.useFakeTimers()` is called, the original `setTimeout` and other methods are replaced with fake methods. When such fake method is spied it's replaced again with another fake. To make sure that `sandbox.restore()` reverts `setTimeout` and other methods to the original state these fakes should be restored in reverse order to make sure each fake is restored to its predecessor when calling fake.restore().

What is more the default sandbox doesn't use `sandbox.useFakeTimers`(see also #1775) and `sandbox.clock` (see #1780). To make the fix work in default sandbox it should pass `useFakeTimers` through to the sandbox method. Using `sandbox.clock` in the default sandbox added for consistency. 

1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon.js
+++ b/lib/sinon.js
@@ -24,10 +24,7 @@ var apiMethods = {
     setFormatter: format.setFormatter,
 
     // fake timers
-    useFakeTimers: fakeTimers.useFakeTimers,
-    clock: fakeTimers.clock,
     timers: fakeTimers.timers,
-
 
     // fake XHR
     xhr: nise.fakeXhr.xhr,
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -17,6 +17,7 @@ var fakeXhr = require("nise").fakeXhr;
 var usePromiseLibrary = require("./util/core/use-promise-library");
 
 // cache original versions, to prevent issues when they are stubbed in user space
+var reverse = Array.prototype.reverse;
 var push = Array.prototype.push;
 var filter = Array.prototype.filter;
 var forEach = Array.prototype.filter;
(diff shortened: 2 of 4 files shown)


lib/sinon.js
```diff
@@ -24,11 +24,8 @@ var apiMethods = {
     setFormatter: format.setFormatter,
 
     // fake timers
-    useFakeTimers: fakeTimers.useFakeTimers,
```
fatso83: So these lines that were hanging around from earlier versions essentially shadowed the new sandbox functions on the prototype. Good find!

fatso83:
Thanks for the fix and the work put in! Out as 5.0.10

mroderick:
Thank you for the contribution @ifrost 💯

Issue #1805, opened 2018-05-24
Build and export Sinon as an ES Module


brettz9:
**Is your feature request related to a problem? Please describe.**

I don't like cluttering HTML with dependencies when it is clearest to have them modularly in the particular JavaScript test files that depend on them. I don't like "magic" of not knowing where variables are supposedly coming from.

**Describe the solution you'd like**

Within each test file:

```js
import sinon from './node_modules/sinon/dist/index-es6.js';
```

or, with the following in `package.json`:

```json
"module": "dist/index-es6.js"
```
...testers could Rollup their test files with just:

```js
import sinon from 'sinon';
```

**Describe alternatives you've considered**

I could build my own module exporting this without globals but it would be a maintenance headache.

**Additional context**

Nothing.

(I would also like to see the same ES6 module option available to `sinon-test`.)

fatso83:
Hi, thanks for chiming in. This issue is just too unclear for me. I can't see that anything is not working, and I don't understand why you use a import path that doesn't exist.

1. Depending on specific submodules within Sinon is not supported, as they're not part of the official API. Thus there is only one module to depend on, and that is the one you get when doing `import sinon from 'sinon'` today. Therefore I am not seeing how this would be any different.
2. We don't have a `dist` folder. There is a `pkg` folder.
3. We don't build an ES Module and I can't see how that would enable anything we can't do today.
4. Sinon is explicitly built to be compatible with ES5.1 engines.
5. We have talked about transpilation from ES6 (for convenience) to ES5, but found it's not worth it, both in terms of more problematic debugging and the potential performance hit you can get.

brettz9:
Please see my comment that this is a feature request. I was just trying to fill in all the fields as far as what is "not working".

One can only use `import sinon from 'sinon'`, AFAICT, by treating it as a CommonJS plugin (and configuring this oneself rather than Sinon distributing such a version for ready consumption).

The current situation also has the disadvantage of not working out of the box in browsers that support ES6 modules--one must use Rollup to use Sinon in this manner. For testing, it is convenient to be able to just work without any build (or watch) step--just be able to work on the live code.

As far as debugging and performance, I guess that's the one issue I can't speak to, but Babel seems to be quite widely tested and employed by a great many projects.

However, in any case, a Rollup routine wouldn't need you to use Babel; like browserify, you could keep your source as is; you'd just need to change it to use `export`/`import` statements and use Rollup to bundle it into either UMD/CommonJS and ES versions.

[...]

PR #1809, opened 2018-05-25
Export Sinon as an ES module

fatso83:
#### Purpose 
Implements #1805 by using Rollup to export the common js bundle. Enables Sinon to be used in `<script type="module>` blocks without polluting the global namespace.

1. Check out this branch
2. `npm link`
3. `cd /tmp; mkdir foo; cd foo; npm install esm; npm link sinon;`
3.  `node -r esm -e 'import sinon from "sinon"; console.log("sinon.stub()", sinon.stub().returns(42)())'`
4. Verify that it outputs: `sinon.stub() 42`

#### Checklist for author

- [x] `npm run lint` passes

One might wonder if there is a need for having both browserify and rollup, but IMHO it adds little maintenance weight. We could get rid of one of them, but perhaps not in this PR.


rollup.config.js
```diff
@@ -0,0 +1,47 @@
+import commonjs from "rollup-plugin-commonjs";
+import nodeResolve from "rollup-plugin-node-resolve";
+
+export default {
+    input: "./lib/sinon.js",
+    output: {
+        file: "pkg/bundle-esm.js",
+        format: "esm"
```
brettz9: Just taking a quick look before getting a little time to test, but should the format be "es"? I don't see "esm" as an alias at https://rollupjs.org/guide/en

fatso83: Worked accidentally, I guess :smile: Still produced an EcmaScript module with `export default sinon`. Fixed in later commit. Thanks for noticing.

fatso83: What I suspect perhaps will cause trouble is the use of the `os` module ...

```
(!) Unresolved dependencies
https://github.com/rollup/rollup/wiki/Troubleshooting#treating-module-as-external-dependency
os (imported by node_modules/supports-color/index.js, commonjs-external:os)
```

fatso83:
Not totally sure how to automate the last step of verifying in the browser, but I have made a manual test that shows we're not quite there yet ...

I added a `index.html` file in `pkg` with this content:

```html
<script type="module">
    import sinon from '/sinon-esm.js';

    console.log('sinon is here', typeof sinon);
</script>
```

I then fired up a webserver on port 800, and pointed chrome to `localhost:8000`. The remaining problem now is the `module` global keyword. Chrome chokes when encountering:

```
    module.exports = nextTick;
```

I guess this is due to the `rollup-plugin-commonjs` not handling the conditional nature of these exports:

```javascript
if (!hasNextTick && !hasSetImmediate) {
    module.exports = nextTick;
}
```

fatso83:
https://github.com/rollup/rollup-plugin-commonjs/issues/131

fatso83:
the modules as they are currently not possible to statically convert to esm. through some minimally invasive changes I think we should be able to make it work, though by hiding the runtime detection of what is exported behind a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).  Not doing this today, and for what I know, some of that logic might be in external dependencies (would have to manually inspect - no time today). If people could delve into this, it would be great.

mroderick:
This is an exciting PR to follow!

fatso83:
For some reason - and I have no idea why - the bug with the unresolved export seems to have vanished. I have tried the bundle in both Chrome and Firefox.

My only "problem" is that I haven't been able to automatically create a verification step, as I can't find a way of [making Puppeteer verify that a test script has been run](https://gist.github.com/fatso83/b5d7d3aa366fc4e5d6b622a5b8c6c338). But manual inspection seems to indicate this is working.

mroderick:
That is what I am seeing too:

```text
$ node test/es2015/test-module-is-runnable.js
Starting browser ...
server responded with sinon module
Failed:  Error: waiting for function failed: timeout 5000ms exceeded
    at Timeout.WaitTask._timeoutTimer.setTimeout (/Users/mroderick/code/sinonjs/sinon/node_modules/puppeteer/lib/FrameManager.js:844:60)
    at ontimeout (timers.js:475:11)
    at tryOnTimeout (timers.js:310:5)
    at Timer.listOnTimeout (timers.js:270:5)
```

fatso83:
That my non-working script didn't work was already established :laughing: Everything _seems_ to work when testing with a browser, though, so in a way it's complete, but I am not comfortable merging this until we have an automatic verification step. I resorted to trying [StackOverflow](https://stackoverflow.com/questions/50568223/running-a-script-after-a-module-has-been-loaded) ... Let's see how that pans out.

mantoni:
@fatso83 Can you share the test file you're using? It's not in this PR ... or am I overlooking something?

mantoni:
@fatso83 Ah, ok. Maybe you want to try `evaluate` like mochify does this:

https://github.com/mantoni/mochify.js/blob/e993f8b37dfb1c27f9ca8dcd91ab450b127e5a95/lib/chromium.js#L155

fatso83:
@mantoni  That wouldn't work, AFAIK (and I already tried). ES Modules are async and non-blocking. If you evaluate immediately after loading the page the script hasn't had time to load the Sinon module. 

Also, I started out with an `evaluate` call: a `setTimeout` wrapped in a `Promise`, but on timeout it still couldn't find any changed globals. Doing the `waitForFunction` is essentially that (or, more like `setInterval` with a timeout). On the other hand, if I start a real Chrome instance and point to the server I can evaluate manually in the console almost immediately after page load and see that `window.sinonLoaded` is defined.

mantoni:
Yeah, I evaluate the script synchronously and then monitor the `console` events from the page. This way I get notified asynchronously without polling.

fatso83:
@mantoni Thanks for the tip on `console.log`. In the end, it was this that made me realize I was missing the right mime type on the resources provided by the http server, making the browser refuse to execute the contents of the module. 

I have now implemented a check that does a simple test of the stubbing functionality. It works and builds as expected, meaning this should be ready for review :smiley:

mroderick:
These tests are failing in older node

fatso83:
:boom: 

:heavy_check_mark: Tests passing

PR #1811, opened 2018-05-26
Add in matcher

mhmoudgmal:
#### Purpose (TL;DR)

Add `sinon.match.in` matcher

I had a case where I wanted to match a computed value (a value that I don't know exactly what it is), but I knew the possibilities of that value.

so I had to do something like this:
`sinon.match.same('something').or(sinon.match.same('something_else'))`

```js
const expectedPayload = {
    experiment_uuid: 'target1',
    uuid: 'fake-uuid',
    variation: sinon.match.same('control').or(sinon.match.same('group1'))
};

```

I know also there is array.contains matcher.. but the case above, that payload is used in different places where I check if a function called with it, and I thought it would be much easier if I have  something like this: `sinon.match.in([v1, v2, v3])`

#### Solution
*sinon.match.in* matcher.
Allows to match a value to one of the elements of an array.
And it can be useful when the the value is not known,
but could be one of few possibilities provided as an array.

Usage:
sinon.match.in([p1, p2, p3, ..])

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/matchers.md
+++ b/docs/release-source/release/matchers.md
@@ -181,6 +181,10 @@ Requires the value to be a `Date` object.
 
 Requires the value to be a `Symbol`.
 
+#### `sinon.match.in(array)`
+
+Requires the value to be in the `array`.
+
 *Since `sinon@2.0.0`*
 
 #### `sinon.match.same(ref)`
--- a/docs/release-source/release/matchers.md
+++ b/docs/release-source/release/matchers.md
@@ -49,18 +49,15 @@ Requires the value to be == to the given number.
 
 Requires the value to be a string and have the expectation as a substring.
 
-
 #### `sinon.match(regexp);`
 
 Requires the value to be a string and match the given regular expression.
-
 
 #### `sinon.match(object);`
 
 Requires the value to be not `null` or `undefined` and have at least the same properties as `expectation`.
 
 This supports nested matchers.
-
 
 #### `sinon.match(function)`
 
@@ -146,7 +143,6 @@ Requires the value to be a `Map`.
 
 Requires a `Map` to be deep equal another one.
 
-
 #### `sinon.match.map.contains(map)`
 
 Requires a `Map` to contain each one of the items the given map has.
@@ -185,12 +181,9 @@ Requires the value to be a `Symbol`.
 
 Requires the value to be in the `array`.
 
-*Since `sinon@2.0.0`*
-
 #### `sinon.match.same(ref)`
 
 Requires the value to strictly equal `ref`.
-
 
 #### `sinon.match.typeOf(type)`
 
@@ -211,7 +204,6 @@ Requires the value to be of the given type, where `type` can be one of
 #### `sinon.match.instanceOf(type)`
 
 Requires the value to be an instance of the given `type`.
-
 
 #### `sinon.match.has(property[, expectation])`
 
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -179,7 +179,7 @@ match.same = function (expectation) {
 };
 
 match.in = function (arrayOfExpectations) {
-    if (!(arrayOfExpectations instanceof Array)) {
+    if (!Array.isArray(arrayOfExpectations)) {
         throw new TypeError("array expected");
     }
 
--- a/lib/sinon/match.js
+++ b/lib/sinon/match.js
@@ -178,6 +178,18 @@ match.same = function (expectation) {
     }, "same(" + valueToString(expectation) + ")");
 };
 
+match.in = function (arrayOfExpectations) {
+    if (!(arrayOfExpectations instanceof Array)) {
+        throw new TypeError("array expected");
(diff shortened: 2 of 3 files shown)


lib/sinon/match.js
```diff
@@ -178,6 +178,18 @@ match.same = function (expectation) {
     }, "same(" + valueToString(expectation) + ")");
 };
 
+match.in = function (arrayOfExpectations) {
+    if (!(arrayOfExpectations instanceof Array)) {
```
mroderick: Please use [`Array.isArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) to accurately detect instances of Array.

See http://web.mit.edu/jwalden/www/isArray.html for a detailed explanation.

test/match-test.js
```diff
@@ -440,6 +440,52 @@ describe("sinonMatch", function () {
...
+            assert(sinonMatch.isMatcher(inMatcher));
+        });
+
+        context("when the test is called with non array argument", function () {
+            it("throws TypeError (array expected)", function () {
+                var arg = "not-array";
+
+                assert.exception(function () {
+                    sinonMatch.in(arg);
+                }, {name: "TypeError"});
```
mroderick: It would be good to also test for the message, so we can be sure that it's actually expected error. Otherwise, another error might go undetected by this test for a long time.

test/match-test.js
```diff
@@ -440,6 +440,52 @@ describe("sinonMatch", function () {
...
+                [{ a: "a" }, { b: "b"}],
+                [function () {}, function () {}],
+                [null, undefined]
+            ];
+
+            context("and one of the elements is the same as the actual", function () {
+                it("returns true", function () {
+                    arrays.forEach(function (array) {
+                        var inMatcher = sinonMatch.in(array);
+                        assert(inMatcher.test(array[0]));
```
mroderick: Referee's [`assert` checks for truthy, not `true`](https://github.com/sinonjs/referee/blob/master/docs/index.md#assert).

You'll want to use [`assert.isTrue`](https://github.com/sinonjs/referee/blob/master/docs/index.md#istrue) to test for `true`.

test/match-test.js
```diff
@@ -440,6 +440,52 @@ describe("sinonMatch", function () {
...
+                    });
+                });
+            });
+
+            context("and none of the elements is the same as the actual", function () {
+                it("returns true", function () {
```
mroderick: Copy-pasta, this should read ` it("returns false", function () {`

mhmoudgmal:
@mroderick Thanks a lot 👍
I updated the code and added the documentation, please rephrase it if it's not clear.

test/match-test.js
```diff
@@ -440,6 +440,52 @@ describe("sinonMatch", function () {
         });
     });
 
+    describe(".in", function () {
+        it("returns matcher", function () {
+            var inMatcher = sinonMatch.in([]);
+            assert(sinonMatch.isMatcher(inMatcher));
+        });
+
+        context("when the test is called with non array argument", function () {
```
fearphage: 2 things here:

* Use `describe` instead _when necessary_
* Only nest tests when there is additional/unique setup

In this instance, the nesting is not warranted. Just change the label/name on the test below.

test/match-test.js
```diff
@@ -440,6 +440,52 @@ describe("sinonMatch", function () {
...
+            it("throws TypeError (array expected)", function () {
+                var arg = "not-array";
+
+                assert.exception(function () {
+                    sinonMatch.in(arg);
+                }, {name: "TypeError", message: "array expected"});
+            });
+        });
+
+        context("when the test is called with an array", function () {
```
fearphage: Use `describe`.

test/match-test.js
```diff
@@ -440,6 +440,52 @@ describe("sinonMatch", function () {
...
+            var arrays = [
+                [1, 2, 3],
+                ["a", "b", "c"],
+                [{ a: "a" }, { b: "b"}],
+                [function () {}, function () {}],
+                [null, undefined]
+            ];
+
+            context("and one of the elements is the same as the actual", function () {
+                it("returns true", function () {
```
fearphage: Rename the tests instead of nesting them here and on the following tests.

docs/release-source/release/matchers.md
```diff
@@ -181,6 +181,10 @@ Requires the value to be a `Date` object.
 
 Requires the value to be a `Symbol`.
 
+#### `sinon.match.in(array)`
+
+Requires the value to be in the `array`.
+
 *Since `sinon@2.0.0`*
```
fatso83: The line about Sinon 2 belongs to the symbol matcher above. I'll fix it in my browser, if my crappy Moto G5 can take the load 😁

mroderick:
This has been published to NPM as `sinon@5.1.0`

PR #1812, opened 2018-05-29
Updated CircleCI test environment

fearphage:
* removed node v4
* added node v10
* running esm tests
* updated `check-esm-bundle-is-runnable.js` to allow chrome path overrides (via env variables)


fearphage:
Done.

mroderick:
> Mind squashing those commits?

You can squash when merging, if you just need one commit

PR #1823, opened 2018-06-01
Fixed coveralls integration

fearphage:
Chrome was failing to execute due to chrome executing as root. I added the `--allow-chrome-as-root` flag to `mochify`.


mroderick:
Thanks

Issue #1796, opened 2018-05-17
Cannot stub `Array.prototype.sort`


artyhedgehog:
**Describe the bug**
Applying stub to `Array.prototype.sort` results in RangeError: Maximum call stack size exceeded.

**To Reproduce**
Run test like this:

```js
const sinon = require('sinon');

describe('Cannot stub Array.prototype.sort', () => {
  it('should not fall with call stack size exceeded error', () => {
    // Even making it always call through would not help:
    const stub = sinon.stub(Array.prototype, 'sort').callThrough();
    [1, 2, 3].sort();
    stub.restore();
  });
});
```

**Expected behavior**
Test should pass.

**Actual behavior**
Falls with error:
```
     RangeError: Maximum call stack size exceeded
      at Function.invoke (node_modules/sinon/lib/sinon/spy.js:1:1)
      at Array.proxy (node_modules/sinon/lib/sinon/spy.js:78:54)
      at Array.functionStub (node_modules/sinon/lib/sinon/stub.js:101:36)
      at Function.invoke (node_modules/sinon/lib/sinon/spy.js:209:51)
      at Array.proxy (node_modules/sinon/lib/sinon/spy.js:78:54)
      at Array.functionStub (node_modules/sinon/lib/sinon/stub.js:101:36)
      at Function.invoke (node_modules/sinon/lib/sinon/spy.js:209:51)
      at Array.proxy (node_modules/sinon/lib/sinon/spy.js:78:54)
      ...
```

**Context:**

* Library version: 5.0.7
* Environment: nodejs v8.11.1
* Command: /usr/bin/node ./node_modules/mocha/bin/_mocha sinon.spec.js | head -n20
* Other libraries you are using: mocha@5.1.1

**Additional context**
Looks similar to: https://github.com/sinonjs/sinon/issues/1521

fatso83:
I am sure the fix to this is very similar to #1521. If you look grep the source for '.sort()' you should find it quickly. A PR is most welcome ;-)

fatso83:
I think Phred fixed this in in #1826. Could you verify?

fatso83:
Verified. Still present :(

mroderick:
This was fixed in #1826, a test to verify it was added in #1876

mroderick:
The fix for this has been published to NPM in `sinon@6.1.5`

PR #1826, opened 2018-06-06
Refactored native calls

fearphage:
So I've seen this method used in [jQuery](https://github.com/jquery/jquery/tree/master/src/var) to handle native method calls. I haven't checked here, but they all get compiled away cleanly in jQuery. I just touched a few files at first to see what everyone thought of the methodology. This would be a global change to combat things like #1796 from popping up.

Note: It's incomplete, but the tests are green.

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -1,19 +1,25 @@
 "use strict";
 
+var arrayProto = require("./var/array");
 var calledInOrder = require("./util/core/called-in-order");
 var orderByFirstCall = require("./util/core/order-by-first-call");
 var timesInWords = require("./util/core/times-in-words");
 var format = require("./util/core/format");
 var sinonMatch = require("./match");
+var stringSlice = require("./var/string").slice;
 
-var slice = Array.prototype.slice;
+var arraySlice = arrayProto.slice;
+var concat = arrayProto.concat;
+var forEach = arrayProto.forEach;
+var join = arrayProto.join;
+var splice = arrayProto.splice;
 
 var assert;
 
 function verifyIsStub() {
-    var args = Array.prototype.slice.call(arguments);
+    var args = arraySlice(arguments);
 
-    args.forEach(function (method) {
+    forEach(args, function (method) {
         if (!method) {
             assert.fail("fake is not a spy");
         }
@@ -64,7 +70,7 @@ function mirrorPropAsAssertion(name, method, message) {
     assert[name] = function (fake) {
         verifyIsStub(fake);
 
-        var args = slice.call(arguments, 1);
+        var args = arraySlice(arguments, 1);
         var failed = false;
 
         verifyIsValidAssertion(name, args);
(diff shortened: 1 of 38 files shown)


fearphage:
> Today, bugs with native functions happen because we forget to check/fail to see that native functions are used directly.

The problem today is we assume the native is still functional/not broken. If we just stored a copy beforehand, we wouldn't face this problem... unless it was broken before sinon is loaded and then we're hosed unless we do something fancier like getting clean `Array`s and `Object`s from an iframe... which is definitely doable.

fatso83:
At least that's one of the problems :smiley: It wouldn't prevent stuff like #1796 from popping up in future PRs, but at least it would serve as a reminder and also avoid the whole "is this native function still functional?" thing in the cases where we [remember to use a cached reference](https://github.com/sinonjs/sinon/blame/master/.github/PULL_REQUEST_TEMPLATE.md#L40).

fearphage:
> It wouldn't prevent stuff like #1796 from popping up in future PRs

I think it would make it easier to see. Calling `randomArray.sort` is what's broken in that issue. If we had cached `Array#sort`, it would be fine. We should essentially disallow those calls in our code. Maybe there's an eslint plugin for this? I think this also makes it a bit more distinct to see visually (`sort(array, sorter)` vs `array.sort(sorter)`) and I think it compiles away cleanly. I haven't tested the later, but I think so.

mroderick:
I like where this PR is going, more defensive code is a good approach for a library.
An ESLint plugin would be very helpful to get contributors to do the right thing. Not sure if such a thing exists yet

fatso83:
I think I have found all we need for such an eslint plugin!

# Use the skeleton for `no-use-extend-native`
[This module](https://github.com/dustinspecker/eslint-plugin-no-use-extend-native/blob/master/src/no-use-extend-native.js) has all the structure in place and most of the code we need. It's basically a matter of removing some checks/inverting the conditions.

## It uses `is-proto-prop`
[This lib](https://www.npmjs.com/package/is-proto-prop) checks  if the prop used is a property on the prototype, but, unlike my initial thoughts based on its name, _it only checks for native props_.

It does this using [`proto-props`](https://www.npmjs.com/package/proto-props), which essentially is a json file listing all the native js props - exactly what we pondered!

mroderick:
That looks promising!

mroderick:
I like the approach of using static analysis better than introducing more code. What do you think @fearphage?

fearphage:
> I like the approach of using static analysis better than introducing more code.

Is it an either or situation? An eslint-style rule will not make the code work correctly.

mroderick:
> An eslint-style rule will not make the code work correctly.

My understanding was that it could be used for preventing us from merging code that was vulnerable to end users overriding native methods. If that was incorrect, then I agree, it won't.

akdor1154:
Maybe `arrayPrototype = Object.freeze(Object.assign({}, Array.prototype))` could be used to avoid having a separate file for every array method you want to use internally?

fearphage:
@akdor1154 `Object.assign` only sees enumerable properties so it misses everything on the prototype. I did use this to try a different direction though. Thank you.

fearphage:
FYI his branch now passes with the new custom eslint rule (#1858).

fatso83:
Great, then I suggest merging this first and the eslint rule second to see the build pass. Thanks for the work!

mroderick:
👍

PR #1831, opened 2018-06-07
 Document/test call ordering checks

DullReferenceException:
#### Purpose
Add documentation for call-level methods that check for relative call ordering

#### Background
Based on the documentation, I was struggling to find ways to determine call ordering of specific calls. I noticed `callId` could be used by looking through the code for `spy.prototype.calledBefore` and thought it'd be useful to add a similar set of `called(Immediately)?(Before|After)` methods for spy calls. To my surprise, my tests passed without me having to add any code, so apparently these methods already exist and just aren't documented.

#### Checklist for author
- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/spy-call.md
+++ b/docs/release-source/release/spy-call.md
@@ -73,6 +73,28 @@ Returns `true` if call threw exception of provided type.
 Returns `true` if call threw provided exception object.
 
 
+### `spyCall.calledBefore(otherCall)`
+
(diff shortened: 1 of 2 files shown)


fatso83:
Screw waiting; looked through the code, and this just seems like we forgot to document at some point. Thanks for your work!

mroderick:
Thank you @DullReferenceException

PR #1836, opened 2018-06-13
Fix mocha test file pattern to match subdirectories correctly

ifrost:
#### Purpose (TL;DR) 

The current test file pattern passed to mocha is not picking up subdirectories correctly (e.g. `issues-test.js` is not included). See also https://github.com/mochajs/mocha/issues/1828.

Running `npm test` before: `1260 passing (515ms)`
Running `npm test` after: `1503 passing (754ms)`

TODO:
- [x] use escaped double quotes

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   },
   "license": "BSD-3-Clause",
   "scripts": {
-    "test-node": "mocha --recursive -R dot test/**-test.js",
+    "test-node": "mocha --recursive -R dot 'test/**/*-test.js'",
     "test-dev": "npm run test-node -- --watch -R min",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,9 +14,9 @@
   },
   "license": "BSD-3-Clause",
   "scripts": {
-    "test-node": "mocha --recursive -R dot test/**-test.js",
+    "test-node": "mocha --recursive -R dot 'test/**/*-test.js'",
```
fatso83: This probably won't fly with Windows, if I am not mistaken. I think it needs to use escaped double quotes.

mroderick:
Thank you @ifrost

Issue #1840, opened 2018-06-15
Default sandbox: useFakeXMLHttpRequest is not restored correctly


ifrost:
**Describe the bug**
`sinon.useFakeXMLHttpRequest()` replaces global XMLHttpRequest object, but it doesn't restore it when `sinon.restore()` is called. The problem is that sinon object overrides useFakeXMLHttpRequest method instead of using the one from Sandbox class and the fake is not added to the collection that is being restored when `sinon.restore()` is called.

**To Reproduce**
```
it("should restore XMLHttpRequest and ActiveXObject", function () {
    var globalXHR = global.XMLHttpRequest;
    var globalAXO = global.ActiveXObject;
    sinon.useFakeXMLHttpRequest();

    sinon.restore();

    assert.same(global.XMLHttpRequest, globalXHR);
    assert.same(global.ActiveXObject, globalAXO);
});
```

**Expected behavior**
The test above should pass.

**Context (please complete the following information):**

* Library version: 6.0.0
* Environment: OSX, node 8.10.0

mroderick:
This was fixed with #1841 and published to NPM as `sinon@6.1.1`

PR #1841, opened 2018-06-15
Default sandbox fake request

ifrost:
Fix for #1840

#### Solution  - optional

Use useFakeXMLHttpRequest from Sandbox class.

1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -371,7 +371,8 @@ function Sandbox() {
 
     sandbox.useFakeXMLHttpRequest = function useFakeXMLHttpRequest() {
         var xhr = fakeXhr.useFakeXMLHttpRequest();
-        return push.call(collection, xhr);
+        push.call(collection, xhr);
+        return xhr;
     };
 
     sandbox.usingPromise = function usingPromise(promiseLibrary) {
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -483,7 +483,6 @@ describe("issues", function () {
     });
 
     describe("#1840 - sinon.restore useFakeXMLHttpRequest", function () {
-
         it("should restore XMLHttpRequest and ActiveXObject", function () {
             sinon.useFakeXMLHttpRequest();
             sinon.restore();
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -5,6 +5,8 @@ var sinon = require("../../lib/sinon");
 var createStub = require("../../lib/sinon/stub");
 var assert = referee.assert;
 var refute = referee.refute;
+var globalXHR = global.XMLHttpRequest;
+var globalAXO = global.ActiveXObject;
 
 describe("issues", function () {
     beforeEach(function () {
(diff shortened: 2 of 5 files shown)


test/sinon-test.js
```diff
@@ -113,10 +113,5 @@ describe("sinon module", function () {
             });
         });
 
-        describe("useFakeXMLHttpRequest", function () {
-            it("should be the fakeXhr.useFakeXMLHttpRequest export from nise", function () {
```
mantoni: Instread of removing this test, you could change it to verify that the nise implementation is invoked :)

ifrost: Updated.

lib/sinon/sandbox.js
```diff
@@ -349,7 +349,8 @@ function Sandbox() {
 
     sandbox.useFakeXMLHttpRequest = function useFakeXMLHttpRequest() {
         var xhr = fakeXhr.useFakeXMLHttpRequest();
-        return push.call(collection, xhr);
+        push.call(collection, xhr);
+        return xhr;
```
fearphage: Why was this change made? There should be a test to confirm this change in functionality.

test/sinon-test.js
```diff
@@ -115,7 +115,18 @@ describe("sinon module", function () {
...
             it("should be the fakeXhr.useFakeXMLHttpRequest export from nise", function () {
-                assert.equals(sinon.useFakeXMLHttpRequest, fakeNise.fakeXhr.useFakeXMLHttpRequest);
+                // use full sinon for this test as it compares sinon instance
+                // proxyquire changes the instance, so `actual instanceof Sandbox` returns `false`
+                // see https://github.com/sinonjs/sinon/pull/1586#issuecomment-354457231
+                sinon = require("../lib/sinon");
+
+                var nise = require("nise");
+
+                sinon.spy(nise.fakeXhr, "useFakeXMLHttpRequest");
```
fearphage: Everything here and above should be in the setup (`beforeEach`).

test/sinon-test.js
```diff
@@ -115,7 +115,18 @@ describe("sinon module", function () {
...
+                sinon.useFakeXMLHttpRequest();
+                assert.isTrue(nise.fakeXhr.useFakeXMLHttpRequest.called);
+
+                nise.fakeXhr.useFakeXMLHttpRequest.restore();
```
fearphage: This should be in an `afterEach`. If the assertion above fails, this line will never be called in its current form.

test/issues/issues-test.js
```diff
@@ -479,4 +481,16 @@ describe("issues", function () {
             assert.same(originalSetTimeout, global.setTimeout, "fakeTimers restored");
         });
     });
+
+    describe("#1840 - sinon.restore useFakeXMLHttpRequest", function () {
+
```
fearphage: MINOR: Remove empty lines around the test below.

lib/sinon/sandbox.js
ifrost: This is to keep behaviour of `sinon.useFakeXMLHttpRequest` the same. 

Sinon object has been made the default sandbox but it doesn't use `sandbox.useFakeXMLHttpRequest` method, making it impossible to restore fake xhr object using `sinon.restore`.

This PR changes it so sinon object uses `sandbox.useFakeXMLHttpRequest`. The behaviour of the latter was, however, slightly different as it would return: 

`return push.call(...)`

It means it returns *the length of the internal fake collection* (what doesn't make any sense) instead of fake xhr object returned by nise.

I've added a test to sandbox-test.js.

test/sinon-test.js
ifrost: Setup moved to `beforeEach`

test/sinon-test.js
ifrost: Clean up moved to `afterEach`.

test/issues/issues-test.js
ifrost: Empty line removed.

mroderick:

Your branch needs a rebase before it can be merged.

mroderick:
This has been published to NPM as `sinon@6.1.1`

PR #1843, opened 2018-06-16
Add SemVer compatibility badge to README

greysteil:
#### Purpose (TL;DR)

Add a README badge that displays SemVer compatibility, calculated using the test suites of sinon users.

First up, thanks for sinon!

Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so threw together the below:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=sinon&package-manager=npm_and_yarn&version-scheme=semver)

Would love your opinion - is it useful? Anything you'd change about it? If you click it there's an explanation of how it's calculated.

Any feedback super appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
 <a href="#sponsors"><img src="https://opencollective.com/sinon/sponsors/badge.svg" alt="OpenCollective"></a>
 <a href="https://www.npmjs.com/package/sinon" target="_blank"><img src="https://img.shields.io/npm/dm/sinon.svg" alt="npm downloads per month"></a>
 <a href="https://cdnjs.com/libraries/sinon.js" target="_blank"><img src="https://img.shields.io/cdnjs/v/sinon.js.svg" alt="CDNJS version"></a>
+<a href="https://dependabot.com/compatibility-score.html?dependency-name=sinon&package-manager=npm_and_yarn&version-scheme=semver" target="_blank"><img src="https://api.dependabot.com/badges/compatibility_score?dependency-name=sinon&package-manager=npm_and_yarn&version-scheme=semver" alt="SemVer compatibility"></a>
 </p>
 
 <p align=center>


fearphage:
> Would love your opinion - is it useful? Anything you'd change about it? If you click it there's an explanation of how it's calculated.

While the badge itself is not very useful, the explanation isn't very helpful either in my opinion. Are you saying we are doing semver correctly 90% of the time? If so, that's not completely clear. If that's not what that number represents, then I may be completely confused.

I'm going to close this PR for now, but feel free to open it if you add additional information or clarity to the issue.

greysteil:
Sounds like I'm doing a bad job of explaining! Lots of projects have said they really like the badge, and it's now included on the jest, webpack and mocha readmes,  but you're not the only one who hasn't immediately "got it". (I could be wrong of course, and you just don't think the data is very interesting!)

The data in the badge is the percentage of sinon update PRs that Dependabot has created where the end user's test suite passes without any code changes required (filtering for only non-major updates, and excluding repos with flaky or initially failing tests). So the text if you click through is:

> Dependabot has updated sinon between SemVer compatible versions 688 times across 86 projects so far. 96% of those updates passed CI.

Since Dependabot PRs isolate a single dependency update, and don't make any other changes (example PR [here](https://github.com/nextcloud/mail/pull/956)), they can be used as a measure of whether a new version contains any bugs or breaks compatibility. I think that's useful information for users.

The folks at webpack suggested a slightly different badge using the same data: they wanted to have a badge for updates to the latest version only, so I created this one:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=sinon&package-manager=npm_and_yarn&version-scheme=semver&new-version=latest)

^^^ At the moment, that badge will only include updates from 6.0.0 to 6.0.1, but when 6.0.2 is released it will change to include updates from `6.0.0 -> 6.0.1`, `6.0.1 -> 6.0.2` and `6.0.0 -> 6.0.2`. The idea there is that if 6.0.2 introduces any bugs it should be immediately obvious from the badge, and releasing a version 6.0.3 that fixed those issues would immediately get the badge back to 100%.

Does that help clarify, or have I got the wrong end of the stick and you just think that data about whether users can upgrade without their tests breaking just isn't interesting?

fatso83:
I think this is a quite nice addition!

PR #1845, opened 2018-06-16
Add documentation for stub.yieldsRight

nivsherf:
Adding missing documentation for `stub.yieldsRight`.

1. Check out this branch
2. `npm install`
3. Review the stub docs.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/_releases/v1.17.6/stubs.md
+++ b/docs/_releases/v1.17.6/stubs.md
@@ -250,12 +250,17 @@ Similar to `callsArg`.
 
 Causes the stub to call the first callback it receives with the provided arguments (if any).
 
-If a method accepts more than one callback, you need to use `callsArg` to have the stub invoke other callbacks than the first one.
+If a method accepts more than one callback, you need to use `yieldsRight` to call the last callback or `callsArg` to have the stub invoke other callbacks than the first or last one.
+
+
+#### `stub.yieldsRight([arg1, arg2, ...])`
+
+Like `yields` but calls the last callback it receives.
 
 
 #### `stub.yieldsOn(context, [arg1, arg2, ...])`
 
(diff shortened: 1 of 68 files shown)


mroderick:
Thank you for your pull request. This looks good to me. Would you mind also folding the changes into the older versions for which this applies?

nivsherf:
@mroderick Done. (Better late than never...)

fatso83:
Thanks, @nivsherf. We know the "apply the change to older releases" step is troublesome. Did you find a way that made the process more slick for you? Would like to find some (semi-)automatic way of applying a patch to the older release docs.

nivsherf:
@fatso83 Well, in this case a simple replace did the trick. But since you asked, I also wrote a bash script that given a doc section and a semver number (e.g "stubs" and "4.1.1") will open the editor to edit the .md file for that version, then recursively do a 3-way merge on subsequent versions, stopping on conflicts so you can resolve them. A bit like "rebasing" the versioned docs. It's very basic and very poorly tested - mostly just a POC for myself... but I can share it if you're interested.

fatso83:
This is our weakest point, and increasingly a pain point, so if you could share the script in gist or something, that would be great.

nivsherf:
@fatso83 Here's a [gist](https://gist.github.com/nivsherf/ae01f0b377bd24b10102e99cdf3a2e0d).

nivsherf:
[Here](https://gist.github.com/nivsherf/66cb49196ad4683652318164bf83ead7) is another version which works the other way around - it assumes changes were already made in `release-source/` and propagates them to older versions. I guess this better fits your use-case, although it makes the conflict-resolution a bit less intuitive. Also handles editing the `release.md` file, which the previous version did not.

PR #1848, opened 2018-06-25
Made callsArg, returnsArg, and throwsArg more strict

fearphage:
All will now throw if not called with enough arguments when before they would return undefined.

Example:

`stub.returnsArg(5)('not', 'enough', 'params')` would previously return `undefined`. Now it throws.

`callsArg` would previously throw a `TypeError` complaining that the param (`undefined`) was not a `function`.

They all throw the same `TypeError` now with a consistent message.


fearphage:
Inspired by a suggestion in #1846. If this merges first, that PR should be assimilated as well.

fearphage:
Note: This could be considered a change to people who were expecting the silent fall through (failure?). I'm not sure. I'll leave that up to the semver gods. :wink:

test/stub-test.js
```diff
@@ -402,14 +417,11 @@ describe("stub", function () {
                 function () {
                     stub("only", "two arguments");
                 },
-                function (error) {
-                    return error instanceof TypeError
-                        && error.message ===
-                        "throwArgs failed: 3 arguments required but only 2 present"
```
fearphage: Of note: This message changed: `throwArgs` (typo?) became `throwsArg`.

huttli:
There is an error at the number vs. arg index in the old error messages for throwsArg. According to the doc, the index starts at 0. I think the check should be done between these two numbers: `args.length` and `throwArgAt +1`.
Please use the correct number in the error message thrown.

fearphage:
@huttli Great catch! Thanks. I pushed the fix.

mroderick:
> Inspired by a suggestion in #1846. If this merges first, that PR should be assimilated as well.

Does that mean that we should not merge this, until #1846 is ready to merge?

fearphage:
It doesn't matter which one goes first honestly. It would be easier and more convenient (for the contributor) to merge that PR first and then I can update this one to include it.

mantoni:
Regarding versioning, I'd classify this as a bug. If it's breaking someones test, it's highlighting an issue that was always there and also needs fixing. Neither the semantics nor the API has changed.

mroderick:
This has been published to NPM as `sinon@6.1.2`

PR #1853, opened 2018-07-06
Fix cyclic deep equal

mroderick:
This PR replaces #1764, which had stalled.

The pull request is a solution for #1709 .

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/util/core/deep-equal.js
+++ b/lib/sinon/util/core/deep-equal.js
@@ -1,54 +1,14 @@
 "use strict";
 
-var div = typeof document !== "undefined" && document.createElement("div");
+var samsam = require("@sinonjs/samsam");
 
 function isReallyNaN(val) {
     return val !== val;
 }
 
-function isDOMNode(obj) {
-    var success = false;
-
-    try {
-        obj.appendChild(div);
-        success = div.parentNode === obj;
-    } catch (e) {
-        return false;
-    } finally {
-        try {
-            obj.removeChild(div);
-        } catch (e) {
-            // Remove failed, not much we can do about that
-        }
-    }
-
-    return success;
-}
-
-function isElement(obj) {
-    return div && obj && obj.nodeType === 1 && isDOMNode(obj);
-}
-
 var deepEqual = module.exports = function deepEqual(a, b) {
     if (typeof a !== "object" || typeof b !== "object") {
         return isReallyNaN(a) && isReallyNaN(b) || a === b;
-    }
-
-    if (isElement(a) || isElement(b)) {
-        return a === b;
-    }
-
-    if (a === b) {
-        return true;
-    }
-
-    if ((a === null && b !== null) || (a !== null && b === null)) {
-        return false;
-    }
-
-    if (a instanceof RegExp && b instanceof RegExp) {
-        return (a.source === b.source) && (a.global === b.global) &&
-            (a.ignoreCase === b.ignoreCase) && (a.multiline === b.multiline);
     }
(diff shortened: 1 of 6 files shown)


test/util/core/deep-equal-test.js
```diff
@@ -352,4 +352,51 @@ describe("util/core/deepEqual", function () {
...
+        var b = {
+            foo: "bar",
+            key: { }
+        };
+        b.key.value = b;
+
+        assert(deepEqual(a, b));
+    });
+
+    it.skip("handles cyclic objects when a matcher provided", function () {
```
fearphage: Is this intentional? You meant to add a skipped test?

fearphage: Added a skipped tests is weird to me. It sounds like this test needs to be added when the bug is fixed or you need to assert that the functionality is broken as expected. Adding a skipped test is like adding a commented out test to me. It doesn't do anything meaningful for the current code base.

mroderick:
Before declaring this a victory, I thought I'd try out the original reproducible scenario from #1709.
I guess there's still some work to do on this one

test/util/core/deep-equal-test.js
```diff
@@ -352,4 +352,53 @@ describe("util/core/deepEqual", function () {
...
+            key: { }
+        };
+        b.key.value = b;
+
+        assert(deepEqual(a, b));
+    });
+
+    // deepEqual still doesn't handle cyclic objects when a matcher is providedd
+    // this test will verify (near) future implementations
+    it.skip("handles cyclic objects when a matcher provided", function () {
```
fearphage: Adding a skipped tests is weird to me. It sounds like this test needs to be added when the bug is fixed or you need to assert that the functionality is broken as expected. Adding a skipped test is like adding a commented out test to me. It doesn't do anything meaningful for the current code base.

mroderick:
@fearphage @mantoni I think I've managed to fix all the known scenarios. Would you mind giving it another look before I merge it?

lib/sinon/util/core/deep-equal.js
```diff
@@ -1,99 +1,46 @@
...
-            (a.ignoreCase === b.ignoreCase) && (a.multiline === b.multiline);
-    }
-
     if (a instanceof Error && b instanceof Error) {
         return a === b;
     }
 
-    var aString = Object.prototype.toString.call(a);
-    if (aString !== Object.prototype.toString.call(b)) {
+    if (Object.prototype.toString.call(a) !== Object.prototype.toString.call(b)) {
```
fearphage: This should be cached on import instead of used "live" like this.

lib/sinon/util/core/deep-equal.js
```diff
@@ -1,99 +1,46 @@
...
-    if (aString === "[object Date]") {
-        return a.valueOf() === b.valueOf();
-    }
-
-    var prop;
-    var aLength = 0;
-    var bLength = 0;
-
-    if (aString === "[object Array]" && a.length !== b.length) {
+    var haveDifferentKeys = Object.keys(a).sort().join() !== Object.keys(b).sort().join();
```
fearphage: I think this is straight-forward enough to put inline in the `if` statement without the single-use variable in my opinion.

mroderick:
Thanks @fearphage

mroderick:
This has been published to NPM as `sinon@6.1.3`

PR #1855, opened 2018-07-08
Fix spy documentation and fix contributing typo

emmabukacek:
* Fix `matchers` link in `spies.md` and `spy-call.md` across all versions
* Fix typo in `docs/CONTRIBUTING`

When I click `see matchers`, I would expect the link to take me to the `matchers` page, but this instead sets an anchor that doesn't exist on the page. Correct me if I'm missing something here.

In addition, the links don't work in Github's markdown viewer, but I assume something with how the doc paths are reconciled makes this work in the actual doc viewer. To try and ensure the correct path, I took a look at the `[calls]: ../spy-call` link which appears to work in the actual documentation. Let me know if this is not the case and I'll be happy to update. :smile: 

1. Check out this branch
2. `npm install`
3. Validate docs for `spies.md` and `spy-call.md` from the `v1.17.6-v6.1.3`

#### Checklist for author

- [x] `npm run lint` passes
  - NOTE: One warning, but it looks unrelated: 
```
sinon/test/sandbox-test.js
  217:34  warning  Do not create function within loops (IE11 compatibility)  ie11/no-loop-func
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contribruting to documentation
+# Contributing to documentation
 
 ## Documenting the next release
 
--- a/docs/_releases/v1.17.6/spies.md
+++ b/docs/_releases/v1.17.6/spies.md
@@ -231,7 +231,7 @@ Returns `true` if the spy was called after `anotherSpy`
 
 #### `spy.calledOn(obj);`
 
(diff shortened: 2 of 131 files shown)


mroderick:
Thank you for your very first pull request to Sinon!

emmabukacek:
Thanks for merging, @mroderick ! :smile:

PR #1856, opened 2018-07-08
Additional test coverage

fearphage:
Mostly test changes, but I did refactor `next-tick` to make it more easily testable. I was dancing with proxyquire and even looked at pulling in [rewire](https://github.com/jhnns/rewire), but dependency injection seemed like a more simplistic solution.

This is mainly ticking up the coverage numbers and not much else. :upside_down_face: 

[Ignoring whitespace](https://github.com/sinonjs/sinon/pull/1856/files?w=1) in the assert tests makes it easier to see.

### DIFF
--- a/lib/sinon/util/core/next-tick.js
+++ b/lib/sinon/util/core/next-tick.js
@@ -1,25 +1,8 @@
 "use strict";
 
+var getNextTick = require("./get-next-tick");
+
+/* istanbul ignore next */
 var root = typeof window !== "undefined" ? window : global;
 
-function nextTick(callback) {
-    /* istanbul ignore next */
-    setTimeout(callback, 0);
-}
-
-function getNextTick(process, setImmediate) {
-    if (typeof process === "object" && typeof process.nextTick === "function") {
-        return process.nextTick;
-    }
-
-    if (typeof setImmediate === "function") {
-        return setImmediate;
-    }
-
-    return nextTick;
-}
-
 module.exports = getNextTick(root.process, root.setImmediate);
-
-// exposed for testing purposes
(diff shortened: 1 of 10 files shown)


lib/sinon/util/core/next-tick.js
```diff
@@ -1,20 +1,25 @@
...
 
-if (!hasNextTick && !hasSetImmediate) {
-    module.exports = nextTick;
+    return nextTick;
 }
+
+module.exports = getNextTick(root.process, root.setImmediate);
+
+// exposed for testing purposes
+module.exports.getNextTick = getNextTick;
```
mroderick: Exposing private methods for testing is an anti-pattern in my opinion. Sometimes there is no (feasible) way around it. I don't think this is one of those times.

A couple of other approaches to consider

* put `getNextTick` in it's own file, test that and test that the result of that is used by `next-tick.js`(link seam)
* make `getNextTick` the export from this module, and let consumers call it with arguments (dependency injection)

In this case, my preference would be to use dependency injection for a couple of reasons:

* it creates fewer files
* this is not a method on the public API, called by end users

fearphage: My only complaint with the dependency injection route (in this specific case) is that it is not clean. Because I'm referring to node things (`process` and `setImmediate`), I have to do the "which global" dance (`window` vs `global`). That seems messy to spread to every file (2 currently) that needs `nextTick`. I've never heard the term link seam used before, but I prefer that method in this case. It keeps the messiness contained to one file.

mroderick:
Nice work 👍

PR #1858, opened 2018-07-11
Custom eslint rule to detect prototype method usage

fearphage:
This will help to prevent errors going forward. This was inspired by discussions in #1826. The code for the rule started life as the built-in [no-prototype-builtins](https://github.com/eslint/eslint/blob/master/lib/rules/no-prototype-builtins.js). I modified it slightly to make it work for our purposes. I didn't do all the things required to make it a real big boy rule/plugin. This seemed like a valid shortcut for now.

Note: This new file could contain other custom rules as well.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,4 +1,4 @@
-extends: 'eslint-config-sinon'
+extends: sinon
 
 globals:
   ArrayBuffer: false
@@ -9,9 +9,11 @@ globals:
 
 plugins:
   - ie11
+  - local-rules
 
 rules:
   ie11/no-collection-args: error
   ie11/no-for-in-const: error
   ie11/no-loop-func: warn
   ie11/no-weak-collections: error
+  local-rules/no-prototype-methods: error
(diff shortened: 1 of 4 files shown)


fearphage:
FYI The rule is intentionally failing, because the code does not adhere to the rules.

fatso83:
Any experience in writing a codemod that will turn `myarr.filter( myfunc )` into `filter.call(myarr, myfunc)`, etc.? Seems like quite a lot of mundane work to make our code comply. Amazed at the amount of breaking code (281 errors).

fearphage:
> Any experience in writing a codemod that will turn myarr.filter( myfunc ) into filter.call(myarr, myfunc), etc.?

That transform is easy. There still needs to be an include added in the head of the file though... in theory. That's the hard (impossible?) part. I haven't seen eslint rules perform changes to lines other than the active/target lines/code.

> Seems like quite a lot of mundane work to make our code comply. Amazed at the amount of breaking code (281 errors).

I did that in another branch already.

fatso83:
If one writes a codemod that does the transform, manually going through the files to add the include is a small thing. But since you've already done it, I guess the situation is somewhat moot :)

Issue #1857, opened 2018-07-10
createStubInstance - optional overrides


ifrost:
**Is your feature request related to a problem? Please describe.**
Stubbing multiple properties require to write the var name multiple times, for example:

```js
var stubbedObject = sinon.createStubInstance(MyClass);
stubbedObject.foo = sinon.stub().returns(2);
stubbedObject.bar = sinon.stub().callsFake(function() { return 1; });
stubbedObject.xyzzy = sinon.stub().returns('test');
```

**Describe the solution you'd like**
It'd be nice to have a shortcut syntax to provide optional overrides for createStubInstance, for example:

```js
var stubbedObject = sinon.createStubInstance(MyClass, {
  foo: sinon.stub().returns(2),
  bar: sinon.stub().callsFake(function() { return 1; }),
  xyzzy: sinon.stub().returns('test')
});
```

**Describe alternatives you've considered**
The same result could be achieved with "with":

```js
with (stubbedObject = sinon.createStubInstance(MyClass)) {
  foo = sinon.stub().returns(2);
  bar = sinon.stub().callsFake(function() { return 1; });
  xyzzy = sinon.stub().returns('test');
}
```

"with" expression, however, is marked as deprecated and probably it's not the best idea to use it.

Another alternative I thought about was to provide functions to give more flexibility:

```js
var stubbedObject = sinon.createStubInstance(MyClass, {
  foo: function(stub, property) { 
    stub[property] = sinon.stub().returns(2);
  }
});
```

The version above would be easier to reuse. For example, sometimes we use fluent interfaces:

```js
var stubbedObject = sinon.createStubInstance(MyClass);
stubbedObject.foo = sinon.stub().returnsThis();
stubbedObject.bar = sinon.stub().returnsThis();
stubbedObject.xyzzy = sinon.stub().returnsThis();
```

Could have even shorter syntax:

```js
var returnsThis = function(stub, propertyName) {
  stub[propertyName] = sinon.stub().returnsThis();
}

var stubbedObject = sinon.createStubInstance(MyClass, {
  foo: returnsThis,
  bar: returnsThis,
  xyzzy: returnsThis
});
```

ifrost:
Hi, the PR for it is still open. Any chance of getting it merged?

mroderick:
This was implemented in #1864 and published to NPM as `sinon@6.3.0`

PR #1864, opened 2018-07-23
Allow providing stubs overrides for sinon.createStubInstance

ifrost:

Implements #1857. The solution is slightly different from what I suggested #1857 - it allows simple stub overrides or providing returned value as described in docs:

```javascript
var stub = sinon.createStubInstance(MyConstructor, overrides);
```

`overrides` is an optional map overriding created stubs, for example:

```javascript

var stub = sinon.createStubInstance(MyConstructor, {
    foo: sinon.stub().returnsThis()
});
```

is the same as:

```javascript
var stub = sinon.createStubInstance(MyConstructor);
stub.foo.returnsThis();
```

If provided value is not a stub, it will be used as the returned value:

```javascript
var stub = sinon.createStubInstance(MyConstructor, {
    foo: 3
});
```
is the same as:

```javascript
var stub = sinon.createStubInstance(MyConstructor);
stub.foo.returns(3);
```

1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -111,7 +111,7 @@ is the same as:
 
 ```javascript
 var stub = sinon.createStubInstance(MyConstructor);
-stub.foo = sinon.stub().returnsThis()
+stub.foo.returnsThis();
 ```
 
 If provided value is not a stub, it will be used as the returned value:
@@ -125,7 +125,7 @@ is the same as:
 
 ```javascript
 var stub = sinon.createStubInstance(MyConstructor);
-stub.foo = sinon.stub().returns(3)
+stub.foo.returns(3);
 ```
 
 #### `stub.withArgs(arg1[, arg2, ...]);`
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -95,7 +95,37 @@ Stubbing individual methods tests intent more precisely and is less susceptible 
 If you want to create a stub object of `MyConstructor`, but don't want the constructor to be invoked, use this utility function.
 
 ```javascript
-var stub = sinon.createStubInstance(MyConstructor)
+var stub = sinon.createStubInstance(MyConstructor, overrides);
+```
+
+`overrides` is an optional map overriding created stubs, for example:
+
+```javascript
+
+var stub = sinon.createStubInstance(MyConstructor, {
+    foo: sinon.stub().returnsThis()
+});
+```
+
+is the same as:
+
+```javascript
+var stub = sinon.createStubInstance(MyConstructor);
+stub.foo = sinon.stub().returnsThis()
+```
+
+If provided value is not a stub, it will be used as the returned value:
+
+```javascript
+var stub = sinon.createStubInstance(MyConstructor, {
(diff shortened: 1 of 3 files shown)


mroderick:
Things have been a bit busy lately, so I am a bit behind on my open source work.
I will get around to take a closer look at this, when I can give it the attention it deserves.
Thank you for your patience

ifrost:
Hi, the PR for it is still open. Any chance of getting it merged?

mroderick:
Thank you for your patience ❤️ 
This has been published as `sinon@6.3.0`

fatso83:
Hmm ... isn't the pre-test checks running on the PRs? This failed after merging to master.

mroderick:
> Hmm ... isn't the pre-test checks running on the PRs? This failed after merging to master.

I think we changed the linting since the PR was created:

https://travis-ci.org/sinonjs/sinon/jobs/427844368

mroderick:
I can fix this lint violation and update the `preversion.sh` to include linting

fatso83:
The linting failed because of [not using a cached reference to `forEach`](https://github.com/sinonjs/sinon/pull/1864/files#diff-048fb1194de4b43f7f7e45758f900749R76). This makes it susceptible to bugs where people stub out `Array.prototype.forEach` in their own tests.

Even if we have just recently added the explicit check for this (good!), we _did_ have a point in our [pull request template](https://github.com/sinonjs/sinon/blob/master/.github/PULL_REQUEST_TEMPLATE.md) just for this:
> - [ ] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

There is a reason we have this checklists, @ifrost, so removing them should only be done if you understand why they are there in the first place and if they are irrelevant.

mroderick:
@fatso83 https://github.com/sinonjs/sinon/pull/1902

mroderick:
@fatso83 in fairness, [those requirements were added](https://github.com/sinonjs/sinon/commit/052a2d0b26aa7ccff1193c16b5828ed7c7293d5d#diff-195a635ad245ded9076330e31134bd80) after this PR was opened.

fatso83:
@mroderick No, they were added one year ago (2017).

mroderick:
> @mroderick No, they were added one year ago (2017).

Oh!

ifrost:
> The linting failed because of [not using a cached reference to `forEach`](https://github.com/sinonjs/sinon/pull/1864/files#diff-048fb1194de4b43f7f7e45758f900749R76). This makes it susceptible to bugs where people stub out `Array.prototype.forEach` in their own tests.
> 
> Even if we have just recently added the explicit check for this (good!), we _did_ have a point in our [pull request template](https://github.com/sinonjs/sinon/blob/master/.github/PULL_REQUEST_TEMPLATE.md) just for this:
> 
> > * [ ]  References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).
> 
> There is a reason we have this checklists, @ifrost, so removing them should only be done if you understand why they are there in the first place and if they are irrelevant.

Sorry, my bad! Thanks for fixing it!

PR #1867, opened 2018-07-27
Use https everywhere and remove legacy links

mantoni:
Latest Google Chrome now shows a non-https pages as "not secure". Since we now have a valid certificate, we should link to the `https` version of the site everywhere. I already changed the repo description link and the org homepage URL.

While going through the docs I noticed that the <http://legacy.sinonjs.org> URL does not work anymore. Since nobody complained about that, I think it's time to remove it.


mroderick:
I've ticked the Enforce HTTPS everywhere button, it should take effect soon.

2018-07-30 at 19 33

mantoni:
👍 Thank you!

PR #1869, opened 2018-07-28
Fixed remaining ESLint failures

fearphage:
This PR is mostly for tracking purposes (instead of committing directly to master).

* Disabled new ESLint rule for `tests` directory and `build.js`
* Disabled an *unrelated* warning (`ie11/no-loop-func`) that has existed for a long time

Now `eslint .` is perfectly silent. \o/ :fireworks: :champagne: \o/


fearphage:
Related to #1826 and #1858.

Issue #1863, opened 2018-07-18
Sandbox.usingPromise does not set promise library for mocks


nhinds:
**Describe the bug**
Calling `.usingPromise(promiseLibrary)` on a sandbox works for stubs created from the sandbox, but does not appear to work for mocks created from the sandbox. Mocks continue to use the node.js default Promise library

**To Reproduce**
Create a sandbox using e.g. Bluebird as its promise library, and attempt to create stubs/mocks which return Promises using `.resolves`:
```js
const sinon = require('sinon');

// Set up a sandbox using bluebird as the promise library
const sandbox = sinon.createSandbox().usingPromise(require('bluebird'));

// Call a stub which returns a resolved promise
const stubPromise = sandbox.stub().resolves('value for stub')();
// Call bluebird API methods on the returned Promise object
stubPromise.tap(console.log).return('final value for stub').then(console.log);
// Works :)

// Call a mock which returns a resolved promise
const mockPromise = sandbox.mock().resolves('value for mock')();
// Call bluebird API methods on the returned Promise object
mockPromise.tap(console.log).return('final value for mock').then(console.log);
// TypeError: mockPromise.tap is not a function

// Call a method on a mock which returns a resolved promise
const objectToMock = { method: function() {} };
sandbox.mock(objectToMock).expects('method').resolves('value for mock method');
const mockMethodPromise = objectToMock.method();
// Call bluebird API methods on the returned Promise object
mockMethodPromise.tap(console.log).return('final value for mock method').then(console.log);
// TypeError: mockMethodPromise.tap is not a function
```

**Expected behavior**
I would expect mocks returned from the sandbox to return Promises using the library passed to `sandbox.usingPromise()`.

As a workaround, `.usingPromise(...)` can be called on each stubbing chain on each mock, e.g.:
```js
const mockPromise = sandbox.mock().usingPromise(require('bluebird')).resolves('value for mock')();

sandbox.mock(objectToMock).expects('method').usingPromise(require('bluebird')).resolves('value for mock method');
```

**Context (please complete the following information):**

* Library version:  6.1.3
* Environment: node.js v8.10.0
* Other libraries you are using: bluebird 3.5.1

**Additional context**
N/A

mroderick:
Thank you for your bug report.

That does indeed look like a defect.

Do you want to make a pull request to fix it?

nhinds:
Sure, opened #1872.

fatso83:
Merged #1862 in 6.2

PR #1872, opened 2018-07-31
Using promise for sandbox mocks

nhinds:
Fix issue #1863 by adding `mock.usingPromise(promiseLibrary)` and calling this from `sandbox.mock`

#### Solution  - optional
There did not appear to be an existing way to influence the Promise library for all expectations on a mock, so I introduced `mock.usingPromise(promiseLibrary)` as an API method and used this to pass the promise library from the sandbox through to mock expectations.

It would be possible to do this without introducing an API method by setting fields on `mock` directly from `sandbox`, if introducing this new API is not okay.

1. Check out this branch
2. `npm install`
3. `npm test`

or original steps from #1863:

```js

// Set up a sandbox using bluebird as the promise library
const sandbox = sinon.createSandbox().usingPromise(require('bluebird'));

// Call a stub which returns a resolved promise
const stubPromise = sandbox.stub().resolves('value for stub')();
// Call bluebird API methods on the returned Promise object
stubPromise.tap(console.log).return('final value for stub').then(console.log);
// Works :)

// Call a mock which returns a resolved promise
const mockPromise = sandbox.mock().resolves('value for mock')();
// Call bluebird API methods on the returned Promise object
mockPromise.tap(console.log).return('final value for mock').then(console.log);
// Works :)

// Call a method on a mock which returns a resolved promise
const objectToMock = { method: function() {} };
sandbox.mock(objectToMock).expects('method').resolves('value for mock method');
const mockMethodPromise = objectToMock.method();
// Call bluebird API methods on the returned Promise object
mockMethodPromise.tap(console.log).return('final value for mock method').then(console.log);
// Works :)
```

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/mocks.md
+++ b/docs/release-source/release/mocks.md
@@ -83,6 +83,14 @@ If any expectation is not satisfied, an exception is thrown.
 
 Also restores the mocked methods.
 
+#### `mock.usingPromise(promiseLibrary);`
+
+Causes all expectations created from the mock to return promises using a specific
+Promise library instead of the global one when using `expectation.rejects` or
+`expectation.resolves`. Returns the mock object to allow chaining.
+
+*Since `sinon@6.2.0`*
+
 
 ### Expectations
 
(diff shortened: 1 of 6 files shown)


mroderick:
Thank you for your contribution 🏆 
It's summer time and the maintainers are busy enjoying their lives. We will be looking into your pull request as soon as we can find the time to give it the attention it deserves.
Thank you for your patience

fatso83:
Sinon 6.2.0 is out! Thanks for your contribution!

PR #1876, opened 2018-08-06
Add test for #1796, failing to stub Array.prototype.sort

mroderick:
This PR shows that the issue described in #1796 has been fixed (in c2c78143ffd381418b9847e8518cadd0741dfddc)

Once this is merged, we should create a new release.

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -508,4 +508,12 @@ describe("issues", function () {
             sinon.assert.calledWith(spy, secondObj);
         });
     });
+
+    describe("#1796 - cannot stub Array.prototype.sort", function () {
+        it("it should not fail with RangeError", function () {
+            var stub = sinon.stub(Array.prototype, "sort");
+            [1, 2, 3].sort();
+            stub.restore();
+        });
+    });
 });
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -511,9 +511,11 @@ describe("issues", function () {
 
     describe("#1796 - cannot stub Array.prototype.sort", function () {
         it("it should not fail with RangeError", function () {
-            var stub = sinon.stub(Array.prototype, "sort");
-            [1, 2, 3].sort();
-            stub.restore();
+            refute.exception(function () {
+                var stub = sinon.stub(Array.prototype, "sort");
+                [1, 2, 3].sort();
+                stub.restore();
+            });
         });
     });
 });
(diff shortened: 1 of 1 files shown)


test/issues/issues-test.js
```diff
@@ -508,4 +508,12 @@ describe("issues", function () {
             sinon.assert.calledWith(spy, secondObj);
         });
     });
+
+    describe("#1796 - cannot stub Array.prototype.sort", function () {
+        it("it should not fail with RangeError", function () {
+            var stub = sinon.stub(Array.prototype, "sort");
+            [1, 2, 3].sort();
```
fearphage: All tests should have assertions. Assert that this doesn't throw.

test/issues/issues-test.js
```diff
@@ -508,4 +508,12 @@ describe("issues", function () {
             sinon.assert.calledWith(spy, secondObj);
         });
     });
+
+    describe("#1796 - cannot stub Array.prototype.sort", function () {
+        it("it should not fail with RangeError", function () {
+            var stub = sinon.stub(Array.prototype, "sort");
```
fearphage: This stub needs to `.callThrough()` to trigger the `RangeError`, right?

test/issues/issues-test.js
mroderick: Good point. I've added an extra commit. Once happy with this PR, I'll fold the commits together

test/issues/issues-test.js
mroderick: Nope, `.callThrough()` is not necessary to trigger the `RangeError`.

test/issues/issues-test.js
```diff
@@ -508,4 +508,14 @@ describe("issues", function () {
             sinon.assert.calledWith(spy, secondObj);
         });
     });
+
+    describe("#1796 - cannot stub Array.prototype.sort", function () {
+        it("it should not fail with RangeError", function () {
+            refute.exception(function () {
+                var stub = sinon.stub(Array.prototype, "sort");
+                [1, 2, 3].sort();
```
fearphage: This should be the only line that we are concerned with not throwing an exception.

mroderick: Correct, I've pushed a commit that fixes that

mroderick:
Dang, I merged without squashing ...

Issue #1885, opened 2018-08-25
Decorator Function


nicosierra:
**Is your feature request related to a problem? Please describe.**
There are times that I have found myself with the need of decorating or modifying the response from a method, while still stubbing the method (like a mix of `callsFake` and `callThrough`), so as to be able to use sinon methods to assess different test scenarios.

A practical example:
Stripe is a payment system that provides test tokens to make requests and test their API.
Unfortunately, they don't provide all test tokens (e.g. apple pay, google pay), so in order to be able to test those scenarios, we need to decorate such responses with a respective parameter that reflects those type of payments (i.e. `tokenization_method`). This is one of several scenarios in which I have been on the need of modifying the response from a library. One may argue that for these tests scenarios I could implement a decorator service myself, instead of relying on sinon, but I find quite useful to stub methods with sinon, and then be able to check the times they were called, the arguments, etc.

**Describe the solution you'd like**
Implement a sinon method in between `callsFake` and `callThrough`, i.e. a method that allows you to execute the original method (with its respective arguments), while being able to attach a function that receives such result as an argument.

`stub.callsThroughFake(decoratorFunction)`

**Describe alternatives you've considered**
So far, this is the way I have been able to do so

```javascript
const Service = {
  aMethod: (a, b) => a + b,
};

const decoratorFunction = result => result + 10;

const originalAMethod = Service.aMethod;
const modifiedAMethod = (...args) => {
  const result = originalAMethod(...args);
  return decoratorFunction(result);
};
const stub = sinon.stub(Service, 'aMethod').callsFake(modifiedAMethod);
Service.aMethod(1, 2); // 13
```

not really compact 😕

PR #1886, opened 2018-08-25
add callsThroughFake(decoratorFunction)

nicosierra:
a method which allows to decorate the result from the
wrapped method. It behaves like callThrough plus
callsFake, i.e. the provided function will be
executed by receiving as an argument
the result from the original wrapped method

Example: Implement feature #1885 by adding the `callsThroughFake` method

There are times that I have found myself with the need of decorating or modifying the response from a stubbed method (like an extended version of callsFake), while still being able to use all of the sinon methods to assess different test scenarios.

A practical example:
Stripe is a payment system that provides test tokens to make requests and test their API.
Unfortunately, they don't provide all test tokens (e.g. apple pay, google pay), so in order to be able to test those scenarios, we need to decorate such responses with a respective parameter that reflects those type of payments (i.e. tokenization_method). This is one of several scenarios in which I have been on the need of modifying the response from a library. One may argue that for these tests scenarios I could implement a decorator service myself, instead of relying on sinon, but I find quite useful to stub methods with sinon, and then been able to check the times they were called, the arguments, etc.

I added the functionality using as reference the `callThrough` and `callsFake` implementation.
Please add comments for my PR. First time I see this codebase.

1. Check out this branch
2. `npm install`
3. run the tests, then try the function by yourself.

```javascript
const Service = {
  aMethod: (a, b) => a + b,
};

const decoratorFunction = result => result + 10;

const stub = sinon.stub(Service, 'aMethod').callsThroughFake(decoratorFunction);
Service.aMethod(1, 2); // 13
```
😄 
#### Checklist for author

- [ ] `npm run lint` passes
- [ ]  documentation is proper
- [ ] test the function

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -243,6 +243,22 @@ sinon.stub(myObj, 'prop').callsFake(function fakeFn() {
 myObj.prop(); // 'bar'
 ```
 
+#### `stub.callsThroughFake(decoratorFunction);`
+Makes the stub call the provided `decoratorFunction` with the result from the call of the original method wrapped into the stub, when none of the conditional stubs are matched.
+
+```javascript
+var myObj = {};
+myObj.prop = function propFn() {
+    return 'foo';
+};
+
+sinon.stub(myObj, 'prop').callsThroughFake(function decoratorFunction(result) {
+    return result + '-bar';
+});
+
+myObj.prop(); // 'foo-bar'
+```
+
 #### `stub.returns(obj);`
 Makes the stub return the provided value.
 
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -160,6 +160,8 @@ var proto = {
             throw args[this.throwArgAt];
         } else if (this.fakeFn) {
             return this.fakeFn.apply(context, args);
+        } else if (this.callsThroughFakeFn) {
+            return this.callsThroughFakeFn.apply(context, args);
         } else if (typeof this.resolveArgAt === "number") {
             ensureArgs("resolvesArg", this, args);
             return (this.promiseLibrary || Promise).resolve(args[this.resolveArgAt]);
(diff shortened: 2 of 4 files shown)


fatso83:
@nicosierra First of all, thanks for your contribution! It seems like a nice utility method. I am just wondering if we need yet another utility method ... My biggest concern ATM with regards to maintainability of Sinon is the sheer size of the API. 

**A small digression**
I have to admit, that even after using Sinon for half a decade I really don't use more than a handful of its methods, and I have to lookup the docs every time I come across some use of even simple methods such as `yield`. I just find it to be more readable for the next one coming along to see `sinon.stub(obj, 'method').callsFake( cb => cb('some error') )` rather than needing to look up the docs for Sinon's own method made for the purpose. 

Likewise, I'd probably just write the following for simplicity, rather than use the one you made just for this purpose.

```javascript
const orgMethod = Service.aMethod;
const decoratedMethod = arg => orgMethod(arg) + 10
const stub = sinon.stub(Service, 'aMethod').callsFake(decoratedMethod);
```
It's just a matter of weighing saved LOC versus the mental load of keeping a vast API surface in your head. I'm a goldfish, hence one extra line over one extra method :smile: 

That being said, others probably don't mind adding yet another, and I do like the resulting code which is readable and shows good intent. Also, your PR seems otherwise flawless :+1: I'll leave it up to Morgan or any of the others in @sinonjs/sinon-core to decide. Great work, anyhow!

nicosierra:
hi @fatso83, thanks for your words! and your solution!
It is actually what I do in my tests (I had added it as an alternative solution in the related issue https://github.com/sinonjs/sinon/issues/1885).

I had always thought that it would be useful to have an hybrid of `callThrough` and `callsFake` in one method, although you are right, it is better to keep the project as succinct as possible.
Anyway, thanks to you guys for bringing such an useful tool.
Cheers!

fearphage:

fatso83:
For the reasons mentioned, me and Morgan are both in favour of not adding yet another utility method.

Issue #1900, opened 2018-09-11
calledWith returning false positives when using dates


mrwillihog:
**Describe the bug**
When calling the `calledWith` or any variant with a date parameter it is returning true, even though the dates clearly do not match. 

**To Reproduce**
```
const sinon = require('sinon');

const stub = sinon.stub();
const dateOne = new Date('2018-09-11');
const dateTwo = new Date('1999-12-31');

stub(dateOne);

// Expect stub.calledWith to return false (different dates)
console.log(stub.calledWith(dateTwo));
```

**Expected behavior**
We would expect `stub.calledWith()` to return false in this instance.

**Context (please complete the following information):**

* Library version: 6.2.0 (latest release)

**Additional context**
We believe this bug was introduced in this commit: https://github.com/sinonjs/sinon/commit/014aecff40025b74e1bb27edeaceb5d5beb3ff6d. Specifically, this check here: https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/core/deep-equal.js#L41-L43 is returning true. In our specific use case, `Object.keys(a)` is returning an empty array, which when passed to `Array.every` is also returning true.

mrwillihog:
The use case above works with version 6.1.0

fatso83:
Great bug report!

mrwillihog:
Probably worth mentioning that `samsam.deepEqual` works as expected, it returns `false` for this use case.

PR #1901, opened 2018-09-11
Adds guard for empty properties in deepEqual when a matcher is provided

mrwillihog:
Fix issue #1900 by guarding against empty properties in `deepEqual` when a matcher is provided.

My big assumption here is that the behaviour of `Array.every` when passed an empty array is not expected. I've tried to keep the pattern of only exiting the `deepEqual` function early when we find equality. If we end up at the bottom of the function and we have a matcher (due to being called via `calledWith`) then we should not assume we are an object with properties. Guarding against this by checking the length of `Object.keys` fixes this bug and does not break any other tests so I assume it is the correct solution.

Happy to hear alternative suggestions though as I am not too familiar with the code base.

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -520,4 +520,17 @@ describe("issues", function () {
             stub.restore();
         });
     });
+
+    describe("#1900 - calledWith returns false positive", function () {
+        it("should return false when call args don't match", function () {
+            var stub = sinon.stub();
+            var dateOne = new Date("2018-07-01");
+            var dateTwo = new Date("2018-07-31");
+
+            stub(dateOne);
+
+            var calledWith = stub.calledWith(dateTwo);
+            assert.same(calledWith, false);
+        });
+    });
 });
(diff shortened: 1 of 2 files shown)


test/issues/issues-test.js
```diff
@@ -520,4 +520,17 @@ describe("issues", function () {
             stub.restore();
         });
     });
+
+    describe("#1900 - calledWith returns false positive", function () {
+        it("should return false when call args don't match", function () {
+            var stub = sinon.stub();
```
mroderick: In order to eliminate potential for regressions, it would be better to verify this using `spy`, as a `stub` is a `spy`, but a `spy` is not a `stub` (the same is true for `spy` and `fake`).

Something like this:

```js
var spy = sinon.spy();
var dateOne = new Date("2018-07-01");
var dateTwo = new Date("2018-07-31");

spy(dateOne);

var calledWith = spy.calledWith(dateTwo);
assert.same(calledWith, false);
```

fearphage:

mroderick:
This has been published to NPM as `sinon@6.3.2`

Issue #1882, opened 2018-08-22
sinon.stub().withArgs() matches too often 


tbiesemann:
Sinon 6.1.5
This test fails:

```js
it('sinon use constructor name when checking deepEquality', function() {
    class ClassWithoutProps {};
    const arg1 = new ClassWithoutProps(); //arg1.constructor.name === ClassWithoutProps
    const arg2 = new Proxy({}, {});	//arg2.constructor.name === Object
    const stub = sinon.stub();
    stub.withArgs(arg1).returns(5);
    const result = stub(arg2);
    assert.strictEqual(result, undefined); //[ERR_ASSERTION]: 5 === undefined
});
```
**Describe the bug**
withArgs() has too many matches.
Empty objects, instances of empty classes or proxies are all treated the same.
The deepEqual function in sinon returns true when comparing those.

One could use the constructor.name for further comparison in deepEqual, to solve some of these false matches.

mgred:
@tbiesemann, @mroderick 
Just found out that the [`samsam.deepEqual()`](https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/core/deep-equal.js#L36) is the function in charge here that returns `true`.
I opened an issue in the `samsam` repo sinonjs/samsam#37

mroderick:
@mgred I don't think we're quite done yet

* `npm install @sinonjs/samsam@2.1.0`
* Adding a slightly modified version if the provided test to `issues-test.js`

```js
describe("#1882", function () {
    it("should use constructor name when checking deepEquality", function () {
        class ClassWithoutProps {};
        const arg1 = new ClassWithoutProps(); //arg1.constructor.name === ClassWithoutProps
        const arg2 = new Proxy({}, {}); //arg2.constructor.name === Object
        const stub = sinon.stub();
        stub.withArgs(arg1).returns(5);
        const result = stub(arg2);
        assert.same(result, undefined); //[ERR_ASSERTION]: 5 === undefined
    });
});
```

Produces `AssertionError: [assert.same] 5 expected to be the same object as undefined`.

Would you take another look at this one?

mgred:
@mroderick Sure I will.

Also: sinonjs/samsam#39

mgred:
@mroderick, In the PR i added your test and it passes for me.

mroderick:
@tbiesemann can you verify that the issue has been fixed in latest version of Sinon?

fatso83:
[...]

PR #1905, opened 2018-09-16
test(1882): add test case

mgred:

See: #1882, sinonjs/samsam#37

* add test case provided in #1882
* change some eslint config to match test:
** class definitions are not enabled in ecmaVersion 5,
** addint Proxy as global

#### Solution  - optional

This test is passing for me locally

1. Check out this branch
2. `npm install`
3. `npm run test-node`
#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,10 +6,14 @@ globals:
   Promise: false
   Set: false
   Symbol: false
+  Proxy: false
 
 plugins:
   - ie11
   - local-rules
+
+parserOptions:
+  ecmaVersion: 6
 
 rules:
   ie11/no-collection-args: error
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,14 +6,10 @@ globals:
   Promise: false
   Set: false
   Symbol: false
-  Proxy: false
 
 plugins:
   - ie11
   - local-rules
-
-parserOptions:
-  ecmaVersion: 6
 
 rules:
   ie11/no-collection-args: error
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -536,12 +536,15 @@ describe("issues", function () {
 
     describe("#1882", function () {
         it("should use constructor name when checking deepEquality", function () {
-            class ClassWithoutProps {}
-            const arg1 = new ClassWithoutProps(); //arg1.constructor.name === ClassWithoutProps
-            const arg2 = new Proxy({}, {}); //arg2.constructor.name === Object
-            const stub = sinon.stub();
+            function ClassWithoutProps() {}
+            function AnotherClassWithoutProps() {}
+            ClassWithoutProps.prototype.constructor = ClassWithoutProps;
+            AnotherClassWithoutProps.prototype.constructor = AnotherClassWithoutProps;
+            var arg1 = new ClassWithoutProps(); //arg1.constructor.name === ClassWithoutProps
+            var arg2 = new AnotherClassWithoutProps(); //arg2.constructor.name === Object
+            var stub = sinon.stub();
             stub.withArgs(arg1).returns(5);
-            const result = stub(arg2);
+            var result = stub(arg2);
             assert.same(result, undefined); //[ERR_ASSERTION]: 5 === undefined
         });
     });
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -533,4 +533,16 @@ describe("issues", function () {
             assert.same(calledWith, false);
         });
     });
+
+    describe("#1882", function () {
+        it("should use constructor name when checking deepEquality", function () {
+            class ClassWithoutProps {}
+            const arg1 = new ClassWithoutProps(); //arg1.constructor.name === ClassWithoutProps
+            const arg2 = new Proxy({}, {}); //arg2.constructor.name === Object
+            const stub = sinon.stub();
+            stub.withArgs(arg1).returns(5);
+            const result = stub(arg2);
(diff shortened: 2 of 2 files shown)


.eslintrc.yml
```diff
@@ -6,11 +6,15 @@ globals:
   Promise: false
   Set: false
   Symbol: false
+  Proxy: false
 
 plugins:
   - ie11
   - local-rules
 
+parserOptions:
```
mroderick: Sinon is an ES5.1 project, as [described in CONTRIBUTING.md](https://github.com/sinonjs/sinon/blob/master/CONTRIBUTING.md#compatibility).

If we allow ES6 syntax, we drop support for IE11 and older Safari. I don't think we're quite there yet.

test/issues/issues-test.js
```diff
@@ -533,4 +533,16 @@ describe("issues", function () {
             assert.same(calledWith, false);
         });
     });
+
+    describe("#1882", function () {
+        it("should use constructor name when checking deepEquality", function () {
+            class ClassWithoutProps {}
```
mroderick: `class` is an ES6 convenience over using `prototype`. This means that you can achieve the same using `prototype`.

```js
function ClassWithoutProps(){}
ClassWithoutProps.prototype.constructor = ClassWithoutProps;
```

test/issues/issues-test.js
```diff
@@ -533,4 +533,16 @@ describe("issues", function () {
...
+            const arg1 = new ClassWithoutProps(); //arg1.constructor.name === ClassWithoutProps
+            const arg2 = new Proxy({}, {}); //arg2.constructor.name === Object
```
mroderick: Can this test be done without using [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), which is from ES6 and is not supported in IE11?

Perhaps create `AnotherClassWithoutProps`?

test/issues/issues-test.js
mgred: Ahhh, yeah let me change this...

test/issues/issues-test.js
mgred: That's true I will rewrite with to your suggestion, which makes the most sense I think

mgred:
Thanks for all your suggestions, and sorry for ignoring the ES5 restriction for IE11. I keep learning.

mroderick:
Thank you 👍

fatso83:
[WebWorker test crashed on Node 10](https://circleci.com/gh/sinonjs/sinon/801?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for some unknown reason after merge to `master`.

fatso83:
This test started failing on IE 11 in our Sauce Labs suite last Friday after merge. It was "hidden" from view due to the webworker test also failing on Circle CI@Node 10  (for some unknown reason I am trying to look into).

See the [logs](https://saucelabs.com/beta/tests/ce1b40752b59482e9aae5c7e7f144b8e/logs#16) ([overview](https://saucelabs.com/u/sinonjs)):
```
    "result": "....\n\n  1572 passing (9s)\n  8 pending\n  1 failing\n\n  1) issues\n       #1882\n         should use constructor name when checking deepEquality:\n     AssertionError: [assert.same] 5 expected to be the same object as undefined\n     at referee.fail (Unknown script code:6419:9)\n     at ctx.fail (Unknown script code:6280:17)\n     at assertion (Unknown script code:6288:13)\n     at referee[type][name] (Unknown script code:6315:9)\n     at Anonymous function (Unknown script code:24983:13)\n     at callFn (Unknown script [truncated]",
```

fatso83:
I found more information on this and detailed this out in this comment https://github.com/sinonjs/samsam/pull/41#issuecomment-423980994. It seems like samsam already failed 4 tests when running in IE11 (not known to us) before merging sinonjs/samsam#38. That didn't matter much, as it seems we did not rely on those parts of `samsam`. 

But the change to constructor equality broke 3 additional tests on IE 11, and this _did_ break things in Sinon :crying_cat_face: Would you like to have a go at trying to get them passing, @mgred? I can supply you with Sauce Labs credentials (mail in profile).

fatso83:
I used `mochify --consolify index.html lib/*test.js` to produce a html file with the tests and set up a web server to serve it. Then used Sauce Labs "Live Testing" feature and the `sc` binary to live test in the IE11 debugger. 

bug samsam ie11

Turns out no `stack` or `message` props are created for errors when they are not thrown in IE, meaning two errors are otherwise identical if not thrown.

@mroderick Suggestions for fix? They _are_ deeply equal, but maybe it's ok to assume that all errors are thrown? Then I could simply replace `new Error()` with `createError()` which does a try/catch/return.

mgred:
@fatso83 puh that escalated quickly, Sorry guys. Sure, I will have a look and see what I can find out.

mroderick:
> @mroderick Suggestions for fix? They _are_ deeply equal, but maybe it's ok to assume that all errors are thrown? Then I could simply replace `new Error()` with `createError()` which does a try/catch/return.

Where does `createError()` come from? I can't seem to find it in `sinon` source.

fatso83:
@mroderick I made it up. I guess "with createError() which does a try/catch/return." should be "with **a** sort of `createError()` function **that could** do a try/catch/return".

fatso83:
ps. suggest we continue any discussion in https://github.com/sinonjs/samsam/issues/42

mgred:
@fatso83 let me see what I can find out

PR #1906, opened 2018-09-17
Add context to wrapped function

### LINKED ISSUE #1850: Unsure how / no way to `replace` methods with `fakes`
**Describe the bug**
If I fake a method (a function with a `this` reference) with `sinon.fake`, it looks like its `this` reference will become broken and unbindable.

This makes fakes entirely useless for stubbing methods of OOP style code.

**To Reproduce**
```js
const method = function() { return this.a; }
const fakeMethod = sinon.fake(method);

const o = {
  a: 'asdf',
  method: method,
  fakeMethod: fakeMethod,
  replacedMethod: method,
  explicitlyBoundMethod: method
}

sinon.replace(o, 'replacedMethod', sinon.fake(o.replacedMethod));
sinon.replace(o, 'explicitlyBoundMethod', sinon.fake(o.explicitlyBoundMethod).bind(o)); // i would hope this workaround would be considered unnecessary, even if it did work...

o.method() // 'asdf'
o.fakeMethod() // undefined
o.replacedMethod() // undefined
o.explicitlyBoundMethod() // undefined
```
**Expected behavior**
Setup as above.
```js
o.method() // 'asdf'
o.fakeMethod() // 'asdf'
o.replacedMethod() // 'asdf'
o.explicitlyBoundMethod() // 'asdf'
```

* Library version: 6.0.1
* Environment: nodejs 8 on linux from nodesource

fatso83:
# Summary
- Fix #1850 by keeping this context 
- Added a regression test to verify it.

# How to verify
1. Check out this branch
2. `npm install`
3. `npm test` - see the test added in the first commit pass

- [x] `npm run lint` passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
               wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
               echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' >> /etc/apt/sources.list.d/google.list
               apt-get -qq update
-              apt-get -qq install -y --no-install-recommends google-chrome-unstable
+              apt-get -qq install -y --no-install-recommends google-chrome-stable
             fi
       - run:
           name: Pre-Test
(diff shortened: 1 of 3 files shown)


fatso83:
For some reason crashed in the webworker test on Node 10, but no other versions ... Weird, since it doesn't actually run on Node, but in chromium.

fatso83:
@fearphage This fails on the unstable Chrome version. Works fine on stable Google Chrome 65.0.3325.162 (Linux). Any idea why we are not running on the stable branch? I remember you setting it up, but I don't remember any details. I suspect this being the culprit:

 `npm run test-webworker -- --chrome $(which google-chrome-unstable) --allow-chrome-as-root`

mantoni:
@fatso83 Stable chrome install doesn't succeed in docker containers. Not sure if / when this will be supported in the regular version.

fatso83:
Made it work by installing both stable and unstable. Needed both - unstable for `headless` and stable for WebWorker ...

PR #1915, opened 2018-10-04
fix sinon.assert link in sandbox.assert doc

zsoldosp:
#### Purpose (TL;DR)
the https://sinonjs.org/releases/v6.3.5/sandbox/#sandboxassert section 
`sinon.assert` link leads to 404

in the https://sinonjs.org/releases/v6.3.5/sandbox/#sandboxassert section,
`sinon.assert` is linked to ./assertions, which translates to 
https://sinonjs.org/releases/v6.3.5/sandbox/assertions which gives a 404, as
`.` in the url is interpreted to be under `/sandbox/`, while the assertions documentation
is on the same hierarchy level

1. visit https://sinonjs.org/releases/v6.3.5/sandbox/#sandboxassert
2. click on the `sinon.assert` link
3. verify it is not a 404, but the right documentation page

### DIFF
--- a/docs/_releases/v6.3.5/sandbox.md
+++ b/docs/_releases/v6.3.5/sandbox.md
@@ -124,7 +124,7 @@ sinon.config = {
 
 #### `sandbox.assert();`
 
-A convenience reference for [`sinon.assert`](./assertions)
+A convenience reference for [`sinon.assert`](../assertions)
 
 *Since `sinon@2.0.0`*
 


mroderick:
@zsoldosp thank you for your pull request.

We merged it a bit too quickly, as there are many `sandbox.md` files (documenting the other versions) that also have broken links, and more importantly the [`sandbox.md` in `release-source`](https://github.com/sinonjs/sinon/blob/master/docs/release-source/release/sandbox.md), which is documenting future versions.

Would you be up for fixing the other links with a new pull request?

fatso83:
Sorry - my bad ...

PR #1926, opened 2018-10-11
Update to Lolex 3

fatso83:
Lolex has some nice improvements in major 3 with regards to the Performance API and nanosecond precision.


fatso83:
The changes in Lolex validated a new major version for Lolex, so that implicitly means a new major version for us as well, given that we expose Lolex timer API.

PR #1927, opened 2018-10-12
SEO improvements and banner on old documentation

fatso83:
#### Purpose
 Fixes #1923 by adding banner on pages containing docs for older versions of Sinon.
 The banner notifies that the page is old and points to the newest version.
old-version

 It also improves SEO by obliterating all the hits for old versions of Sinon. Instead,
 using rel=canonical, all those search hits now leads to /releases/latest/<page>.

 #### Background
When searching for "Sinon spy" on Google, the most common hits are for version 1.17 and version 2.0.0. This leads to people reporting errors. We have discussed ways of improving the situation, and the solution of having all the doc pages point to the latest version of the same page using rel=canonical seemed like the most pragmatic way of getting better search hits while still keeping the old docs around for reference.
skjermdump fra 2018-10-11 21-53-30

 #### Solution  - optional
The lack of a clear way of seperating programming logic from the view logic
meant some bits have a slightly hacky feel to them. You might suggest other
ways of doing so :-)

 #### How to verify - mandatory
1. Check out this branch
2. `bundle exec jekyll serve`
3. Browse the [/releases page](http://localhost:4000/releases/)
4. Observe the banner being there on v6.3.4 but not 6.3.5
5. Type http://localhost:4000/releases/latest/ and see that it shows the docs for 6.3.5

PR #1930, opened 2018-10-13
Add Google Analytics

fatso83:
Add analytics to our site to analyze search traffic patterns.

Right now we are a little bit in the dark with regards to how people come to our site, but we do know they sometimes end up at the wrong docs.
Having analytics would enable us to make better decisions for how to improve our docs and also verify that our changes work as intended

Ref #1927 and #1929

With increasing number of infrastructure points and logons we might need some central password store for the Sinon members, like 1Password.
Right now, I added this to my personal google account, as a start. I am not too intimately aware of Google Search Console, but
I am quite certain I can just add the rest of the team as well.

No verification for us - it's a file for Google to verify :-)

### DIFF
--- a/docs/google04dc5213e8f949a9.html
+++ b/docs/google04dc5213e8f949a9.html
@@ -1,0 +1,1 @@
+google-site-verification: google04dc5213e8f949a9.html


mroderick:
Ew. I guess it's necessary, but I am still not a fan of giving more data to Google.

fatso83:
We could use any other tracker. I don't care that much which.

Issue #1898, opened 2018-09-09
Document how we can control Promises


fatso83:
There have been a couple of longstanding issues and pull requests regarding Promises on the lolex issue tracker.

I am assuming we will probably implement https://github.com/sinonjs/lolex/issues/114 in time, for a totally synchronous way to control Promises, but for it to work fully in true ES2017+ environments using asynchronous/await it requires explicit support from the underlying engine.

Before it's implemented we should document for people how they can control Promises today in "normal" cases. We can't cover all cases (as mentioned above), but we could perhaps make a tutorial to show a possible path. I don't care about pretty in this regard, but results, so if it involves overriding Promise.prototype, installing Bluebird or other non-core Sinon libs that's ok. It's just about covering how to achieve a common desire, like the proxyquire article shows.

I am not totally sure how or what, but something ...

fatso83:
I was thinking something along the lines of @mroderick's [link seams article](https://sinonjs.org/how-to/link-seams-commonjs/). So in the typical "tell ’em what you are going to tell ’em;tell ’em; and then tell ’em what you told ’em.” style I'd propose

- a quick intro to what the article is trying to cover
- case 1: problem, solution
- case 2: problem, solution
- summary

I am not totally sure about the cases, though... Might find something in our issue tracker? I am guessing people have trouble with stubbing time when relying on some method that relies on Promises. Some cases might when Promise's have been implemented internally in a library using `setTimeout`, `setImmediate` and similar, which are timers overridden by `lolex`. 

With regards to goals:
A reader should be able to stub out time while still controlling some method that resolves to a promise?

benjamingr:
Ok, do we want to show the examples with a particular library in mind (other than sinon or lolex)? 

For example - in the context of testing async functions in mocha tests?

fatso83:
Mocha is more or less the default, so use that. People use other stuff, but they have always seen that, and if not, it's easy enough to understand.

And I don't mind if we pull in stuff like Bluebird or stub `global.Promise` if it gets the job done for the test. Right now, there is _nothing_, and a working example is better than perfect example. Let's just get _something_ to point people to when they ask questions. Whatever we do, we should strive to make the examples as simple and pedagogical as possible without introducing too many concepts in each case, as that easily confuses more than helps.

benjamingr:
I'd rather focus on writing an async test (for promises) and using fake timers - there are some interesting patterns based on promises waiting for a tick.

[...]

PR #1931, opened 2018-10-13
Add a how-to guide for speeding up time with lolex

orlangure:
This PR adds a how-to guide about current ways to control promises/async functions with lolex (https://github.com/sinonjs/sinon/issues/1898).

In a nutshell, there are a few examples: a regular function, an async function, and a Promise (_walk into a bar_). These things are tested in different ways to demonstrate what we can and cannot control.

1. Check out this branch
2. Follow the steps described [here](https://github.com/sinonjs/sinon/blob/master/docs/CONTRIBUTING.md) to build the documentation website locally (using jekyll)
3. Navigate to http://localhost:4000/ -> "How to" -> "How to speed up time with lolex"


docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+it('should return 42 after one second', async () => {
+    const result = await maker.asyncReturnAfterOneSecond();
+    assert.equal(result, 42); // PASS
+});
+
+// or using Mocha's promises
+it('should return 42 after one second', () => {
+    const promise = maker.asyncReturnAfterOneSecond();
+
+    // this call does not really speed up anything
```
fatso83: This just creates confusion, if it is supposed to show a way of doing the same as above, but without ES2015 syntax. Line 86 and 87 can be deleted.

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+    clock.tick(1000);
+
+    // the Promise was not resolved,
+    // even though we moved the time forward
+    assert.equal(result, 42); // FAIL
+});
+```
+
+The above test fails, since `asyncReturnAfterOneSecond` is an `async` function
+that returns a Promise, and it is currently impossible to control the Promises'
```
fatso83: _the Promise's_ (definite singular) or _promises'_ (indefinite plural)

benjamingr:
Hey, thank you so much for following up on our email discussion - will take a look promptly :)

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
+---
+layout: page
+title: How to speed up time with lolex
```
benjamingr: Should the title be "How to test async functions with fake timers" perhaps?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+---
+
+With lolex, testing code that depends on timers is easier, as it sometimes
```
benjamingr: With lolex -> "With fake timers (lolex)"?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+    // callback is not called immediately
+    assert.ok(!spy.called);
+
+    // but it is called synchronously after the clock is fast forwarded
+    clock.tick(1000);
+    assert.ok(spy.called); // PASS
+});
+```
+
+We could expect similar behavior from a promisified timeout:
```
benjamingr: We could - "one might"?

Also - maybe "One might expect to be able to use a similar approach to test an async function using promises:"?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+const util = require('util');
+const setTimeoutPromise = util.promisify(setTimeout);
+
+module.exports.asyncReturnAfterOneSecond = async () => {
+    await setTimeoutPromise(1000);
+    return 42;
+};
+```
+
+Trying a naive approach:
```
benjamingr: Not sure about "naive", maybe "One may try"?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+// maker.js
+module.exports.fulfillAfterOneSecond = () => {
+    return new Promise(fulfill => {
+        setTimeout(() => fulfill(42), 1000);
+    });
+};
+```
+
+`fulfillAfterOneSecond` resolves to 42 after one second, just like
+`asyncReturnAfterOneSecond`, but it is testable in a synchronous way:
```
fatso83: no, it's not synchronous. the test is still asynchronous, as the `assert` call is done in a microtick after the function exits. The reason `fulfillAfterOneSecond` resolves (almost) immediately using `clock.tick` and `asyncReturnAfterOneSecond` is not affected by `clock.tick` has to do with how `setTimeout` is resolved. In `asyncReturnAfterOneSecond` you bind the original value of `setTimeout` using `util.promisify()`. That means it has a reference to the original `setTimeout` which _will not be affected by anything `lolex` does_. `async` really has nothing to do with it.

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+    const promise = maker.fulfillAfterOneSecond();
+
+    // this call actually makes the resolution quicker
+    clock.tick(1000);
+    return promise.then(result => assert.equal(result, 42)); // PASS
+});
+```
+
+The above test passes immediately, without the 1 second delay. This is because
+we do not try to hijack `then()` call, but use Mocha's native ability to test
```
fatso83: I don't understand this section. You didn't try to hijack the `then` call in earlier tests? Mocha's built-in promise support is just syntactic sugar to avoid passing in a callback and there isn't much `native` stuff going on. I'd replace "native ability to test Promises" with "built-in support for Promises"

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+Promises, while speeding up their resolution.
```
fatso83: Here it seems like it is Mocha that is speeding up their resolution. I think it would be clearer if you changed it to "while using `lolex` to speed up the test".

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+```js
+// test.js
+
+// using async await
+it('should return 42 after one second', async () => {
+    const result = await maker.asyncReturnAfterOneSecond();
+    assert.equal(result, 42); // PASS
+});
+
+// or using Mocha's promises
```
benjamingr: Maybe "Or without using an async function" since both versions are using Mocha's promise return value support?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+    clock.tick(1000);
+    return promise.then(result => assert.equal(result, 42)); // PASS
+});
+```
+
+The above test passes immediately, without the 1 second delay. This is because
+we do not try to hijack `then()` call, but use Mocha's native ability to test
+Promises, while speeding up their resolution.
+
+Although it is currently impossible to speed up `async` functions, or even
```
fatso83: This isn't correct. You can speed up `async` functions. It's just syntactic sugar-coating for promises. Anything you can do with promises is possible with async and vice versa. Unless you can inform me otherwise?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+    const promise = maker.asyncReturnAfterOneSecond();
+
+    // this call does not really speed up anything
+    clock.tick(1000);
+
+    return promise.then(result => assert.equal(result, 42)); // PASS
+});
+```
+
+Although `async` functions cannot be tested synchronously, we can test Promises
```
benjamingr: Add "(yet)" after "cannot"? We're talking about it :)

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+that are resolved in `setTimeout`. Consider the following function, that has the
+same functionality as `asyncReturnAfterOneSecond`:
+
+```js
+// maker.js
+module.exports.fulfillAfterOneSecond = () => {
+    return new Promise(fulfill => {
```
benjamingr: Nit: this is `resolve` and not `fulfill`

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+        setTimeout(() => fulfill(42), 1000);
+    });
+};
+```
+
+`fulfillAfterOneSecond` resolves to 42 after one second, just like
```
benjamingr: Maybe something like: By returning the promise from mocha's test and making it an async test we can still progress the internal timers using lolex's fake timers - so we only have to wait the time it takes the promise to resolve which is almost instant.

Or something of the sort?

docs/_howto/lolex-async-promises.md
benjamingr: "we do not try" "while we do not"?

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,128 @@
...
+`asyncReturnAfterOneSecond`, but it is testable in a synchronous way:
+
+```js
+// test.js
+it('should be fulfilled after one second', () => {
```
benjamingr: Add example with an async function?

Very similar to what you have for a regular promise where you create the promise without immediately `await`ing it and then waiting for it after calling `tick`.

docs/_howto/lolex-async-promises.md
benjamingr: It's not currently impossible - just very very hacky :) I would add here the conclusion maybe:

"Although it is currently impossible to speed up the `then`s async functions run it is still possible to speed up timers and write meaningful sped-up tests with code using async functions." or something of the sort.

Up to you.

benjamingr:
Thank you for working on this!

docs/_howto/lolex-async-promises.md
fatso83: To nitpick further: I _think_ what Benjamin is saying is not that you have to call your callbacks "resolve", but that it's a somewhat misleading name, as a promise is _also_ fulfilled if you reject it :-)

docs/_howto/lolex-async-promises.md
benjamingr: I think there is an assumption `lolex.install()` was called prior to anything - but I think it makes a ton of sense to make it explicit as part of the test (that is, add a `lolex.install()` part above)

docs/_howto/lolex-async-promises.md
benjamingr: Agreed about "built-in support for promises"

docs/_howto/lolex-async-promises.md
benjamingr: It's easier to speed up promises because you can override the promise library to one that uses a scheduler

 Speeding up async functions without invoking the native `%RunMicrotasks` (which requires an addon) or using a `require('binding')` with an undocumented internal API requires transpiling your async functions to coroutines.

fatso83: True, but I didn't mean overriding the scheduler for the Promises, but rather the statement that implied that one cannot speedup `async` tests that use timers. It seemed as if the statement said that if you have an `async` test using `setTimeout` (for instance), then it would not be affected by `clock.tick` calls, which would be incorrect.

Just the fact that we are arguing the meaning might be a hint that this sentence should be rephrased somehow :-)

benjamingr: Agreed

.gitignore
```diff
@@ -5,3 +5,4 @@ coverage/
 .idea
 .sass_cache
 _site
+*.swp
```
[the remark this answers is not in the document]
benjamingr: I also think it makes sense to either allow these or remove them all but no strong feelings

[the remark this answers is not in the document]
fatso83: No strong feelings either. If `.swp` doesn't belong here, neither does `.idea`.

orlangure:
Thanks for your comments, I'm in the middle of fixing the issues you guys raised.
Will post a second draft sometime soon.

benjamingr:
@orlangure thanks, we're looking forward to it :)

.gitignore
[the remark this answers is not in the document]
fearphage: I didn't review the initial commits and I have no idea what `.idea` is for in the first place. I think it makes sense for the files in a repo to be specific to the repo instead of specific to the people that edited the repo. If `.idea` is specific to some editor and/or OS, then I think it should be removed as well.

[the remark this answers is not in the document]
fatso83: .idea -> [IntelliJ IDEA](https://www.jetbrains.com/idea/)

docs/_howto/lolex-async-promises.md
```diff
@@ -0,0 +1,108 @@
...
+    clock.tick(1000);
+    assert.ok(spy.called); // PASS
+});
+```
+
+The same approach can be used to test an `async` function:
+
+```js
+module.exports.asyncReturnAfterOneSecond = async () => {
+    // util.promisify is not used deliberately
```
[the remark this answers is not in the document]
benjamingr: > When it is used though, the test fails.

 Make it a proxy to `util.promisify` or call clock.install() before `util.promisify` is called (because then it holds a reference to the original setTimeout):

```js
module.exports. asyncReturnAfterOneSecond = async () => {
  const setTimeoutPromise = util.promisify(setTimeout); //
  // ...
}

```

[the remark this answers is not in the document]
orlangure: Nope, even with `lolex.install` call before `util.promisify`, tests of this function fail.

What did you mean by proxying somehting to `util.promisify`? I'll check it as well.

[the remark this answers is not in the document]
benjamingr: I think this needs to be fixed in lolex. I'll make an issue.

orlangure:
Can this guide be reviewed again before the [issue](https://github.com/sinonjs/lolex/issues/223) is fixed? I guess most of it does not depend on that bug in lolex.

orlangure:
Still waiting for https://github.com/sinonjs/lolex/pull/227 to be reviewed and merged. Bumping for now.

fatso83:
@benjamingr Any chance of looking at the PR blocking getting this in? Yuri has kind of been waiting since October.

docs/_howto/lolex-async-promises.md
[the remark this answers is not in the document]
fatso83: For reference, the issue is sinonjs/lolex#227

fatso83:
```suggestion
    // Using util.promisify would look nicer, but there is a lolex issue blocking this at the moment: https://github.com/sinonjs/lolex/pull/227
```

orlangure:
@fatso83 Done, thanks for pushing this forward.

fatso83:
Thank you! This should go live at https://sinonjs.org/how-to in a few minutes.

Issue #1852, opened 2018-07-05
"TypeError: Cannot read property 'now' of undefined" when sinon.useFakeTimers()


moraveyo:
**Describe the bug**
Using sinon@**6.1.0** when I do `sinon.useFakeTimers()` - I got an error:
```
TypeError: Cannot read property 'now' of undefined
      at mirrorDateProperties (dist/webpack:/node_modules/sinon/pkg/sinon-esm.js:3644:1)
      at hijackMethod (dist/webpack:/node_modules/sinon/pkg/sinon-esm.js:3947:1)
      at Object.install (dist/webpack:/node_modules/sinon/pkg/sinon-esm.js:4337:1)
      at useFakeTimers (dist/webpack:/node_modules/sinon/pkg/sinon-esm.js:4385:1)
      at Sandbox.useFakeTimers (dist/webpack:/node_modules/sinon/pkg/sinon-esm.js:15390:1)
      at Context.<anonymous> (dist/webpack:/tests/unit/Countdown.spec.js:23:1)
```
When I downgrade to sinon@**5** - everything is allright!

My repo:
https://github.com/moraveyo/chess-clock/blob/bf764f3ed9026366de9375bbc870109e4f007099/tests/unit/Countdown.spec.js#L23

**Context (please complete the following information):**

* Library version: 6.1.0
* Environment: Ubuntu 16.04
* Example URL:
* Other libraries you are using:
  - "vue": "^2.5.16",
  -  "vue-router": "^3.0.1"
  -  "@vue/cli-plugin-babel": "^3.0.0-beta.15",
  -  "@vue/cli-plugin-eslint": "^3.0.0-beta.15",
  -  "@vue/cli-plugin-unit-mocha": "^3.0.0-beta.15",
  -  "@vue/cli-service": "^3.0.0-beta.15",
  -  "@vue/eslint-config-airbnb": "^3.0.0-rc.3",
  -  "@vue/test-utils": "^1.0.0-beta.16",
  -  "chai": "^4.1.2",
  -  "vue-template-compiler": "^2.5.16"
  -  Mocha 5.2.0
  -  Webpack 4.15.1

iraklisg:
Same problem with `sinon@6.0.3` :disappointed: 
`sinon@5.1.1` works as expected

rgroothuijsen:
[...]

PR #1935, opened 2018-10-22
Issue #1852: Add a way to pass a global context to lolex when calling useFakeTimers

LouisBrunner:
#### Purpose (TL;DR)

Partial fix of issue #1852: Add a way to pass a global context to lolex when calling useFakeTimers

 #### Background

As discussed in #1852, some environments make `sinon.useFakeTimers` crash, because of `lolex`, when some properties are missing in the `global` object. The issue has been improved by triggering a more explicit error in `lolex`, however this PR adds a way to provide the global context directly and forward it to `lolex`.

#### Solution 

This PR adds an extra property when using `sinon.useFakeTimers` with an object argument: by giving the `global` property, `lolex.withGlobal(global)` will be called and used to install the clock instead of using the default implementation.

 #### How to verify
Checkout [this repositoy](https://github.com/LouisBrunner/sinon-issue-1852) for an example of an environment that triggers this bug.

### DIFF
--- a/lib/sinon/util/fake-timers.js
+++ b/lib/sinon/util/fake-timers.js
@@ -3,8 +3,12 @@
 var extend = require("./core/extend");
 var llx = require("lolex");
 
-function createClock(config) {
-    var clock = llx.install(config);
+function createClock(config, globalCtx) {
+    var llxCtx = llx;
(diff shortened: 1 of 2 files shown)


mroderick:
This has been published as `sinon@7.1.0`. Thank you :)

Issue #1936, opened 2018-10-25
Make spy properties non-enumerable


ehmicky:
**Is your feature request related to a problem? Please describe.**
When a Sinon spy assertions fails during unit testing, some assertion libraries like `power-assert` (used by Ava) will print all the spy properties (including the spy methods) which is very verbose and makes it harder to see why the test failed. 

**Describe the solution you'd like**
Make Sinon-specific spy properties non-enumerable.

**Describe alternatives you've considered**
  - customizing the assertion library `power-assert` but it's hard to do from Ava
  - customizing Ava but it's not possible because it directly relies on `power-assert` for this case
  - overriding `spy[Symbol.toStringTag]` or `spy.toJSON()` but this does not work since `power-assert` does not use either. Instead it iterates over enumerable properties.

**Additional context**
The feature request is about spies, but it most likely can be related to other Sinon objects, such as stubs and fakes.

Here's a screenshot of how verbose a single test Sinon assertion can become with Ava:

![enumerable_properties](https://user-images.githubusercontent.com/8136211/47487808-6bee7c00-d843-11e8-9f2a-ed7b1ef798ab.png)

fatso83:
How would this affect REPL usage and API discoverability? One of my favorite ways of using new tech is through exploratory testing on  the REPL. Doing `var spy = sinon.spy(); spy.` and then <tab>ing my way through the various methods. Would this still work?

ehmicky:
Yes, non-enumerable properties are still shown on tab autocomplete in Node.js REPL.

mroderick:
This has been implemented with #1936 and released as `sinon@7.1.1`

ehmicky:
Thanks for this.

However I do not think this worked:

```js
> sinon.spy()
{ [Function: proxy]
  isSinonProxy: true,
  called: false,
  notCalled: true,
  calledOnce: false,
  calledTwice: false,
  calledThrice: false,
  callCount: 0,
  firstCall: null,
  secondCall: null,
  thirdCall: null,
  lastCall: null,
  args: [],
  returnValues: [],
  thisValues: [],
  exceptions: [],
  callIds: [],
  errorsWithCallStack: [],
  displayName: 'spy',
  toString: [Function: toString],
  instantiateFake: [Function: create],
  id: 'spy#0' }

> Object.keys(sinon.spy())
[ 'isSinonProxy',
  'called',
  'notCalled',
  'calledOnce',
  'calledTwice',
  'calledThrice',
[...]

PR #1941, opened 2018-10-28
Issue #1936: Make the spy functions non enumerable so that printing it is more concise

LouisBrunner:
#### Purpose

Improve issue #1936: Make the spy functions non enumerable so that printing it is more concise

 #### Background

When printing a spy (which is often done by assertion libraries), the output takes a long chunk of space and make debugging really difficult. This PR removes the internal functions from this output, making the list smaller and more useful.

 #### Solution

This PR expands the `extend` function to support extending using `Object.defineProperty` with `enumerable: false`. As `configurable` and `writable` are set to `true`, the rest of the codebase will still work as intended. Note that as indicated by @ehmicky in the original issue, the node REPL is not affected.

 #### How to verify

screen shot 2018-10-28 at 18 38 51

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -217,9 +217,9 @@ var spyApi = {
         var length = spyLength || funk.length;
         var proxy = createProxy(funk, length);
 
-        extend(proxy, spy);
+        extend.nonEnum(proxy, spy);
         delete proxy.create;
-        extend(proxy, funk);
+        extend.nonEnum(proxy, funk);
 
         proxy.resetHistory();
         proxy.prototype = funk.prototype;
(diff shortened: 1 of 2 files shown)


LouisBrunner:
I am not too familiar with JSDoc, are the comments fine like this?

I tested a bit and `eslint`'s `valid-jsdoc` seems to only check comments that start with `/**`.

mroderick:
> I am not too familiar with JSDoc, are the comments fine like this?

If ESLint doesn't complain, then I guess they're good :)

I think that JSDoc comments are supposed to start with `/**`,  which is why it didn't complain about the previous comments.

Thanks!

mroderick:
This has been published as `sinon@7.1.1`

Issue #1919, opened 2018-10-08
Remove duplication: create async methods


mroderick:
While working on the codebase tonight, I discovered some unnecessary duplication:

https://github.com/sinonjs/sinon/blob/master/lib/sinon/default-behaviors.js#L266-L272

```js
// create asynchronous versions of callsArg* and yields* methods
forEach(Object.keys(module.exports), function (method) {
    // need to avoid creating anotherasync versions of the newly added async methods
    if (method.match(/^(callsArg|yields)/) && !method.match(/Async/)) {
        module.exports[method + "Async"] = createAsyncVersion(method);
    }
});
```

and

https://github.com/sinonjs/sinon/blob/master/lib/sinon/behavior.js#L216-L222

```js
// create asynchronous versions of callsArg* and yields* methods
forEach(Object.keys(proto), function (method) {
    // need to avoid creating anotherasync versions of the newly added async methods
    if (method.match(/^(callsArg|yields)/) && !method.match(/Async/)) {
        proto[method + "Async"] = createAsyncVersion(method);
    }
});
```

We should refactor the code to only have one copy of this, and re-use it.

LEQADA:
@mroderick I'd like to pick up this task. Any suggestions on where to put a new util function?

mroderick:
@LEQADA that would be great 👍 

I think you could create a file in [`lib/sinon/util/core`](https://github.com/sinonjs/sinon/tree/master/lib/sinon/util/core)

morgsmccauley:
@LEQADA did you make any progress on this? I'd like to pick this up if it's not being worked on 😄

LEQADA:
@morgsmccauley sorry for a long inactivity. I already started to work on the issue but didn't have time to write tests. I plan to finish it next week if it's not urgent. If it's urgent then sure take it over. 
Otherwise I'd ask to give me some more days. :slightly_smiling_face:

mroderick:
There are plenty [more issues across the sinonjs repositories with "Help wanted" label](https://github.com/search?q=org%3Asinonjs+is%3Aopen+label%3A%22Help+wanted%22&unscoped_q=is%3Aopen+label%3A%22Help+wanted%22).

mroderick:
LIke this one to find a solution for an unmaintained dependency: https://github.com/sinonjs/nise/issues/66

morgsmccauley:
Great, thanks @mroderick!

mroderick:
Fixed with #1942

PR #1942, opened 2018-10-28
Fix issue #1919 by adding a new util function to prevent duplicate code

khamiltonuk:
#### Purpose
Fix issue #1919 by adding a new util function to prevent duplicate code`

We have two file containing behaviours, some behaviours are exports in async form. 
[https://github.com/sinonjs/sinon/issues/1919](https://github.com/sinonjs/sinon/issues/1919)

 #### Solution  - optional
- Remove duplicate code, make a new util function in core
- Cover the function with a test to ensure the new functions are renamed correctly and the correct and already aync functions are not converted 

 #### Questions 
- Is it ok to use lodash reduce?
- Why has the version bumped up? i think it should only be a patch update
- Any ideas for more tests

1. test pass

 #### Checklist for author
- [x] `npm run lint` passes

### DIFF
--- a/test/util/core/export-async-behaviors-test.js
+++ b/test/util/core/export-async-behaviors-test.js
@@ -1,0 +1,35 @@
+"use strict";
+
+var referee = require("@sinonjs/referee");
+var assert = referee.assert;
+var exportAsyncBehaviors = require("../../../lib/sinon/util/core/export-async-behaviors");
+
+describe("util/core/exportAsyncBehaviors", function() {
+    describe("for methods with names starting with 'callsArg' or 'yields'", function() {
+        it("should create an async version", function() {
+            var methods = {
+                yieldsOn: function yieldsOn() {
+                    return "2";
+                },
+                callsArg: function callsArg() {
+                    return "3";
+                }
+            };
+            assert.equals(Object.keys(exportAsyncBehaviors(methods)), ["yieldsOnAsync", "callsArgAsync"]);
+        });
+    });
+
+    describe("for methods with names not starting with 'callsArg' or 'yields'", function() {
+        it("should not add any new methods", function() {
+            var methods = {
+                callsFake: function callsFake() {
+                    return "1";
+                },
+                resolvesThisAsync: function resolvesThisAsync() {
+                    return "4";
+                }
+            };
+            assert.equals(Object.keys(exportAsyncBehaviors(methods)), []);
+        });
+    });
+});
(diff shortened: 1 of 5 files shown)


mroderick:
> Is it ok to use lodash reduce?

It is if it is somehow superior to using [`Array.prototype.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) and [`Object.keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/Keys). Is it?

mroderick:
> Why has the version bumped up? i think it should only be a patch update

I think if your rebase on to (unto?) `master` that will resolve itself

test/util/core/export-async-behaviors-test.js
```diff
@@ -0,0 +1,27 @@
...
+    callsArg: function callsArg() {
+        return "3";
+    },
+    resolvesThisAsync: function resolvesThisAsync() {
+        return "4";
+    }
+};
+
+describe("util/core/exportAsyncBehaviors", function() {
+    it("should rename functions to Async names and should avoid creating another async versions of the newly added async methods", function() {
```
mroderick: I think it would be good to show both the positive and negative with tests.

```js
context("for methods with names starting with 'callsArg' or 'yields'", function(){
    it("should create an async version", function(){
    });
});

context("for methods with names not starting with 'callsArg' or 'yields'", function(){
    it("should not add any new methods", function(){
    });
});
```

khamiltonuk: 👍 Thanks for the idea showing both give clearer intention

khamiltonuk:
> > Is it ok to use lodash reduce?
> 
> It is if it is somehow superior to using [`Array.prototype.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) and [`Object.keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/Keys). Is it?
Not superior, managed to do it without now

khamiltonuk:
> > Why has the version bumped up? i think it should only be a patch update
> 
> I think if your rebase on to (unto?) `master` that will resolve itself

Perfect worked a charm

mroderick:
Lovely!

Would you mind removing the changes to `package-lock.json`, now that you're not making changes to dependencies?

mroderick:
Thank you 🎸

PR #1947, opened 2018-11-17
Upgrade ~browserify~ cached-path-relative

mroderick:
This also upgrades cached-path-relative, which has a security warning.

See https://nvd.nist.gov/vuln/detail/CVE-2018-16472

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -814,9 +814,9 @@
       }
     },
     "cached-path-relative": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/cached-path-relative/-/cached-path-relative-1.0.1.tgz",
-      "integrity": "sha1-0JxLUoAKpMB44t2BqGmqyQ0uVOc=",
+      "version": "1.0.2",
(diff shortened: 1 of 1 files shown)


mantoni:
Interestingly, this causes a browser test to fail. 🤔

mroderick:
That IS interesting!

mroderick:
If I uninstall `mochify` and `npm link mochify` to the branch of https://github.com/mantoni/mochify.js/pull/184, then I am seeing the same error.

```
```

I will try to isolate the problem further

mroderick:
I've changed the PR to only upgrade `cached-path-relative`, so we can be rid of the security warning. Upgrading dependencies can be done later

mantoni:
While working on #1950, I noticed the same issue when I freshly installed all dependencies. However, I was able to upgrade `cached-path-relative` there without the issue happening.

However, I noticed that lolex is in the tree twice now. There is `lolex@3` as a direct Sinon dependency and `lolex@2` via Nise. Maybe we should upgrade Nise to align the Lolex versions.

Issue #1948, opened 2018-11-17
Remove security warning from old lodash


mroderick:
When running `npm audit`, there is a warning about `lodash`.

1. Run `npm audit`
1. Observe the warning about lodash (example below)

```
┌───────────────┬────────────────────────────────────────┐
│ Low           │ Prototype Pollution                    │
├───────────────┼────────────────────────────────────────┤
│ Package       │ lodash                                 │
├───────────────┼────────────────────────────────────────┤
│ Patched in    │ >=4.17.5                               │
├───────────────┼────────────────────────────────────────┤
│ Dependency of │ mochify-istanbul [dev]                 │
├───────────────┼────────────────────────────────────────┤
│ Path          │ mochify-istanbul > lodash              │
├───────────────┼────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/577 │
└───────────────┴────────────────────────────────────────┘
```

The old `lodash` that has the issue  is used by `mochify-istanbul`, which is part of the `test-coverage`task.

I've tried [upgrading `lodash` in `mochify-istanbul`](https://github.com/ferlores/mochify-istanbul), but that was not successful. To me that project looks pretty dead.

We should consider other solutions for coverage. I have a couple of ideas:

1. use `nyc` to instrument the files for mochify (this might take some work because of `proxyquire-universal`)
1. use `jsdom` to fill in the missing apis, so all tests can be run in node ... and then use `nyc` to instrument files for mocha

Other ideas welcome

mantoni:
The project isn't dead. It worked just fine all the time. I'm happy to resurrect it.

mantoni:
Ok, I had a quick look at `mochify-istanbul`, and although it works, the tests don't and the whole project is in a terrible state. The owner seems to have abandoned GitHub entirely.

The way forward seems indeed to use `nyc`. There is a babel plugin for instrumentation that could be used. I have good experience with using `babelify` to get any babel plugin to work with the browserify pipeline. @mroderick You know `nyc` better than me. Is there an easier way?

mroderick:
I did some experiments yesterday, and couldn't get `nyc` working with `mochify` in the first try.

I don't know if there's an easy way out of this. Option 1 requires studying how all the pipelining works, option 2 requires changing some of the tests. I do see option two to have the added benefit of being able to easily use the debugger with node for those tests also.

Any `nyc` experts out there who can offer advice?

mantoni:
Fixed by #1950.

[...]

PR #1950, opened 2018-11-19
Use nyc for coverage, update mochify and browserify

mantoni:
#### Purpose (TL;DR)

As discussed in #1948, we want to replace mochify-istanbul since it's basically unmaintained and has outdated dependencies.

 #### Solution

This change upgrades Mochify to v6 which comes with `nyc` support, generating the coverage report with up-to-date istanbul instrumenter libraries.

More dependency updates:

- Mochify 6 uses Browserify 16 now, so the dependency was also updated here.
- Mochify 6 uses Puppeteer 1.10 now,, cause two Chromium downloads here, so this was also aligned.
- The obsolete mochify-istanbul dependency was removed.

1. Check out this branch
2. `npm install`
3. `npm run test-coverage`

 #### Checklist for author

- [x] `npm run lint` passes
- [x] `npm test` passes – ~~__NOTE:__ This currently fails due to `npm run test-webworker` failing.~~


package.json
```diff
@@ -17,7 +17,7 @@
     "test-node": "mocha --recursive -R dot \"test/**/*-test.js\"",
     "test-dev": "npm run test-node -- --watch -R min",
     "test-headless": "mochify --recursive -R dot --grep WebWorker --invert --plugin [ proxyquire-universal ] \"test/**/*-test.js\"",
-    "test-coverage": "mochify --recursive -R dot --grep WebWorker --invert --plugin [ proxyquire-universal ] --plugin [ mochify-istanbul --exclude '**/test/**' --report text --report lcovonly --dir ./coverage ] \"test/**/*-test.js\"",
+    "test-coverage": "nyc npm run test-headless -- --transform [ babelify --ignore [ test ] --plugins [ babel-plugin-istanbul ] ]",
```
mroderick: At least the line is shorter :P

package.json
```diff
@@ -58,8 +66,11 @@
     "type-detect": "^4.0.8"
   },
   "devDependencies": {
+    "@babel/core": "^7.1.6",
```
mroderick: Ugh! How long do you think it'll be before people start suggesting that we rewrite the entire codebase in stage-0 ES?

mantoni: We've been fighting it successfully so far 😉

mroderick: "But you're already loading babel..."

mantoni: It's actually not such a bad idea to verify that sinon works fine when babelified. However, the one strong argument that we can always pull is that Sinon works very low level and we need complete control over the code that runs. Therefore, the codebase is written soley with language features that are supported by the targeted runtimes natively.

mroderick:
> So I double checked that the webworker is able to load the file `pgk/sinon.js`. There seems to be an error with the script execution within the webworker. Unfortunately, we don't see that error. I'd be very thankful for pointers on how to debug this.

Perhaps @fatso83 can give a pointer, since they spent some time in this part of the codebase (I think 🤔)

mantoni:
Final touches:

- Upgraded Mocha to v5.2, so now Node and Browser tests use the same Mocha version everywhere.
- I had to remove the commit that added a try-catch in the webworker code. The linter said "no".
- The last patch (v6.0.2) in Mochify adds Content-Type headers when serving files with the built-in https server. This fixes the webworker test case.

Build is green again. Coverage even went up a little. This is good to merge from my point of view.

mantoni:
This fixes https://github.com/sinonjs/sinon/issues/1948

package.json
mroderick: If it becomes a problem to hold off the stage-0 hordes, I'll take a hammer to the problem

mroderick:
👍

mantoni:
Thanks!

PR #1959, opened 2018-12-11
#1957 Fix passing null as an override

### LINKED ISSUE #1957: createStubInstance override and null return values
**Describe the bug**
`sinon.createStubInstance` was extended recently to allow overrides such as:
```
sinon.createStubInstance(AClass, {
  aMethod: "a string",
  aChainableMethod: sinon.stub().returnsThis()
});
```
Theres a small issue which prevents a user from returning null values

**To Reproduce**
```
sinon.createStubInstance(AClass, {
  aMethod: null
});
```

**Expected behavior**
The library should correctly create a stub instance with a stub method returning null

**Reason**
Theres an initial check whether the override is a stub by checking if `createStubInstance` property exists on the return value https://github.com/sinonjs/sinon/blob/master/lib/sinon/stub.js#L85

This throws when we try to check a property of null

* Library version: Current latest v7.x
* Environment: OSX

Thanks!
Let me know if it makes sense, I can help create a PR for it

spicalous:
#### Purpose (TL;DR)
Fix passing null as an override when invoking `sinon.createStubInstance(Class, <overrides>)`

Fixes #1957 

 #### How to verify
Unit test added and can be verified with `npm test`

 #### Checklist for author

- [x] `npm run lint` passes
- [ ] ~~References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).~~ N/A

### DIFF
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -2989,6 +2989,21 @@ describe("stub", function() {
             assert.equals(3, stub.method());
         });
 
+        it("allows providing null as a return value", function() {
+            var Class = function() {
+                return;
+            };
+            Class.prototype.method = function() {
+                return;
+            };
+
+            var stub = createStubInstance(Class, {
+                method: null
+            });
+
+            assert.equals(null, stub.method());
+        });
+
         it("throws an exception when trying to override non-existing property", function() {
             var Class = function() {
                 return;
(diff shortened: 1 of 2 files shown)


test/stub-test.js
```diff
@@ -2989,6 +2989,21 @@ describe("stub", function() {
             assert.equals(3, stub.method());
         });
 
+        it("allows providing null as a return value", function() {
+            var Class = function() {
```
mroderick: I am surprised that you're able to use `Class` as a variable name!
It doesn't look right to me ... would you mind renaming it to something that doesn't use a reserved word? ... like `Klass`?

spicalous: Do you mind if I rename it to `TestClass` for all cases in this file?

mroderick:
This has been released as `sinon@7.2.1`

PR #1961, opened 2018-12-12
Fix mock.withArgs using matchers

### LINKED ISSUE #1960: custom matchers value is undefined when using .withArgs(...)
**Describe the bug**
It appears that using custom matchers in combination with .withArgs(...) has broken in release 7.2.0.  It works in release 7.1.1 

**To Reproduce**
Here's a simple example of the usage.  With release 7.2.0, the `value` in the sinon.matcher function is undefined. 
```
class MyClient {
  doTheStuff(myRequest: MyRequest) {
    console.log('you should not see this', myRequest);
  }
}

interface MyRequest {
  theString: string;
  theBool: boolean;
}

const client = new MyClient();

const methodToTest = (option1: boolean, option2: boolean) => {
  let myreq: MyRequest;
  if (option1 && option2) {
    myreq = {
      theString: 'success',
      theBool: true
    };
  } else {
    myreq = {
      theString: 'failure',
      theBool: false
    };
  }
  client.doTheStuff(myreq);
};

describe('methodToTest', () => {
  let mockClient;
  beforeEach(() => {
    mockClient = sinon.mock(client);
  });

  afterEach(() => {
    mockClient.restore();
  });

  it('when both options true, call client with success request', () => {
    mockClient.expects('doTheStuff').once().withArgs(
      sinon.match((value) => {
        console.log('matcherValue:', value);
        return value.theString === 'success' && value.theBool === true;
      })
    );
    methodToTest(true, true);
  });

  it('otherwise, call with failure request', () => {
    mockClient.expects('doTheStuff').once().withArgs(
      sinon.match((value) => {
        console.log('matcherValue:', value);
        return value.theString === 'failure' && value.theBool === false;
      })
    );
    methodToTest(false, true);
  });
});
```

**Expected behavior**
I expect that when the mock is called in the code, it will pass the arg to the sinon matcher to evaluate whether the arg is correct.

**Screenshots**
Test Run with Release 7.2.0 (FAILS)
screen shot 2018-12-11 at 11 13 23 am

Test Run with Release 7.1.1 (SUCCEEDS)
screen shot 2018-12-11 at 11 13 11 am

* Library version: _7.2.0_ (loading using yarn/npm)
* Environment: nodejs8.10
* Example URL: none
* Other libraries you are using: Mocha, Typescript

mantoni:

Fixes #1960

The `deepEqual` semantics in `@sinonjs/samsam` have changed in v3 to be `(actual, expected)` instead of `(expected, actual)`. There was a call left unchanged since no test case existed for this case.

1. Check out this branch
2. `npm install`
3. `npm t`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/mock.js
+++ b/lib/sinon/mock.js
@@ -37,7 +37,7 @@ function arrayEquals(arr1, arr2, compareLength) {
     }
 
     return every(arr1, function(element, i) {
-        return deepEqual(element, arr2[i]);
+        return deepEqual(arr2[i], element);
     });
 }
 
--- a/test/mock-test.js
+++ b/test/mock-test.js
@@ -653,6 +653,20 @@ describe("sinonMock", function() {
                     { name: "ExpectationError" }
                 );
             });
+
+            it("should not throw when expectation withArgs using matcher", function() {
+                var obj = {
+                    foobar: function() {
+                        return;
+                    }
+                };
+                var mock = sinonMock(obj);
+                mock.expects("foobar").withArgs(match.string);
+
+                refute.exception(function() {
+                    obj.foobar("arg1");
+                });
+            });
         });
 
         describe(".withExactArgs", function() {


dueyfx:
@mantoni Thanks for the quick fix!

PR #1968, opened 2019-01-10
Fix stubbing function objects

### LINKED ISSUE #1967: `sinon.stub(obj)` should work even if obj happens to be a function
I'm not sure if this is a bug or a feature request; see below.

I have a class and I want to stub _all and only its **static** methods_. So, I want to treat the class's constructor function like a plain object, where I'm stubbing all the function-valued properties on it and its prototype (as those properties are the static methods).

More generally, I don't see why: 

```
function xyz() { return true; }
xyz.abc = function() { return false; }
xyz.__proto__.def = function() { return false; }

sinon.stub(xyz);
```

Shouldn't produce an object with `abc` and `def` replaced with stub methods. 

The `sinon.stub(obj)` seems to say that it will stub all the function-valued properties on the object. But functions are objects too, and it doesn't stub their function-valued properties. So maybe this is actually a bug and not a feature request.

None; not sure I have enough context to consider alternatives. I'm just telling you what I expected, which didn't seem to work.

mantoni:

Currently `sinon.stub(function () {})` throws `TypeError: Attempted to wrap undefined property undefined as function`. Sinon should either throw a meaningful error explaining why a `function` is not a valid argument, or treat functions like any other object. I can't think of a reason why we shouldn't support this.

Fixes #1967

1. Check out this branch
2. `npm install`
3. `npm t`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -36,10 +36,11 @@ function stub(object, property) {
     }
 
     var actualDescriptor = getPropertyDescriptor(object, property);
-    var isStubbingEntireObject = typeof property === "undefined" && typeof object === "object";
+    var isObject = typeof object === "object" || typeof object === "function";
+    var isStubbingEntireObject = typeof property === "undefined" && isObject;
     var isCreatingNewStub = !object && typeof property === "undefined";
     var isStubbingNonFuncProperty =
-        (typeof object === "object" || typeof object === "function") &&
+        isObject &&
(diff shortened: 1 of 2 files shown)


mantoni:
I don't know what's wrong with the CircleCI build and won't have time to look into it. Everything works fine on Travis.

mantoni:
Released in `v7.2.3`.

PR #1874, opened 2018-08-01
Add example of default sandbox usage

giles-v:
#### Purpose

In reading [the docs for Sandboxes](https://sinonjs.org/releases/v6.1.4/sandbox/), the page states:

> the sinon object is a default sandbox. Unless you have a very advanced setup or need a special configuration, you probably want to just use that one.

However, no example of doing this is given, so the recommended basic usage behavior here feels like it is relegated to a footnote, easily missed. If this is the most common usage, I strongly think there should be an example of that _first_, before proceeding on to the API for creating other sandboxes.

#### How to verify
1. Check out this branch
2. Navigate to `/docs` and run the docs server
3. Inspect the Sandbox API page here: http://localhost:4000/releases/v6.1.4/sandbox/

#### Checklist for author

- [x] `npm run lint` passes
- [x] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

Issue #1969, opened 2019-01-19
CircleCI tests on Node 10 is failing


mantoni:
**Describe the bug**
The build is failing on CircleCI again with Mochify printing `[object Object]` and then failing. The same build works fine on Travis.

**To Reproduce**
Trigger a build on CircleCI. It failed 3 times now, so it seems to be consistent.

If anyone has a good idea how to reproduce locally or otherwise debug, I'd love to hear it.

PR #1973, opened 2019-01-25
docs(sandbox): add example for default sandbox

mgred:
This finishes #1874 to document the example for all releases

1. Check out this branch
2. Run the docs server
3. Inspect the Sandbox API page

 #### Checklist for author

- [x] `npm run lint` passes


mgred:
@mroderick I picked up #1874 here. Since this is only docs and the node 10 builds are failing in general for now #1969. Do you think we can merge this?

mroderick:
> @mroderick I picked up #1874 here. Since this is only docs and the node 10 builds are failing in general for now #1969. Do you think we can merge this?

Yeah, go ahead

PR #1975, opened 2019-02-07
Make all properties non-enumerable in spies, stubs, mocks and fakes

ehmicky:
Fix issue #1936 by making all properties non-enumerable in spies, stubs, mocks and fakes.

1. Check out this branch
2. `npm install`
3. `sinon.spy()`, `sinon.stub()`, `sinon.fake()`, `sinon.mock()` should not return non-enumerable properties.

Very few properties had to be left because removing them was breaking the tests.

Also this is a breaking change (major version) because some users might rely on spy/stub/fake/mock properties being enumerable.

CircleCI on Node 10 fails but this does not seem related to this PR, as it failed on other commits/PRs.

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -227,4 +227,4 @@ proto.createBehavior = createBehavior;
 
 var asyncBehaviors = exportAsyncBehaviors(proto);
 
-module.exports = extend({}, proto, asyncBehaviors);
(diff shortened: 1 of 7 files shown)


mroderick:
Thank you ❤️

PR #1987, opened 2019-03-01
Retain spy function names and fix spy.named(name)

### LINKED ISSUE #250: Named spies
It would help with readability of assertion error messages if spies (other than the object property ones) could be created with names. For example, Jasmine spies allow this.

mantoni:

- Let spies have the same `name` as the function they wrap
- Change the `name` property in `spy.named(name)`, not just the `displayName` (used for sinon error messages)

Fixes #250 

We used to retain the name of spied functions, but this has been lost along the way at some point.

There is https://github.com/sinonjs/sinon/issues/950 which shows in the issue description that the `name` property of a spy reflected the original function name. However, the [test that is verifying this](https://github.com/sinonjs/sinon/pull/1435) was added later and it hard coded the name to `"proxy"`.

I think this dates back to the old times when we used `eval` to generate the spy function.

#### Solution  - optional

This change verifies that a `name` can be configured on the function – unfortunately this is inconsistent across (old but supported) browser versions – and then configures the name using the same property descritor flags as the original value.

1. Check out this branch
2. `npm install`
3. `npm test`
4. `npm run test-cloud`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -152,6 +152,12 @@ function createProxy(func, proxyLength) {
             return p.invoke(func, this, slice(arguments));
         };
     }
+    var nameDescriptor = Object.getOwnPropertyDescriptor(func, "name");
+    if (nameDescriptor && nameDescriptor.configurable) {
+        // IE 11 functions don't have a name.
+        // Safari 9 has names that are not configurable.
+        Object.defineProperty(p, "name", nameDescriptor);
+    }
     extend.nonEnum(p, {
         isSinonProxy: true,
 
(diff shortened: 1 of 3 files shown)


lib/sinon/spy.js
```diff
@@ -152,6 +152,12 @@ function createProxy(func, proxyLength) {
             return p.invoke(func, this, slice(arguments));
         };
     }
+    var nameDescriptor = Object.getOwnPropertyDescriptor(func, "name");
```
mroderick: The code in this change looks very similar to the code in the next change. Would it be wise to extract it to a shared function? Perhaps even put that function in `@sinonjs/commons`?

mantoni: Yes, I had the same thought, but in one case we're just using one descriptor as is, in the other case we modfy the `value`. Extracting a function produces more code and would include a superfluous `value` assignment for one of the cases. So I decided against it for clarity.

I am treating `commons` as a toolbelt for shared functions. This is a Sinon only feature so far. We can move it if we want to reuse it elsewhere.

mantoni:
Released in `v7.2.7`.

PR #1988, opened 2019-03-01
Set `fake.lastArg` to last argument regardless of type

### LINKED ISSUE #1986: `lastArg` is set to undefined when last argument is `false`
**Describe the bug**
When using fakes, `lastArg` is set to `undefined` whenever the last argument passed is `false`.

The primary issue is in `fake.js#wrapFunc` where the assignment is using `&&` -
```js
var lastArg = (arguments.length > 0 && arguments[arguments.length - 1]) || undefined;
```
`lastArg` will always be `undefined` when the last passed argument is `false` not just _falsy_.

**To Reproduce**
Make an expectation of the last argument
_example uses `chai`_
```js
const foo = {
  greet() {
    this.speak("hello, world", false);
  },
  speak(msg, speakTwice) {
    console.log(msg);
    if (speakTwice) {
      console.log(msg);
    }
  }
};

const fake = sinon.fake();
sinon.replace(foo, "speak", fake);

foo.greet();

expect(fake.called).to.equal(true);
expect(fake.lastCall.lastArg).to.equal(false);
// => AssertionError: expected undefined to equal false
```

**Expected behavior**
`fake.lastCall.lastArg` should return `false` when the last argument of a function call is actually `false`.

* Library version: `7.2.5`
* Environment: `nodejs@10.15` `npm@6.4.1` on `macOS 10.14`
* Other libraries you are using:
```json
"chai": "^4.2.0",
"mocha": "^5.2.0",
"sinon": "^7.2.5",
"ts-node": "^8.0.2",
```

davidpaulhunt:
#### Purpose (TL;DR)
"Falsy" last arguments to `fake` were ignored.

Rely on arguments length instead of value comparison.

resolves #1986 

 #### Solution
Change `lastArg` assignment from this
```js
var lastArg = (arguments.length > 0 && arguments[arguments.length - 1]) || undefined;
```
to only trust arguments list length
```js
var lastArg = arguments.length > 0 ? arguments[arguments.length - 1] : undefined;
```

1. Check out this branch
2. `npm install`
3. `npm test`
4. (optional) Use `sinon.fake` and pass `false` as the last argument; assert `lastArg` to equal `false`; it should pass.

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/fake.js
+++ b/lib/sinon/fake.js
@@ -37,7 +37,7 @@ function cleanProxy(f) {
 var uuid = 0;
 function wrapFunc(f) {
     var fakeInstance = function() {
-        var lastArg = (arguments.length > 0 && arguments[arguments.length - 1]) || undefined;
+        var lastArg = arguments.length > 0 ? arguments[arguments.length - 1] : undefined;
         var callback = lastArg && typeof lastArg === "function" ? lastArg : undefined;
 
         /* eslint-disable no-use-before-define */
--- a/test/fake-test.js
+++ b/test/fake-test.js
@@ -143,6 +143,18 @@ describe("fake", function() {
             f(46);
             assert.equals(f.lastArg, 46);
 
+            f(false, true, 47, "string");
+            assert.equals(f.lastArg, "string");
+
+            f("string", false, true, 47);
+            assert.equals(f.lastArg, 47);
+
+            f(47, "string", false, true);
+            assert.equals(f.lastArg, true);
+
+            f(true, 47, "string", false);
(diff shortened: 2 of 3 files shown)


mantoni:
Cool, thank you! ❤️  Can you revert the `package-lock.json` changes? They are unrelated.

davidpaulhunt:
@mantoni I reverted `package-lock.json` and force pushed 👍

mroderick:
Thank you 🎸 

This has been published as `sinon@7.2.6`

PR #1991, opened 2019-03-08
Fix #1974 by upgrading to @sinonjs/samsam@3.3.0

### LINKED ISSUE #1974: calledWith ignores symbols used as property keys
**Describe the bug**
Asserting `calledWith` ignores mismatched properties if they keys are [JavaScript symbols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol).

**To Reproduce**
```
var sinon = require('sinon');
var mock = sinon.mock();

mock({
    [Symbol("a")]: true
});

sinon.assert.calledWith(
    mock,
    {
        [Symbol("a")]: false
    }
);
```

**Expected behavior**
The above script should throw an `AssertError` because the mock was called with an object containing a `true` value when we were expecting a `false` value. It should behave similarly to this example that uses non-symbolic keys:

```
var sinon = require('sinon');
var mock = sinon.mock();

mock({
    a: true
});

sinon.assert.calledWith(
    mock,
    {
        a: false
    }
);
```

image

* Library version: 7.2.3
* Environment: Node.js v10.8.0
* Other libraries you are using: None; I reproduced this issue in a newly initialized npm project.

**Additional context**
I started working on a solution, but I wanted to raise an issue before opening pull requests like the [contributing guide](https://github.com/sinonjs/sinon/blob/master/CONTRIBUTING.md#contributing-to-the-code-base) suggests to make sure I'm going down the right path. Using [this change for samsam](https://github.com/sinonjs/samsam/compare/master...BrandonE:symbolic-keys) and [this change for formatio](https://github.com/sinonjs/formatio/compare/master...BrandonE:symbols), I get the behavior I expected for the script in the _To Reproduce_ section:

image

Because symbols are [unique](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Description), this:

```
var sinon = require('sinon');
var mock = sinon.mock();

mock({
    [Symbol("a")]: true
});

sinon.assert.calledWith(
    mock,
    {
        [Symbol("a")]: true
    }
);
```

will throw an `AssertError` (thanks @pettyalex for pointing this out):

image

This, however, will not throw an `AssertError`:

```
var sinon = require('sinon');
var mock = sinon.mock();

const symbol = Symbol("a");

mock({
    [symbol]: true
});

sinon.assert.calledWith(
    mock,
    {
        [symbol]: true
    }
);
```

Unfortunately, the samsam change breaks [one unit test](https://github.com/sinonjs/samsam/blob/master/lib/deep-equal.test.js#L425):

image

This is because `arguments` has a property symbol called `iterator`:

```
> function test() { return Object.getOwnPropertySymbols(arguments) } test();
[ Symbol(Symbol.iterator) ]
```

mroderick:
1. Check out this branch
2. `npm ci`
3. `npm run test-cloud`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -309,9 +309,9 @@
       }
     },
     "@sinonjs/samsam": {
-      "version": "3.2.0",
-      "resolved": "https://registry.npmjs.org/@sinonjs/samsam/-/samsam-3.2.0.tgz",
-      "integrity": "sha512-j5F1rScewLtx6pbTK0UAjA3jJj4RYiSKOix53YWv+Jzy/AZ69qHxUpU8fwVLjyKbEEud9QrLpv6Ggs7WqTimYw==",
+      "version": "3.3.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/samsam/-/samsam-3.3.0.tgz",
+      "integrity": "sha512-beHeJM/RRAaLLsMJhsCvHK31rIqZuobfPLa/80yGH5hnD8PV1hyh9xJBJNFfNmO7yWqm+zomijHsXpI6iTQJfQ==",
       "requires": {
         "@sinonjs/commons": "^1.0.2",
         "array-from": "^2.1.1",
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
   "dependencies": {
     "@sinonjs/commons": "^1.4.0",
     "@sinonjs/formatio": "^3.2.1",
-    "@sinonjs/samsam": "^3.2.0",
+    "@sinonjs/samsam": "^3.3.0",
     "diff": "^3.5.0",
     "lolex": "^3.1.0",
     "nise": "^1.4.10",


mroderick:
> I assume we have a test for this in samsam. Otherwise feel a bit hesitant to add a fix without verification :-)

We do: https://github.com/BrandonE/samsam/blob/cef5959ce99377c9ce08a70629ecf648af7e9c4f/lib/deep-equal.test.js#L368-L400

BrandonE:
🎉 

@mroderick @mantoni Thoughts on when to release this change? I'm looking forward to trying this out in my project!

mroderick:
I meant to push a release it seems. Am away from computer this evening. @mantoni @fatso83 could you check if I did a release for this, and if not, make one?

mantoni:
This change doesn't need a release. Installing Sinon freshly will pull down the latest `samsam`. This PR was useful to verify that the tests are still passing and to make sure everyone else is running tests against the latest `samsam`.

PR #1994, opened 2019-03-12
Inject fake functionality and createStubInstance

fatso83:
#### Purpose
Fix sinonjs/sinon-test#102 by exposing the newer props on object we are
injecting the sandbox functionality into.

Also removes the unused `injectIntoThis` prop from the config.

1. Check out the `master` branch
1. `npm install sinon-test`
1. Run the mocha test case under this block. It should fail.
1. Check out this branch
1. Re-run the test case. It should now work.

```javascript
import * as sinon from "sinon";

const sinonTest = require("sinon-test");
const test = sinonTest(sinon);

describe("sandbox", function() {

    class MyClass {

    }

    it("test 1", test(function() {
        this.createStubInstance(MyClass);
    }));

});
```

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -71,6 +71,26 @@ function Sandbox() {
             return sandbox.mock.apply(null, arguments);
         };
 
+        obj.createStubInstance = function() {
+            return sandbox.createStubInstance.apply(sandbox, arguments);
+        };
+
+        obj.fake = function() {
+            return sandbox.fake.apply(null, arguments);
+        };
+
+        obj.replace = function() {
+            return sandbox.replace.apply(null, arguments);
+        };
+
+        obj.replaceSetter = function() {
+            return sandbox.replaceSetter.apply(null, arguments);
+        };
+
+        obj.replaceGetter = function() {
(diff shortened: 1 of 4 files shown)


fatso83:
The test fails on Node 10 with `[object Object]`. I thought this was fixed? I remember something about using a beta version or something of Chrome in Puppeteer ...

Also, not super happy [about the added test](https://github.com/sinonjs/sinon/pull/1994/files#diff-536d6d65592b132d984d4f646cb0448eR1567), but thought I should demonstrate that they actually could be called without crashing (something which did happen in the first iterations).

mroderick:
> The test fails on Node 10 with `[object Object]`. I thought this was fixed? I remember something about using a beta version or something of Chrome in Puppeteer ...

It could be a new error, or it could be that the error that previous unstable has become the unstable, without fixing that error.

mantoni:
I'm trying to get more information out of this with https://github.com/mantoni/mochify.js/pull/197.

fatso83:
Well, I did run the `test-cloud` script locally, and it passed, so I am pretty sure this is good to go. Merging.

PR #1995, opened 2019-03-13
Don't install Chrome unstable in Circle

mroderick:
We've stopped using it for builds, so we may as well save everyone some
bandwidth and time and not install it any longer


mantoni:
@mroderick Can you update Mochify to v6.1.1? We should get more information about the error now.

mroderick:
> @mroderick Can you update Mochify to v6.1.1? We should get more information about the error now.

I did, let's see what that gives us

mroderick:
@mantoni I've made you a collaborator on my fork of this repo, feel free to push commits to this branch

mantoni:
Too bad. Looks like it was something else. Unfortunately, I don't have more time to look into this today.

mroderick:
This has been obviated by https://github.com/sinonjs/sinon/pull/1999

PR #1999, opened 2019-03-20
Fix Circle CI headless builds breaking

fatso83:
All builds on the `node:10` image running the headless Puppeteer tests have been [failing](https://circleci.com/gh/sinonjs/sinon/1248) for the last two weeks (last good build is [1223](https://circleci.com/gh/sinonjs/sinon/1223).

This PR fixes that build by sticking to a known good version of `circleci/node:10-browsers`, which has a vetted version of Chrome (v72).

As https://github.com/sinonjs/sinon/pull/1995#issuecomment-472358296 details, the biggest problem with the failures was that we were left blind with just an `[object Object]` as the only output from the `test-headless` run.

So to get any further I needed to be able to
1. reproduce the issue
1. get some diagnostic input of value

 #### Solution  - optional
First I needed to be able to reproduce this locally to get a quicker feedback loop and not overload Circle CI. That story is [documented in a branch of its own](https://github.com/fatso83/sinon/commits/docker-experiments) and basically gave me a reproducible setup for local debugging. I could now do `docker-compose up` to run the tests. Now for the debugging output.

I dug into the documentation of Puppeteer and found it accepted an options object, which among other things had a field `dumpio`. By manually editing `node_modules/mochify/lib/chromium.js` to add the `dumpio:true` option, I could finally see some useful output! (I guess this should could exposed as a user option in `mochify`, @mantoni?)

At some point during the run, Chrome just gave up and exited with the following error:
```
    node_1 | [0320/100336.123457:FATAL:gpu_data_manager_impl_private.cc(892)] The display compositor is frequently crashing. Goodbye.
```

Googling that error I found GoogleChrome/puppeteer#894, which showed a lot of people having that issue, along with some proposed fixes (basically setting a version).

This has happened in other versions of Chrome as well, so by always downloading the latest version we are likely to break the build again. This is why I propose to explicitly use vetted image versions in this PR.

1. Check out this branch
2. `npm run test-docker-image`

 #### Checklist for author
- [x] `npm run lint` passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,15 +20,6 @@ jobs:
           command: |
             npm config set strict-ssl false
             npm install
-
-            # install latest chrome unstable version
-            if node --version | grep -q '^v10'; then
-              wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
-              echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' >> /etc/apt/sources.list.d/google.list
-              apt-get -qq update
-              apt-get -qq install -y --no-install-recommends google-chrome-stable
-              apt-get -qq install -y --no-install-recommends google-chrome-unstable
-            fi
       - run:
           name: Pre-Test
           # ESLint only supports Node >=4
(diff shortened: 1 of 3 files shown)


mantoni:
Thank you! I've opened an issue in Mochify to support this option. Didn't know about it so far, but could have used it myself in the past.

fatso83:
Already started on it (and just finished it).

PR #2007, opened 2019-04-11
improve error message for stub object property

wilk:


mroderick:
I agree that we should try to keep the recommendations in the documentation.

Issue #1358, opened 2017-03-22
Add documentation for how to stub ES6 module dependencies


gajus:
It is a common use case to mock ES modules.

e.g. I have a module that imports `getEventByEventId`

```js
import getEventByEventId from '../queries/getEventByEventId';

export default async (
  {
    connection,
    session
  }: ResolverContextType
) => {
  if (!session || !session.userId) {
    throw new Error('User must be authenticated.');
  }

  const event = await getEventByEventId(connection, eventId);

  // ..
}

```

In my test I want to mock `getEventByEventId`, e.g.

```js
import test from 'ava';
import sinon from 'sinon';
import moment from 'moment';
import createReservation from '../../../src/mutators/createReservation';

test('throws an error if the event is in the past', async (t) => {
  const parameters: any = {};

  const context: any = {
    session: {
      userId: 1
    }
  };

  const stub = sinon.stub().returns({
    date: moment().format('YYYY-MM-DD'),
    time: moment(new Date().getTime() - 1000 * 60).format('HH:mm')
  });

  // How to use the stub to mock `getEventByEventId`?

  await t.throws(createReservation(context), 'Cannot create a reservation for a past event.');
});

```

gajus:
To contribute to my own question, the only way I found this to work is by using [`babel-plugin-rewire`](https://github.com/speedskater/babel-plugin-rewire), e.g.

```js
import test from 'ava';
import sinon from 'sinon';
import moment from 'moment';
import createReservation from '../../../src/mutators/createReservation';

test('throws an error if the event is in the past', async (t) => {
  const parameters: any = {};

  const context: any = {
    session: {
      userId: 1
    }
  };

  const stub = sinon.stub().returns({
    date: moment().format('YYYY-MM-DD'),
    time: moment(new Date().getTime() - 1000 * 60).format('HH:mm')
  });

  createReservation.__Rewire__('getEventByEventId', stub);

  await t.throws(createReservation(context), 'Cannot create a reservation for a past event.');
});

```

This solution is far from ideal, though. For one, it breaks the Flow type annotations.

gajus:
One way to improve a DX would be to provide an ability to define a getter method when stubbing an object, e.g.

```js
const stub = sinon.stub(createReservation, '__Rewire__', 'getEventByEventId');

```

which would be equivalent to:

```js
const stub = sinon.stub();

createReservation.__Rewire__('getEventByEventId', stub);

```

gajus:
[...]

PR #2010, opened 2019-04-26
Improve dependency mocking documentation.

theKashey:
#### Purpose
Provide a better how-to, background and explanation for a dependency mocking. Make difference between different patterns a bit more clear.

Page changed: https://sinonjs.org/how-to/link-seams-commonjs/
> PS: Not renaming the file to make a diff view possible.

#### Background
"Add documentation for how to stub ES6 module dependencies" - https://github.com/sinonjs/sinon/issues/1358#issuecomment-474190978

#### Solution
I am not proposing a better solution, just providing a more rich explanation for the existing one.

### DIFF
--- a/docs/_howto/link-seams-commonjs.md
+++ b/docs/_howto/link-seams-commonjs.md
@@ -5,7 +5,7 @@ title: Dependency mocking with Sinon
 This page describes how to isolate your system under test, by stubbing out dependencies.
 
 While in other languages you might use [link seams][seams], [Dependency Injection(DI)][di] or
-[Inversion or Control(IoT)][IoT], which are not just _patterns_ but also requires some pattern
+[Inversion of Control (IoC)][IoC], which are not just _patterns_ but also requires some pattern
 implementation(a framework) - CommonJS(aka nodejs) module system provides a way better to replace
 any module, or package you need - _require-time_ dependency replacement.
 
--- a/docs/_howto/link-seams-commonjs.md
+++ b/docs/_howto/link-seams-commonjs.md
@@ -1,17 +1,80 @@
 ---
 layout: page
-title: How to use Link Seams with CommonJS
+title: Dependency mocking with Sinon
 ---
+This page describes how to isolate your system under test, by stubbing out dependencies.
 
-This page describes how to isolate your system under test, by stubbing out dependencies with [link seams][seams].
+While in other languages you might use [link seams][seams], [Dependency Injection(DI)][di] or
+[Inversion or Control(IoT)][IoT], which are not just _patterns_ but also requires some pattern
+implementation(a framework) - CommonJS(aka nodejs) module system provides a way better to replace
+any module, or package you need - _require-time_ dependency replacement.
 
-This is the CommonJS version, so we will be using [proxyquire][proxyquire] to construct our seams.
+## Why
+Before we started, let's clarify why you could not use _just Sinon_ to mock.
+### Source file:  `file.js`
+A short script, which does a quite important and a very dangerous action. Hopefully only for a `root` user.
+```javascript
+const fs = require('fs');
+const isRoot = require('./isRoot');
 
-To better understand the example and get a good description of what seams are, we recommend that you read the [seams (all 3 web pages)][seams] excerpt from [Working Effectively with Legacy Code][legacy] before proceeding.
+const wasRoot = isRoot(); // lets just precache it! Why not?
 
-Read it?
+exports.deleteEverything = function () {
+  if (wasRoot) {
+    throw new Error('only for a true root, ' + isRoot() ? 'sudoer' : 'user')
+  }
+  fs.removeAllFiles();
+}
+```
+### Test file: `file.test.js`
+Keep in mind - the test should be repeatable.
+```javascript
+const file = require('./file');
 
-Great! Let's continue.
+// you can stub deleteEverything
+sinon.stub(file, 'deleteEverything');
+// you can "safely" call it
+file.deleteEverything();
+// but how could it help?
+```
+In short - you can't test this file at all if you are not a root. And if you are, well, say goodbye to your file system.
+
+To solve tasks like this, `fs` and `./isRoot`, dependencies, not local or exported variables, should be stubbed and controlled.
+Sinon could stub only something you export, not the way module is assembled.
+
+### Why not Rewire
+Another popular solution for this problem is to use [rewire][rewire], which may _rewire_ module from inside.
+```js
+const file = rewire('./file');
+file.__set__('fs', sinon.stub())
+```
+Actually, the code above would not work, by design, as long as is not compatible with `const` declaration we used to define `fs`.
+Yes, you may use `let` or `var`, but is it a good solution?
+From another point of view, tests would require changing `wasRoot` value, which is derived from `isRoot` and should be the same, while
+you will have to mock them separately. This library is for _monkey-patching_, not for mocking.
+
+### Mock-fs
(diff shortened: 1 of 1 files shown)


docs/_howto/link-seams-commonjs.md
```diff
@@ -1,17 +1,80 @@
...
 layout: page
-title: How to use Link Seams with CommonJS
+title: Dependency mocking with Sinon
 ---
+This page describes how to isolate your system under test, by stubbing out dependencies.
 
-This page describes how to isolate your system under test, by stubbing out dependencies with [link seams][seams].
+While in other languages you might use [link seams][seams], [Dependency Injection(DI)][di] or
+[Inversion or Control(IoT)][IoT], which are not just _patterns_ but also requires some pattern
+implementation(a framework) - CommonJS(aka nodejs) module system provides a way better to replace
```
fatso83: You don't need a framework to do DI. I do it all the time by having optional constructor parameters. If using links seams really is "better" is debatable, as it introduces a new dependency, apis, complexity and mental overhead, but at least it doesn't require you to change the [SUT](https://en.wikipedia.org/wiki/System_under_test).

In general, you don't need any custom framework to use IoC (of which DI is one pattern). Constructor injection or service locators are simple to implement in any language. Less mystery is one of our tenants, in any case, so we wouldn't want to give people the impression you need special tools, unless you are doing "special" stuff like hooking into the module loading system (as we are doing here).

docs/_howto/link-seams-commonjs.md
```diff
@@ -66,19 +129,36 @@ describe('example', function () {
...
 });
 ```
+Look at the example above - code coverage is 100%. We have tested all branches we need,
+and even tested that `fs.existsSync` was called with a right arguments.
+
+Dependency mocking is a powerful tool, but sometimes you have to extract some functionality
+to another module to be able to mock it.
+
 
+[IoT]: https://en.wikipedia.org/wiki/Inversion_of_control
```
fatso83:
```suggestion
[IOC]: https://en.wikipedia.org/wiki/Inversion_of_control
```

docs/_howto/link-seams-commonjs.md
```diff
@@ -1,17 +1,80 @@
 ---
 layout: page
-title: How to use Link Seams with CommonJS
+title: Dependency mocking with Sinon
 ---
+This page describes how to isolate your system under test, by stubbing out dependencies.
 
-This page describes how to isolate your system under test, by stubbing out dependencies with [link seams][seams].
+While in other languages you might use [link seams][seams], [Dependency Injection(DI)][di] or
+[Inversion or Control(IoT)][IoT], which are not just _patterns_ but also requires some pattern
```
fatso83:
```suggestion
[Inversion of Control (IoC)][IoC], which are not just _patterns_ but also requires some pattern
```

docs/_howto/link-seams-commonjs.md
```diff
@@ -1,17 +1,80 @@
...
+  }
+  fs.removeAllFiles();
+}
+```
+### Test file: `file.test.js`
+Keep in mind - the test should be repeatable.
+```javascript
+const file = require('./file');
 
-To better understand the example and get a good description of what seams are, we recommend that you read the [seams (all 3 web pages)][seams] excerpt from [Working Effectively with Legacy Code][legacy] before proceeding.
```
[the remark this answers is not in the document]
theKashey: Sorry, but I shall disagree.
- "Seams" are a quite rare thing, probably defined only at the mentioned article, and that a quite _broad_ term.
- while seams are  _"alter behaviour in your program without editing in that place"_, they are not about dependency mocking
- provided examples are using or C preprocessor(not exists in JS), or Java class inheritance(might not exist in js). All examples are 100% alien to JS dev

I've read and wrote many articles related to this subject, and this one is probably most less useful.

docs/_howto/link-seams-commonjs.md
theKashey: - DI. Accepting your dependencies via arguments, ie using some sort of a factory pattern(and able to produce multiple objects). That's more about runtime.
- dependency mocking. Statically declaring your dependencies, without the ability to factor different versions. That's only about testing.

I would say - DI needs some framework, as long as you have to structure and design your code in a specific way, while dependency mocking operates on a bit different level and in a different time(tests only).

To be honest - I am quite happy that we have such ability in JS as a dependency mocking, so I don't have to implement 3 tier architecture to be able to mock `fs`. That makes code simpler and tests simpler.

There are more Java developers around me than JS, with all that _Enterprise stuff_ from the fairy tales, and I like to show them how simple the code could be. (Sometime they do the same)

docs/_howto/link-seams-commonjs.md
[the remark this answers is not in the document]
theKashey: About the articles. To say the truth - I don't know any good one, I truly want to recommend. I could share a few ones of mine, but they are also not great, and bound to rewiremock, which is a conflict of interest of some sort.
- https://itnext.io/how-to-mock-dependency-in-a-node-js-and-why-2ad4386f6587
- https://medium.com/techtrument/unit-tests-are-production-code-d256d86f073e
- https://itnext.io/unit-tests-for-skynet-written-in-js-6704265858a4
- https://itnext.io/mocking-was-a-code-smell-7f93d8a5d6f2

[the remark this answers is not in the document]
mroderick: I stand by my recommendation of that book. Please drop this change.

docs/_howto/link-seams-commonjs.md
mroderick: I agree with the opinion that dependency injection does not need a framework in JavaScript. I often use Sinon's spies, stubs and fakes in combination with pure functions. This makes my code easy enough to reason about, that I can get my colleagues to 👍 my pull requests.

docs/_howto/link-seams-commonjs.md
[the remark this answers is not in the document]
theKashey: Sure. But sooner or later it have to be replaced by a more js-related example, as long as the goal is to _improve_. Someone has to write it first.

The goal of the every article is to be useful for the reader, and provide some information their could use directly or indirectly, but just after reading. Unfortunately, patterns from the "3 pages" are not quite applicable to JS, or just don't exist in the same form:
- page 2 seam - is portable to js class inheritance, almost as-is.
- page 3 example 1(preprocessor seams) - there is no analog for preprocessor, except babel macros or require hooks, but that's not an option. The closest analog is using `process.env.NODE_ENV==='test'`, but, again, that's not the same.
- page 3 example 2(link seams) - seam at the linking stage is equal module dependency injection pattern.
- page 3 example 3(object seams) - no analog, except _inheritance seam_ from page 2, which is not the same.

So - if you have to idea how linking in C works, or what's the difference between C and JS class system(vast majority of devs) - then __0(zero) seams are applicable__ or understandable.

By fact there are 2 legit ways to use _seams_ in js:
- DI/providing _callbacks_ into the constructor, so creating an object/calling a function, which shall be "pure", and call only provided callbacks in a response, not perform a real action. Especially not to do anything before "the call".
- sandboxing - via class inheritance, module dependency replacement, and sinon.sandbox.
- 😬 while devs are usually just overriding class methods or module variables "when it's too late", and that's what I am fighting with.

fatso83:
I'll try to let the good work put into this come to use by adding some additional commits on top of it to bring it more inline with our general thoughts.

theKashey:
Sorry guys. Every day I am trying to get some free time to finish this PR and every day there is less and less free time left.

aiodh:
`theKashey:master`

PR #1549, opened 2017-09-03
fix #1442

killmenot:
#### Purpose (TL;DR) - mandatory
Fix issue #1442 by extending expectation with wrapped method props

#### Background (Problem in detail)  - optional
See issue #1442

#### Solution
Extend expectation with wrapped method's props

#### How to verify - mandatory
1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes
- [x] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

Issue #1442, opened 2017-05-31
Using callThrough with a mock expectation throws an 'apply' of undefined error.


jamespedid:
* Sinon version : 2.3.2
* Environment   : Node
* Example URL   :  N/A
* Other libraries you are using: Mocha, Chai, Chai as Promised

**What did you expect to happen?**
The mock expectation should support the full stub API, including callThrough. The mocked expectation should invoke the actual app code.

**What actually happens**
An error is thrown trying to use the callThrough method because this.stub.wrappedMethod appears to be undefined.

**How to reproduce**
```
describe('test', function () {
    it('should be ok', function () {
        const transaction = Object.create({
            commit: function () { return Promise.reject(); },
            rollback: function () { return Promise.resolve(); }
        });

        function appCode() {
            return transaction.commit();
        }

        const mock = sinon.mock(transaction);
        mock.expects('commit').once().callThrough();

        return expect(appCode()).to.eventually.be.fulfilled.then(() => mock.verify())
    });
});
```
<details>
    <summary>Stack Trace</summary>

    TypeError: Cannot read property 'apply' of undefined
      at Object.invoke (node_modules/sinon/lib/sinon/behavior.js:146:43)
      at Object.functionStub (node_modules/sinon/lib/sinon/stub.js:88:53)
      at Function.invoke (node_modules/sinon/lib/sinon/spy.js:194:51)
      at Function.invoke (node_modules/sinon/lib/sinon/mock-expectation.js:80:26)
      at Object.proxy (node_modules/sinon/lib/sinon/spy.js:97:22)
      at Object.invokeMethod (node_modules/sinon/lib/sinon/mock.js:132:43)
      at Object.commit (node_modules/sinon/lib/sinon/mock.js:66:35)
      at appCode (test/unit/graphql/models/Workflow.test.js:40:36)
</details>

tonylukasavage:
I've encountered the same error (sinon 2.3.2), as well as an additional one to go with it, so I distilled it down to as small an example as I could. The following example shows the error @jamespedid cites above, as well as shows that `callThrough()` is not behaving as expected. I'll leave it to the maintainer to decide whether these are related:

```
let sinon = require('sinon');

let obj = { query: function() { return 'foo'; } };

sinon.stub(obj, 'query')
	.withArgs(123)
	.returns('stubbed')
	.callThrough();

console.log(obj.query());     // prints "undefined"
console.log(obj.query(123));  // throws error 
```

stack trace

```
[...]

PR #2012, opened 2019-04-30
Fix issue #1964

### LINKED ISSUE #1964: callThrough doesn't work on a withArgs sub-stub
**Describe the bug**

#1442 was not fully fixed by #1549.  The case of using stubs directly instead of with mocks and expectations, as shown in https://github.com/sinonjs/sinon/issues/1442#issuecomment-306330402, is still broken.

**To Reproduce**
This code will throw an exception `TypeError: Cannot read property 'apply' of undefined` due to the fake returned from `withArgs` not having a `wrappedMethod` member.
```javascript
let stub = sinon.stub(obj, 'method')
	.withArgs(123)
	.returns('bar')
	.callThrough();
obj.method('bar')
```

**Expected behavior**
The `callThrough` behavior should be available on `withArgs` fakes.

**Screenshots**

* Library version: Verified with `6.3.4` and `master` at `97a12178`
* Environment: Linux Node 10.x
* Example URL: N/A
* Other libraries you are using: N/A

**Additional context**

I've prepared a test case demonstrating the issue:
https://github.com/sinonjs/sinon/compare/master...mgabeler-lee-6rs:issue-1964

mgabeler-lee-6rs:
#### Purpose

Fix #1964 by adding a helper method to walk the stub parent links to find the `wrappedMethod` value for the `callsThrough` behavior.

 #### Background

If you try to use the `callsThrough` behavior on a `withArgs` stub, calling the stub with the matching args will fail, because the behavior fails to locate the original wrapped method to call through to.

This is effectively a followup to #1442 / #1549.

 #### Solution  - optional

A simple helper method walks up `stub.parent` links until it finds a `stub` that has a `wrappedMethod` member.

1. Check out this branch
2. `npm install`
3. `npm test`, or just `npm run test-node`, or just `npm run test-node --grep 1964` to run the issue-specific test case

 #### Checklist for author

- [x] `npm run lint` passes
  - N/A -- no standard library functions used

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -170,7 +170,8 @@ var proto = {
         } else if (this.reject) {
             return (this.promiseLibrary || Promise).reject(this.returnValue);
         } else if (this.callsThrough) {
-            return this.stub.wrappedMethod.apply(context, args);
+            var wrappedMethod = this.effectiveWrappedMethod();
+            return wrappedMethod.apply(context, args);
         } else if (typeof this.returnValue !== "undefined") {
             return this.returnValue;
         } else if (typeof this.callArgAt === "number") {
@@ -178,6 +179,15 @@ var proto = {
         }
 
         return this.returnValue;
+    },
+
+    effectiveWrappedMethod: function effectiveWrappedMethod() {
+        for (var stubb = this.stub; stubb; stubb = stubb.parent) {
+            if (stubb.wrappedMethod) {
+                return stubb.wrappedMethod;
+            }
+        }
+        return undefined;
     },
 
     onCall: function onCall(index) {
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -187,7 +187,7 @@ var proto = {
                 return stubb.wrappedMethod;
             }
         }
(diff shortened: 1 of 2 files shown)


lib/sinon/behavior.js
```diff
@@ -180,6 +181,15 @@ var proto = {
         return this.returnValue;
     },
 
+    effectiveWrappedMethod: function effectiveWrappedMethod() {
+        for (var stubb = this.stub; stubb; stubb = stubb.parent) {
+            if (stubb.wrappedMethod) {
+                return stubb.wrappedMethod;
+            }
+        }
+        return undefined;
```
fatso83: I'd much prefer failing with an explit `Error` here telling us which assumption was wrong, rather than failing in the next step when ` wrappedMethod.apply` results in a null pointer exception.

```suggestion
        throw new Error("Unable to find wrapped method");
```

mgabeler-lee-6rs: I had thought about that, but hadn't done it at first in the interest of minimizing changes to existing behavior.  But if you're in favor of it, then so am I :)

fatso83:
Thanks!

Issue #2014, opened 2019-05-07
Copyright Year in Footer (2010-2018)


vinitshahdeo:
The copyright year in footer is `2010-2018`. But now it should be `2010-2019`.

<img width="444" alt="Screenshot 2019-05-07 at 6 15 00 PM" src="https://user-images.githubusercontent.com/20594326/57300423-688e5400-70f4-11e9-8053-60cbd9ad1508.png">

fatso83:
Closed in 62937cc

PR #2015, opened 2019-05-07
Updated Copyright year in Footer

vinitshahdeo:
Solved issue #2014 

Earlier it was `2010-2018`. I have changed it to `2010-2019`

### EARLIER

Screenshot 2019-05-07 at 6 15 00 PM

### NOW

Screenshot 2019-05-07 at 6 20 28 PM

### DIFF
--- a/docs/_includes/footer.html
+++ b/docs/_includes/footer.html
@@ -4,7 +4,7 @@
   <div class="container text-center">
     <a href="http://sinonjs.org/" target="blank"><img class="grow" src="{{ "/assets/images/logo.png" | prepend: site.baseurl }}" alt="Sinon.JS"></a>
     <p><a target="blank" class="join-btn" href="https://stackoverflow.com/questions/tagged/sinon">Join the discussion on Stack Overflow!</a></p>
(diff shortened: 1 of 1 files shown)


fatso83:
Thanks for the fix. I am just wondering if we should do this differently: our work should be copyrighted by default. Wouldn't it be sufficient to change the text to drop the years altogether? "All copyright is reserved the Sinon committers", or something? Don't see what the years is bringing to the table.

vinitshahdeo:
@fatso83 
Yes!
_**All copyright is reserved the Sinon committers**_ seems fine. There will be no longer changes required.

Should I update the PR with the same?

mroderick:
Thank you for your pull request and your patience with the slow response from me. My paid work has been very demanding in the last few months, so I haven't had much time to dedicate towards unpaid work.

I agree, I think we should remove the year from the copyright statement.

I still think we should keep the statement, as it acknowledges the contributions of the committers.

mroderick:
@vinitshahdeo will you update the PR to remove the year entirely?

vinitshahdeo:
> @vinitshahdeo will you update the PR to remove the year entirely?

@mroderick Sure, will update soon!

Issue #2026, opened 2019-05-22
spy should support the same signature as stub for objects


benawhite:
* Sinon version : v7.3.2
* Environment   : node
* Example URL   : n/a
* Other libraries you are using: n/a

**What did you expect to happen?**
I would expect that I could **spy** on all methods of an object without explicitly calling out each method by name... the same a **stub**

**What actually happens**
TypeError: Attempted to wrap undefined property undefined as function

**How to reproduce**
```javascript
// create sandbox
const sandbox = sinon.createSandbox();
// all methods should be spied
sandbox.spy(console); // actually throws error

// ... do some stuff here ...

// restore back to original
sandbox.restore();
```

benawhite:
I added a pull request to address this https://github.com/sinonjs/sinon/pull/2027

PR #2027, opened 2019-05-22
2026 - Updated to support spying all methods on an object

benawhite:
#### Purpose
Fix issue #2026 by adding **spyEntireObject** to the supported signature of sinon.spy
added tests and updated docs to reflect the change

 #### How to verify
1. Check out this branch
2. `npm install`
3. npm run test
4. note: **assert-test** / **supports no properties** uses the newly added syntax and no other tests are broken
```
sinonSpy(api);
```

 #### Checklist for author

- [X] `npm run lint` passes
- [X] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

### DIFF
--- a/docs/release-source/release/spies.md
+++ b/docs/release-source/release/spies.md
@@ -32,6 +32,40 @@ handles a callback, as in the following simplified example:
 }
 ```
 
+### Using a spy to wrap all object method
+
+`sinon.spy(object)`
+
+Spies all the object's methods.
+
+Note that it's usually better practice to spy individual methods, particularly on objects that you don't understand or control all the methods for (e.g. library dependencies).
+
+spying individual methods tests intent more precisely and is less susceptible to unexpected behavior as the object's code evolves.
+
+ The following is a slightly contrived example:
+
+```javascript
+{
+    sandbox: sinon.createSandbox(),
+
+    setUp: function () {
+        this.sandbox.spy(jQuery);
+    },
+
+    tearDown: function () {
+        this.sandbox.restore(); // Unwraps all spied methods
+    },
+
+    "test should inspect jQuery.getJSON's usage of jQuery.ajax": function () {
+        jQuery.getJSON("/some/resource");
+
+        assert(jQuery.ajax.calledOnce);
+        assertEquals("/some/resource", jQuery.ajax.getCall(0).args[0].url);
+        assertEquals("json", jQuery.ajax.getCall(0).args[0].dataType);
+    }
+}
+```
+
 
 ### Using a spy to wrap an existing method
 
--- a/lib/sinon/spy-entire-object.js
+++ b/lib/sinon/spy-entire-object.js
@@ -1,0 +1,22 @@
+"use strict";
+
+var getPropertyDescriptor = require("./util/core/get-property-descriptor");
+var walk = require("./util/core/walk");
+
+function spyEntireObject(spy, object) {
+    walk(object || {}, function(prop, propOwner) {
+        // we don't want to spy things like toString(), valueOf(), etc. so we only spy if the object
+        // is not Object.prototype
+        if (
+            propOwner !== Object.prototype &&
+            prop !== "constructor" &&
+            typeof getPropertyDescriptor(propOwner, prop).value === "function"
+        ) {
+            spy(object, prop);
+        }
+    });
+
+    return object;
+}
+
+module.exports = spyEntireObject;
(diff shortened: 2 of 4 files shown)


lib/sinon/spy-entire-object.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+var getPropertyDescriptor = require("./util/core/get-property-descriptor");
+var walk = require("./util/core/walk");
+
+function spyEntireObject(spy, object) {
+    walk(object || {}, function(prop, propOwner) {
+        // we don't want to spy things like toString(), valueOf(), etc. so we only spy if the object
```
lucasfcosta: Is there any particular reason for us no to spy on those? IMO these should also be spied on; otherwise it feels like we're making decisions on behalf of our users. I agree most users wouldn't need these spied on, but that's up to them.

I'd go for the least astonishment principle here and spy on every single function.

People can also be attaching methods to `Object.prototype` — which is weird, but valid nonetheless.

lib/sinon/spy-entire-object.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+var getPropertyDescriptor = require("./util/core/get-property-descriptor");
+var walk = require("./util/core/walk");
+
+function spyEntireObject(spy, object) {
+    walk(object || {}, function(prop, propOwner) {
```
lucasfcosta: Maybe we could have this callback as an arrow function?

lib/sinon/spy-entire-object.js
```diff
@@ -0,0 +1,22 @@
...
+        // we don't want to spy things like toString(), valueOf(), etc. so we only spy if the object
+        // is not Object.prototype
+        if (
+            propOwner !== Object.prototype &&
+            prop !== "constructor" &&
```
lucasfcosta: We may also want to spy on the `constructor` for the same reasons above. I think reducing edge-cases would be productive here.

docs/release-source/release/spies.md
```diff
@@ -32,6 +32,40 @@ handles a callback, as in the following simplified example:
...
 
+### Using a spy to wrap all object method
+
+`sinon.spy(object)`
+
+Spies all the object's methods.
+
+Note that it's usually better practice to spy individual methods, particularly on objects that you don't understand or control all the methods for (e.g. library dependencies).
+
+spying individual methods tests intent more precisely and is less susceptible to unexpected behavior as the object's code evolves.
```
lucasfcosta: Sorry for nitpicking, but I think the first character here needs to be in uppercase.

docs/release-source/release/spies.md
```diff
@@ -32,6 +32,40 @@ handles a callback, as in the following simplified example:
 }
 ```
 
+### Using a spy to wrap all object method
```
lucasfcosta: Very well written docs, I especially like the example, thanks!

lib/sinon/spy-entire-object.js
fatso83: the entire codebase is ES5.1, so arrow functions are out.

lib/sinon/spy-entire-object.js
benawhite: This is a blatant copy of stub-entire-object.js as I wanted to keep the API for the two in sync.

benawhite: FYI, I also agree with the way it was originally written. Spying on native object methods seems out of scope...

lucasfcosta: I see. That's fine then, I didn't know that, sorry 😅

fatso83:
This is all very nice work, @benawhite. Any chance of getting those ✅ nailed down before the Summer gets to us all? 🌞

fatso83:
Having a new look at this, seeing it has been stalling for two months. Right now, there are two things I don't understand:
1. Which tests/edge cases are you pointing to, @mantoni? I scanned the changed files several times without seeing any todos or special comments. The only thing I see that hasn't been adressed is a minor spelling/letter casing thing mentioned by Lucas.
2. What is the fourth point in the PR description really saying, @benawhite? I don't understand this:
> note: assert-test / supports no properties uses the newly added syntax and no other tests are broken

I could fix the spelling issue and merge this if someone could enlighten me on the above two points.

mantoni:
There is a comment saying

> we don't want to spy things like toString(), valueOf(), etc.

It‘s documenting and if statement. There are no tests covering these cases.

Sorry, I should have been more specific.

fearphage:
As it stands, you either have to use this with a `sandbox` or you have to go to each property of the related object to call `.restore()`. It would be nice (read: good user experience) if `spyOnEntireObj` returned a way to revert everything back to normal.

fatso83:
I created a new branch that addresses all of the points above, except Phred's suggestion. I feel that improvement belongs in a separate PR, as it should touch both the stubs and spies API.

@mantoni, you'll find the tests you requested as `does not replace non-enumerable properties from Object.prototype`, which are now shared between the stubs and spies test suite.

Issue #1553, opened 2017-09-07
Add sinon.restore(obj) to restore all stubbed methods on the object


tsclaus:
Without using sandboxes, there is no quick reversal to `sinon.stub(obj)`. `sinon.restore(obj)` and/or `sinon.stub.restore(obj)` could be added to serve as the inverse to `sinon.stub(obj)`. With the `walk` utility function, this should be pretty easy; with approval, I can add the feature.

An example of its usefulness (using Karma):

```
const sinon = require('sinon');

const testedLibrary = require('testedLibrary'),
      dependency    = require('dependency');

describe('testedLibrary', function() {
  beforeEach(function() {
    sinon.stub(dependency);
  });

  it('should test something', function() {
    // changing dependency.method behavior here automatically changes it in `testedLibrary.js` as well
    // since the same object is used in both files
  });

  afterEach(function() {
    // without this you either have to clear the cache or iterate over dependency yourself
    sinon.restore(dependency);
  });
});
```

tsclaus:
Apologies, I missed this is in `util.js` as it's own function. This issue should definitely be closed.

fatso83:
@tsclaus We always appreciate [contributions to the docs](https://github.com/sinonjs/sinon/blob/master/docs/CONTRIBUTING.md) :-)

PR #2045, opened 2019-06-14
Explain how to restore a wholly stubbed object

samlandfried:
#### Purpose (TL;DR)

To help those trying to restore a wholly-stubbed object via `sinon.stub(obj)`.

 #### Background

* Solution and doc change proposed in #1553 
* Whoever is reading this will still need to figure out how to access the utility. It functions as a clue more than a real solution. There are a few reasons I'm not explaining how to use it:
    1. This doesn't feel like the right place to explain how to import a utility.
    1. I'm using `restore` by directly importing the util via `import restore from 'sinon/lib/sinon/util/core/restore.js'`. I wonder if there's a more direct way to access it that would be better to recommend.
    1. [This page](https://sinonjs.org/releases/latest/utils/) makes me unsure whether or not it's something you want to draw attention to at all.

NA. It's a documentation change.

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/_releases/latest/stubs.md
+++ b/docs/_releases/latest/stubs.md
@@ -86,7 +86,7 @@ A [codemod is available](https://github.com/hurrymaplelad/sinon-codemod) to upgr
 
 #### `var stub = sinon.stub(obj);`
 
-Stubs all the object's methods.
+Stubs all the object's methods. Use the `restore` utility to restore an object stubbed this way.
 
 Note that it's usually better practice to stub individual methods, particularly on objects that you don't understand or control all the methods for (e.g. library dependencies).
 


mroderick:
Thank you for your contribution 🚀

AlmogCohen:
That's a great contribution :) Thank you! I've been looking for these missing docs...

AlmogCohen:
Was this already made public?

fatso83:
@AlmogCohen No, we haven't made a new release since 7.5.0. We probably should, though :)

AlmogCohen:
Thanks for the quick response :) Just see this comment as a +1 for even a mini release as this docs fix is something I was looking for :)

Thanks!!

Issue #1538, opened 2017-08-18
Document dangers of using destructuring to require a function that was stubbed.


dllndv:
when stubbing a function like this:
```
const requiredThing = require('/path/to/required/thing');

sinon.stub(requiredThing, 'requiredFunction').callsFake(func);
```
and then that function is required with destructuring by another file that uses it like this:
```
const { requiredFunction } = require('/path/to/required/thing');

requiredFunction();
```
the stub is not used because the reference is different. This isn't necessarily a bug with sinon, rather a quirk in how references and destructuring works in JS, and it should be added to the documentation somewhere so that another dev doesn't lose a whole day trying to figure out what's going on like I did. Thanks for the help and support!

mroderick:
It falls outside the focus of open source libraries such as Sinon.JS to document all the quirks in JavaScript and it's many implementations, no matter how frustrating they are. 

A better place for such documentation would be at [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript), where it is much more likely to be seen by a wider audience.

jhnlsn:
yes, we also spent quite some time trying to debug this only to find that this was the problem!  Wish we knew about this before hand

fatso83:
We might put such a thing on our how-to section, but I don't think people read those until googling for an answer. And once they do, they'll be just as likely to find this issue or [some Stack Overflow question](https://stackoverflow.com/a/52591287/200987), which probably is a better fit any how for such things.

StuAtGit:
Yet another person running into this.

> ```It falls outside the focus of open source libraries such as Sinon.JS to document all the quirks in JavaScript```

Very true!! Javascript has a lot of quirks, and that would be hard task.

But irrelevant to this issue - it's about a quirk in *sinon* not javascript. Destructuring is not a "quirk" it's an established language feature in common use. (Still not sure whether I personally like it, but I have to live it!). The fact that sinon stubs of a method fail when one imports that method via a destructure statement vs a reference to the object is a quirk of *sinon's underlying implementation of stubs*.

Unless, of course, you define stubs as "The thing sinon does when you call stub() - read the code if you want to know what that means", and not the definition given in the sinon docs or in wide use in many different mocking/testing libraries in many different languages.

PR #2049, opened 2019-06-19
Documentation for stubbing modules

kobbikobb:
It is not obvious from the documentation how you stub external dependencies (modules that are imported from the modules under test).
I feel we should add something to help users with this common use case.
What do you think?

1. Check out this branch
2. `npm install`
3. <your-steps-here>

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/_howto/stub-dependency.md
+++ b/docs/_howto/stub-dependency.md
@@ -1,0 +1,54 @@
+---
+layout: page
+title: How to stub a dependency of a module?
+---
+
+Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](./link-seams-commonjs) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
+
+To stub a dependency (imported module) of a module under test you have to import it explicitly in your test and stub the desired method. For the stubbing to work, the stubbed method cannot be [destructured](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), neither in the module under test nor in the test.
+
+## Example: dependencyModule.js
+```javascript
+
+function getSecretNumber() {
+  return 44;
+}
+
+module.exports = {
+  getSecretNumber
+};
+```
+
+## Example: moduleUnderTest.js
+
+```javascript
+const dependencyModule = require("./dependencyModule");
+
+function getTheSecret() {
+  return `The secret was: ${dependencyModule.getSecretNumber()}`;
+}
+
+module.exports = {
+  getTheSecret
+};
+```
+
+## Example: test.js
+
+```javascript
+const assert = require("assert");
+const sinon = require("sinon");
+
+const dependencyModule = require("./dependencyModule");
+const { getTheSecret } = require("./moduleUnderTest");
+
+describe("moduleUnderTest", function() {
+  describe("when the secret is 3", function() {
+    it("should be returned with a string prefix", function() {
+      sinon.stub(dependencyModule, "getSecretNumber").returns(3);
+      const result = getTheSecret();
+      assert.equal(result, "The secret was: 3");
+    });
+  });
+});
+```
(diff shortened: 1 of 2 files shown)


fatso83:
I think this should go here: https://github.com/sinonjs/sinon/tree/master/docs/_howto

With a link from the stub docs.

kobbikobb:
@fatso83 good idea, I've pushed changes :)

docs/_howto/stub-dependency.md
```diff
@@ -0,0 +1,21 @@
+---
+layout: page
+title: How to stub a dependency of a module?
+---
+
+To stub dependency(imported module) of a module under test you have to import it explicitly in your test and stub the desired method.
```
fatso83: missing a space.

docs/_howto/stub-dependency.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Example
+
+```javascript
+import * as importedModule from "./importedModule";
+import {getTheSecret} from "./moduleUnderTest";
+
+"test should get secret": function () {
```
fatso83: This isn't actually valid javascript as it is and will result in a runtime error (`Unexpected token, expected ;`). Could you change it to the standard BDD style Mocha test style we use everywhere else?

docs/_howto/stub-dependency.md
fatso83: You might add that Sinon is simply a stubbing library, not a module intercepting library, and how you stub out dependencies is highly dependant on your environment. For Node environments, we usually recommend solutions targetting [link seams](https://github.com/sinonjs/sinon/blob/8683af388a2b46b2b3e7764fe6fcfe91b8543e35/docs/_howto/link-seams-commonjs.md) (replace with relative link) or explicit dependency injection. Though in _some simple cases_, you can get away with just using Sinon by modifying the module exports of the dependency. 

(_And then comes your text_)

docs/_releases/latest/stubs.md
```diff
@@ -76,6 +76,8 @@ Replaces `object.method` with a stub function. An exception is thrown if the pro
 
 The original function can be restored by calling `object.method.restore();` (or `stub.restore();`).
 
+A dependency of a module can be stubbed [see example](../../_howto/stub-dependency.md).
```
fatso83:
```suggestion
Stubs can be used to replace a module's dependencies, ensuring it is tested in isolation. [For simple cases](../../_howto/stub-dependency.md) you might not need additional helper libraries, but for more complex cases you might need to [target the module loading mechanisms](../../link-seams-commonjs.md).
```

kobbikobb:
@fatso83 thank you for your comments and suggestions. I have done your suggestions and added also the modules under test. Mocking modules this way is highly dependant upon the implementation, so adding a complete example might help the user understand better.

Any counter proposals?
I also rebased the code...

docs/_howto/stub-dependency.md
```diff
@@ -0,0 +1,54 @@
+---
+layout: page
+title: How to stub a dependency of a module?
+---
+
+Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](./link-seams-commonjs) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
+
+To stub a dependency (imported module) of a module under test you have to import it explicitly in your test and stub the desired method. For the mocking to work, the mocked method cannot be deconstructed, neither in the module under test nor in the test.
```
fatso83: I'd replace "mocking" with "stubbing" and "mocked" with "stubbed", since we actually differ "mocks" from "stubs" in our docs and API. Otherwise, people might get a bit confused.

Also, replace "deconstructed" with "destructured", using a link to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). That's assuming I understand it's actually destructuring you refer to? As in
```javascript
const {a} = require('/module-with-a-export');
```

kobbikobb: @fatso83 Makes sense, yes you are right about destructed, and the link to MDN makes this more clear. Good feedback :)

I've updated the code accordingly.

BTW, I see there have already been discussions about this topic.
https://github.com/sinonjs/sinon/issues/1538

Hopefully this documentation we are adding will help developers having problems with stubbing dependencies this way :)

fatso83:
I'm assuming this is good enough for merge, since no-one has chimed in. It adds value and if anyone thinks it can be improved, we now have a good foundation to be build on for further additions :-)

Thanks for the work and sticking it through!

PR #2052, opened 2019-06-26
Update broken link

lineus:
currently, clicking this link results in a 404. This should fix it.

### DIFF
--- a/docs/_howto/stub-dependency.md
+++ b/docs/_howto/stub-dependency.md
@@ -3,7 +3,7 @@ layout: page
 title: How to stub a dependency of a module?
 ---
 
-Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](./link-seams-commonjs) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
+Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](./link-seams-commonjs/) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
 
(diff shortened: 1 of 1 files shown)


docs/_howto/stub-dependency.md
```diff
@@ -3,7 +3,7 @@ layout: page
 title: How to stub a dependency of a module?
 ---
 
-Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](./link-seams-commonjs) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
+Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](./link-seams-commonjs/) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
```
fatso83: It's not the slash missing, it's at the wrong level. Currently you end up at https://sinonjs.org/how-to/stub-dependency/link-seams-commonjs/.

```suggestion
Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](../link-seams-commonjs/) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
```

lineus:
Thanks @fatso83. You're absolutely right :)

kobbikobb:
@fatso83 sorry, this is on me. I did not run this :(

But it seems to be still broken, seems to me the original PR had the right path but was missing the .md

```
./link-seams-commonjs.md
```

?

fatso83:
@kobbikobb Unfortunately, while that might seem to work when just working with Markdown, that's not the case for the produced docs. This is working now (after this has been merged), as [you can see on the Sinon homepage](https://sinonjs.org/how-to/stub-dependency/) (try clicking the link seams link).

kobbikobb:
@fatso83 thank you I see it now :)

PR #2054, opened 2019-06-28
Add documentation on how to test methods that return promises, linked…

kobbikobb:
Issue https://github.com/sinonjs/sinon/issues/1898 is stale and talks about missing documentation for promises. Is this something we would like?

I can change the code to use dependency injection if we think more developers would use that. If we want that we need to decide constructor or property injection... :)

1. Check out this branch
2. `npm install`
3. <your-steps-here>

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/_howto/stub-dependency.md
+++ b/docs/_howto/stub-dependency.md
@@ -1,15 +1,17 @@
 ---
 layout: page
-title: How to stub a dependency of a module?
+title: How to stub a dependency of a module
 ---
 
 Sinon is simply a stubbing library, not a module interception library. Stubbing dependencies is highly dependant on your enviroment and the implementation. For Node environments, we usually recommend solutions targetting [link seams](../link-seams-commonjs/) or explicit dependency injection. Though in some simple cases, you can get away with just using Sinon by modifying the module exports of the dependency.
 
 To stub a dependency (imported module) of a module under test you have to import it explicitly in your test and stub the desired method. For the stubbing to work, the stubbed method cannot be [destructured](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), neither in the module under test nor in the test.
 
-## Example: dependencyModule.js
-```javascript
+## A simple example
 
+### Source file: dependencyModule.js
+
+```javascript
 function getSecretNumber() {
   return 44;
 }
@@ -19,7 +21,7 @@ module.exports = {
 };
 ```
 
-## Example: moduleUnderTest.js
+### Source file: moduleUnderTest.js
 
 ```javascript
 const dependencyModule = require("./dependencyModule");
@@ -33,7 +35,7 @@ module.exports = {
 };
 ```
 
-## Example: test.js
+### Test file: test.js
 
 ```javascript
 const assert = require("assert");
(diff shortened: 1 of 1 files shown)


fatso83:
Coincidentally, I looked into all of these issues yesterday and actually followed up on this! Weird timing :)

@orlangure already made a significant effort on this in #1931, which have unfortunately stalled for some time, but I think I'm merging this within a week. I'd rather have this build-upon that work, if that is fine with you? I'll ensure it doesn't drag on any longer than it needs to. Working beats perfect any day :)

fatso83:
OK, #1931 has been merged. Could you see if you have stuff that is uncovered in https://github.com/sinonjs/sinon/blob/master/docs/_howto/lolex-async-promises.md and copy it in?

kobbikobb:
@fatso83 my examples are realistic and show how to test simple async code with stubs and different arguments.

Lolex is intended for more complex scenarios, f.x. setTimeout, setInterval...

I think some developers might benefit in some opinionated and simple ways of testing async code.
The developers I work with thought you needed to call a done callback but that is not the case and they tend to not to test async code because they think it is too complicated.

I am open to anything if we want to move it, change it or combine this with other documentation.
You might think this is too general for our documentation and maybe it is.

What do you think? :D

fatso83:
Right now, it's a demo of Mocha's API, along with some examples on doing async tests - which happen to use Promises. I think _this is fine_, but it's covering a bit of different ground than we were originally thinking of with #1898. I think your doc is basically the complement to your previous addition: https://github.com/sinonjs/sinon/blob/master/docs/_howto/stub-dependency.md.

Maybe it would fit better there, as a section on how to stub out dependencies that are async? It could have your existing stuff, but possibly also an example on doing the same for classic callback passing async code.

kobbikobb:
@fatso83 Good idea, I was thinking the same thing. I added them together and improved some texts.

The thing that is now bothering me is that the examples: simple and asynchronous are not working on the same codebase(one is more real like...). Maybe it does not matter.

What do you think, any feedback?

docs/_howto/stub-dependency.md
```diff
@@ -52,3 +54,128 @@ describe("moduleUnderTest", function() {
   });
 });
 ```
+
+## A complex example with asynchronous code
+
+In some cases you might need to stub a dependency that returns a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). To test it you can add the [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) keyword to the test method and call the method being tested with the [await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) keyword.
```
fatso83: @mroderick How do you feel about this? I know our codebase is strictly ES5.1, but this is probably how most people write new async code these days, so the examples are more relevant in that regard by using ES2015 syntax.

kobbikobb: @fatso83 Good point, I didn't even notice I wrote it differently :P It should at least be consistent... Maybe it should be ES5 but very few write code like that...

fatso83: I'm saying ES2015+ is fine in example docs, given that there was no feedback from Morgan.

Whether or not there are few people writing code like that is debatable ... there are very few modern tutorials not emplying a kitchen sink of a build pipeline, but there are definitely looooads of web developers that don't do that, possibly most. But the ones who only make little interactive snippets here and there are not the ones usually employing unit testing in any case :)

mroderick:
I don't think the `echo` file was really necessary here ...

Issue #1951, opened 2018-11-19
IE 11 not supported


eps1lon:
**Describe the bug**
IE 11 environment is not supported with regards to syntax and methods. For example sinon has transitive dependencies to `has-flag` which is an arrow function and uses `String.prototype.startsWith`. Missing syntax support wouldn't be as bad since it can be included in babel-loader but adding polyfills is kind of rough since this would also mean that it wouldn't warn us if we use `startsWith`. Maybe there is also an option to only include polyfills for certain node_modules but both points can't really be attributed to IE 11 support.

**To Reproduce**
1. Build `sinon-esm.js`
2. search for `hasFlag` which is an arrow function and uses `startsWith`

**Expected behavior**
Support for IE 11 out of the box.

**Screenshots**
None

**Context (please complete the following information):**
We're trying run unit tests for our react components in IE 11.

* Library version: 7.1.1
* Environment: IE 11
* Example URL: none
* Other libraries you are using: babel, enzyme, karma, webpack

I guess enzyme will be another bottleneck which requires `airbnb-browser-shims` anyway and I couldn't find a solution to provide polyfills only to certain dependencies.

fatso83:
We are running Sinon in IE11 in our own browser tests. IE11 does not support Ecmascript Modules, so it makes no sense to use the sinon-esm build.

Just use the Sinon.js bundle. Works everywhere.

oliviertassinari:
@fatso83 The issue we are hitting with the `esm` version isn't with Ecmascript Modules, it's with the ES6 syntax, like arrow functions.

fatso83:
https://saucelabs.com/beta/tests/475df00710394fc0bd6890fad7051580/commands#0

https://api.travis-ci.org/v3/job/456960446/log.txt

fatso83:
@oliviertassinari, I get that, but if you are consuming an Ecmascript module you shouldn't expect it to be ES5 to begin with (as it's not). You *will* require some sort of transpilation step in your build pipeline if you build a bundle from ES modules, as this is not valid ES5, but it IS a ES module:
`export { _createFakeServerWithClock as createFakeServerWithClock };`

I suspect that your webpack config excludes node_modules from undergoing Babel transplantation, so you either need to selectively include Sinon or just import the non-esm bundle instead.

Chances are I just had the same issue with the Material UI React lib (what are the chances, Olivier!), so you could employ my fix: https://stackoverflow.com/q/53154986/200987

[...]

PR #2060, opened 2019-07-12
add browser field in package.json

Feiyang1:

It makes bundlers(webpack, rollup) to pick up the es5 version by default when targeting browsers. It solves issues like https://github.com/sinonjs/sinon/issues/1951

1. Check out this branch
2. `npm install`
3. use webpack to bundle tests without compiling sinon
4. run tests in non-es6 ready browsers, e.g. IE11 
5. tests should run successfully

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
     "LICENSE",
     "README.md"
   ],
+  "browser": "./lib/sinon.js",
   "main": "./lib/sinon.js",
   "module": "./pkg/sinon-esm.js",
   "cdn": "./pkg/sinon.js",


mantoni:
The browser field can be used to override the “main” for browsers. This change point browsers to the same file so I don’t think it makes a difference.

Feiyang1:
webpack checks field in `package.json` using `['browser', 'module', 'main']`, in that order, when targeting browsers. Without the `browser` field, `module` field will be used which is the esm + es6 version, so issues like #1951 will happen.

People can change webpack configuration to make it work, but it's extra burden for them and sometimes is not obvious to people who don't know webpack well. Having `browser` field defined will enable `sinon` to work in the browser environment without additional configuration.

mantoni:
Thank you for clarifying. Now I understand.

fatso83:
@Feiyang1 Why did we not choose the prebuilt file, `./pkg/sinon.js` with this change? That is guaranteed to work, whereas using the `lib/sinon.js` makes producing a working Sinon build for tests up to the ones bundling sinon into their builds.

I don't think we would have issues like https://github.com/sinonjs/sinon/issues/2218#issuecomment-590046624 if we used build package, but I might be forgetting something ...

mantoni:
Well, when I'm running my test cases in a browser, I usually browserify my tests along with Sinon. If the `browser` field would point to a pre-bundled version this would have unwanted effects. For example, if a library is used that is also used by Sinon, it would be bundled multiple times, causing issues with `instanceof` and `Object.isPrototypeOf` checks.

mantoni:
The way I see it, the pre-bundled Sinon packages are for `<script>` tags. The `browser` field is explicitly there for bundlers and therefore shouldn't point to a pre-bundled version.

PR #2063, opened 2019-07-19
add callThroughWithNew method

dustinfarris:
This mimics `callThrough`'s behavior but uses the `new` operator as
would be appropriate for classes.

The added `else if` clause to fake.invoke increases that function's
complexity to 20 which is an eslint error by default.  Added an
exception to increase the maximum allowed complexity.

Let me know if this looks like a welcome candidate and I'll add docs.

### DIFF
--- a/lib/sinon/behavior.js
+++ b/lib/sinon/behavior.js
@@ -8,7 +8,6 @@ var valueToString = require("@sinonjs/commons").valueToString;
 var exportAsyncBehaviors = require("./util/core/export-async-behaviors");
 
 var concat = arrayProto.concat;
-var flat = arrayProto.flat;
 var join = arrayProto.join;
 var reverse = arrayProto.reverse;
 var slice = arrayProto.slice;
@@ -176,8 +175,12 @@ var proto = {
 
             return wrappedMethod.apply(context, args);
         } else if (this.callsThroughWithNew) {
+            // Get the original method (assumed to be a constructor in this case)
             var WrappedClass = this.effectiveWrappedMethod();
-            var F = WrappedClass.bind.apply(WrappedClass, concat([null], flat(args)));
+            // Turn the arguments object into a normal array
+            var argsArray = slice(args);
+            // Call the constructor
+            var F = WrappedClass.bind.apply(WrappedClass, concat([null], argsArray));
             return new F();
         } else if (typeof this.returnValue !== "undefined") {
             return this.returnValue;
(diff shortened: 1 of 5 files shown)


lib/sinon/behavior.js
```diff
@@ -8,6 +8,7 @@ var valueToString = require("@sinonjs/commons").valueToString;
 var exportAsyncBehaviors = require("./util/core/export-async-behaviors");
 
 var concat = arrayProto.concat;
+var flat = arrayProto.flat;
```
fatso83: We are strictly using ES5.1, predating the existence of this method by half a decade. Thanks for keeping a reference to the original, though :-)

lib/sinon/behavior.js
```diff
@@ -171,7 +173,12 @@ var proto = {
             return (this.promiseLibrary || Promise).reject(this.returnValue);
         } else if (this.callsThrough) {
             var wrappedMethod = this.effectiveWrappedMethod();
+
             return wrappedMethod.apply(context, args);
+        } else if (this.callsThroughWithNew) {
+            var WrappedClass = this.effectiveWrappedMethod();
+            var F = WrappedClass.bind.apply(WrappedClass, concat([null], flat(args)));
```
fatso83: Why are you flattening the arguments array? That would change the parameters sent to the class, like 
`new Stub([a,b,c],[1,2,3])` would end up calling `new Original(a,b,c,1,2,3)`, which is not the same at all.

lib/sinon/behavior.js
fatso83: This is why all tests fail in Node 6, 8 and 10 (as you can see by clicking the failed build), as neither has that method in the runtime.

fatso83:
To me, this looks fine. The current API surface is so big I don't know if this overlaps with anything existing, but others could intercept before we merge in that case. Could you please add some docs?

lib/sinon/behavior.js
dustinfarris: Resolved in 8e00c22.  Using slice instead, which is better anyway.

lib/sinon/behavior.js
dustinfarris: The arguments object is treated as a single entity when concatenated with any other array, so it needs to be converted into a normal array.  You're right, though, flat is not the right tool here.  Changing to slice.  Resolved in 8e00c22.

PR #2068, opened 2019-07-23
Dev dependency updates

mantoni:

A little maintenance on dev dependencies related to our test runner setup and referee.

I found the latest puppeteer to be more reliable. At least upgrading in mochify made the tests pass reliably while the previous release was very flaky and had random failures in any node environment.

The latest referee introduces `assert.isUndefined` in favor of `refute.defined` so that it aligns with `assert.isNull` and friends. A deprecation warning was introduced, so I changed the assertions accordingly.

1. Check out this branch
2. `npm ci`
3. `npm test`
4. `npm run test-coverage`
5. `npm run test-cloud`

 #### Checklist for author

- [x] `npm run lint` passes


mantoni:
One thing I noticed: If I try to upgrade browserify, a test fails:

```
  1) fakeTimers.clock
       .useFakeTimers
         fakes provided methods:
      at Object.fail (node_modules/@sinonjs/referee/lib/create-fail.js:5)
      at Object.fail (node_modules/@sinonjs/referee/lib/define-assertion.js:46)
      at assertion (node_modules/@sinonjs/referee/lib/define-assertion.js:68)
      at Function.referee.<computed>.<computed> [as same] (node_modules/@sinonjs/referee/lib/define-assertion.js:93)
      at Context.<anonymous> (test/util/fake-timers-test.js:1075)
```

Maybe somebody with more experience in lolex known what that could be 🤔

test/call-test.js
```diff
@@ -1215,7 +1215,7 @@ describe("sinonSpy.call", function() {
                 sinonSpy(myObj, "ouch");
             });
 
-            refute.defined(myObj.ouch);
```
fearphage: Was this removed/changed or did the tests fail here? I'm just curious.

mantoni: It logged a deprecation message.

Issue #2073, opened 2019-08-07
sinon.stub().resolves(...) returns undefined when used with createStubInstance


clemmy:
**Describe the bug**
This was working in 7.3.2, but broke in 7.4.1.
Basically, when calling `sinon.stub().resolves(...)` as part of the second argument of `sinon.createStubInstance()`, I get `undefined`. It works as a standalone, however.

**To Reproduce**
Run the following test:
```
class X {
  async test() {
    return Promise.resolve(1)
  }
}

it('is a buggy test', async () => {
  const thing = sinon.createStubInstance(X, {
    test: sinon.stub().resolves(2) as sinon.SinonStub<[], Promise<number>>
  })
  console.log(thing.test()) // undefined
}
```

**Expected behavior**
I would expect the `console.log(thing.test())` to output `2`

**Screenshots**
N/A

**Context (please complete the following information):**

* Library version: 7.4.1
* Environment: Mac OSX
* Example URL: n/a
* Other libraries you are using: n/a

**Additional context**
N/A

fatso83:
Not sure what could cause this in the changelog ... If you could run a `git bisect` run to find out which commit introduced this you can use this template: https://github.com/fatso83/git-bisect-scripts/blob/master/full-template.sh

rgroothuijsen:
`git bisect` reveals that `createStubInstance` was removed from the `sinon` API in [this commit](https://github.com/sinonjs/sinon/commit/ff505fc5d070f06d9ba3893cc86c9b3c09cc627b).

joebowbeer:
I can verify that it was broken in 7.4.0 as well

kevinbosman:
Also related to this, due to `createStubInstance` now using the (different) `Sandbox` implementation, it can no longer be called without explicitly specifying the `Sandbox` object binding.

So the following Typescript import no longer works (it worked in 7.3.2):
```typescript
import { createStubInstance } from "sinon";
...
const thing = createStubInstance(X);
```
which results in `TypeError: Cannot read property 'stub' of undefined`

This could be resolved by replacing `this.stub` with `sandbox.stub` here: https://github.com/sinonjs/sinon/blob/671330cebf271b05afea01fdadcc111130b2b4af/lib/sinon/sandbox.js#L58

fatso83:
[...]

PR #2022, opened 2019-05-15
sinon.resetHistory() does not reset history

rpgeeganage:
#### Purpose (TL;DR) - mandatory

Fix issue: https://github.com/sinonjs/sinon/issues/2016 `sinon.resetHistory()` does not reset history

Remove `createStubInstance` from `sinon.js` so same method from `sandbox` will be used.

 #### How to verify - mandatory
No tests added yet.
But followed below code snippet.
```js
const sinon = require('../../my_sinon_fork/sinon/');
const { assert } = require('referee');

describe('resetHistory', function() {
  var num = null;

  beforeEach(function() {
    num = sinon.createStubInstance(Number);
  });

  afterEach(() => {
    // Restore the default sandbox here
    sinon.restore();
  });

  describe('num.toFixed.resetHistory()', function() {
    it('num.toFixed.resetHistory()', function() {
      num.toFixed();
      assert.isTrue(num.toFixed.called);
      num.toFixed.resetHistory();
      assert.isFalse(num.toFixed.called);
    });
  });

  describe('sinon.resetHistory()', function() {
    it('sinon.resetHistory()', function() {
      num.toFixed();
      assert.isTrue(num.toFixed.called);
      sinon.resetHistory();
      assert.isFalse(num.toFixed.called);
    });
  });
});

```
Need to execute as follows.
`mocha test.js`
Got the following output.
```
  resetHistory
    num.toFixed.resetHistory()
      ✓ num.toFixed.resetHistory()
    sinon.resetHistory()
      ✓ sinon.resetHistory()


  2 passing (27ms)
```
 #### Checklist for author

- [x] `npm run lint` passes
- [x] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

PR #2075, opened 2019-08-14
Restore sinon.createStubInstance() behaviour

kevinbosman:
Fix issue #2073 by reusing `createStubInstance()` behaviour from `stub.js` while maintaining the sandbox `collection`.

PR #2022 redirected `sinon.createStubInstance()` to use the Sandbox implementation thereof. This introduced a breaking change due to the sandbox implementation not supporting property overrides.

#### Solution  - optional
This PR extends `sandbox.createStubInstance()` to mimic the behaviour of the original `sinon.createStubInstance()` method.

Instead of duplicating the original `createStubInstance()` behaviour from `stub.js` into `sandbox.js`, it now calls through to the `stub.js` implementation, then adds all the stubs to the sandbox `collection` as usual.

The extra tests for property overrides in `stub-test.js` have also been added to `sandbox-test.js` to ensure the  sandbox implementation now works the same as the original `createStubInstance()`, while maintaining backward compatibility with all existing sandbox tests.

EDIT: Added some specific tests for issue #2073 in `issues-test.js`.

1. Check out this branch
2. `npm install`
3. `npm test`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -51,11 +51,18 @@ function Sandbox() {
         return fakeRestorers;
     };
 
-    sandbox.createStubInstance = function createStubInstance(constructor) {
-        if (typeof constructor !== "function") {
-            throw new TypeError("The constructor should be a function.");
-        }
-        return this.stub(Object.create(constructor.prototype));
+    sandbox.createStubInstance = function createStubInstance() {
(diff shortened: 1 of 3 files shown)


spalger:
This works for me, and seems good to go 😁

mroderick:
This has been published as `sinon@7.4.2` 🚢

PR #1990, opened 2019-03-07
Added Github Actions

fearphage:
Setting up [Github Actions](https://github.com/features/actions).

PR #2076, opened 2019-08-20
Added Github Workflow

fearphage:
Using UI this time since CLI doesn't seem to work.

This replaces #1990.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,34 @@
+name: Node CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [6.x, 8.x, 10.x, 12.x]
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


fatso83:
Eh, I didn't mean to do that (change from draft to ready). #bigthumbs 

Anyway, this _looks_ fine (maybe not run on Node 6). How can we tell if it works before merging?

fearphage:
> Anyway, this _looks_ fine (maybe not run on Node 6). 

Removed.

> How can we tell if it works before merging?

image

One thing of note is it's not running `npm run test-cloud`. I believe it needs some extra setup (secrets) that we can add later.

mroderick:
> One thing of note is it's not running `npm run test-cloud`. I believe it needs some extra setup (secrets) that we can add later.

I don't think we can run that on anything other than `master`, so that shouldn't stop us from moving forwards.

How about adding the coveralls integration (or switching to `codecov` like in `@sinonjs/referee`)?

.github/workflows/main.yml
```diff
@@ -0,0 +1,34 @@
+name: Node CI
+
+on: [push]
+
+jobs:
+  build:
```
mgred: Nice Job @fearphage. Would maybe `test` fit better here?

Issue #2006, opened 2019-04-04
Add Sinon.assert.calledOnceWithExactly 


dpmott:
**Is your feature request related to a problem? Please describe.**
Closely related to https://github.com/sinonjs/sinon/issues/1247, but in this case I'd like to be able to type `Sinon.assert.calledOnceWithExactly(stub, arg1, ... argN)`

**Describe the solution you'd like**
I believe that there just needs to be an entry in `assert.js` to facilitate this.  I'm not entirely sure what the failure message (including interpolated variables) should be though.

Possibly just this addition to assert.js?
`mirrorPropAsAssertion("calledOnceWithExactly", "expected %n to be called once with exact arguments %D");
`

**Describe alternatives you've considered**
The equivalent functionality with today's Sinon is:
```
Sinon.assert.calledOnce(stub);
Sinon.assert.calledWithExactly(stub, arg1, ...argN);
```

If I could get some agreement on what the assertion message should be, then I'd be happy to generate a PR.

brandonau24:
@dpmott This is possible with a [spy](https://sinonjs.org/releases/latest/spies/) assuming you're on the latest version. The beginning of the page says, spies can record arguments which is what you seem to be looking for. [Stubs](https://sinonjs.org/releases/latest/stubs/) are only for pre-programming behavior like forcing a function to return a specific value. Hope this helps!

fatso83:
@brandonau24 I believe you are talking about something else. @dpmott is addressing the feature disparity between the assertions available on the spies and the standalone assertions that can be invoked on spies. Furthermore, everything a spy can do, a stub can do as well: a stub simply extends the spy api. So all the spy assertions are already available on a stub, so you can always use a stub in place of a spy - it's more of a semantical discussion whether or not to explicitly use a spy.

@dpmott I'd try to make the assertion message even clearer (ref 88a7150 and the whole discussion in https://github.com/sinonjs/sinon/issues/1971#issuecomment-496150383).

```diff
+ What about "expected %n to be called once and with the exact arguments %D"?
- What about "expected %n to be called once with exact arguments %D"?
```

dpmott:
Thanks for the input. Will do.

fatso83:
Closed by #2080 !

PR #2080, opened 2019-08-28
Add Sinon.assert.calledOnceWithExactly()

dpmott:
#### Purpose

Fix issue https://github.com/sinonjs/sinon/issues/2006 by implementing `Sinon.assert.calledOnceWithExactly()`

 #### Solution
This solution implements `Sinon.assert.calledOnceWithExactly()`
Note that no other `calledOnce*` methods appear to work with spyCall, so no such tests have been added (e.g. "works with spyCall" or "fails when spyCall failed").  This seems appropriate because the end-user would just use `calledWithExactly` on a spyCall.

 #### How to verify
1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -205,6 +205,7 @@ mirrorPropAsAssertion("calledWithMatch", "expected %n to be called with match %D
 mirrorPropAsAssertion("alwaysCalledWith", "expected %n to always be called with arguments %D");
 mirrorPropAsAssertion("alwaysCalledWithMatch", "expected %n to always be called with match %D");
 mirrorPropAsAssertion("calledWithExactly", "expected %n to be called with exact arguments %D");
+mirrorPropAsAssertion("calledOnceWithExactly", "expected %n to be called once and with exact arguments %D");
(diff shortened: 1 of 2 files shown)


fatso83:
Thanks, David! Great work.

mroderick:
This has been published as `v7.5.0`

rhodgkins:
I'm not getting any autocomplete for this in VS code - I'm using sinon v8.1.1. Is there anything I need to do to pick up this extra signature? Thanks!

mroderick:
> I'm not getting any autocomplete for this in VS code - I'm using sinon v8.1.1. Is there anything I need to do to pick up this extra signature? Thanks!

You may need to configure your VSCode to pick up JavaScript as well as TypeScript: https://medium.com/@trukrs/type-safe-javascript-with-jsdoc-7a2a63209b76

fatso83:
@rhodgkins VS Code (probably) makes use of the `@types/sinon` package, which needs to be in sync (up to date) with Sinon's advances. It probably lacks signatures for this method, but this is not a package we maintain, so I would simply supply a PR to Definitely Typed.

Issue #2085, opened 2019-08-07
Upgrade of Browserify fails tests


fatso83:
* Sinon version : `master` (before updating) or branch `bug-updating-deps`
* Environment   : Chromium

**What did you expect to happen?**
All tests pass in all environments.

**What actually happens**
One test fails.

```
  1360 passing (919ms)
  4 pending
  1 failing

  1) fakeTimers.clock
       .useFakeTimers
         fakes provided methods:
     AssertionError: [refute.same] function onNextTick() {} expected not to be the same object as function onNextTick() {}
      at Object.fail (node_modules/@sinonjs/referee/lib/create-fail.js:5)
      at Object.fail (node_modules/@sinonjs/referee/lib/define-assertion.js:46)
      at assertion (node_modules/@sinonjs/referee/lib/define-assertion.js:68)
      at Function.referee.<computed>.<computed> [as same] (node_modules/@sinonjs/referee/lib/define-assertion.js:93)
      at Context.<anonymous> (test/util/fake-timers-test.js:1075)
```

**How to reproduce**

Check out `master` (671330ce). Run `npm update && npm test`.

Also pushed the updates: https://github.com/sinonjs/sinon/commit/c8ff2b8f6f438ccac7c5c3abd9a106dc84914610

fatso83:
As this bug manifested itself in the fake timers code, I assumed the bug was in Lolex, but alas, it turns out adding the same tests to lolex and upgrading Browserify (=>Chokidar=>fsevents=>process-nextick-args)  does _not_ manifest the bug. The same process-nextick-args version (2.0.0) was there (in Lolex), as well. Which points to this being the little code that is specific to Sinon's own `fake-timers.js`.  My guess in on the `module.exports` code.

<details>
<summary>output_pre_all_passing.txt</summary>

```
# chromium:


  fakeTimers.clock
    .setTimeout
      ✓ throws if no arguments
      ✓ returns numeric id or object with numeric id
      ✓ returns unique id
      ✓ sets timers on instance
      ✓ evals non-function callbacks
      ✓ passes setTimeout parameters
      ✓ calls correct timeout on recursive tick
    .setImmediate
      ✓ returns numeric id or object with numeric id
      ✓ calls the given callback immediately
      ✓ throws if no arguments
      ✓ manages separate timers per clock instance
      ✓ passes extra parameters through to the callback
    .clearImmediate
      ✓ removes immediate callbacks
    .tick
      ✓ triggers immediately without specified delay
[...]

PR #2086, opened 2019-09-24
Add failing tests for chromium immediate timers

mantoni:

Chromium does not implement `setImmediate` and `clearImmediate`.
See https://bugs.chromium.org/p/chromium/issues/detail?id=146172

Sinon should not install these functions when faking the clock.

See discussion in #2085

1. Check out this branch
2. `npm install`
3. `npm run test-node` passes
4. `npm run test-headless` fails

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/test/util/fake-timers-test.js
+++ b/test/util/fake-timers-test.js
@@ -102,56 +102,62 @@ describe("fakeTimers.clock", function() {
             this.clock = fakeTimers.clock.create();
         });
 
-        it("returns numeric id or object with numeric id", function() {
-            var result = this.clock.setImmediate(function() {
-                return;
+        if (typeof setImmediate === "function") {
+            it("returns numeric id or object with numeric id", function() {
+                var result = this.clock.setImmediate(function() {
+                    return;
+                });
+
+                if (typeof result === "object") {
+                    assert.isNumber(result.id);
+                } else {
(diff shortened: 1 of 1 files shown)


test/util/fake-timers-test.js
```diff
@@ -102,72 +102,84 @@ describe("fakeTimers.clock", function() {
             this.clock = fakeTimers.clock.create();
         });
 
-        it("returns numeric id or object with numeric id", function() {
-            var result = this.clock.setImmediate(function() {
-                return;
-            });
+        if (typeof setImmediate === "function") {
```
fatso83: I think we elsewhere in the codebase agreed to use the more Mocka-ish way of using `this.skip()` (`grep -C5 this.skip test/*.js` for hits). Could be used in both `before*()` and in `it()`. A la

```
252-        before(function() {
253-            if (!global.Promise) {
254:                this.skip();
255-            }
256-        });
```

mantoni: Yes, I think we agree on that. We should probably refactor the file to use skip by grouping the corresponding tests in contexts.

However, the rest of this specific file is using if-statements and I aligned with that.

fatso83:
Included in #2131.

PR #2088, opened 2019-09-26
Restore object implementation

### LINKED ISSUE #2087: UX: Easy way to restore after spying on all methods of an object
In #2027 we gained the ability to spy on all methods of an object. However the only way to restore the object is to use a sandbox or visit every method of the object. It would be a convenient and a better user experience if `spyOnEntireObj` returned a way to revert everything back to normal.

fearphage:
Took a quick stab at this one, I probably forgot a lot of corner cases. Feel free to help me flesh this out.

Fixes #2087


test/restore-object.js
```diff
@@ -0,0 +1,85 @@
...
+        assert.isFunction(restoreObject);
+    });
+
+    it("works on empty objects", function() {
+        refute.exception(function() {
+            restoreObject({});
+        });
+    });
+
+    it("quietly ignores falsey input", function() {
```
mroderick:
```suggestion
    it("quietly ignores falsy input", function() {
```

lib/sinon/restore-object.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+
+var getPropertyDescriptor = require("./util/core/get-property-descriptor");
+var walk = require("./util/core/walk");
+
+function restoreObject(object) {
+    walk(object || {}, function(prop, propOwner) {
+        if (
```
mroderick: You can reduce the indentation by using the return early pattern here.

This is very much a matter of style/preference, feel free to ignore.

fearphage: This feels better to me. If there was more than one line of code inside the `if`, I would be more compelled to switch this around. Also this was copied from one of the other files so it follows the same pattern.

fearphage:
> You might want to add some documentation for it

I added a ticket for the docs. #2089

test/restore-object.js
```diff
@@ -0,0 +1,85 @@
...
+var restoreObject = require("../lib/sinon/restore-object.js");
+var assert = referee.assert;
+var refute = referee.refute;
+
+describe("restore-object", function() {
+    it("is defined", function() {
+        assert.isFunction(restoreObject);
+    });
+
+    it("works on empty objects", function() {
```
mantoni: Why should it? Sinon usually throws a meaninful exception in cases where passed arguments are not "Sinon objects". This helps finding issues while writing or refactoring test cases. Think of passing a non-fake object to `sinon.assert.called(xyz)`, for example. In case of this method I would prefer having an exception thrown if nothing is being restored. Other thoughts?

test/restore-object.js
```diff
@@ -0,0 +1,85 @@
...
+        refute.exception(function() {
+            restoreObject({});
+        });
+    });
+
+    it("quietly ignores falsy input", function() {
+        refute.exception(function() {
+            restoreObject(false);
```
mantoni: Same as comment above. This should throw a meaningful exception. The user is doing something dodgy. The argument is clearly not a result of calling `sinon.stub(...)`.

test/restore-object.js
```diff
@@ -0,0 +1,85 @@
...
+            restoreObject(null);
+        });
+    });
+
+    it("works with no spies or stubs", function() {
+        refute.exception(function() {
```
mantoni: Should throw in my opinion. Again, not the result of calling `sinon.stub` being passed.

test/restore-object.js
fearphage: @mantoni This is the way both `spyOnEntireObject` and `stubEntireObject` are defined. Should those change as well?

```shell
➜  sinon git:(restore-object) ✗ ag -Q 'walk(object || {}' lib
lib/sinon/stub-entire-object.js
7:    walk(object || {}, function(prop, propOwner) {

lib/sinon/spy-entire-object.js
7:    walk(object || {}, function(prop, propOwner) {

lib/sinon/restore-object.js
7:    walk(object || {}, function(prop, propOwner) {
```

test/restore-object.js
fearphage: Same as my comment above.

test/restore-object.js
mantoni: I lean towards failing in the new API. We can make it consistent now, or later as a breaking change. I have no specific opinion here. Do @fatso83 and @mroderick have additional arguments?

fearphage: What does failing do/mean? 

From my perspective, the user asks to restore all mocks, spies, and stubs. We return an object with 0 mocks, spies, and stubs.

The error case makes sense for 1-to-1 mappings of "do X to this spy/stub". It feels a bit more ambiguous here though.

fatso83: Break things! Yeah! I always like to land on the strict side of things, so :+1: to not allowing stubbing or restoring `undefined`.

fatso83: But ... we could do it in a separate PR just to not bloat this one.

fearphage: I like the idea of doing them outside of this fix. 

We should make them all congruent. That sounds like a very breaking change.

mantoni: It makes sense to keep the implementations symetric for now add strictness later. I have created #2090 summarizing what was discussed here. 👍

fearphage: I started working on #2090 in a branch(based off of this one).

fearphage:
@mroderick If you're happy with this, I'll merge and move on to making this stricter in the next PR.

fearphage:
I'm addressing issues addressed here in #2092.

PR #2092, opened 2019-09-28
Stricter object iteration

### LINKED ISSUE #2090: Make spying, stubbing and restoring entire objects more strict
As discussed in #2088, spying and stubbing entire objects should be more strict. The motivation is to fail early with meaningful exceptions to highlight superfluous or unintended invocations.

These cases should throw exceptions:

- falsy values being passed, currently ignored
- spying and stubbing should fail if there are no functions to replace on the object
- `sinon.restoreObject` should fail if there are no function to restore on the object

Since this might be breaking for some users, it has to land in a major release.

The relevant code can be found in these files:

- lib/sinon/stub-entire-object.js
- lib/sinon/spy-entire-object.js
- lib/sinon/restore-object.js

fearphage:
This fixes #2090 by making object iteration more strict an less error-prone.


fearphage:
@fatso83 @mantoni I pushed the last commit to show you the [15 failing tests](https://travis-ci.org/sinonjs/sinon/jobs/590905762#L269) I would need to delete/update to make this stricter change. Just making sure we're fine with those changes before I move forward.

mantoni:
I like it 👍

fearphage:
This needs a major revision bump I believe. If you had `sinon.spy({})` in your code previously, it would have silently done nothing. Now it will throw an error.

mantoni:
Yes, I agree. The linked issue already had the semver:major tag.

fatso83:
Seems this causes some IE11 issues https://circleci.com/gh/sinonjs/sinon/2203 :-/

fatso83:
@fearphage Morgan created #2134 after this broke `master`.

PR #2122, opened 2019-10-04
Add runnable example for Spies documentation - Resolves #2040

### LINKED ISSUE #2040: Add a "PubSub" description to document
For example, "spies" does not describe "PubSub" in document.
Please add a description

adhyapranata:

Add runnable example with Runkit for Spies documentation for issue #2040

1. `sh ./scripts/copy-documentation-for-new-release.sh {version}`
2. `npm run serve-docs`
3. open the spies doc page (http://127.0.0.1:4000/releases/v{version}/spies/)
4. when done, git reset


fatso83:
I worked so long [for the code in this comment](https://github.com/sinonjs/sinon/issues/2040#issuecomment-538442805), so I didn't see you posted this in the meantime. 

We probably don't want to merge a demo page to our docs, but I can test this out to see how it works and you can perhaps just update the PR based on what I wrote?

fatso83:
This worked really well! Would you try and "go all the way" using the example code I posted? I think it's pretty close to runnable. Maybe some syntax issues, since I wrote it from the top of my head in vim without testing.

adhyapranata:
I wasn't sure either when making this PR because this is just a demo. I'll have a look at the comment and update this PR accordingly

adhyapranata:
I have updated all of the static examples to Runkit. All examples are running just fine except for the second one (`spies-2-wrap-object-methods`). I'm not quite sure what's wrong.

I also make some minor adjustments from the comment:
- use .stub extension instead of js to skip linting
- use `scriptTemplate.innerHTML` instead of `scriptTemplate.content` to get the text (not the node)

docs/_includes/head.html
```diff
@@ -9,6 +9,7 @@
     <script>
         var site = {}; // site settings
     </script>
+    <script src="https://embed.runkit.com"></script>
```
fatso83:
```suggestion
```
I understand you probably moved this here so that the function in `embed-scripts.html` above wouldn't crash due to a undefined `RunKit` variable, but this will block rendering, so we should try to find a way of keeping this in the footer.

docs/release-source/release/spies.md
```diff
@@ -470,3 +385,5 @@ Returns the passed format string with the following replacements performed:
 </dl>
 
 [call]: ../spy-call
+
+{% include embed-scripts.html %}
```
fatso83: By moving this into the `footer.html`, after with the script to load runkit, you avoid blocking rendering of the content above and avoid crashing at the same time.
```suggestion
```

docs/_includes/footer.html
```diff
@@ -1,5 +1,4 @@
 <script src="{{ "/assets/js/prism.js" | prepend: site.baseurl }}"></script>
-<script src="https://embed.runkit.com" data-element-id="spy-example" data-theme="untilted-c3qcs6lijfuc"></script>
```
fatso83: Just put the script at the end of this file, along with the include
```
    <script src="https://embed.runkit.com"></script>
```

adhyapranata:
Done! Render blocking JavaScript and the embed scripts have been moved to the footer

fatso83:
Thank you! Great job.

PR #2125, opened 2019-10-05
refactor: Move behavior setup from spy to stub

mantoni:

The behavior setup code is not required for spies. It should be done in `stub.js` where it's actually being used.

1. Check out this branch
2. `npm install`
3. `npm t`
4. Observe that all tests still pass.

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -196,6 +196,15 @@ var proto = {
 
     onThirdCall: function onThirdCall() {
         return this.onCall(2);
+    },
+
+    withArgs: function withArgs() {
+        var fake = spy.withArgs.apply(this, arguments);
+        if (this.defaultBehavior && this.defaultBehavior.promiseLibrary) {
+            fake.defaultBehavior = fake.defaultBehavior || behavior.create(fake);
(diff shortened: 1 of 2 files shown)


lib/sinon/stub.js
```diff
@@ -196,6 +196,15 @@ var proto = {
 
     onThirdCall: function onThirdCall() {
         return this.onCall(2);
+    },
+
+    withArgs: function withArgs() {
+        var fake = spy.withArgs.apply(this, arguments);
```
mroderick: Since this is being moved to another file, we have the opportunity to also rename the variable from `fake` (which can easily be confused with `fake`) into something that is unambiguous.

Perhaps `spy` or `instance`?

mantoni: Hmmm, the `spy.withArgs` implementation calls it a `fake` and there is the `this.fakes` array where the newly created `fake` is being kept. I think it should either have this name or we should remove the ambiguity from the codebase by renaming it everywhere in `spy.js` and stub.js`.

mroderick: Fair enough, we can rename them all in a later PR ... all at the same time

mroderick: Go ahead and merge this

PR #2130, opened 2019-10-06
Retain stubbed function name and make additional properties non-enumerable

mantoni:

This aligns stubbed functions on objects with the behavior of spies where the function name is retained and the additional properties added by Sinon are made non-enumerable.

Properties on stubs where added in a way so that they where returned by `Object.keys(stub)`. Also, the function name wasn't retained.

1. Check out this branch
2. `npm install`
3. `npm t`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/spy.js
+++ b/lib/sinon/spy.js
@@ -26,7 +26,7 @@ var bind = Function.prototype.bind;
 var callId = 0;
 var uuid = 0;
 
-function createSpy(func, arity) {
+function createSpy(func) {
     var name;
     var funk = func;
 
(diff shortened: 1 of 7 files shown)


lib/sinon/spy.js
```diff
@@ -38,7 +38,7 @@ function createSpy(func, arity) {
         name = functionName(funk);
     }
 
-    var proxy = createProxy(funk, arity);
+    var proxy = createProxy(funk, funk);
```
mroderick: This reads like a mistake ... is it though?

mantoni: No, it's on purpose. `createProxy` now takes two function arguments, the fake and the original function. In case of a spy there is no difference between the two.

mroderick: 👍

mroderick:
Go ahead and merge it 🚀

PR #2131, opened 2019-10-07
Avoid installing setimmediate

### LINKED ISSUE #2085: Upgrade of Browserify fails tests

fatso83:
This was an attempt at fixing #2085 (it doesn't), based on a suggestion in #2086.
Includes additional tests from #2086 by @mantoni (just rebased).

Even though it doesn't fix #2085, I guess it's still an improvement ... 

See #2085, #2086, sinonjs/lolex#268, sinonjs/lolex#269

 #### Solution  - optional
An underlying issue is fixed in Lolex: it tries to avoid the issues in #2085 simply by not installing setImmediate when the environment doesn't support it. Unfortunately it doesn't fix #2085, as it seems as this is actually caused by a browserify transform: https://github.com/browserify/insert-module-globals/releases/tag/v7.1.0. 

1. Check out this branch
2. `npm install`
3. `npm t`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/test/util/fake-timers-test.js
+++ b/test/util/fake-timers-test.js
@@ -102,56 +102,62 @@ describe("fakeTimers.clock", function() {
             this.clock = fakeTimers.clock.create();
         });
 
-        it("returns numeric id or object with numeric id", function() {
-            var result = this.clock.setImmediate(function() {
-                return;
+        if (typeof setImmediate === "function") {
+            it("returns numeric id or object with numeric id", function() {
+                var result = this.clock.setImmediate(function() {
+                    return;
+                });
+
+                if (typeof result === "object") {
+                    assert.isNumber(result.id);
+                } else {
+                    assert.isNumber(result);
+                }
             });
 
-            if (typeof result === "object") {
-                assert.isNumber(result.id);
-            } else {
-                assert.isNumber(result);
-            }
-        });
+            it("calls the given callback immediately", function() {
+                var stub = sinonStub();
 
-        it("calls the given callback immediately", function() {
-            var stub = sinonStub();
+                this.clock.setImmediate(stub);
+                this.clock.tick(0);
 
-            this.clock.setImmediate(stub);
-            this.clock.tick(0);
-
-            assert(stub.called);
-        });
-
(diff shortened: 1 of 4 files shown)


test/util/fake-timers-test.js
```diff
@@ -102,72 +103,84 @@ describe("fakeTimers.clock", function() {
             this.clock = fakeTimers.clock.create();
         });
 
-        it("returns numeric id or object with numeric id", function() {
-            var result = this.clock.setImmediate(function() {
-                return;
-            });
+        if (typeof setImmediate === "function") {
```
fearphage: I wish we had a rule to detect this. This should be using `this.skip()` in the `before` instead of this.

mroderick: It shouldn't be that difficult to write an ESLint rule that detects when an `if` statement wraps an `it` invocation

fatso83: We had our reasons 🤷‍♂  https://github.com/sinonjs/sinon/pull/2086#discussion_r327539951

PR #2144, opened 2019-10-18
Separate fake from spy

mantoni:

Another internal refactoring, finally separating the `fake` implementation from `spy` by extracting the common logic into separate files.

This is done for two reasons:

1. It was a little hacky to derive `fake` from `spy` and then remove unwanted properties.
2. I would like to modularize Sinon so that I can use the fakes standalone. It's currently confusion to introduce all of Sinon into a new project with all these very similar APIs.

1. Check out this branch
2. `npm install`
3. `npm t`

 #### Checklist for author

- [x] `npm run lint` passes


lib/sinon/fake-api.js
```diff
@@ -0,0 +1,184 @@
...
+var forEach = arrayProto.forEach;
+var concat = arrayProto.concat;
+var ErrorConstructor = Error.prototype.constructor;
+var bind = Function.prototype.bind;
+var filter = Array.prototype.filter;
+
+var callId = 0;
+
+// Public API
+var fakeApi = {
```
mroderick: Since this is used by both `spy` and `fake` and there's a name overlap with `fake`, perhaps now is the time to rename this to something different?

`commonApi`, `sharedApi`, `your-suggestion-here`...

mantoni: Good point. I'm not sure. We've been using the term `fake` as the generic term for `spy` and `stub` before, e.g. `instantiateFake`.

mroderick: In this particular case, none of the names including `fake` is part of the public api. Is that correct?

mantoni: I think so, yes.

mantoni: Thinking a bit more about this releaved even more cleanup that I'm about to add.

lib/sinon/spy.js
```diff
@@ -362,97 +198,40 @@ var spyApi = {
...
 });
-delegateToCalls("yieldToOn", false, "yieldToOn", true, function(property) {
+delegateToCalls(spyApi, "yieldToOn", false, "yieldToOn", true, function(property) {
     throw new Error(
         this.toString() + " cannot yield to '" + valueToString(property) + "' since it was not yet invoked."
     );
 });
 /* eslint-enable local-rules/no-prototype-methods */
 
+extend(spy, fakeApi);
```
mantoni: We're adding all the functions to the exported API here just for `stub.js` to have access to it. Now that the `fakeAPI` was moved into `proxy.js` and is applied there to each proxy instance, there's no need to do this anymore.

lib/sinon/fake-api.js
mroderick: Excellent!

lib/sinon/proxy.js
```diff
@@ -12,9 +119,7 @@ function createProxy(func, originalFunc) {
 
     proxy.prototype = func.prototype;
 
-    extend.nonEnum(proxy, {
-        toString: functionToString
```
mantoni: Moved into `proxyApi`.

lib/sinon/stub.js
```diff
@@ -164,8 +164,6 @@ var proto = {
         });
     },
 
-    resetHistory: spy.resetHistory,
```
mantoni: This is now inherited from `proxy.js`, the `spy` module does not expose `resetHistory` anymore, so this would assign `undefined`.

mantoni:
We've been using the exposed `spy` function to transport the public API to `stub`. In my first attempt here I replicated this pattern in `proxy.js`. However, it turns out that this is not needed and these functions are not part of the public API anyway. I've internalized as much possible into the individual modules now.

mantoni:
I found an issue with this refactoring that is currently not covered by tests. Investigating, please do not merge yet.

mantoni:
@mroderick The call order functions where missing in `fake` and there wheren't any tests for them. I'm now also "porting" `.named(...)`. There used to be `matches(args, strict)` and `printf(format)` that where implicitly inherited by fakes. I don't even recall the use case for them.

mantoni:
Here's an interesting case about `matches`:

For spies it's needed to figure out which "sub-fake instances" are matching the given arguments. These instances are created when we're calling `.withArgs` on the spy. Makes sense, however, I don't think this function should be exposed (it's also not documented).

But here's the thing: The new proxy API needs to implement `matchingFakes`, or else `invoke` fails because it wants to find all fakes matching the call. In case of a `fake` this will always return an empty array, because we create the internal `fakes` array on each instance, but it's never populated because `fake.withArgs` doesn't exist.

I'll clean this up as well now.

lib/sinon.js
```diff
@@ -15,7 +15,7 @@ var apiMethods = {
     assert: require("./sinon/assert"),
     match: require("@sinonjs/samsam").createMatcher,
     restoreObject: require("./sinon/restore-object"),
-    spyCall: require("./sinon/call"),
+    spyCall: require("./sinon/proxy-call"),
```
mantoni: We should probably remove this. It's very old and I don't know why it is exposed. Maybe along with the deprecated `sinon.sandbox.create` in the next major.

mantoni:
I went over the namings once more and made them more consistent.

It wasn't possible to easily remove the `fakes` entirely from the new `proxy` base implementation. That would be an even larger refactoring. I hope the intention is now a bit clearer.

The `matches` function is now internalized and not exposed on the spy API anymore.

Ready for review.

lib/sinon.js
mroderick: Good idea

fatso83:
I can't review this at the time. I trust you guys.

mantoni:
I'm really glad this is done now. The entanglement between the implementations was disturbing me for quite a while.

PR #2150, opened 2019-10-30
Add firstArg to spy calls and fakes.

GreeKatrina:
### Purpose - Add `firstArg` convenience property to fakes and spy calls.
Currently, we allow getting the last argument for spy calls and fakes. It would be nice to be able to get the first arg without doing `args[0]`. I couldn't find any open issues or feature requests for this.

#### `spyCall.firstArg`

This property is a convenience for the first argument of the call.

```js
var spy = sinon.spy();
var date = new Date();

spy(date, 1, 2);

spy.lastCall.firstArg === date;
// true
```
#### `fake.firstArg`

This property is a convenient way to get a reference to the first argument passed in the last call to the fake.

```js
var f = sinon.fake();
var date1 = new Date();
var date2 = new Date();

f(date1, 1, 2);
f(date2, 1, 2);

f.firstArg === date2;
// true
```

 #### How to verify
1. Check out this branch
2. `npm install`
3. `npm test`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/proxy.js
+++ b/lib/sinon/proxy.js
@@ -123,6 +123,7 @@ var proxyApi = {
         this.thirdCall = null;
         this.lastCall = null;
         this.args = [];
+        this.firstArg = null;
         this.lastArg = null;
         this.returnValues = [];
         this.thisValues = [];
@@ -272,6 +273,7 @@ function wrapFunction(func, originalFunc) {
         calledThrice: false,
         callCount: 0,
         firstCall: null,
+        firstArg: null,
         secondCall: null,
         thirdCall: null,
         lastCall: null,
--- a/test/proxy-call-test.js
+++ b/test/proxy-call-test.js
@@ -553,6 +553,24 @@ describe("sinonSpy.call", function() {
         });
     });
 
+    describe(".firstArg", function() {
+        it("should be the first argument from the call", function() {
+            var spy = sinonSpy();
+
+            spy(41, 42, 43);
+            assert.equals(spy.getCall(0).firstArg, 41);
+
+            spy(44, 45);
+            assert.equals(spy.getCall(1).firstArg, 44);
+
+            spy(46);
+            assert.equals(spy.getCall(2).firstArg, 46);
+
+            spy();
+            assert.equals(spy.getCall(3).firstArg, undefined);
+        });
+    });
+
     describe(".lastArg", function() {
         it("should be the last argument from the call", function() {
             var spy = sinonSpy();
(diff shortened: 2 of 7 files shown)


GreeKatrina:
@mantoni Ah, gotcha. That makes sense. There are a lot of files in that folder so I probably missed it haha. I'll update it and let you know.

test/proxy-call-test.js
```diff
@@ -553,6 +553,24 @@ describe("sinonSpy.call", function() {
...
+            assert.equals(spy.getCall(0).firstArg, 41);
+
+            spy(44, 45);
+            assert.equals(spy.getCall(1).firstArg, 44);
+
+            spy(46);
+            assert.equals(spy.getCall(2).firstArg, 46);
+
+            spy();
+            assert.equals(spy.getCall(3).firstArg, undefined);
```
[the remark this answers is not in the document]
GreeKatrina: I just copied the tests for `lastArg`. I can change it though if you'd like.

lib/sinon/proxy.js
```diff
@@ -123,6 +123,7 @@ var proxyApi = {
         this.thirdCall = null;
         this.lastCall = null;
         this.args = [];
+        this.firstArg = null;
```
fearphage: It's a bit weird/inconsistent that the values start out as `undefined`, but become `null` after being `reset()`.

Not sure this needs to be fixed, but just making note of it.

GreeKatrina: I noticed this too. I was going to change it, but didn't want to assume whether the values should be `undefined` or `null`. (My vote is `undefined`.)

mroderick: I agree with the idea of returning to `undefined`.  Do you want to make the change, before we merge this in?

GreeKatrina: Sure. Did you just want me to change this line to `undefined`? I feel like it'd be weird to have `firstArg` and `lastArg` default to different values, but changing `lastArg`'s default value is a big change. Maybe we could do that in a separate PR and just change `firstArg` to default to `undefined` for now?

mroderick: I agree. Let's leave `lastArg` unchanged in this PR.

A separate PR to tidy up the initial and reset values would be most welcome, and will trigger a new major release.

fearphage:
> Nice work. The minor suggestions to change the assertions shouldn't block merging.

Should we merge this and fix the minor assertion issues in another PR?

mroderick:
This has been published as part of `sinon@9.0.0`

PR #2151, opened 2019-10-31
Improve sandbox documentation

### LINKED ISSUE #1429: Incomplete and incorrect sandbox documentation
When researching sinonjs/sinon-test#57 I found that the [documentation for sandboxes](http://sinonjs.org/releases/v2.3.1/sandbox/) was incomplete and somewhat misleading.

### Incorrect default config
It says the sandbox is configured with the following config per default, which is incorrect:
```javascript
sinon.defaultConfig = {
    // ...
    injectInto: null,
    properties: ["spy", "stub", "mock", "clock", "server", "requests"],
    useFakeTimers: true,
    useFakeServer: true
}
```
I do not think there is something called `sinon.defaultConfig` and a quick test will show that there is no server configured per default:
```javascript
        var sandbox = sinonSandbox.create();
        console.log(sandbox.server) // => undefined
```

### References removed fields
`sinon.defaultConfig` was removed in Sinon 2. Is still used in the docs.

### Incomplete list of properties
There is no list of properties one can expose on the sandbox, but a list is given:
```javascript
properties: ["spy", "stub", "mock", "clock", "server", "requests"]
```
One important field that is missing is `sandbox` which will expose the actual sandbox object, along with all of its methods. This is for instance useful when using `sinon-test` and you want to invoke `sandbox.reset()`.

### No example of how to use `inject()` or the config `injectInto`
The code and [set of tests](https://github.com/sinonjs/sinon/blob/c89fb21b00cd4f3d99fdeb48d2c5193f96d5e9fb/test/sandbox-test.js#L258) were pretty useful. No other examples exist in the docs. Not even mention of `inject()`

### Incomplete API description
`inject()` is missing from the docs

* Sinon version : 2.3.1

rodrigo-garcia-leon:
Adds a list of sandbox properties that can be exposed, and an example on how to do it.

Fixes #1429

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -118,6 +118,19 @@ What properties to inject. Note that simply naming "server" here is not
 sufficient to have a `server` property show up in the target object, you also
 have to set `useFakeServer` to `true`.
 
+The list of properties that can be injected are the ones exposed by the object
+returned by the function `inject`, namely:
+
+```javascript
+{
+    //...
+    properties: [
+        "spy", "stub", "mock", "createStubInstance", "fake", "replace",
+        "replaceSetter", "replaceGetter", "clock", "server", "requests", "match"
+    ]
+}
+```
+
 ##### useFakeTimers
 
 If set to `true`, the sandbox will have a `clock` property. You can optionally pass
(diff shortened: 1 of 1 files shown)


mantoni:
Thank you 👍

Issue #2062, opened 2019-07-19
Improve documentation on Stubs


michaelisvy:
Hi,
first of all I'd like to thank you for the great work on Sinon. I've been using it for a few weeks and it helps us a lot having better tests.
I would be happy to work on this issue if you're interested.

**Is your feature request related to a problem? Please describe.**
I was discussing with my colleagues about the fact that the `sinon` documentation is often not enough and that we need to search on the web / go on stackoverflow to get more examples before we can understand things.

**Describe the solution you'd like**
Have better examples for the [stub documentation page](https://sinonjs.org/releases/v7.3.2/stubs/). Add some examples that would be less theoretical, closer to what we use on a real project.
For instance, the example on stubs uses Spies (which will make people confused if they don't know what a Spy is). 

I would be happy to draft something out if you are interested.

**Describe alternatives you've considered**
NA

**Additional context**
NA

fatso83:
Improvements to our docs are welcome. We have been discussing a different take on the API docs as well, more in line with how the Lodash docs work, but that's a huge undertaking not likely to happen anytime soon. Working beats perfect :-)

For extensive tutorial-like examples, we do have the [how-to section](https://sinonjs.org/how-to/), which have been in need of expansion for some time.

michaelisvy:
thanks @fatso83 for the quick feedback. I see the example in the `How to Section` for stubs is pretty good. If that's ok for you, we will start by proposing a small contribution so we get used to the process (and you can see if that's suitable to your expectations). 
I'll be working on it with my friend and colleague @lootster

michaelisvy:
hi, apology on the delay. I was playing with Sinon and will try to propose some changes by Friday night.
Quick question: all the things I've tried with Sinon stubs work well, except in one case:
When using Sinon together with Jasmine, I understand that `sinon.restore()` should always be called before the `expect()` methods. Otherwise a failure to meet the `expect()` close will cause the next tests to fail.
However, it doesn't work when my stub throws an Error. I need to do `expect()` first and `restore()` after. 
 This is shown in the code samples I've created here: https://github.com/michaelisvy/javascript-samples/blob/master/src/test-doubles-sinon/stub/secret-number/secretSentence.spec.js

Would you know if I'm missing something? 

Great thanks,
Michael.

michaelisvy:
[...]

PR #2152, opened 2019-11-05
improve-stubs-doc.md

andrewdroz:
https://github.com/sinonjs/sinon/issues/2062

Improvement to stubs doc

- Focus more on the basics of stubbing an object's function (stub pre-programmed to throw exception/return value)
- Less discussion on spies
- Provide very simple code to test as an example, instead of external library's (PubSub) code
- Show why restoration of a stub is necessary

N.A. because no code change

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -2,6 +2,23 @@
 layout: page
 title: Stubs - Sinon.JS
 breadcrumb: stubs
+examples:
+- stubs-1-pubsub
+- stubs-2-different-args
+- stubs-3-sequential-interactions
+- stubs-4-sequential-with-args
+- stubs-5-reset-behaviour
+- stubs-6-reset-history
+- stubs-7-call-fake
+- stubs-8-call-through
+- stubs-9-call-through-with-new
+- stubs-10-use-promise-library
+- stubs-12-yield-to
+- stubs-14-add-behavior
+- stubs-15-replace-getter
+- stubs-16-define-new-setter
+- stubs-17-define-new-value
+- stubs-18-restore-values
 ---
 
 ### What are stubs?
@@ -24,24 +41,8 @@ Use a stub when you want to:
 
 The following example is yet another test from [PubSubJS][pubsubjs] which shows how to create an anonymous stub that throws an exception when called.
 
-```javascript
-"test should call all subscribers, even if there are exceptions" : function(){
-    var message = 'an example message';
-    var stub = sinon.stub().throws();
-    var spy1 = sinon.spy();
-    var spy2 = sinon.spy();
+<div data-example-id="stubs-1-pubsub"></div>
 
-    PubSub.subscribe(message, stub);
-    PubSub.subscribe(message, spy1);
-    PubSub.subscribe(message, spy2);
-
-    PubSub.publishSync(message, undefined);
-
-    assert(spy1.called);
-    assert(spy2.called);
-    assert(stub.calledBefore(spy1));
-}
-```
 
 Note how the stub also implements the spy interface. The test verifies that all
 callbacks were called, and also that the exception throwing stub was called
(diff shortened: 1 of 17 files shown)


docs/release-source/release/stubs.md
```diff
@@ -16,36 +16,95 @@ wrapping an existing function with a stub, the original function is not called.
...
+    it('should handle EmailClientError when emailService throws', function() {
+
+        // SETUP
+        let sendStub = sinon.stub(emailService, 'send').throws('EmailClientError');
+        // EXERCISE
+        let result = orderService.confirm();
+        // VERIFY
+        expect(result).toBe("Error occurred in email client!");
+        // TEARDOWN
+        sendStub.restore();
```
fatso83: As tests have a tendency to fail, it's better to have restore steps that will run irrespective of the failure status. This is better left to a `afterEach` hook.

docs/release-source/release/stubs.md
```diff
@@ -16,36 +16,95 @@ wrapping an existing function with a stub, the original function is not called.
...
+```
+
+The 2 tests above pass when run in isolation. However, when these tests are run in a single test suite, the second test would fail because the function was already stubbed. In order to change the stub for the same function, the stub needs to be restored:
+
+```javascript
+describe('orderService', function() {
+    it('should handle EmailClientError when emailService throws', function() {
+
+        // SETUP
+        let sendStub = sinon.stub(emailService, 'send').throws('EmailClientError');
```
fatso83: Avoid `let` and other ES2015+ syntax in the examples. We try to have all produced be ES5.1 compatible.

fatso83:
Is there something I can do to get the train back on its tracks?

andrewdroz:
Hello, sorry for not giving a reply thus far. Okay, I understand your points, and probably there's no need to remove the use of PubSub (I didn't really know what PubSub was about). I plan try out implementing runnable code this weekend, so I'll keep you posted.

fatso83:
One shouldn't really need to know anything about the implementation of pubsub either. It's just an example of a very common pattern used practically everywhere: Publish/Subscribe (which is [more or less the same as the Observer Pattern](https://hackernoon.com/observer-vs-pub-sub-pattern-50d3b27f838c)). As it usually just deals with two methods; `publish({topic,msg})` and `subscribe(topic)` the concept is usually known to most, which is why I assume it was used to begin with.

andrewdroz:
I have replaced almost all the examples to be runnable in my latest change. There are 2 examples (yieldsTo, callArg) for which I had trouble making tests that would pass, perhaps due to me not fully understanding the two functions.

fatso83:
Wow, this seems great. I will need to check it out to understand what (if anything) is wrong with the 2 examples you mention, but in any case, this is a huge improvement.

fatso83:
Ooh, bad me. I have forgotten about this ...

fatso83:
I have looked into this and there was a test I could not get passing, `spies-2-wrap-object-methods.stub`. It always failed due to the sandbox not replacing all the fields with spies. This seems like a bug within Sinon, somehow, as changing the beforeEach to this made it pass:

```javascript
           sandbox.spy(jQuery, 'getJSON');
           sandbox.spy(jQuery, 'ajax');
           //sandbox.spy(jQuery);
```

I will look some more into it.

fatso83:
This is not abandoned, btw. I am actively working on a fork of this branch to get everything working, but I am struggling with [a last test](https://github.com/fatso83/sinon/blob/037872dbe31a6d0a9d3b44eca6b6aac94e81c325/docs/release-source/release/examples/stubs-1-pubsub.stub), as `pubsub` is throwing exceptions after the event loop has finished the current work (the test). So we are just talking amongst us about how to rewrite this to not cause Node (or Mocha) to exit with a non-zero exit code.

fatso83:
Included this in #2152 (leaving your commit intact!), so when/if that gets merged, then so will this :)

fatso83:
Thanks for all the work, @andrewdroz! I had some additions as you can see in the PR that built upon this, but this should be live as soon as the next release is published.

Issue #2155, opened 2019-11-07
fake timers don't work with vue (sinon 7.5.0, @vue/cli 3.12.0)


AyanamiSan:
**Describe the bug**
sinon.useFakeTimers does not affect new Date() when used with vue-cli webpack setup.

**To Reproduce**
* checkout https://github.com/AyanamiSan/sinon-fake-timers-test-vuecli
* `npm run test:unit`

first commit is clean vue cli setup, second is my minimal reproducer.

**Expected behavior**
Test passes - no matter how much real time passes, `new Date().getTime()` always returns the same value

**Screenshots**
n/a
**Context (please complete the following information):**

* Library version: 7.5.0
* Environment: tested on linux, node v10.10.0 
* Example URL: https://github.com/AyanamiSan/sinon-fake-timers-test-vuecli
* Other libraries you are using: setup from @vue/cli 3.12.0 with babel, linter, mocha

**Additional context**
vue cli version itself is outdated in this setup, but I also briefly tried with newest vue cli, with the same results.
Depending on the machine it might be needed to tweak number of iterations in my reproducer.

fatso83:
Of course, Sinon works fine. It's your setup that doesn't. Now I took 20 minutes to dig into this for you, as I assume there are plenty of people with some combination of JSDOM and Webpack issues that will stumble upon this.

Your setup has a lot of "magic", using a lot of implicit dependencies. The first thing I was thinking about was that you probably had some kind of DOM-thing going on that would shadow the original Date function in some way or mess up what environment Sinon thinks it's running in. I was right and I found out like this, by just printing out some values from your test. I replaced the contents of your test with the following:

```
    console.log(typeof global);
    console.log(typeof window);
    console.log("window = global", window === global);
    console.log("window.Date = global.Date", window.Date === global.Date);

    const orgWinDate = window.Date;
    const orgGlobalDate = global.Date;

    const clock = sinon.useFakeTimers();

    console.log("window");
    console.dir(window);
    console.log("global");
    console.dir(global);

    console.log("orgWinDate replaced?", orgWinDate !== window.Date);
    console.log("orgGlobalDate replaced?", orgGlobalDate !== global.Date);
```

The important takeaway are the last two lines, which prints
```
orgWinDate replaced? true
orgGlobalDate replaced? false
```

OK, so I know there are two `Date` objects, one for `window` and one for `global`. Originally they point to the same object. This is what your implicit dependency `jsdom-global` sets up for you. 

[...]

PR #2156, opened 2019-11-07
Bring `config.target` into the light

fatso83:
Remove unneeded branch logic and improve docs.

When looking into a user question (#2155) it became obvious that the
it's not obvious at all that the docs for Lolex install config can be
used when trying to configure `useFakeTimers`.

For one, we added a prop, `config.global`, to Sinon's `useFakeTimers` config in #1935 that
was quite superfluous: it supported a use case that was already there using `config.target`!

In this PR, I remove some extra branching logic and also update the docs
to highlight the lolex docs and the `target` prop.

1. `npm test`
1. `npm run build-docs`

### DIFF
--- a/docs/release-source/release/fake-timers.md
+++ b/docs/release-source/release/fake-timers.md
@@ -60,6 +60,7 @@ You can also pass in a Date object, and its `getTime()` will be used for the sta
 As above, but allows further configuration options, some of which are:
 
 - `config.now` - *Number/Date* - installs lolex with the specified unix epoch (default: 0)
+- `config.target` - *Object* - use `target` instead of the usual global object. This is useful if you use JSDOM along with Node.
 - `config.toFake` - *String[ ]* - an array with explicit function names to fake. By default lolex will automatically fake all methods *except* `process.nextTick`. You could, however, still fake `nextTick` by providing it explicitly
 - `config.shouldAdvanceTime` - *Boolean* - tells lolex to increment mocked time automatically based on the real system time shift (default: false)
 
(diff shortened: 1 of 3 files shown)


fatso83:
Basically fails due to sinonjs/lolex#276

fatso83:
@mroderick This kind of stalled because I don't know how to act around @SimenB's comment in Lolex, as this turned out to have a little catch: https://github.com/sinonjs/lolex/issues/276#issuecomment-557020270: Is there a reason one might want to use the timers from a second environment and "pass them" into a third? That's basically the use case the current behavior supports that simplifying this in Lolex would remove.

SimenB:
FWIW, I don't think it's necessarily a valid use case. Maybe you want a working `queueMicrotask` into some other env that doesn't support it, but that's essentially polyfilling and not Lolex's responsibility at all

mroderick:
@fatso83 @SimenB what do you think we should do with this PR?

SimenB:
I think that for now this can land as is, and if anything changes in lolex we can revisit. But the current implementation matches what's documented in this PR

fatso83:
Sorry for keeping this hanging. Will redo once https://github.com/sinonjs/fake-timers/pull/318 is merged.

fatso83:
How would we _like_ the API to look here? Keep the `config.target` (which is now proposed to be removed in the fake timers project) or expose `withGlobal` somehow? :confused: 

Might be worth considering that [nise could also use such a thing](https://github.com/sinonjs/nise/issues/149).

mroderick:
@fatso83 I think there's nothing left blocking this ... is there?

fatso83:
As sinonjs/fake-timers#318 removed `config.target` (my doing), this PR no longer makes sense. What makes sense is document the `global` property introduced in #1935 and highlight that in our docs (missing) to avoid stuff like https://github.com/sinonjs/sinon/issues/2155#issuecomment-551301490.

PR #2159, opened 2019-11-20
Update Lolex (and Nise) for new async timer methods

fatso83:
Updates Lolex for new async methods made by @kylefleming
(async versions of all timer-executing calls).

See sinonjs/lolex#105 (repackaged as sinonjs/lolex#237)

> The goal of these functions is to provide those
> writing tests with tools to test code that uses
> promises without having to work around the fact
> that lolex executes all timers synchronously in a row.

1. `npm install && npm test`
1. Try to use some of the new methods from Lolex

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3947,9 +3947,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.1.2",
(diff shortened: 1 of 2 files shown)


mroderick:
I guess this will go into the next `semver:major`, which will be a very nice upgrade 👍

fatso83:
It's not breaking, though. Just wanted to signal that :)

PR #2171, opened 2019-12-04
update source docs to document spying on accessors

### LINKED ISSUE #2160: current state of setter/getter spies
7.2.3:
https://sinonjs.org/releases/v7.2.3/spies/#using-a-spy-to-wrap-property-getter-and-setter

7.2.4 and onwards don't have this section of the docs.

For a patch release to remove a piece of functionality would be very questionable in terms of versioning so i assume the logic is still there in the source.

However, can someone please clear up what the situation is?

* Can we still spy on setters/getters this way?
* Where did the docs go?
* What is returned, if it works? a property descriptor? so it has a special signature different to all other spy calls?

Maybe the docs just went walkies by accident?

43081j:
Fixes #2160 

 #### Purpose

Updating the source docs because only the generated ones were modified when the documentation for spying on accessors was introduced.

I also went ahead recently and fixed the typescript types again over on DefinitelyTyped so spying on an accessor returns the descriptor rather than a spy its self. Let me know if thats wrong in any way.

cc @fatso83

### DIFF
--- a/docs/release-source/release/spies.md
+++ b/docs/release-source/release/spies.md
@@ -104,13 +104,14 @@ assert(spy.get.calledOnce);
   </dd>
   <dt><code>var spy = sinon.spy(object, "property", types);</code></dt>
   <dd>
-    Creates a spy for <code>object.property</code> descriptor and replaces the
-    original accessor methods (`get`, `set`) listed in the <code>types</code>
-    array with a spy. The spies act exactly like the original accessors in all
-    cases. The original accessors can be restored by calling
-    <code>spy.restore()</code>. The returned spy is the object which replaced
-    the original property descriptor. <code>spy.get ===
-    Object.getOwnPropertyDescriptor(object, 'property').get</code>.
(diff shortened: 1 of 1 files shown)


fatso83:
Looks great. Care to also add the new info on the return types?

43081j:
It does already seem to mention that it returns the descriptor but im happy to change it a little if needed.

> spy.restore(). The returned spy is the object which replaced
> the original property descriptor. spy.get ===
> Object.getOwnPropertyDescriptor(object, 'property').get</code>.

fatso83:
Yeah, it is a bit too cryptic and terse. Verboseness FTW.

43081j:
@fatso83 i updated the docs re the above. sorry for the long delay

mroderick:
Pull requests like these are very valuable for projects like Sinon. Thank you 🏆

mroderick:
The updated documentation will become visible, when the next release is out

PR #2175, opened 2019-12-18
Upgrade nise and browserify

mantoni:

- Upgrade nise to v2 which includes lolex v5 so that we don't have two lolex versions in the tree.
- Use mochify `--no-detect-globals` which needed a few adjustments to global detection in test cases.
- Upgrade browserify to latest
- Set `detectGlobals` to `false` when bundling Sinon.

~~The last step currently fails the Worker test because we're doing the global detection wrong everywhere. I suggest we create a utility for this in `commons`, use that in `nise` and `lolex` and create patch releases.~~ Edit: this has been done.

Fixes #2119

1. Check out this branch
2. `npm install`
3. `npm test`
4. `npm run test-cloud`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/build.js
+++ b/build.js
@@ -34,7 +34,9 @@ makeBundle(
         // Add inline source maps to the default bundle
         debug: true,
         // Create a UMD wrapper and install the "sinon" global:
-        standalone: "sinon"
+        standalone: "sinon",
+        // Do not detect and insert globals:
+        detectGlobals: false
     },
     function(bundle) {
         var script = preamble + bundle;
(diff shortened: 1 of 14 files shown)


mantoni:
PR to have a common way to detect the correct global: https://github.com/sinonjs/commons/pull/46

build.js
```diff
@@ -34,29 +34,47 @@ makeBundle(
...
-            .join("\n");
-
-    var script = preamble + intro + bundle + outro;
-    fs.writeFileSync("pkg/sinon-esm.js", script);
-});
+makeBundle(
+    "./lib/sinon.js",
+    {
+        // Create a UMD wrapper and install the "sinon" global:
+        standalone: "sinon",
```
mantoni: Note that this wasn't done before, which makes this bundle essentially unusable. I guess nobody actually uses this.

Issue #2180, opened 2019-12-23
@sinonjs/commons version in Sinon v8 is out of date


forehalo:
**Describe the bug**

After bump up sinon version to v8, the dependency @sinonjs/commons version is still kept at v1.4.0 which has incompatible api required by sinon

[https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/core/next-tick.js#L3](https://github.com/sinonjs/sinon/blob/master/lib/sinon/util/core/next-tick.js#L3)

**Screenshots**
![image](https://user-images.githubusercontent.com/8281226/71335923-f897e580-257f-11ea-9689-139222262897.png)

**Context (please complete the following information):**

* Library version: v8.0.0

mantoni:
This has been fixed with Sinon 8.0.1. Thank you for reporting ❤️

forehalo:
@mantoni @mroderick Thank you all for the quick fixing

PR #2181, opened 2019-12-23
Force upgrade @sinonjs/commons

mroderick:
Fixes https://github.com/sinonjs/sinon/issues/2180

#### Background

I guess `@sinonjs/commons` should have had a major version bump somewhere along the way, when the  API changed.

#### Solution 

* Force `@sinonjs/commons` to latest

1. Observe tests pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "docs/**/*.md": "markdownlint"
   },
   "dependencies": {
-    "@sinonjs/commons": "^1.4.0",
+    "@sinonjs/commons": "^1.7.0",
     "@sinonjs/formatio": "^4.0.1",
     "@sinonjs/samsam": "^4.0.1",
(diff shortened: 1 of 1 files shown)


mantoni:
Here is how this happened:

This is the commit that introduced the new commons feature in Sinon and it should have had the commons upgrade: f34a13ec8c878188f805a57a0c1c1de4176c2765

In the same commit, I upgraded commons in the sinon tree here: https://github.com/sinonjs/sinon/commit/f34a13ec8c878188f805a57a0c1c1de4176c2765#diff-32607347f8126e6534ebc7ebaec4853dR188

This was necessary, because I ran into this npm issue: https://npm.community/t/cannot-read-property-match-of-undefined/203

Otherwise tests would have failed and it would have been obvious that commons needed an upgrade.

PR #2185, opened 2019-12-30
Upgrade nise to 3.0.1

mroderick:
This version has fixed an issue reported as

`TypeError: GlobalTextEncoder is not a constructor`

See https://github.com/sinonjs/nise/issues/123


mroderick:
This has been published as `sinon@8.0.2`

PR #2186, opened 2019-12-31
Move printf to proxy

mroderick:
This PR is a solution for the failing tests in https://github.com/sinonjs/referee-sinon/pull/64

 #### Purpose

Make `fake` as useful in `referee` as `spy` and `stub` are.

 #### Background

When I originally created `fake`, I didn't realise that we needed a `printf` method on them, and omitted it.

 #### Solution

* Move `printf` to `proxy`, making it available on `fake` as well as `spy` and `stub`
* Validate via tests that `fake`, `spy` and `stub` all delegate their `printf` method to `proxy`

1. Check out this branch
2. `npm install`
3. `npm test`


mroderick:
This has been published as `sinon@8.0.3`

PR #2193, opened 2020-01-04
Restore fully spied objects in the sandbox

### LINKED ISSUE #2192: restore() && reset() non-functional if wrapping all object methods
**Describe the bug**
When using `sinon.spy(<obj>)` the method `sinon.restore()` does not restore the object back to the original state.

This is very useful for mocha (other) tests which use beforeEach/afterEach to restore the object to previous states.

A workaround I'm using is `sinon.spy(<obj>, "method_name")` as this works as expected and `sinon.restore()` removes the wrapped methods.

**To Reproduce**
See [https://runkit.com/olirogers/sinon-restore-issue/2.0.0](https://runkit.com/olirogers/sinon-restore-issue/2.0.0)

1. Wrap an object
2. (Optionally) run some tests on it
3. Restore the sandbox/sinon
4. Wrap the object again
5. Receive an error that `Attempted to wrap <method_name> which is already wrapped`

**Expected behavior**
Script to run successfully, restore removes the wrapping of the methods on the object.

**Screenshots**
N/A - See [https://runkit.com/olirogers/sinon-restore-issue/2.0.0](https://runkit.com/olirogers/sinon-restore-issue/2.0.0)

* Library version: 8.0.2
* Environment: Node 10.18.0 (from runkit)
* Example URL: [https://runkit.com/olirogers/sinon-restore-issue/2.0.0](https://runkit.com/olirogers/sinon-restore-issue/2.0.0)
* Other libraries you are using: @sinonjs/referee (version 4.0.0)

**Additional context**
This issue was encountered following the official guides at - https://sinonjs.org/releases/latest/spies/, version 1.0.0. of the runkit has an example that should also exhibit this bug (but runkit was being slow!).

The same behaviour is experienced with `sinon.reset()`. It appears to be calling reset on the object itself (`new Foo{}`) rather than each of the methods.

I found #1553 but assumed that this should be called automatically from the sandbox/environment restore()?

Let me know if you need any more information.

Oli

fatso83:
- Align the behaviour of spies with that of stubs to fix #2192 

Additionally, this PR also fixes a misleading error message.

See #2192. It is currently not possible to use the sandbox restore functionality on objects that have been spied on like this `sinon.spy(obj)`.

 #### Solution  
The only thing I want to point out as strange is the fixes to the error message. I found it somewhat strange that we allow fakes to replace stuff on the prototype, even though we have had a discussion about this earlier regarding spies and stubs where @mroderick came to the opposite conclusion: do not allow stubbing on fields up the prototype chain (hence the error messages). This makes using fakes a different beast from using spies and stubs, which I think is a bit unfortunate.

1. Notice the additional test
2. See `npm test` pass

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -210,7 +210,7 @@ function Sandbox() {
         var descriptor = getPropertyDescriptor(object, property);
 
         if (typeof descriptor === "undefined") {
-            throw new TypeError("Cannot replace non-existent own property " + valueToString(property));
+            throw new TypeError("Cannot replace non-existent property " + valueToString(property));
         }
 
         if (typeof replacement === "undefined") {
@@ -243,7 +243,7 @@ function Sandbox() {
         var descriptor = getPropertyDescriptor(object, property);
 
         if (typeof descriptor === "undefined") {
-            throw new TypeError("Cannot replace non-existent own property " + valueToString(property));
+            throw new TypeError("Cannot replace non-existent property " + valueToString(property));
         }
 
         if (typeof replacement !== "function") {
@@ -271,7 +271,7 @@ function Sandbox() {
         var descriptor = getPropertyDescriptor(object, property);
 
         if (typeof descriptor === "undefined") {
-            throw new TypeError("Cannot replace non-existent own property " + valueToString(property));
+            throw new TypeError("Cannot replace non-existent property " + valueToString(property));
         }
 
         if (typeof replacement !== "function") {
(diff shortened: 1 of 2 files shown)


fatso83:
Coincidentally, the exact discussion about own vs prototype values was just revived (unrelated to this issue): https://github.com/sinonjs/sinon/pull/1557#issue-140637498

mroderick:
I agree that is is confusing and unfortunate. An improvement would be to clarify in the documentation when fakes, spies and stubs can/cannot replace prototype methods. That won't solve the inconsistency, but at least make it less confusing for people using the library.

Since the discussion is here (or linked from here), should we do that as part of this pull request?

fatso83:
I was about to say yes to that, but I see won't have time to update this PR with any meaningful amount of additional work. I say yes to creating an additional PR to deal with the texts. Of course, I would have like to settle the matter, but that is probably an even greater issue ...

mroderick:
This has been published as `sinon@8.0.4`

PR #2197, opened 2020-01-12
docs: add `wrappedMethod` line

### LINKED ISSUE #2182: Document stub.wrappedMethod property
Thanks to #988 stub.wrappedMethod of a stub was exposed.
But, as long as this is not documented anywhere no one would use it.
I suggest you should document this feature.
Pay attention to the fact that the property is only defined when wrapping an existing method,
not when using sinon.stub() (with no parameters).

43081j:
Fixes #2182 

This is a tiny one but it means we can unblock @NoamDev over in DefinitelyTyped/DefinitelyTyped#41219.

There was a typo'd header too i fixed in there. If you want any more info adding to it, let me know.

cc @fatso83

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -335,7 +335,7 @@ Causes the stub to throw the provided exception object.
 Causes the stub to throw the exception returned by the function.
 
 
-### `stub.throwsArg(index);`
+#### `stub.throwsArg(index);`
 
 Causes the stub to throw the argument at the provided index.
 
@@ -684,3 +684,7 @@ stub.restore()
 myObj.example; // 'oldValue'
 ```
 
+#### `stub.wrappedMethod`
+
+Holds a reference to the original method/function this stub has
+wrapped.
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -3484,4 +3484,29 @@ describe("stub", function() {
             assert.same(stub.printf, proxy.printf);
         });
     });
+
+    describe(".wrappedMethod", function() {
+        it("should return the original method being stubbed", function() {
+            var myFn = function() {
+                return "foo";
+            };
+            var myObj = {
+                fn: myFn
+            };
+            createStub(myObj, "fn");
+            assert.same(myFn, myObj.fn.wrappedMethod);
+        });
+
+        it("should not exist for accessors", function() {
+            var myObj = {
+                get prop() {
+                    return "foo";
+                }
+            };
+            createStub(myObj, "prop").get(function() {
+                return "bar";
+            });
+            assert.isUndefined(myObj.prop.wrappedMethod);
+        });
+    });
 });


NoamDev:
from my testings, stub.wrappedMethod will be undefined when stubbing a property
example for stubbing getter/setter
https://runkit.com/embed/0u8nznkm2nu9
example for stubbing raw property
https://runkit.com/embed/mzqczaq4g5xt

NoamDev:
In addition wrappedMethod is available for spies too.
However, it seems incomplete when you try it on properties.
Consider this example:
```
var sinon = require("sinon")
var sinon = require("sinon")
var object = {
  get test() {
    return 1;
  },
  set test(value) {
  }
};

var spy= sinon.spy(object, "test",['get','set']);
```
spy.get.wrappedMethod is undefined, so is spy.set.wrappedMethod.
however, spy.wrappedMethod is equal to the original setter.
It seems like first, it does spy.wrappedMethod = TheOriginalGetter
and then spy.wrappedMethod = TheOriginalSetter (overriding the previous value)
changing the spy to `sinon.spy(object, "test",['set','get']);`
leads the opposite (spy.wrappedMethod is the original getter).

43081j:
yep seems that stubbing an accessor doesn't result in `wrapMethod` being called so we have no `wrappedMethod` set.

lets see what @fatso83 thinks in case it should have one.

NoamDev:
Note that adding a call to wrapMethod might result with what happens with
spies and accessors, see #2198

fatso83:
This isn't a feature we support. This is an internal detail. If we relied on this for the public interface we would have covered it with tests. Alas:
```
$ ag -l wrappedMethod test/ | wc -l
0
```

There are 0 lines of test code covering that field, so I am not going to be the one giving a :+1: for documenting that feature. @mroderick might remember more about the history of this internal property, but from what I can gather, this is part of some internal bookkeeping machinery for spies and stubs.
```
$ ag -l wrappedMethod lib/
lib/sinon/behavior.js
lib/sinon/util/core/wrap-method.js
lib/sinon/mock.js
```

--- 

That being said, I can understand if this field has some value, but I would rather suggest creating an explicit extension of the existing API with a new method that has some clearly defined behavior, preferably also working for getters and setters.

fatso83:
Reopening, as I read #988, which kind of implies that we have explicitly added this field _as a feature_. I am not totally sure about how to go about this, though, for the reasons mentioned in the previous comment. At least I think we need to cover this feature with some tests documenting how it works, before adding it to the docs.

43081j:
yep i made this PR because i had seen #988 which claimed it was exposed on purpose, and because i saw that we had a blocked PR over on definitelytyped for the same thing.

you're definitely right it should probably be a method which works for accessors, too. but in its current state, a decision should be made really. 

im happy to add tests, but are we certain it should be part of the public api before i do?

43081j:
here's some tests so we can get this merged hopefully

NoamDev:
I believe that if it was declared as on official feature, then we should solve #2198

mroderick:
> I believe that if it was declared as on official feature, then we should solve #2198

It's in the documentation now, so it's official!

If it gets removed in future version(s), then that will be a breaking change and a new `MAJOR` version.

Issue #2196, opened 2020-01-11
Proposal: Support negative indices for getCall


RealZogger:
The API currently provides:
`getCall(n)`
`getFirstCall`
`getSecondCall`
`getThirdCall`
`getLastCall`

Some functions in js support negative indices to get elements from the end of the array (e.g. `arr.slice(-2)` returns the last 2 elements of an array).

Currently, `getCall(-n)` returns null

The proposal is to support negative indices to get previous calls, for example
`getCall(-1)` would be equivalent to `getLastCall`
`getCall(-2)` would be the second last call.
`getCall(-n)` where n > the number of calls would return `null`

fatso83:
Have a look at the implementation and see to what degree it would complicate the implementation. You would essentially need to create a circular buffer so that a pointer could be shifted to keep things in sync for each call. That's quite a different beast from simple arrays, and I am not all that sure it's worth it given that this is the first time I heard someone have a need for it in the ten years Sinon existed :-)

RealZogger:
well, from what I've seen I think it's a one line change - At the start of the getCall function, if a negative index is supplied, add the length of the array to the passed in index and get that index instead.

I'm wrapping around getCall in my tests to do effectively this and it works fine.

My use case is that I am running async tests on a vscode extension using the vscode-test package. I'm stubbing a particular function in the vscode API, and I know that the function should be called twice just before resolving.

However the vscode test package runs up the ide so it's technically possible for someone running the text to interfere with the test by accident, perhaps clicking something.

Obviously the IDE should be left alone but for the sake of being a bit more robust against this kind of thing I'm getting the last two calls to the function as these must be the calls I'm interested in.

It's not a majorly important feature for me and quite a niche case, but I thought I would raise it to see if anyone thought it was worthwile

fatso83:
If it seems relatively simple to implement, just have a go at it :smiley_cat:

PR #2199, opened 2020-01-15
Support negative indices in getCall

RealZogger:
This allows getting calls indexed from the end of the array
e.g. -1 gets the last call, -2 gets the penultimate call etc.

 #### Purpose (TL;DR)
Implements the proposal in #2196

Allows the user to get the -nth call when the total number of calls is arbitrary, without having to wrap around getCalls() function.

In an ideal world, the full set of calls should be known and this function would not be necessary, but for some use cases this can be useful (see #2196 for my use case). I understand this is a bit of a niche case so don't mind if you choose to close instead of merge!

 #### How to verify
1. Check out this branch
2. `npm install`
3. `npm test`

### DIFF
--- a/lib/sinon/proxy.js
+++ b/lib/sinon/proxy.js
@@ -43,6 +43,7 @@ var proxyApi = {
     getCall: function getCall(index) {
         var i = index;
         if (i < 0) {
+            // Negative indices means counting backwards from the last call
             i += this.callCount;
         }
         if (i < 0 || i >= this.callCount) {
--- a/lib/sinon/proxy.js
+++ b/lib/sinon/proxy.js
@@ -40,7 +40,11 @@ var proxyApi = {
         return emptyFakes;
     },
 
-    getCall: function getCall(i) {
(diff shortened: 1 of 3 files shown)


lib/sinon/proxy.js
```diff
@@ -40,7 +40,11 @@ var proxyApi = {
         return emptyFakes;
     },
 
-    getCall: function getCall(i) {
+    getCall: function getCall(index) {
+        var i = index;
+        if (i < 0) {
+            i += this.callCount;
```
fatso83: Delightfully simple. Wonderful!

```suggestion
            // Negative indices means counting backwards from the last call
            i += this.callCount;
```

RealZogger: Applied the comment (I'm quite new to GitHub, assume that's the right way to do it, I have the impression there's some automagic to apply comments with a click, but only with the right permissions)

mroderick:
This has been published to `npm` as `sinon@8.1.0`

Issue #2202, opened 2020-01-24
Slowdown with since 8.0.0 for NodeJs 10


Flarna:
**Describe the bug**
Since Sinon v8 the execution of my testsuite is significant slower if I use NodeJs 10.
If I use NodeJs 12 I don't see a significant slowdown between 8.1.1 and 7.5.0.

Sinon 8.1.1:
NodeJs 12.14.1: 15.7 seconds
NodeJs 10.18.1: 50.7 seconds

Sinon 8.0.0:
NodeJs 12.14.1: 15.4 seconds
NodeJs 10.18.1: 50.7 seconds

Sinon 7.5.0:
NodeJs 12.14.1: 19.9 seconds
NodeJs 10.18.1: 21.2 seconds

The only change between above runs is the sinon version I selected in my package.json.

As you can see NodeJs 12 got faster with sinon 8 but NodeJs 10 got significant slower.

**To Reproduce**
Execute mocha testsuite of my project using different versions of node.js and sinon.

**Expected behavior**
No significant slowdown with sinon 8 compared to sinon 7 if NodeJs 10 is used.

**Screenshots**

**Context (please complete the following information):**

* Library version: 8.1.1
* Environment: linux (windows is also effected)
* Example URL:
* Other libraries you are using: mocha 7.0.0

**Additional context**
My project is closed source therefore I can't share it as reproducer.
I plan to investigate this further but any hints where to start would be helpful.
Maybe it's something similar then #1983?

mroderick:
You can use `npm link` to use a local checkout of Sinon, and then use `git bisect` to find exactly the commit that caused the slowdown.

Something along the lines of:

1. `git clone git@github.com:sinonjs/sinon.git`
1. `cd sinon`
1. (sinon):`npm ci`
1. (sinon): `npm link`
1. (your project): `npm link sinon`
1. (your project): npm test, observe speed
1. (sinon): `git bisect start`
1. (sinon): `git bisect good v7.5.0`
1. (sinon): `git bisect bad`
1. (sinon): `npm ci`
1. (your project): npm test, observe speed
1. If slow
    * (sinon): `git bisect bad`
1. If not slow
    * (sinon): `git bisect good`
1. Repeat steps **10 - 14** until you find the first commit, where the tests are running slowly 

Make sure you repeat the `npm ci` step, as the slowdown could be in one of Sinon's dependencies.

Once you know which commit caused the slow down, then we can analyse the changes and see what might have caused it.

[...]

PR #2204, opened 2020-01-30
Merge stub and spy APIs only once

mantoni:

This is an attempt to fix the performance issue described in #2202 that has been introduced by #2130.

Some properties from spy are overridden by the stub implementation. My guess is that re-declaring non-enumerable properties on an object is inefficient in some V8 versions. Doing this just once for the `stubApi` might be faster than doing it on every invocation. @Flarna Can you confirm this with your project?

1. Check out this branch
2. `npm install`
3. `npm test`

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -40,8 +40,6 @@ function createStub(originalFunc) {
     }
 
     proxy = createProxy(functionStub, originalFunc || functionStub);
-    // Inherit spy API:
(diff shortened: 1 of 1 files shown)


Flarna:
Thanks for working on this!
But unfortunately these changes don't help.

mantoni:
Thank you for testing. It was just a quick shot in the dark.

PR #2210, opened 2020-02-01
Drop Node 8 support

mroderick:
As can be seen at https://github.com/nodejs/Release, Node 8 reached
"end" of life on 2019-12-31, and is no longer actively supported.

We will stop testing in Node 8 and start testing in Node 13, which will
become the next LTS release from April 2020.

#### How to verify

1. Observe that we're no longer testing in Node 8
1. Observe that we've started testing in Node 13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,15 +48,6 @@ jobs:
           name: Lint
           command: npm run lint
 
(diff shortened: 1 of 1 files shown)


mshaaban0:
LGTM 👍

PR #2229, opened 2020-02-25
Added missing documentation for sinon.assert.calledOnceWithExactly

rhodgkins:
Added for all versions >=7.5.0 (when it [was introduced](https://github.com/sinonjs/sinon/pull/2080#issuecomment-534948093)).

(Also see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42569)


mroderick:
Thank you for your pull request! These kinds of updates are essential to make using libraries such as Sinon more enjoyable 💯

PR #2251, opened 2020-04-13
Set wrappedMethod on getters/setters

### LINKED ISSUE #2198: spy.wrappedMethod is not ok when spying getter/setter
**Describe the bug**
when spying getter/setter, spy.get.wrappedMethod and spy.set.wrappedMethod are both undefined but spy.wrappedMethod is the original getter/setter whoever was written last (looks like an override).

**To Reproduce**
```var sinon = require("sinon")
var object = {
  get test() {
    return 1;
  },
  set test(value) {
  }
};

var stub = sinon.spy(object, "test",['get','set']);

console.log('1',stub.get.wrappedMethod)
console.log('2', stub.set.wrappedMethod)
console.log('3', stub.wrappedMethod)
stub.restore()
var stub = sinon.spy(object, "test",['set','get']);
console.log('4', stub.wrappedMethod)
```
results can be seen here https://runkit.com/embed/b9qjkhx4ufk7

**Expected behavior**
the first should be the original getter
the second should be the original setter
the third should be the undefined
the fourth should be the undefined

**Actual behavior**
the first is undefined
the second is undefined
the third is the original setter
the fourth is the original getter

rgroothuijsen:
#### Purpose
This PR fixes #2198 by wrapping accessors if they are present, rather than the property they belong to.

 #### Background
Because the current wrapMethod implementation only assumes the presence of a single method on a property, trying to set both a getter and a setter will cause the former to be overwritten by the latter, and not added as separate `get` and `set` methods. Furthermore, `restore()` is implemented on the property itself, which does not allow the `spy.get.restore()` and `spy.set.restore()` outlined in the documentation.

#### Solution

This problem was addressed by changing wrapMethod to handle multiple methods per property. This includes a check whether any particular method is an accessor or not, and a method-wrapping implementation is chosen based on this. If the method in question is not a "get" or "set" method, the existing implementation will be used.

In order to satisfy ESLint code quality requirements, some restructuring of wrapMethod has been performed as well. To the extent possible, however, the existing structure has been left alone.

- restore() was moved because function declarations are not allowed inside a loop
- Parts of the code have been separated into functions because the complexity of wrapMethod had become too high

 #### How to verify
1. Check out this branch
2. `npm install`
3. `npm test` (See included test for details)

### DIFF
--- a/lib/sinon/util/core/wrap-method.js
+++ b/lib/sinon/util/core/wrap-method.js
@@ -4,6 +4,7 @@ var getPropertyDescriptor = require("./get-property-descriptor");
 var extend = require("./extend");
 var hasOwnProperty = require("@sinonjs/commons").prototypes.object.hasOwnProperty;
 var valueToString = require("@sinonjs/commons").valueToString;
+var push = require("@sinonjs/commons").prototypes.array.push;
 
 function isFunction(obj) {
     return typeof obj === "function" || Boolean(obj && obj.constructor && obj.call && obj.apply);
@@ -15,6 +16,18 @@ function mirrorProperties(target, source) {
             target[prop] = source[prop];
         }
     }
+}
+
+function getAccessor(object, property, method) {
+    var accessors = ["get", "set"];
+    var descriptor = getPropertyDescriptor(object, property);
+
+    for (var i = 0; i < accessors.length; i++) {
+        if (descriptor[accessors[i]] && descriptor[accessors[i]].name === method.name) {
+            return accessors[i];
+        }
+    }
+    return null;
 }
 
 // Cheap way to detect if we have ES5 support.
@@ -51,7 +64,9 @@ module.exports = function wrapMethod(object, property, method) {
         }
     }
 
-    var error, wrappedMethod, i, wrappedMethodDesc;
+    var error, wrappedMethods, wrappedMethod, i, wrappedMethodDesc, target, accessor;
+
+    wrappedMethods = [];
 
     function simplePropertyAssignment() {
         wrappedMethod = object[property];
@@ -86,6 +101,7 @@ module.exports = function wrapMethod(object, property, method) {
         for (i = 0; i < types.length; i++) {
             wrappedMethod = wrappedMethodDesc[types[i]];
             checkWrappedMethod(wrappedMethod);
+            push(wrappedMethods, wrappedMethod);
         }
 
         mirrorProperties(methodDesc, wrappedMethodDesc);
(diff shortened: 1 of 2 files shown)


mroderick:
Once this is merged, we should reverse the https://github.com/sinonjs/sinon/pull/2270 changes!

lib/sinon/util/core/wrap-method.js
```diff
@@ -106,49 +122,82 @@ module.exports = function wrapMethod(object, property, method) {
         simplePropertyAssignment();
     }
 
-    extend.nonEnum(method, {
-        displayName: property,
+    extendObjectWithWrappedMethods();
+
+    function extendObjectWithWrappedMethods() {
+        for (i = 0; i < wrappedMethods.length; i++) {
```
mroderick, line 128: Try to avoid using shared variables across functions (specifically `i` in this case), as that can lead to some difficult to spot bugs.

In this case, I think it is perfectly valid to use the ES5 Array Extras, i.e. `forEach`.


```suggestion
        wrappedMethods.forEach(function(wrapped){
```

lib/sinon/util/core/wrap-method.js
```diff
@@ -106,49 +122,82 @@ module.exports = function wrapMethod(object, property, method) {
...
+            accessor = getAccessor(object, property, wrappedMethods[i]);
```
mroderick, line 129:
```suggestion
            accessor = getAccessor(object, property, wrapped);
```

lib/sinon/util/core/wrap-method.js
```diff
@@ -106,49 +122,82 @@ module.exports = function wrapMethod(object, property, method) {
...
+            target = accessor ? method[accessor] : method;
+            extend.nonEnum(target, {
+                displayName: property,
+                wrappedMethod: wrappedMethods[i],
```
mroderick, line 133:
```suggestion
                wrappedMethod: wrapped,
```

lib/sinon/util/core/wrap-method.js
```diff
@@ -17,6 +18,18 @@ function mirrorProperties(target, source) {
     }
 }
 
+function getAccessor(object, property, method) {
+    var accessors = ["get", "set"];
+    var descriptor = getPropertyDescriptor(object, property);
+
+    for (var i = 0; i < accessors.length; i++) {
```
mroderick, line 25:
```suggestion
    accesssors.forEach(function(accessor){
```

lib/sinon/util/core/wrap-method.js
```diff
@@ -17,6 +18,18 @@ function mirrorProperties(target, source) {
...
+        if (descriptor[accessors[i]] && descriptor[accessors[i]].name === method.name) {
```
mroderick, line 26:
```suggestion
        if (descriptor[accessor] && descriptor[accessor].name === method.name) {
```

lib/sinon/util/core/wrap-method.js
```diff
@@ -17,6 +18,18 @@ function mirrorProperties(target, source) {
...
+            return accessors[i];
```
mroderick, line 27:
```suggestion
            return accessor;
```

mroderick (changes requested):
Thank you for diligently fixing this issue and for your patience with waiting for a review 🏆 

I've added a few suggestions around using `forEach` instead of `for` loops.

fatso83:
I'll be taking this across the finish line tonight. Just implementing the feedback basically.

mroderick:
Running the tests in browsers, @hexeberlin and I discovered that this PR fails in IE11 and Safari 9

```
# internet explorer 11:

  1) spy
       sets wrappedMethod on getter and setter:
     AssertionError: [refute.isUndefined] Expected undefined ('undefined') not to be undefined
      at fail (node_modules/@sinonjs/referee/lib/create-fail.js:15)
      at ctx.fail (node_modules/@sinonjs/referee/lib/define-assertion.js:44)
      at assertion (node_modules/@sinonjs/referee/lib/define-assertion.js:62)
      at referee[type][name] (node_modules/@sinonjs/referee/lib/define-assertion.js:89)
      at Anonymous function (test/spy-test.js:350)

  2) stub
       .get
         can restore stubbed setters for functions:
     AssertionError: [assert.equals] undefined expected to be equal to [Function]
      at fail (node_modules/@sinonjs/referee/lib/create-fail.js:15)
      at ctx.fail (node_modules/@sinonjs/referee/lib/define-assertion.js:44)
      at assertion (node_modules/@sinonjs/referee/lib/define-assertion.js:62)
      at referee[type][name] (node_modules/@sinonjs/referee/lib/define-assertion.js:89)
      at Anonymous function (test/stub-test.js:3298)

  3) stub
       .set
         can restore stubbed setters for functions:
     AssertionError: [assert.equals] undefined expected to be equal to [Function]
      at fail (node_modules/@sinonjs/referee/lib/create-fail.js:15)
      at ctx.fail (node_modules/@sinonjs/referee/lib/define-assertion.js:44)
      at assertion (node_modules/@sinonjs/referee/lib/define-assertion.js:62)
      at referee[type][name] (node_modules/@sinonjs/referee/lib/define-assertion.js:89)
      at Anonymous function (test/stub-test.js:3385)


# Safari 9.0

  1) spy
       sets wrappedMethod on getter and setter:
     [refute.isUndefined] Expected undefined ('undefined') not to be undefined
      fail node_modules/@sinonjs/referee/lib/create-fail.js:5
      fail node_modules/@sinonjs/referee/lib/define-assertion.js:44
      assertion node_modules/@sinonjs/referee/lib/define-assertion.js:62
      node_modules/@sinonjs/referee/lib/define-assertion.js:89
      test/spy-test.js:348
```

Since we will drop IE11 with #2324 and subsequently allow for more modern syntax and probably require ES5 support as minimum, I propose that we delay fixing #2198 until we've had a chance to refactor `wrapMethod` to more modern JS.

fatso83:
> I propose that we delay fixing #2198 until we've had a chance to refactor wrapMethod to more modern JS.

Not totally sure what should happen, but we have dropped support for IE11 and Safari 9 with Sinon 10, so ... how do we progress from here? I not totally sure what is supposed to happen first.

mroderick:
I think we should try rebasing it onto the primary branch, and see if that doesn't fix it. I don't currently have the bandwidth to do that, but might be able to look into it on the weeekend.

fatso83:
Rebased: https://github.com/sinonjs/sinon/compare/master...fatso83:SINON-2198?expand=1

Some conflicts, but I think it's ok. All tests pass, at least.

fatso83:
Resubmitted as #2378 to get it in.

PR #2265, opened 2020-06-03
Add example where withArgs of a stub is used in an assertion

tbiesemann:
There was no example whether it is allowed to use withArgs within an assertion.
It was not clear if this is allowed or if sinon.stub().withArgs can only be used to specify the behavior.


mantoni:
Very nice. Thank you!

Issue #2298, opened 2020-10-05
`sandbox.assert.fail` and `sandbox.assert.pass` affect global sinon configuration


kim366:
**Describe the bug**
`sandbox.assert.fail` and `sandbox.assert.pass` seem to not only affect the current sandbox, but the entire sinon environment. This becomes an issue when tests run concurrently, as they do in, e.g., AVA.

**To Reproduce**
I created this function. When used in two AVA tests that run concurrently, AVA will either report `Error: Test finished without running any assertions` or `Error: Assertion passed, but test has already finished`.

```ts
export function useSinon(t: ExecutionContext) {
  const sandbox = createSandbox();
  sandbox.assert.fail = msg => t.fail(msg);
  sandbox.assert.pass = () => t.pass();
  t.teardown(() => sandbox.restore());
  sandbox.useFakeTimers(<any>{global});

  return sandbox;
}
```

**Expected behavior**
The fail and pass bindings should only affect the current sandbox, that way they don't get overridden by concurrent tests.

**Context**

* Library version: 9.1.0*
* Environment: Node.js 14.11.0
* Other libraries you are using: AVA 3.13.0

mroderick:
Is there any chance you can make this a runnable example?
Also, could you make it a JavaScript example and not TypeScript?

kim366:
@mroderick Here you go: https://github.com/kim366/sinon-issue-concurrent. If you mark the tests with `test.serial`, they will pass. While they run concurrently, one does not pass, because `t` is rebound to be the other test's context

fatso83:
Great bug report with repro! Fearing that this might be tightly related to understanding AVA as well.

> While they run concurrently, one does not pass, because t is rebound to be the other test's context
:thinking: ... :exploding_head: Could you elaborate on that? Not sure I understand what you mean wrt how this `t` is being rebound. I don't see `t` being passed around anywhere.

kim366:
@fatso83 Thank you. `t` (AVA's execution context) is being passed from the test case to `useSinon`.

Let's call the `t` from foo `tFoo` and the other one `tBar`. Here is what happens:

1. `foo` binds `sandbox.assert.pass` to `tFoo`.
1. `bar` binds `sandbox.assert.pass` to `tBar` (which overwrites the first assignment).
1. 20ms pass
1. `foo` finishes and calls `sandbox.assert.pass`, which in turn calls `tBar.pass()`.
1. `foo` fais, because `tFoo.pass()` was never called.

[...]

PR #2302, opened 2020-10-09
Make sandboxes each use their own assert object

kim366:

Fix issue #2298 by having assertion objects be created by each sandbox instead of using the global one

Intuitively you would think that sandboxes use isolated assertion objects. But this was not the case. If you overrode the `failException` or `pass`/`fail` functions, they would apply to all sandboxes and all assertions. This particularly was an issue with concurrent testing, such as with AVA.

We no longer create and export a single `assert` object, instead there is a `createAssertionObject` function that is used inside the sandboxes' constructor.

1. Check out this branch
2. `npm install`
3. `npm test`

 #### Checklist for author

- [x] `npm run lint` passes

PS: on github the diff shows very poorly. In actuality only ~3 lines changed and a test case was added, but a buch of lines were indented one step.

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -37,6 +37,8 @@ function Sandbox() {
     var fakeRestorers = [];
     var promiseLib;
 
+    sandbox.assert = sinonAssert.createAssertObject();
+
     sandbox.serverPrototype = fakeServer;
 
     // this is for testing only
@@ -410,7 +412,6 @@ function Sandbox() {
     };
 }
 
-Sandbox.prototype.assert = sinonAssert;
 Sandbox.prototype.match = match;
 
 module.exports = Sandbox;
--- a/test/sandbox-test.js
+++ b/test/sandbox-test.js
@@ -15,7 +15,6 @@ var sinonFake = require("../lib/sinon/fake");
 var sinonSpy = require("../lib/sinon/spy");
 var sinonStub = require("../lib/sinon/stub");
 var sinonConfig = require("../lib/sinon/util/core/get-config");
-var sinonAssert = require("../lib/sinon/assert");
 var sinonClock = require("../lib/sinon/util/fake-timers");
 
 var supportsAjax = typeof XMLHttpRequest !== "undefined" || typeof ActiveXObject !== "undefined";
@@ -45,12 +44,6 @@ describe("Sandbox", function() {
         var sandbox = new Sandbox();
 
         assert.same(sandbox.match, match);
-    });
-
-    it("exposes assert", function() {
-        var sandbox = new Sandbox();
-
-        assert.same(sandbox.assert, sinonAssert);
     });
 
     it("can be reset without failing when pre-configured to use a fake server", function() {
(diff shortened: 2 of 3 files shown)


lib/sinon/sandbox.js
```diff
@@ -37,6 +37,8 @@ function Sandbox() {
     var fakeRestorers = [];
     var promiseLib;
 
+    sandbox.assert = sinonAssert.createAssertObject();
```
fatso83, line 40: 👍

test/sandbox-test.js
```diff
@@ -2077,4 +2070,24 @@ describe("Sandbox", function() {
...
+
+            assert.exception(
+                function() {
+                    sandboxA.assert.fail("Some message");
+                },
+                { name: "CustomErrorA" }
+            );
+
+            sandboxA.restore();
+            sandboxB.restore();
```
fatso83, line 2090: Not really seeing the point of sandboxB here, as it is not being exercised in any way. What purpose does it then serve? Should you not assert an exception on both sandboxes?

fatso83 (changes requested):

test/sandbox-test.js
kim366: This is just testing the behavior I demonstrated in the repro. If we have two tests running concurrently, one initializing A, the other B and *then* A makes an assertion, it uses its own assertion object, which is configured to throw the correct assertion rather than the global assertion object that may be reconfigured in B.

We can add in the second assertion, if you want, but this suffices and the test breaks without the changes

fatso83: Two months from now no one will remember the context for _why_ and will only look at the test implementation and go :thinking: .

If there is one part of the test that _should not throw_ (and does without the fix) or that _does throw_ (and should not) without the fix, then make it clear in the test. Wrap a `refute.exception(...)` around that block which should not block.

fatso83: btw refute is referee's "assert.not*"

kim366: Ok. Please you add it. I get `Expected not to throw but threw 'CustomErrorA' ('Some message')` even though, I only want to refute that it throws `CustomErrorB`. I don't know referee at all.

fatso83: Ah, I'll fix. And I probably told you something wrong, anyway. I'll have to lookup those docs anyhow.

test/sandbox-test.js
```diff
@@ -2077,4 +2070,24 @@ describe("Sandbox", function() {
...
+            sandboxB.assert.failException = "CustomErrorB";
+
+            assert.exception(
+                function() {
+                    sandboxA.assert.fail("Some message");
+                },
+                { name: "CustomErrorA" }
+            );
+
+            sandboxA.restore();
```
fatso83, line 2089:
```suggestion
            assert.exception(
                function() {
                    sandboxB.assert.fail("Some message");
                },
                { name: "CustomErrorB" }
            );

            sandboxA.restore();
```

fatso83 (approved):
All good.

fatso83:
Thanks for the changes: a very welcome improvement, Kim! :100: 

---

Just a tip: if you think the diff is messier than what is actually happening, you can suppress all whitespace changes by passing a query parameter: `?w=1`. This is equivalent to `git log -w` or `git diff -w`. So your diff would look [like this](https://github.com/sinonjs/sinon/pull/2302/files?w=1).

mroderick:
Very nice PR, thank you 🍰

mroderick:

PR #2303, opened 2020-10-12
Update error message on assert when representation of expected and ac…

romanbalayan:
Fixes issue #2084 by adding quotes to diff values if said values are strings, creating distinct string representations and preventing the expected and actual values from being reduced to a single value.

When the expected and actual values of an assertion do not match but they have the same string representation (for example the number 1234 and the string "1234"), the output does not show the difference, but reduces it to a single value. In this example, this output is produced:

expected spy to be called with arguments
<black>1234</black>
while this output was expected:

expected spy to be called with arguments
<red>1234</red>
<green>1234</green>

 #### Solution  - optional
A solution was proposed in the ticket itself which was also implemented here: adding quotes around a diff value if the value in question is a string. This process is skipped if the value is undefined, not a string, or already wrapped in quotes.

Instead, diffs are now displayed as follows, given the example where the string 1234 was expected but the number 1234 was found instead:

expected spy to be called with arguments
<red>1234</red>
<green>'1234'</green>

1. Check out this branch
2. `npm install`
3. Run the test included with this PR

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/spy-formatters.js
+++ b/lib/sinon/spy-formatters.js
@@ -10,6 +10,7 @@ var jsDiff = require("diff");
 var join = arrayProto.join;
 var map = arrayProto.map;
 var push = arrayProto.push;
+var slice = arrayProto.slice;
 
 function colorSinonMatchText(matcher, calledArg, calledArgMessage) {
     var calledArgumentMessage = calledArgMessage;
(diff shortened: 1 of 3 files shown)


fatso83 (approved):

fatso83:
@mroderick Is this a semver major change? If someone parses sinon's output and expects it to be a certain format, I guess this would break it. So maybe just bump the major ... 

@romanbalayan Thanks for the fix!

romanbalayan:
Hi @fatso83,

Also adding the same request here, hopefully it would not be much of a bother for you to add the hacktoberfest-accepted label to this PR so it may be counted towards my progress?

Thank you!

mroderick:

Issue #2203, opened 2020-01-27
`TypeError: Cannot assign to read only property 'name' of function` in `lib/sinon/util/core/extend.js`


jandockx:
**Describe the bug**

When creating a spy for a function that has an explicit `name` property, `lib/sinon/util/core/extend.js` fails on line 88:

```
TypeError: Cannot assign to read only property 'name' of function 'function proxy() {
                return p.invoke(func, this, slice(arguments));
            }'
    at copyValue (node_modules/sinon/lib/sinon/util/core/extend.js:88:20)
    at extendCommon (node_modules/sinon/lib/sinon/util/core/extend.js:62:17)
    at extend (node_modules/sinon/lib/sinon/util/core/extend.js:87:12)
    at createProxy (node_modules/sinon/lib/sinon/proxy.js:196:5)
    at createSpy (node_modules/sinon/lib/sinon/spy.js:122:17)
    at Function.spy (node_modules/sinon/lib/sinon/spy.js:157:45)
    at Sandbox.spy (node_modules/sinon/lib/sinon/sandbox.js:322:35)
    at Context.<anonymous> (test/FlowLogger.test.js:168:44)
    at processImmediate (internal/timers.js:439:21)
```

The reason is that `proxy.name` is not writeable.

**To Reproduce**

Try to spy on a function that has a name as an own property.

```
  it('cannot spy', function () {
    const brandNewName = 'my brand new name'
    function importantFunction () {}

    importantFunction.name.should.equal('importantFunction')

    Object.defineProperty(importantFunction, 'name', {
      configurable: false,
      enumerable: true,
      value: brandNewName,
      writable: false
    })

    console.log(importantFunction)
    importantFunction.name.should.equal(brandNewName)

    function callSomethingImportant () {
      importantFunction()
    }

    const sandbox = sinon.createSandbox()
    const stub = sandbox.spy(importantFunction) // fails

    callSomethingImportant()
    stub.should.be.calledOnce()
  })
```

**Expected behavior**

Make the proxy have the same name, or skip copying the name to the proxy.

**Solution**

I believe you can do this, by not copying with `dest[prop] = source[prop]`, but by _defining_ the property on `dest` as derived:

```
    return extendCommon(target, sources, function copyValue(dest, source, prop) {
        const sourcePropDescriptor = Object.getOwnPropertyDescriptor(source, prop)
        Object.defineProperty(dest, prop, {
            configurable: sourcePropDescriptor.configurable,
[...]

PR #2304, opened 2020-10-18
Skip write to 'name' property if not [[writable]]

sjoseph7:
#### Purpose
Fix issue #2203 by skipping 'name' property if it is not [[writable]].

#### Solution
This solution works by skipping writing a source property to the target object if the property is 'name' _and_ is not [[writable]] on the 'target'.

 #### How to verify
1. Check out this branch
2. `npm install`
3. `npm test`

 #### Checklist for author
- [x] `npm run lint` passes 
> // unable to run npm script on Windows, linted manually with `eslint --fix`.

- [X] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

### DIFF
--- a/lib/sinon/util/core/extend.js
+++ b/lib/sinon/util/core/extend.js
@@ -73,8 +73,8 @@ function extendCommon(target, sources, doCopy) {
     return target;
 }
 
-/** Public: Extend target in place with all (own) properties from sources in-order. Thus, last source will
- *         override properties in previous sources.
+/** Public: Extend target in place with all (own) properties, except 'name' when [[writable]] is false,
+ *         from sources in-order. Thus, last source will override properties in previous sources.
  *
  * @arg {Object} target - The Object to extend
  * @arg {Object[]} sources - Objects to copy properties from.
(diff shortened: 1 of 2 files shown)


lib/sinon/util/core/extend.js
```diff
@@ -85,7 +85,18 @@ module.exports = function extend(target /*, sources */) {
...
+
+        // If the property is 'name' but is not [[writable]], skip it
+        if (prop !== "name" || destOwnPropertyDescriptor.writable) {
+            Object.defineProperty(dest, prop, {
+                configurable: sourceOwnPropertyDescriptor.configurable,
+                enumerable: sourceOwnPropertyDescriptor.enumerable,
+                writable: sourceOwnPropertyDescriptor.writable,
+                value: sourceOwnPropertyDescriptor.value
+            });
+        }
```
mroderick, line 99: The  comment explains the idea, the implementation uses inversed logic, making it difficult to follow.

Instead, I suggest dropping the comment and make the implementation say what the comment said:

```suggestion
        if (prop === "name" && !destOwnPropertyDescriptor.writable) {
        	return;
        }

        Object.defineProperty(dest, prop, {
            configurable: sourceOwnPropertyDescriptor.configurable,
            enumerable: sourceOwnPropertyDescriptor.enumerable,
            writable: sourceOwnPropertyDescriptor.writable,
            value: sourceOwnPropertyDescriptor.value
        });
```

test/extend-test.js
```diff
@@ -77,4 +77,30 @@ describe("extend", function() {
...
+        var result = extend(object1, object2);
+
+        var expected = {
+            prop1: null,
+            prop2: "hey",
+            name: "not-writable"
+        };
+
+        assert.equals(result, expected);
+    });
```
mroderick, line 105: Mocha's [`context` keyword in the BDD interface](https://mochajs.org/#bdd) is a convenient way of organising code. It allows us to only specify the expectation in the `it` block.

```suggestion
    context("when 'name' property is not writable", function(){
        it("does not attempt to write to the property", function() {
            var object1 = { prop1: null };

            Object.defineProperty(object1, "name", {
                configurable: false,
                enumerable: true,
                value: "not-writable",
                writable: false
            });

            var object2 = {
                prop2: "hey",
                name: "write-attempt"
            };

            var result = extend(object1, object2);

            var expected = {
                prop1: null,
                prop2: "hey",
                name: "not-writable"
            };

            assert.equals(result, expected);
        });
    });
```

mroderick (changes requested):
Thank you for your pull request to fix this issue 💯 

I have a couple of stylistic suggestions, trying out the multiline suggestions feature in the process.

sjoseph7 (commented):
I've committed the changes - thanks for the feedback 👍🏾

sjoseph7 (commented):
~ Linted ~

mroderick:

PR #2324, opened 2021-01-03
Use prettier v2

mroderick:
This PR upgrades `prettier` to latest and removes local configuration, adopting `prettier` default options.

#### Solution  - optional

* Upgrade `prettier`
* Remove `.prettierrc`, accepting the default options
* Re-format all JavaScript files using `prettier`
* Configure editors via `.editorconfig`

1. Check out this branch
1. `npm install`
1. `npm run prettier:check`
1. `npm run lint`
1. `npm run test-cloud`
1. Observe that there are no warnings or errors

#### Checklist for author

- [x] `npm run lint` passes


mantoni (approved):
👍

mroderick:
I'd like to merge https://github.com/sinonjs/sinon/pull/2251 before this one, so IE users can have one final fix, before a new major release that drops support for that runtime.

mroderick:
Working on this PR with @hexeberlin today, we discovered that my assumptions about the Prettier upgrade were flawed.

At https://prettier.io/docs/en/options.html#trailing-commas, it is specified that _Default value changed from none to es5 in v2.0.0_

As in [the MDN article about trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas), it's clear that ES5 trailing commas are supported in IE11

Therefore, upgrading to `prettier@v2` doesn't need to trigger a `MAJOR` revision, as ES5 trailing commas are supported in the runtimes that we currently support.

This can be verified by checking out this branch and running `npm run test-cloud`, which will pass in IE11 and Safari 9.

I've updated the label for this PR, to use `semver:patch`.

This means that we can leave the dropping of IE11 to when we import an upgraded `eslint-config-sinon`, that allows for more modern syntax, which actually breaks in legacy runtimes.

Every day is a school day...

mroderick:
I still haven't figured out how to get `prettier --check` to show what's actually wrong, leading to a poorer development experience than when using `eslint-plugin-prettier`

mantoni:
It doesn't matter what is wrong. If a file fails the check, it needs to be formatted with prettier. Then it will pass the check. The whole point is that you don't make formatting changes manually anymore.

mroderick:
> It doesn't matter what is wrong. If a file fails the check, it needs to be formatted with prettier. Then it will pass the check. The whole point is that you don't make formatting changes manually anymore.

That pretty much destroys staging parts of files, which I do all the time, to be able to craft a good git narrative

mantoni:
> That pretty much destroys staging parts of files

Use a prettier plugin. There is one for every editor. I configured mine to detect prettier configs to have it enabled only for projects that have one. It then formats on save.

PR #2327, opened 2021-01-04
Remove npm run all

mroderick:
In 0d1e40b the last use of `run-p` was removed, which is arguably the whole justification for this package.

It is trivial to run commands *serially* using `npm run` and `&&`, making this package superfluous.

Removing it, reduces the maintenance burden slightly and will speed up package install
slightly.

1. Check out this branch
2. `npm install`
3. `npm test`
4. Observe that the test command still works

 #### Checklist for author

- [x] `npm run lint` passes


mantoni:
I never understood why `run-s` even exists. AFAIK `&&` was always handled by npm cross platform.

mantoni (approved):

fatso83:
@mantoni Succinctness, I would guess. NPM scripts are supposed to fit in a single line, and there is often less than 60 chars remaining, so the difference between `npm run foo && npm run bar && npm run baz` vs `run -s foo bar baz` can have a meaningful difference (18 vs 41 chars).

mantoni:
@fatso83 Aaah, that makes sense. Thank you.

PR #2329, opened 2021-01-05
Replace @sinonjs/formatio with util.inspect

mroderick:
Replace `@sinonjs/formatio` with Node's `util.inspect`.

 #### Background

As part of a larger effort, we are retiring `@sinonjs/formatio`. When it was created Node was
quite immature and there were no other community projects for solid object formatting.

Ten years later, Node is mature and has great support for formatting all sorts of values.

Removing the use of `@sinonjs/formatio` in favour of Node's `util.inspect` removes an entire
project's worth of maintenance burden from the organisation.

In practial terms, it means that some output from `sinon` will look slightly different. This won't
cause a new major release as this output is only intended to be read by humans and not be part of
an API promise.

See:

* https://github.com/sinonjs/referee/pull/183
* https://github.com/sinonjs/referee-sinon/pull/98
* https://github.com/sinonjs/samsam/pull/182
* https://github.com/sinonjs/nise/pull/139

1. Check out this branch
2. `npm ci`
3. Observe that tests pass
4. `npm ls @sinonjs/formatio`
5. Observe that the dependency has been removed

 #### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/util/core/format.js
+++ b/lib/sinon/util/core/format.js
@@ -1,12 +1,6 @@
 "use strict";
 
-var formatio = require("@sinonjs/formatio");
-
-var formatter = formatio.configure({
-    quoteStrings: false,
-    limitChildrenCount: 250
-});
-
+var inspect = require("util").inspect;
 var customFormatter;
 
 function format() {
@@ -14,7 +8,7 @@ function format() {
         return customFormatter.apply(null, arguments);
     }
 
-    return formatter.ascii.apply(formatter, arguments);
+    return inspect.apply(inspect, arguments);
 }
 
 format.setFormatter = function(aCustomFormatter) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -495,15 +495,6 @@
         "@sinonjs/commons": "^1.7.0"
       }
     },
-    "@sinonjs/formatio": {
-      "version": "5.0.1",
-      "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-5.0.1.tgz",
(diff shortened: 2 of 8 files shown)


fatso83:
What has happened to prompt the change?

mantoni:
@fatso83 We've been experimenting with `util.inspect` as an alternative in referee (see https://github.com/sinonjs/referee/pull/183) and I personally find that it delivers better results, and it's consistent with `console.log` output on node. We would also have one less project to maintain.

mroderick:
I can't wrap my head around how to write an elegant test for `it("assert.calledWith exception message with large object arguments", function() {` in `test/assert-test.js` using `util.inspect` in the test.

@mantoni can you see your way through that one?

mantoni:
@mroderick That's a really tricky one. I played with it a little, but couldn't find a cross-environment solution either. How about having two different test cases, one for node and one for browsers?

mroderick:
> @mroderick That's a really tricky one. I played with it a little, but couldn't find a cross-environment solution either. How about having two different test cases, one for node and one for browsers?

`util.inspect` does different things in different environments. I've pushed a commit that detects the different styles and uses slightly different expectations.

Do you think that is good enough?

mantoni:
Feature detection of sorts 🙂. That seems like the best we can do. 👍

mroderick:
Credit to @hexeberlin for assisting me in the investigation into how the object formatting was coming out differently between Node and Chromium (and the other browsers) 🙇

mantoni (approved):
🎉

mroderick:
This has been published as `sinon@9.2.3`

PR #2343, opened 2021-02-23
Upgrade @sinonjs/eslint-config

mroderick:
This PR upgrades ~`eslint-config-sinon`~ `@sinonjs/eslint-config` and fixes all uncovered lint violations.

#### Purpose

Adopt the latest `@sinonjs/eslint-config` in order to allow modern JavaScript syntax and ease the maintenance burden of keeping all the development tools up to date.

#### Background

https://hackmd.io/@mroderick/sinon-rfc-drop-legacy-browsers

#### Solution

* Add latest version of `@sinonjs/eslint-config` as `devDependency`
* Fix all lint violations
* Configure pre-commit hook and CI to ignore lint warnings. This allows contributions before we add all the missing JSDoc comments. We can still see all warnings by running `npm run lint`
* Uninstall no longer needed linting dependencies

#### TO DO

- [x] Install `@sinonjs/eslint-config`
- [x] Remove `eslint`
- [x] Remove `eslint-config-prettier`
- [x] Remove `eslint-plugin-ie11`
- [x] Remove `eslint-plugin-mocha`
- [x] Remove `eslint-plugin-prettier`

#### How to verify

1. Check out this branch
2. `npm install`
3. `npm run lint`
4. Observe that the only warnings are `jsdoc/*`
5. Try creating an `async` function without an `await` statement
6. Observe that the linter complains about it
7. Add an `await` statement
8. Observe that the linter allows `async` functions

#### Checklist for author

- [x] `npm run lint` passes


mantoni:
I was able to verify this locally according to the provided instructions. Nice work 👍

The things I noticed folllowing the instructions:

- `npm link @sinonjs/eslint-config` has to be executed after installing the new dependencies, because `npm install` removes the link.
- `eslint-plugin-mocha` doesn't need to be installed because the dependency already exists.
- `eslint-plugin-compat` and `eslint-plugin-jsdoc` should be installed with `-D`.

package.json
```diff
@@ -36,7 +36,7 @@
     "test-headless": "mochify --no-detect-globals --recursive -R dot --grep WebWorker --invert --plugin [ proxyquire-universal ] \"test/**/*-test.js\"",
     "test-coverage": "nyc npm run test-headless -- --transform [ babelify --ignore [ test ] --plugins [ babel-plugin-istanbul ] ]",
     "test-cloud": "npm run test-headless -- --wd",
-    "test-webworker": "mochify --no-detect-globals --https-server 8080 --no-request-interception test/webworker/webworker-support-assessment.js",
+    "test-webworker": "mochify --no-detect-globals --https-server 8080 --no-request-interception test/webworker/webworker-support-assessment-test.js",
```
mantoni, line 39: Wasn't this named without `-test` on purpose to exclude it from the other test runs?

mantoni (approved):
Just one minor question. I'm very happy to see this land. Thank you ❤️

package.json
fatso83: Ooh, I don't remember exactly, but I think you are right. We should not run it as part of a Mocha suite, just when explicitly triggered, and if running Mocha recursively, then this might be included.

mroderick: You're right! Thank you for spotting this.

I've updated the affected files to use inline configuration of ESLint, so they will not trip up with the new configuration and also won't be matched and run with the other tests.

🙏

mroderick: That replaced the silly commit with two new ones, at the top of the PR

mantoni (approved):
👍🚀

mroderick:
This has been published to the npm registry as `sinon@10.0.0` 🎉

PR #2354, opened 2021-04-13
.matching function implemented

VaclavObornik:
#### Purpose (TL;DR)
To be able distinguish calls in cases, where exact arguments are not known. For example functions where the first argument is dynamic object with a constant part - we cannot use withArgs() behaviour in such case, because of its deep comparsion.

1. Check out this branch
2. `npm install`
3. run test(s) introduced in this PR

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/_releases/v3.1.0/spies.md
+++ b/docs/_releases/v3.1.0/spies.md
@@ -202,6 +202,23 @@ Creates a spy that only records [calls][call] when the received arguments match 
 }
 ```
 
+#### `spy.matching((arg1[, arg2, ...]) => boolean);`
+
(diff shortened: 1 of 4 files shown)


mantoni:
Thanks for contributing 🙂

Isn't this use case already covered by `spy.calledWithMatch(arg1, arg2, ...)`?

It's a shortcut for `spy.calledWith(sinon.match(arg1), sinon.match(arg2), ...)`.

The Sinon `match` API also accepts a function to create a custom matcher.

fatso83:
Thanks for taking the time, but I _think_ this is superfluous. Your example test could be written using the existing matcher functionality, like in the following code, which is basically the same (somewhat shorter even): https://runkit.com/fatso83/6076d62d0d80fa00130c7f29

VaclavObornik:
Thank you guys! 
Yes you are right, I didn't notice the possibility of matchers in docs. 

Closing the PR.

Issue #2356, opened 2021-04-21
Emit a warning if sandbox restore has not been called or have a non-sandbox API


DullReferenceException:
**Is your feature request related to a problem? Please describe.**
Our team has gone many many years using this wonderful software unaware that sinon retained references to fakes in a global sandbox. [It's documented well](https://sinonjs.org/releases/v10.0.1/general-setup/), but either we didn't find our way to that specific page before or we just didn't RTFM properly. It wasn't until our OOM issues got really frustrating with a large test suite that we finally figured out that we weren't restoring this global sandbox.

**Describe the solution you'd like**
There are two things that could have helped us here:
- Outputting a warning if the global sandbox accumulates some threshold of object fakes, notifying the user that you may have a memory leak due to not releasing the global sandbox
- Have a non-sandboxed API available. We have never needed/used the sandbox because we don't usually stub/spy live objects, and when we do we retain a reference to the stub and restore it ourselves. It is not intuitively obvious that if you have `const someStub = sinon.stub().returns('foo');` that it's retained elsewhere in case you wanted to do a bulk operation on all your fakes.

mantoni:
I like the idea of a (configurable?) threshold for registered fakes in the default sandbox.

Would you like to send a PR for this to help avoid this frustration for others?

DullReferenceException:
Sure, will try to make the time for that.

mroderick:
Wouldn't it be better if manually restoring stubs didn't cause memory leaks in the first place?

Did you look into a solution that removes the reference from the sandbox that is tracking the stub?

DullReferenceException:
It makes sense that if you're stubbing live objects (`sinon.stub(someObject, 'prop')`) within the context of a sandbox that you would want the convenience of restoring all stubbed objects at the sandbox level, bringing those objects back to their original state. But why does sinon need to keep track of standalone `sinon.stub()` or `sinon.spy()` instances? Regular garbage collecting should be sufficient. Needing to call `sandbox.restore()` to dereference those standalone fakes is counter-intuitive.

mroderick:
What I meant was that we could have found a solution under the hood, that would remove the stub from the collection when you call it's own `.restore()` method.

Nothing new for the users to worry about and no messages in the console either.

I think #2357 got merged too soon, before other options had been explored.

Ping @fatso83

DullReferenceException:
[...]

PR #2357, opened 2021-04-21
Warn of potential memory leaks

DullReferenceException:

- Warn on the console if a potential memory leak is detected
- Allow adjusting of the leak detection threshold

Fixes https://github.com/sinonjs/sinon/issues/2356

1. Check out this branch
2. `npm install`
3. Call sinon.stub() 10,001 times in a loop
4. Verify that a console warning is output

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -334,3 +334,7 @@ Verifies all mocks created through the sandbox.
 #### `sandbox.verifyAndRestore();`
 
 Verifies all mocks and restores all fakes created through the sandbox.
+
+#### `sandbox.leakThreshold`
+
+Gets/sets the threshold at which memory leak detection warnings are logged.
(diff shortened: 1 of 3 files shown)


mantoni:
Great! Thank you 👍 

Can I ask you to use `printWarning` from our commons library?
https://github.com/sinonjs/commons/blob/e909e9746c32ba6c3c67e64b9ca5b62644cf343e/lib/deprecated.js#L48

This is safer and tests across various platforms to behave nicely.

I'm also not sure if the default threshold of `1000` isn't a bit low, since APIs like `sinon.createStubInstance` create lots of instances at once. What do you think @mroderick?

DullReferenceException:
Updated, bumping default threshold to `10000` and using `deprecated.printWarning` instead.

fatso83:
Great stuff and something that should have been there long ago. Not totally sold on the `track()` naming, though. I think it was easier to reason about what `push(collection, spy)` did than `track(spy)`, so I would have preferred if the tracking bits were internal details of "pushing into internal collection", as the latter is the main point. Maybe a simple rename is all that's needed (`addToCollection(spy)`?). These is all quite minor details, though, and we could merge it as is.

DullReferenceException:
Good point on naming. Renamed to `addToCollection` as suggested.

fatso83:
Thanks!

PR #2363, opened 2021-04-28
Split job into more steps

mroderick:
This PR improves the use of GitHub actions by splitting the workflow into several jobs that can be run in parallel as well as removing the redundant work done in CircleCI.


fatso83:
Ha, I trust your word on this 😄

Issue #2367, opened 2021-05-03
sinon.promise()


mantoni:
When testing callback based code, we can create fakes that yield (e.g. `sinon.fake.yield(null, "result")`), and we can also programatically invoke the callback somewhere in a test (e.g. `myFake.callback(null, "result")`).

When testing promise based code, we only have the first option with `sinon.fake.resolves("result")`, and there is no convenient way to have a fake return a promise and control when it is invoked.

I'm currently helping myself with something like this:

```js
const resolver = sinon.fake();
const promise = new Promise(resolver);
const fake = sinon.fake.returns(promise);

// trigger something

resolver.firstCall.firstArg("result");
await promise;
```

## Proposal

Here is an example using fakes, but this would work the same way with stubs:

```js
const promise = sinon.promise();
const fake = sinon.fake.returns(promise);

// trigger something

await promise.resolve("result");
```

The sinon promise could be a regular promise and expose these additional functions:

- `resolve(value)`
- `reject(value)`

and these properties:

- `resolvedValue`
- `rejectedValue`

Please let me know your thoughts.

fatso83:
This seems like a great addition

mroderick:
I don't understand the examples ...

mantoni:
Here is a simple, runnable example:

```js
'use strict';

const fs = require('fs/promises');

async function readPackageJson() {
  const json = await fs.readFile('package.json', 'utf8');
  return JSON.parse(json);
}

// ----------------------------------------------------------------

const assert = require('assert');
const sinon = require('sinon');

describe('readPackageJson', () => {
  afterEach(() => {
    sinon.restore();
  });

  it('reads package.json before invoking JSON.parse with result', async () => {
    const resolver = sinon.fake();
    const promise = new Promise(resolver);
    sinon.replace(fs, 'readFile', sinon.fake.returns(promise));
    sinon.replace(JSON, 'parse', sinon.fake.returns({}));

    readPackageJson();

    sinon.assert.calledWith(fs.readFile, 'package.json', 'utf8');
    sinon.assert.notCalled(JSON.parse);

    const json = '{"version":"1.0.0"}';
    // The following causes the promise to resolve and the await to complete in the implementation.
    // It would read nicer as `await promise.resolve(json)`.
    resolver.firstCall.firstArg(json);
    await promise;

    sinon.assert.calledWith(JSON.parse, json);
  });
});
```

mroderick:
Now I get it. Thank you!

PR #2369, opened 2021-05-06
Add sinon.promise() implementation

mantoni:
#### Purpose (TL;DR)

This is an implementation of the feature request described in #2367 (`sinon.promise()`).

#### Background

The state of a promise can only be controlled by the promise itself. Also, promises do not expose their state, which makes it difficult to work with them in tests.

#### Solution

The `sinon.promise([executor])` function returns a new promise with this additional API:

- `status`: The internal status of the promise, one of `pending`, `resolved` or `rejected`.
- `resolvedValue`: The promise resolved value.
- `rejectedValue`: The promise rejected value.
- `resolve(value)`: Resolves the promise with the given value. Throws if the promise is not `pending`.
- `reject(value)`: Rejects the promise with the given value. Throws if the promise is not `pending`.

A new promise can be created with a custom executor. This may be a `fake` or `stub` instance, or a custom function. If the custom executor resolves or rejects the promise, the above fields are populated accordingly.

__Example:__

```js
var promise = sinon.promise(function (resolve) {
  resolve(42);
});

assert.equals(promise.status, "pending");
await promise;
assert.equals(promise.status, "resolved");
assert.equals(promise.resolvedValue, 42);
```

Furthermore the `promise.resolve(value)` and `promise.reject(value)` functions return promises, so that the caller can `await` until the sinon promise has settled. These promises always resolve successfully.

__Example:__

```js
var promise = sinon.promise();
var error = new Error();

await promise.reject(error);

assert.equals(promise.status, "rejected");
assert.same(promise.rejectedValue, error);
```

#### How to verify

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/promise.js
+++ b/lib/sinon/promise.js
@@ -1,0 +1,84 @@
+"use strict";
+
+var fake = require("./fake");
+var isRestorable = require("./util/core/is-restorable");
+
+var STATUS_PENDING = "pending";
+var STATUS_RESOLVED = "resolved";
+var STATUS_REJECTED = "rejected";
+
+/**
+ * Returns a fake for a given function or undefined. If no functino is given, a
+ * new fake is returned. If the given function is already a fake, it is
+ * returned as is. Otherwise the given function is wrapped in a new fake.
+ *
+ * @param {Function} [executor] The optional executor function.
+ * @returns {Function}
+ */
+function getFakeExecutor(executor) {
+    if (isRestorable(executor)) {
+        return executor;
+    }
+    if (executor) {
+        return fake(executor);
+    }
+    return fake();
+}
+
+/**
+ * Returns a new promise that exposes it's internal `status`, `resolvedValue`
+ * and `rejectedValue` and can be resolved or rejected from the outside by
+ * calling `resolve(value)` or `reject(reason)`.
+ *
+ * @param {Function} [executor] The optional executor function.
+ * @returns {Promise}
+ */
+function promise(executor) {
+    var fakeExecutor = getFakeExecutor(executor);
+    var sinonPromise = new Promise(fakeExecutor);
+
+    sinonPromise.status = STATUS_PENDING;
+    sinonPromise
+        .then(function (value) {
+            sinonPromise.status = STATUS_RESOLVED;
+            sinonPromise.resolvedValue = value;
+        })
+        .catch(function (reason) {
+            sinonPromise.status = STATUS_REJECTED;
+            sinonPromise.rejectedValue = reason;
+        });
+
+    /**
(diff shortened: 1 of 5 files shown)


lib/sinon/promise.js
```diff
@@ -0,0 +1,84 @@
...
+     * @param {*} value
+     * @param {Function} callback
+     */
+    function finalize(status, value, callback) {
+        if (sinonPromise.status === STATUS_PENDING) {
+            sinonPromise.status = status;
+            callback(value);
+            return;
+        }
+        throw new Error(`Promise already ${sinonPromise.status}`);
```
mroderick, line 64: You can reduce indentation slightly by checking for error conditions at the top of the function

```suggestion
        if (sinonPromise.status !== STATUS_PENDING) {
            throw new Error(`Promise already ${sinonPromise.status}`);
        }
    
        sinonPromise.status = status;
        callback(value);        
```

test/promise-test.js
```diff
@@ -0,0 +1,225 @@
...
+    var value = null;
+    promise
+        .then(function (val) {
+            status = "fulfilled";
+            value = val;
+        })
+        .catch(function (reason) {
+            status = "rejected";
+            value = reason;
+        });
```
mroderick, line 17: Is there any reason not to use `await` wrapped in `try/catch` instead of `then/catch` on the promise?

```suggestion
    try {
        status = "fulfilled";
        value = await promise;
    } catch (reason) {
        status = "rejected";
        value = reason; 
    }
```

mroderick (changes requested):
This looks good 👍 

Would you mind adding some documentation in the Jekyll site, so users can learn how to use it?

mantoni:
Thanks for reviewing. Before spending time on writing documentation, I wanted to get some feedback.

We can keep iterating on this PR for a bit until everyone is happy with the API and implementation.

test/promise-test.js
mantoni: Yes, there is a good reason: We don't want to `await` the promise, because we want to know if the promise is still `pending`. I committed your suggestion, but then there's a test failure and I reverted.

mantoni:
I have made some minor changes and added a documentation page. I cannot see the new documentation page locally for reasons I don't understand.

mroderick (approved):

fatso83:
@mantoni This is a problem with the docs setup: it does not have a way of showing `release-source` (not copied to _site_), so you will not see your changes until they have been versioned ... This is less than great, but I have fixed it locally by manualyl including release-source in the build and not commiting the change.

fatso83 (approved):

PR #2360, opened 2021-04-23
Fake docs improvement

srknzl:
#### Purpose (TL;DR) - mandatory

Tries to improve documentation for the fake API. Resolves #2070 

#### Solution

I went over this [comment](https://github.com/sinonjs/sinon/issues/2070#issuecomment-823090909):

- [x] Make sure the list of static factory methods on sinon.fake is complete

* yes, the list is complete except for usingPromise

- [ ] Convert and/or add some runnable Runkit examples (very easy, but takes a bit of time)

* On it, I wanted open this so that you can check out current status.

- [x] Update the link in Instance properties to point to the same section in Spies.

* Only `firstArg` was missing a reference, added that. Changed others' links so that they are linked to the exact same header.

- [x] State that the listed props are just one of the many from the spies documentation and that you can refer to that for the complete list

- [x] Clarify the wording so that it becomes clear that 1. The Fakes API is an alternative to the Stubs and Spies API that can fully replace all such uses. 2. It is intended to be simpler to use, maintain and has a lot smaller API surface and avoids many bugs by having quazi-immutable Fake instances (the behavior cannot be changed after creation (unlike Stubs), but they do have internal state that changes when used, like the callCount and other stats) 3. A Fake instance has the API of a Spy (

* I've done my best to make it clear. 

#### How to verify - mandatory

1. Check out this branch
2. `npm install`
3. cd docs
4. gem install bundler
5. bundle install
6. bundle exec jekyll serve
7. Check the website

#### Checklist for author

- [x] `npm run lint` passes
- [x] References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

PR #2381, opened 2021-06-01
Use runkit snippets in fake documentation

srknzl:

In #2360, I forgot to replace snippets with runkit samples. This pr adds runkit samples to fakes.md and makes small changes to the fakes.md required due to using runkit examples 

1. remove `_releases/latest` and copy `release-source/release` to `_releases` folder with the name `latest`
2. cd docs
3. gem install bundler
4. bundle install
5. bundle exec jekyll serve
6. go to http://localhost:4000/releases/latest/fakes/ to see new version of fakes.md

#### Checklist for author

- [x] `npm run lint` passes(only docs change is done, and no linting error in docs folder. there are some errors in `scripts` folder but they are unrelated)

cc @fatso83

### DIFF
--- a/docs/release-source/release/examples/fakes-12-adding-fake-to-system-under-test.test.js
+++ b/docs/release-source/release/examples/fakes-12-adding-fake-to-system-under-test.test.js
@@ -4,7 +4,7 @@ const referee = require("@sinonjs/referee");
 const assert = referee.assert;
 
 describe("FakeTest", function () {
-    it("should have working lastArg property", function () {
+    it("should be able to be added to the system under test", function () {
         const fake = sinon.fake.returns("42");
 
         sinon.replace(console, "log", fake);
--- a/docs/release-source/release/fakes.md
+++ b/docs/release-source/release/fakes.md
@@ -2,6 +2,19 @@
 layout: page
 title: Fakes - Sinon.JS
 breadcrumb: fakes
+examples:
+  - fakes-1-using-fakes-instead-of-spies
+  - fakes-2-using-fakes-instead-of-stubs
+  - fakes-3-creating-without-behaviour
+  - fakes-4-creating-with-custom-behaviour
+  - fakes-5-returns
+  - fakes-6-throws
+  - fakes-7-yields
+  - fakes-8-yields-async
+  - fakes-9-callback
+  - fakes-10-firstArg
+  - fakes-11-lastArg
+  - fakes-12-adding-fake-to-system-under-test
 ---
 
 ### Introduction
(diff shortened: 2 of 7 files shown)


docs/release-source/release/examples/fakes-12-adding-fake-to-system-under-test.test.js
```diff
@@ -4,7 +4,7 @@ const referee = require("@sinonjs/referee");
 const assert = referee.assert;
 
 describe("FakeTest", function () {
-    it("should have working lastArg property", function () {
```
srknzl, line 7: this was leftover

docs/release-source/release/fakes.md
```diff
@@ -129,33 +89,15 @@ If an `Error` is passed as the `value` argument, then that will be the value of
...
-// file content
-// end of this event loop
-```
+<div data-example-id="fakes-7-yields"></div>
 
 #### `sinon.fake.yieldsAsync([value1, ..., valueN]);`
 
 Similar to `yields`, `yieldsAsync` also returns a function that when invoked, the function expects the last argument to be a callback and invokes that callback with the same previously given values. However, the returned function invokes that callback asynchronously rather than immediately, i.e. in the next event loop.
 
-Compare the output of the code example below with the output of the code example above for `yields` to see the difference.
```
srknzl, line 147: no more output comparison, so the change is needed

docs/release-source/release/fakes.md
```diff
@@ -172,106 +114,32 @@ The instance properties are the same as those of a [`sinon.spy`][spies]. The fol
...
-//
-f.lastCall.callback === cb2;
-// true
-```
+<div data-example-id="fakes-9-callback"></div>
 
 #### `f.firstArg`
 
 This property is a convenient way to get a reference to the first argument passed in the last call to the fake.
+The same convenience has been added to [spy calls](../spy-call#spycallfirstarg).
```
srknzl, line 124: runkit samples have the spy call part indicated with a comment

fatso83:
Perfect! The only thing that is a bit of a shame with RunKit is that you don't know that it is loading until it has finished loading. And if RunKit is down for some reason, we have no examples to show. That's why I would like to replace the 

```html
<div data-example-id="fakes-11-lastArg"></div>
```

with something like
```html
@includeExample("fakes-11-lastArg")
```

that would produce something like 
```html
<div data-example-id="fakes-11-lastArg">
<!-- some syntax colouring tag here -->
require("@fatso83/mini-mocha").install();
const sinon = require("sinon");
const referee = require("@sinonjs/referee");
const assert = referee.assert;

describe("FakeTest", function () {
    it("should have working lastArg", function () {
        const f = sinon.fake();
        const date1 = new Date();
        const date2 = new Date();

        f(1, 2, date1);
        f(1, 2, date2);

        assert.isTrue(f.lastArg === date2);
        // spy call methods:
        assert.isTrue(f.getCall(0).lastArg === date1);
        assert.isTrue(f.getCall(1).lastArg === date2);
        assert.isTrue(f.lastCall.lastArg === date2);
    });
});
</div>
```

That would make it show working examples even if javascript should crash or the RunKit service being down.

fatso83:
P.S. That is a different PR, btw :smile:

srknzl:
So `includeExample` macro(is it called macro?) does not exists and should be implemented in a new PR?

fatso83:
> Does it copy from examples folder and use the code as fallback mechanism to Runkit?

Exactly! 

TBH, it's probably not worth the (not so big) effort of creating a Liquid Tag (a Tag is the name the Liquid templating system calls what you called a macro). All I was thinking about was something like just using [`relative_include`](https://jekyllrb.com/docs/includes/) to include the source verbatim, but if you want to have a look at how it works under the hood, it's not that heavy: https://jekyllrb.com/docs/plugins/tags/

The least effort way (without creating a Tag) is probably just doing something manual  like
```
<div data-example-id="fakes-11-lastArg">
{% include_relative examples/fakes-11-lastArg.test.js %}
</div>
```

As you see, there is some repetition, making it easy to make mistakes, and it does not solve how to add syntax highlighing for the fallback, but maybe it's good enough for a fallback?

Issue #2361, opened 2021-04-26
Add support for destructured import definitions (esm module imports)


iambumblehead:
Add support for destructured definitions in esm-module-only environments.

The requested feature is to support this style of import
```javascript
import { stub } from 'sinon';
```

Currently, an esm-module-only environment must import sinon as shown,
```javascript
import sinon from 'sinon';

sinon.stub();
```

fatso83:
OK, what is the use case? Purely ergonomics? Sinon is a test utility so there is no bundling performance improvement to be had and doing member imports makes it even less obvious that all the fakes and stubs created belong to the Sinon sandbox.

iambumblehead:
I only wanted to let you know sinon does not support destructured imports. Personally, I prefer the supported way of importing one default namespace, but am aware of those who prefer destructured imports.

All core modules and most public packages (I've used) support _both_ default and destructured import assignments.

A few days ago, I removed babel from an application using destructured sinon imports. Those no longer worked using the un-transpiled sources and it was necessary to update each one to import the default sinon namespace.

fatso83:
It is a bit weird that this should not work, seeing that #1833 was supposed to do exactly that. I'll have a quick look.

fatso83:
OK, I see now that this does not work without modifications in the standard Node case:
```javascript
$ cat test.mjs
import { spy } from "sinon";

console.log(spy);

$ node test.mjs
file:///tmp/test1/test.mjs:1
import { spy } from "sinon";
         ^^^
SyntaxError: Named export 'spy' not found. The requested module 'sinon' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'sinon';
const { spy } = pkg;
```

Explicitly importing the esm  export file and changing the file ending "fixes" it:
```javascript
$ cat test.mjs
import { spy } from "sinon/pkg/sinon-esm.mjs";

console.log(spy);

$ cp node_modules/sinon/pkg/sinon-esm.js  node_modules/sinon/pkg/sinon-esm.mjs
$ node test.mjs
[Function: spy]
```

Reading up [on the Node docs](https://nodejs.org/api/packages.html#packages_type) now. Also found [an interesting article](https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html) that seems to cover the various other use cases (Webpack, browsers, etc) for hybrid modules.

fatso83:
[...]

PR #2382, opened 2021-06-10
Enable esm named exports

perrin4869:
#### Adds support for esm named exports

Fixes issue #2361

I tried to keep this as backwards compatible as possible, but some minor (undocumented) changes were made which may break people importing from the `pkg` directory directly.
Specifically we can't do `require('sinon/pkg/sinon.js')` or `require('sinon/pkg/sinon-no-sourcemaps.js')` anymore, as these were changed into `cjs` files.
An additional `pkg/sinon.js` file is produced, identical to `pkg/sinon.cjs`, to keep the WebWorker functional. `importScripts` seems to fail on `cjs` files.

#### How to verify

1. Check out this branch
2. `npm ci`
3. `npm test`
4. `import { stub } from "sinon"` from an esm javascript file on node >= 12 should work

Edit: this is more of a band-aid, since the preferred solution would be an esm rewrite and using rollup to produce a cjs and esm build.

### DIFF
--- a/build.cjs
+++ b/build.cjs
@@ -18,6 +18,11 @@ try {
     // We seem to have it already
 }
 
+/**
+ * @param entryPoint
+ * @param config
+ * @param done
+ */
 function makeBundle(entryPoint, config, done) {
     browserify(entryPoint, config).bundle(function (err, buffer) {
         if (err) {
@@ -39,7 +44,8 @@ makeBundle(
     },
     function (bundle) {
         var script = preamble + bundle;
-        fs.writeFileSync("pkg/sinon.js", script);
+        fs.writeFileSync("pkg/sinon.cjs", script);
+        fs.writeFileSync("pkg/sinon.js", script); // WebWorker can only load js files
     }
 );
 
@@ -53,7 +59,7 @@ makeBundle(
     },
     function (bundle) {
         var script = preamble + bundle;
-        fs.writeFileSync("pkg/sinon-no-sourcemaps.js", script);
+        fs.writeFileSync("pkg/sinon-no-sourcemaps.cjs", script);
     }
 );
 
--- a/lib/package.json
+++ b/lib/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "commonjs"
+}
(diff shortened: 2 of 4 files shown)


fatso83:
Hi, Perrin. Thanks for this. As you probably saw in https://github.com/sinonjs/sinon/issues/2361#issuecomment-846581991, I had a go at this and found that, as you say, the best solution would just be to rewrite into ESM. I did have a go, but I found that the stuff I did always broke some kind of use case in some environment, so curious how this checks out.

Thanks!

perrin4869:
Honestly I was thinking that even if it's a band-aid, this PR would still require a major version bump just to be on the safe side.
That said, we could still tweak around the `exports` field to try to make it as backwards compatible as possible.
Dropping support for node 10 would make things slightly simpler I think, it's a good chance as it has reached EOL.
I tested this on node 12 through 16 loading as `commonjs` and `esm`, and got positive results, the rest of the use-cases I wouldn't really know... but I did make sure tests are passing...

fatso83:
I was just thinking of these use cases:
1. bundlers should not start having problems
1. node in ESM mode (i.e. loading usign `import` in `*.mjs` scripts) should work
1. node in require mode should work

Does the second point work with this? Or which envs work and how (i.e. which files need to be imported)?

perrin4869:
1 I haven't tested neither `webpack` not `rollup`, but in theory they should work
2 and 3 I tested the following way:

```
❯ cat package.json
{
  "type": "module",
  "dependencies": {
    "sinon": "github:perrin4869/sinon#esm-named-exports"
  }
}
❯ cat index.js
import sinon from "sinon"
import { stub } from "sinon"

console.log(sinon)
console.log(stub)
❯ cat index.cjs
const { stub } = require('sinon');
console.log(stub);
```

Node >= 12 behaved as you'd expect

mroderick:
Related to the discussion in this PR, I added some detail on converting the whole codebase to ESM: https://github.com/sinonjs/sinon/issues/2361#issuecomment-887454521

perrin4869:
Well, while I can totally get being hesitant about merging this in, this does get the job done for me now, and I am not sure how realistic this esm rewrite really is...
Maybe we could release this as a beta for people to test?

fatso83:
I think this just stalled on me (and or others) just forgetting about it pre-holiday. Not intentional from my side, in any way. I do not mind merging this (AFAI can remember).

fatso83:
Released as Sinon 12! Thanks @perrin4869 <3

fatso83:
@perrin4869 Seems this broke webpack bundling for CJS. See #2411. Any quickfix?

perrin4869:
ah dammit ><;; Let me look into this!

PR #2389, opened 2021-07-12
Refactor fake

mroderick:
This PR has some refactorings and JSDoc for `fake`.

#### Purpose

Have JSDoc so we can generate API docs in the future

#### Discussion

* Is there a way to document members of _functions_, as opposed to _namespaces_?
* How do we document the returned function `f` from most of these methods?

This is not complete, but I think it is a good start. Suggestions for improvements welcome.

1. Check out this branch
2. `npm install jsdoc`
3. `jsdoc lib/sinon/fake.js`
4. `open out/index.html`
5. Navigate to the docs for `fake`
6. Observe that all methods are documented

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/fake.js
+++ b/lib/sinon/fake.js
@@ -9,6 +9,38 @@ var promiseLib = Promise;
 
 module.exports = fake;
 
+/**
+ * Returns a `fake` that records all calls, arguments and return values.
+ *
+ * When an `f` argument is supplied, this implementation will be used.
+ *
+ * @example
+ * // create an empty fake
+ * var f1 = sinon.fake();
+ *
+ * f1();
+ *
+ * f1.calledOnce()
+ * // true
+ *
+ * @example
+ * function greet(greeting) {
+ *   console.log(`Hello ${greeting}`);
+ * }
+ *
+ * // create a fake with implementation
+ * var f2 = sinon.fake(greet);
+ *
+ * // Hello world
+ * f2("world");
+ *
+ * f2.calledWith("world");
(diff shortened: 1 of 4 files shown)


lib/sinon/fake.js
```diff
@@ -5,117 +5,283 @@ var createProxy = require("./proxy");
...
-            lastArg && typeof lastArg === "function" ? lastArg : undefined;
+    return wrapFunc(f);
+}
 
-        proxy.firstArg = firstArg;
-        proxy.lastArg = lastArg;
-        proxy.callback = callback;
+/**
+ * Creates a `fake` that returns the provided `value`, as well as recording all
+ * calls, arguments and returnvalues.
```
khamiltonuk, line 54:
```suggestion
 * calls, arguments and return values.
```

lib/sinon/fake.js
```diff
@@ -5,117 +5,283 @@ var createProxy = require("./proxy");
...
+module.exports = fake;
 
-        if (arguments.length > 0) {
-            firstArg = arguments[0];
-            lastArg = arguments[arguments.length - 1];
-        }
+/**
+ * Returns a `fake` that records all calls, arguments and return values.
+ *
+ * When an `f` argument is supplied, this implementation will be used.
```
khamiltonuk, line 15:
```suggestion
 * When an `function` argument is supplied, this implementation will be used.
```
is `f` a known shorthand, it might be a barrier to understanding to some

PR #2410, opened 2021-11-03
Check call count type

### LINKED ISSUE #2408: Asserting on callCount with a numeric string results in very confusing failure message
**Describe the bug**
This bug was originally reported in `sinon-chai` by @BebeSparkelSparkel as https://github.com/domenic/sinon-chai/issues/136 but turned out to be an issue with Sinon itself. The bug is that we can get the error message `AssertError: expected spy to be called 10 times but was called 10 times` - this happens when specifying `callCount` as a string.

**To Reproduce**
I made a slightly modified version of the original test case so that it just uses Sinon:
```
const n = '10',
      s = sinon.spy();
for(let i=0; i < n; ++i) s();

sinon.assert.callCount(s, n);
```

Run the above, you will get `AssertError: expected spy to be called 10 times but was called 10 times` which is a very confusing error message.

**Expected behavior**

One of two options:
 1. Sinon could coerce numeric strings into numbers, and pass the assertion
 2. Sinon could surface the actual problem, which is that callCount was not a number.

Both of these seem sensible, the first is a little more user-friendly, the second is a little more correct.

Internally it looks like Sinon raises a `TypeError`, which is I think what is actually triggering this failure, so it could just surface that, or it could attempt to convert to a number.

This actually happens with any string: `sinon.assert.callCount(s, 'banana')` will get you `expected spy to be called banana times but was called 10 times`, which is less confusing, but also it would probably be better to instead fail with an error about `banana` not being a number.

- Library version: All versions (tested with most recent releases for v1-11, including 11.1.2)
- Environment: OS X
- Other libraries you are using: None

cincodenada:
#### Purpose (TL;DR)

Fix #2408 by checking the type of the argument to callCount and emitting a specfic error message if it is not a number

#### Details

I also refactored the assert tests a bit while I was here - the current code is manually removing the stack traces in each individual test, but since we already have a `message()` helper function to extract the message, I moved the stack trace stripping up into that, which streamlines the rest of the test cases quite a bit.

I also modified the regex to be more specific, and avoid accidentally stripping the end of messages that happen to have the word "at" in them.

#### How to verify

1. Check out this branch
2. `npm install`
3. Run the following test case:

```
const n = '10',
      s = sinon.spy();
for(let i=0; i < n; ++i) s();

sinon.assert.callCount(s, n);
```

You should get `expected '10' to be a number but was of type string`, rather than the current message which is `expected spy to be called 10 times but was called 10 times`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/proxy-call.js
+++ b/lib/sinon/proxy-call.js
@@ -220,7 +220,7 @@ var callProto = {
             }
         }
         if (this.stack) {
-            // Omit the error message and the two top stack frames in sinon itself:
+            // Emit the error message and the two top stack frames in sinon itself:
             callStr += (this.stack.split("\n")[3] || "unknown").replace(
                 /^\s*(?:at\s+|@)?/,
                 " at "
--- a/lib/sinon/proxy-call.js
+++ b/lib/sinon/proxy-call.js
@@ -220,7 +220,8 @@ var callProto = {
             }
         }
         if (this.stack) {
-            // Emit the error message and the two top stack frames in sinon itself:
+            // If we have a stack, add the first frame that's in end-user code
+            // Skip the first two frames because they will refer to Sinon code
             callStr += (this.stack.split("\n")[3] || "unknown").replace(
                 /^\s*(?:at\s+|@)?/,
                 " at "
--- a/test/proxy-call-test.js
+++ b/test/proxy-call-test.js
@@ -1176,7 +1176,7 @@ describe("sinonSpy.call", function () {
 
         // https://github.com/sinonjs/sinon/issues/1066
         /* eslint-disable consistent-return */
-        it("does not throw when the call stack is empty", function (done) {
+        it("does not throw when the call stack is empty", async function () {
             if (typeof Promise === "undefined") {
                 this.skip();
             }
@@ -1184,21 +1184,36 @@ describe("sinonSpy.call", function () {
             var stub1 = sinonStub().resolves(1);
             var stub2 = sinonStub().returns(1);
 
-            function run() {
-                return stub1().then(stub2);
-            }
+            await stub2(await stub1());
 
-            run()
-                .then(function () {
-                    assert.equals(
-                        stub2.getCall(0).toString().replace(/ at.*/g, ""),
-                        "stub(1) => 1"
-                    );
-                    done();
-                })
-                .catch(done);
+            assert.equals(
+                stub2.getCall(0).toString().replace(/ at.*/g, ""),
+                "stub(1) => 1"
+            );
         });
         /* eslint-enable consistent-return */
+
+        it("includes first stack entry from end-user code", function () {
+            /* We find the first stack frame that points to end-user code and
+             * add it to the error message. We do this by chopping off a
+             * constant number of stack frames, so if this test fails, you
+             * probably need to chop off a different number of frames
+             */
+            if (typeof __filename === "undefined") {
+                this.skip();
+            }
+
+            var object = { doIt: sinonSpy() };
+            object.doIt();
+
+            const [name, stackFrame] = object.doIt
+                .getCall(0)
+                .toString()
+                .split(" at ");
+
+            assert.equals(name, "doIt()");
+            assert.contains(stackFrame, __filename);
+        });
     });
 
     describe("constructor", function () {
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -159,10 +159,16 @@ function createAssertObject() {
         callCount: function assertCallCount(method, count) {
             verifyIsStub(method);
 
-            if (method.callCount !== count) {
-                var msg = `expected %n to be called ${timesInWords(
-                    count
-                )} but was called %c%C`;
+            var msg;
+            if (typeof count !== "number") {
+                msg =
+                    `expected ${format(count)} to be a number ` +
+                    `but was of type ${typeof count}`;
+                failAssertion(this, msg);
+            } else if (method.callCount !== count) {
+                msg =
+                    `expected %n to be called ${timesInWords(count)} ` +
+                    `but was called %c%C`;
                 failAssertion(this, method.printf(msg));
             } else {
                 assert.pass("callCount");
--- a/test/assert-test.js
+++ b/test/assert-test.js
@@ -1438,7 +1438,8 @@ describe("assert", function () {
                         [].slice.call(arguments, 1)
                     );
                 } catch (e) {
-                    return e.message;
+                    // Strip off the stack frames we append
+                    return e.message.replace(/( at.*\(.*\)$)+/gm, "");
                 }
             };
         });
@@ -1454,10 +1455,7 @@ describe("assert", function () {
             this.obj.doSomething();
 
             assert.equals(
-                this.message("notCalled", this.obj.doSomething).replace(
-                    / at.*/g,
-                    ""
-                ),
+                this.message("notCalled", this.obj.doSomething),
                 "expected doSomething to not have been called but was called once\n    doSomething()"
             );
         });
@@ -1469,10 +1467,7 @@ describe("assert", function () {
             this.obj.doSomething();
 
             assert.equals(
-                this.message("notCalled", this.obj.doSomething).replace(
-                    / at.*/g,
-                    ""
-                ),
+                this.message("notCalled", this.obj.doSomething),
                 "expected doSomething to not have been called " +
                     "but was called 4 times\n    doSomething()\n    " +
                     "doSomething()\n    doSomething()\n    doSomething()"
@@ -1486,10 +1481,7 @@ describe("assert", function () {
             this.obj.doSomething();
 
             assert.equals(
-                this.message("notCalled", this.obj.doSomething).replace(
-                    / at.*/g,
-                    ""
-                ),
+                this.message("notCalled", this.obj.doSomething),
                 "expected doSomething to not have been called " +
                     "but was called 4 times\n    doSomething()\n    " +
                     "doSomething(3)\n    doSomething(42, 1)\n    doSomething()"
@@ -1573,10 +1565,7 @@ describe("assert", function () {
             this.obj.doSomething();
 
             assert.equals(
-                this.message("callCount", this.obj.doSomething, 3).replace(
-                    / at.*/g,
-                    ""
-                ),
+                this.message("callCount", this.obj.doSomething, 3),
                 "expected doSomething to be called thrice but was called once\n    doSomething()"
             );
         });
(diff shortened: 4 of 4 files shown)


cincodenada:
I would like to backport this to all current releases, as this bug exists in all released versions - what's the procedure for doing that? Should I just open a separate PRs for each version once this is approved?

fatso83:
Truth be told, we don't have a procedure for backporting ... 🤔 This would involve creating release branches for every major release to accept PRs, so I think it might be just as fast just cherry picking it manually after merging. What would you suggest?

The biggest "consumer" is probably Sinon, and that is a bit more tricky to get right. Due to creating bundles it won't distribute the fix as efficiently.

fatso83:
>  all current releases,

We don't actually have something like "LTS" releases, so in the sense that we need to support different use cases it would be:
- Sinon 1.x (if you really can't handle the breaking 2.x changes)
- Sinon 9 (if you cannot use ES2017+)
- Latest version

What versions would you like to fix this for? I would probably not bother with anything else than Sinon 7-12. I can probs do it if you are fine with that.

cincodenada:
> EDIT: It was 3AM when I wrote this, so not everything makes sense in hindsight. I somehow mixed this with a PR in the fake-timers project.

Ha! No worries, and thanks, that part was a bit confusing 😅

I didn't have any grand designs on backporting, but I did get carried away ("this seems like the job for a simple bash script...") and backported it to all the versions - my fork [now has a `-backport` branch for each version](https://github.com/cincodenada/sinon/branches/all?query=backport), so feel free to grab from those whatever makes sense! It's certainly not a _critical_ issue, it just seemed like it's would be relatively easy to do, so the cost/benefit ratio was still low. I did end up backporting just the fix/test, and not the test refactoring.

fatso83 (changes requested):
Wondering if the fixup of one comment actually turns right to wrong.

fatso83:
Hi, Joel. I was about to do something about this which has been lingering for way too long, but then realized one of the commits might not be doing the right thing. Could you have a quick look at my comment and decide if I am right?

lib/sinon/proxy-call.js
```diff
@@ -220,7 +220,7 @@ var callProto = {
             }
         }
         if (this.stack) {
-            // Omit the error message and the two top stack frames in sinon itself:
+            // Emit the error message and the two top stack frames in sinon itself:
```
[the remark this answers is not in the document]
cincodenada: Okay, yes upon investigation you're totally right here, thanks, I misunderstood what was going on here. I read "in sinon itself" as "add the stack frame to sinon's output" when I now realize I think what it meant is "omit the top two stack frames, which will be in sinon itself".

After re-orienting myself, I do think it's a little confusing because if this code wasn't here, the stack frames wouldn't show up at all, and also because we're actually just pulling out only the third stack frame, rather than stripping off the first two and leaving the rest, right?

So the comment as it was is certainly better than my replacement, but what about something like:
```
          if (this.stack) {
            // If present, add the third stack frame to the error output for context
            // Skip the first two stack frames, which will be within Sinon code
            callStr += (this.stack.split("\n")[3] || "unknown").replace(
                /^\s*(?:at\s+|@)?/,
                " at "
            );
          }
```
(Code is the same, I just left it there for context)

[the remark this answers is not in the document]
fatso83: Definitely an improvement!

test/proxy-call-test.js
```diff
@@ -1176,29 +1176,44 @@ describe("sinonSpy.call", function () {
...
                 this.skip();
             }
 
             var stub1 = sinonStub().resolves(1);
             var stub2 = sinonStub().returns(1);
 
-            function run() {
-                return stub1().then(stub2);
-            }
+            await stub2(await stub1());
```
cincodenada, line 1187: This could be
```
const val = await stub1();
await stub2(val);
```
if that seems better, I couldn't decide which was less confusing 😄

lib/sinon/proxy-call.js
```diff
@@ -220,7 +220,8 @@ var callProto = {
             }
         }
         if (this.stack) {
-            // Omit the error message and the two top stack frames in sinon itself:
+            // If we have a stack, add the first frame that's in end-user code
+            // Skip the first two frames because they will refer to Sinon code
```
cincodenada, line 224: I put this through another round of editing, hopefully this is still clear

test/proxy-call-test.js
```diff
@@ -1176,29 +1176,44 @@ describe("sinonSpy.call", function () {
...
+            object.doIt();
+
+            const [name, stackFrame] = object.doIt
+                .getCall(0)
+                .toString()
+                .split(" at ");
+
+            assert.equals(name, "doIt()");
+            assert.contains(stackFrame, __filename);
+        });
```
cincodenada, line 1216: This test should make sure we're always returning the first end-user stack, even if the internals change. I use `__filename` here because for the test's purposes `proxy-call-test.js` is the "end-user" code, and the variable ensures we're safe from renames or what have you. Let me know if that seems unreasonable/weird.

test/proxy-call-test.js
```diff
@@ -1176,29 +1176,44 @@ describe("sinonSpy.call", function () {
 
         // https://github.com/sinonjs/sinon/issues/1066
         /* eslint-disable consistent-return */
-        it("does not throw when the call stack is empty", function (done) {
+        it("does not throw when the call stack is empty", async function () {
```
cincodenada, line 1179: This test doesn't have anything to do with the rest of the PR, other than being next to the test I was adding.

I figured since it's just in tests async should be fine to use, but I also just did a search and realized the only other async tests are in `promise-test.js` and an es2015-specific test, so if y'all don't do async tests generally, let me know and I'll revert this bit.

cincodenada:
Okay, thanks for catching the comment - I decided to go ahead and add a test for the frame-appending bit too, since I didn't see one already, and converted a test to async while I was there. At this point the test cleanup outweighs the original fix I think, so if it makes more sense to split this into two PRs, let me know and I can rewrite accordingly, ha!

test/proxy-call-test.js
fatso83: TBH I find Promise chaining much more understandable, as in `await stub1().then(stub2)`, but I do not really care all that much either :p Whatever rocks your boat :)

test/proxy-call-test.js
fatso83: We just changed target browsers to evergreen last year, so it's not all that much that has changed since. Adding `async` is fine, as I believe ES2018 is our current ES level.

fatso83 (approved):

fatso83:
Great stuff, thanks for the addition!

fatso83:
Ajajaj, seems that did not pan out that well as it failed when running the browser tests on SauceLabs. These 11 breaking tests are for Firefox:
```
  1464 passing (3s)
  13 pending
  11 failing

  1) assert
       message
         assert.notCalled exception message one call:

      [assert.equals] 'expected doSomething to not have been called but was called once\n    doSomething() at callFn@about:blank line 2 > injectedScript:47353:21' expected to be equal to 'expected doSomething to not have been called but was called once\n    doSomething()'
      + expected - actual

       expected doSomething to not have been called but was called once
      -    doSomething() at callFn@about:blank line 2 > injectedScript:47353:21
      +    doSomething()
      
  fail@about:blank line 2 > injectedScript:8641:21
  fail@about:blank line 2 > injectedScript:8801:17
  assertion@about:blank line 2 > injectedScript:8819:11
  defineAssertion/referee[type][name]@about:blank line 2 > injectedScript:8846:22
  require<[248]</</</<@about:blank line 2 > injectedScript:25036:20
      -    doSomething(1, 3, 'hey') at callFn@about:blank line 2 > injectedScript:47353:21
      -    doSomething(1, 3) at callFn@about:blank line 2 > injectedScript:47353:21
      +    doSomething(1, 3, 'hey')
      +    doSomething(1, 3)
      
  fail@about:blank line 2 > injectedScript:8641:21
  fail@about:blank line 2 > injectedScript:8801:17
  assertion@about:blank line 2 > injectedScript:8819:11
  defineAssertion/referee[type][name]@about:blank line 2 > injectedScript:8846:22
  require<[248]</</</<@about:blank line 2 > injectedScript:25748:20
  callFn@about:blank line 2 > injectedScript:47353:21
  require<[277]</<[32]</</Runnable.prototype.run@about:blank line 2 > injectedScript:47345:13
  require<[277]</<[33]</</Runner.prototype.runTest@about:blank line 2 > injectedScript:47881:10
  next/<@about:blank line 2 > injectedScript:47999:12
  next@about:blank line 2 > injectedScript:47795:14
  next/<@about:blank line 2 > injectedScript:47805:11
  next@about:blank line 2 > injectedScript:47729:14
  next/<@about:blank line 2 > injectedScript:47768:11
  done@about:blank line 2 > injectedScript:47300:7
  callFn@about:blank line 2 > injectedScript:47376:7
  require<[277]</<[32]</</Runnable.prototype.run@about:blank line 2 > injectedScript:47345:13
  next@about:blank line 2 > injectedScript:47743:10
  require<[277]</<[33]</</Runner.prototype.hook/<@about:blank line 2 > injectedScript:47773:9
  timeslice@about:blank line 2 > injectedScript:42885:27
  

  11) assert
       message
         assert.alwaysThrew exception message:

      [assert.equals] 'doSomething did not always throw exception\n    doSomething(1, 3, \'hey\') at callFn@about:blank line 2 > injectedScript:47353:21\n    doSomething(1, 3) at callFn@about:blank line 2 > injectedScript:47353:21' expected to be equal to 'doSomething did not always throw exception\n    doSomething(1, 3, \'hey\')\n    doSomething(1, 3)'
      + expected - actual

       doSomething did not always throw exception
      -    doSomething(1, 3, 'hey') at callFn@about:blank line 2 > injectedScript:47353:21
      -    doSomething(1, 3) at callFn@about:blank line 2 > injectedScript:47353:21
      +    doSomething(1, 3, 'hey')
      +    doSomething(1, 3)
      
  fail@about:blank line 2 > injectedScript:8641:21
  fail@about:blank line 2 > injectedScript:8801:17
  assertion@about:blank line 2 > injectedScript:8819:11
  defineAssertion/referee[type][name]@about:blank line 2 > injectedScript:8846:22
  require<[248]</</</<@about:blank line 2 > injectedScript:25758:20
  callFn@about:blank line 2 > injectedScript:47353:21
  require<[277]</<[32]</</Runnable.prototype.run@about:blank line 2 > injectedScript:47345:13
  require<[277]</<[33]</</Runner.prototype.runTest@about:blank line 2 > injectedScript:47881:10
  next/<@about:blank line 2 > injectedScript:47999:12
  next@about:blank line 2 > injectedScript:47795:14
  next/<@about:blank line 2 > injectedScript:47805:11
  next@about:blank line 2 > injectedScript:47729:14
  next/<@about:blank line 2 > injectedScript:47768:11
  done@about:blank line 2 > injectedScript:47300:7
  callFn@about:blank line 2 > injectedScript:47376:7
  require<[277]</<[32]</</Runnable.prototype.run@about:blank line 2 > injectedScript:47345:13
  next@about:blank line 2 > injectedScript:47743:10
  require<[277]</<[33]</</Runner.prototype.hook/<@about:blank line 2 > injectedScript:47773:9
  timeslice@about:blank line 2 > injectedScript:42885:27
```

fatso83:
@cincodenada any chance of having a look at the Firefox issue? Otherwise I need to revert the change to unblock releases.

cincodenada:
@fatso83 Yes, sorry I intended to get to this yesterday and then got sidetracked. I'll take a look this evening (Pacific Time) - is there a way for me to run the tests in Firefox myself? If not that's fine, I'm sure I can reproduce manually, but if there's an easier way I'll take it 😄

And as a Firefox user myself, I appreciate SauceLabs' dilligence!

cincodenada:
Okay, yeah I see what's happening here, should hopefully be a pretty simple fix - I wasn't able to test against Firefox yet, but after investigating the issue seems clear enough that I put up #2425, which reverts a small part of this PR that is causing the Firefox tests to fail, due to Firefox having a different stack trace format that breaks some assumptions I had made.

PR #2412, opened 2021-11-04
Fix bundling of cjs module

### LINKED ISSUE #2411: [12.0.0][regression] error when bundling the cjs version
**Describe the bug**
webpack picks either the commonjs or the esm version, depending on the file requesting it and how it does that (require call or dynamic import from a commonjs file).
esm bundles successfully (esm file importing sinon).
cjs, on the other end (cjs file requiring sinon), gets the following bundling error:
```
WARNING in ./node_modules/sinon/pkg/sinon.cjs 1:529-530
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

ERROR in ./node_modules/sinon/pkg/sinon.cjs 1:538-545
Cannot statically analyse 'require(…, …)' in line 1

webpack compiled with 1 error and 1 warning
```

**To Reproduce**
```
mkdir sinon-bundled
cd sinon-bundled
npm init -y
npm i sinon webpack webpack-cli -D
mkdir src
echo "require('sinon');" > src/index.js
npx webpack --mode development
```

**Expected behavior**
Both cjs and esm versions should be bundling-compatible.

- Library version: 12.0.0
- Environment: Fedora 35

perrin4869:
Fix #2411 

Webpack can't bundle `pkg/sinon.cjs`, so going to use the original `lib/sinon.js` which bundles without issues.
I don't remember right now why I introduced `pkg/sinon.cjs` anymore ><;;

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 <p align=center>
 <a href="https://app.saucelabs.com/u/sinonjs">
   <img src="https://app.saucelabs.com/browser-matrix/sinonjs.svg" alt="Sauce Test Status"/>
-</a> 
+</a>
 </p>
 -->
 
--- a/build.cjs
+++ b/build.cjs
@@ -44,7 +44,6 @@ makeBundle(
     },
     function (bundle) {
         var script = preamble + bundle;
-        fs.writeFileSync("pkg/sinon.cjs", script);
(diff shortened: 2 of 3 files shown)


fatso83:
Thanks for this, btw! Published as 12.0.1

perrin4869:
Thanks, and sorry for the trouble!

PR #2413, opened 2021-11-04
Add explicit export for `./*`

ntkme:
#### Purpose

Add an explicit export entry for `package.json` to fix the following error when running karma:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in node_modules/sinon/package.json
```

 #### Background

https://github.com/nodejs/node/issues/33460

There is a lot of discussion around whether declaring `package.json` as "public API" via `exports` is a good or bad idea. My take is simply that we need to expose this as a workaround before node team decide what to do, otherwise it breaks many tools that relies on `package.json`.

#### How to verify

Dependabot update sinon 12.0.0 failed on CI and manually patching `package.json` tested and working locally.

E.g. https://github.com/ntkme/github-buttons/runs/4099992660?check_suite_focus=true

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,8 +118,11 @@
   "main": "./lib/sinon.js",
   "module": "./pkg/sinon-esm.js",
   "exports": {
-    "require": "./lib/sinon.js",
-    "import": "./pkg/sinon-esm.js"
+    ".": {
+      "require": "./lib/sinon.js",
+      "import": "./pkg/sinon-esm.js"
+    },
+    "./*": "./*"
   },
   "type": "module",
   "cdn": "./pkg/sinon.js",


fatso83:
This is great, so thanks for this. But to avoid issues of this kind I would like to not rely on after-the-fact bug reports, but have our own little set of integration tests of sorts to catch these things when they happen. So could you instruct me of what a minimal setup for reproduction of this issue would be?

Something a la 
```
mkdir karma-integration && cd karma-integration && npm init -y
npm i sinon karma
cat > karma.conf.js << EOF
// ??
EOF
```

Having that, we could run this set of integration tests before every release.

ntkme:
Let me try come with a small test that can repro the issue.

ntkme:
Did a little bit more digging and the issue is caused by this line in karma-sinon-chai:

https://github.com/kmees/karma-sinon-chai/blob/c0663bcf150a4f28b0225f91d878a33d2ce8992d/index.js#L34

ntkme:
Opened PR against karma-sinon-chai: https://github.com/kmees/karma-sinon-chai/pull/64

If it get fixed in karma-sinon-chai, I think this PR might not be necessary. Let's see.

perrin4869:
Hey dude, in case you are interested, I setup some `karma` based testing in this repo:
https://github.com/dotcore64/react-stay-scrolled/blob/master/karma.conf.js
It uses `chai`, `sinon` etc

fatso83:
@perrin4869 That was a 404 when I checked.

fatso83:
Since kmees/karma-sinon-chai seems quite dead, I assume that fix is not landing. I had quick glance the karma config and the karma-runner/karma-mocha, but I could not immediately figure out how to create a headless test (though I saw the `HeadlessChrome` bits in the config), so I would appreciate if someone just put the bits together.

ntkme:
Chai.js has recently [merged a change](https://github.com/chaijs/chai/commit/022c2faefc9a86b478392594c9015c3b942892e2) that fixes this issue by adding export for `./*`. This allows all files in this package to be required in traditional way. I just updated this PR to do the same.

fatso83:
This feels a bit scary to merge without any tests of typical consumers ...

ntkme:
The rationale behind this:

- For node <12, `exports` in `package.json` is ignored. All files in the package can be loaded by referencing `package_name/relative/path/in/package.js`.
- For node >=12, `.` inside `exports`  defines the top level exports. `./*` allows all files to be loaded similar to node <12. - The only catch being in node <12 `relative/path` maybe resolve to `relative/path/index.js`, but with new exports mapping `"./*": "./*"` any file named `index.js` can only be resolved explicitly as `relative/path/index.js`, but not `relative/path`. A quick file search shows that sinon does not use `index.js` as filename at all so that it should not be an issue. In the end, this is just restoring the original behavior we had until node 12.

If you like we can wait a week or two and see if any issue arise on chai.js regarding the change on their side.

fatso83:
Thanks for the explanation! Will release a new major version anyway after upgrading the fake-timers package.

PR #2425, opened 2022-01-23
Be more general in stripping off stack frames to fix Firefox tests

cincodenada:
#### Purpose (TL;DR)

This fixes a breaking change to tests in Firefox introduced in #2410, by reverting a small subset of that PR.

We're stripping off stack traces for comparison's sake in tests, and previously to #2410 we used a pretty non-specific regex to do so, which accidentally stripped off part of the assertion message "expected func() to have been called at least once", and thus were asserting that such an assertion results in the message "expected func() to have been called". This is technically incorrect, but also fairly harmless. Amusingly, the message itself is still semantically sound, just less specific than the original.

As part of a larger refactor in #2410, I attempted to make this regex more specific, and thus not incorrectly strip part of the assertion message along with the stack traces. This was successful in not stripping the message, but went too far in the other direction - the new regex is now too specific, and makes assumptions about stack traces that do not hold for Firefox's stack trace format.

#### Solution

It's likely possible to come up with a regex that will match both formats, while still not stripping off the ends of legitimate messages, but in the interest of expedience and simplicity, this PR simply reverts to the previous regex, and adjusts the one test affected to once again assert against the technically incorrect message.

1. Run the tests in this branch under Firefox (with SauceLabs, or elsewise)

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/test/assert-test.js
+++ b/test/assert-test.js
@@ -1440,15 +1440,18 @@ describe("assert", function () {
                 } catch (e) {
                     // We sometimes append stack frames to the message and they
                     // make assertions messy, so strip those off here
-                    return e.message.replace(/( at.*\(.*\)$)+/gm, "");
+                    return e.message.replace(/ at.*/g, "");
                 }
             };
         });
 
         it("assert.called exception message", function () {
+            // The actual message here is "...called at least once but was
+            // never called" but everything after "at" gets stripped out
+            // becuase of how we strip off stack traces for comparison's sake
             assert.equals(
                 this.message("called", this.obj.doSomething),
-                "expected doSomething to have been called at least once but was never called"
+                "expected doSomething to have been called"
             );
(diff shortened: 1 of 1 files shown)


cincodenada:
Okay, I was able to run the tests under Firefox now, verified that I got the failures in master, and that the tests are passing again in this branch.

cincodenada:
Okay, and now that I can test, and after thinking on it a bit, I updated with a more general regex whose only assumption is that a stack frame has a non-word, non-space character somewhere - which seems a much safer assumption to me, and certainly holds for all the browsers we test in.

This gets us back to not chopping off the ends of legitimate messages, and seems like a reasonable regex to me as far as regexes go, but feel free to revert to the previous commit if we'd rather go with the simpler fix.

Issue #2400, opened 2021-09-11
Matching accepts iterators with completely different values


peterflynn:
**Describe the bug**
If a function is called with an iterator, matching expected vs. actual arguments will accept any iterator of the same type – even when the values yielded by the two iterators are completely different. For example, the iterators from two different Sets with unequal contents are considered a match. This leads to tests that _think_ they are checking argument values, yet still pass even when the spy/stub is called with incorrect arguments.

Secondary issue, maybe separable: matching _won't_ accept an iterator with a different backing impl even when it _does_ yield the same values as the actual iterator. This can be challenging because tests may need to assume more about implementation details in other layers of code, or may not be able to instantiate the correct impl type at all. This is a bit less problematic since it's obvious you have a problem (the test you just wrote is failing) and I assume it can be worked around using a custom `sinon.match`-er?

**To Reproduce**

(runnable copy at https://codepen.io/peterflynn/pen/wvedJoK?editors=0012)

```js
const stub = sinon.stub();

// Call stub with iterator that yields ["one", "two"]
const actualSet = new Set(["one", "two"]);
stub(actualSet.keys());


// Passes even though this iterator yields different values!
const wrongSet = new Set(["four"]);
sinon.assert.calledWith(stub, wrongSet.keys());

// Fails even though this iterator yields the SAME values!
const correctButArray = ["one", "two"];
sinon.assert.calledWith(stub, correctButArray.values());


// Or put another way:
sinon.assert.match(actualSet.keys(),      wrongSet.keys());      // succeeds
sinon.assert.match([...actualSet.keys()], [...wrongSet.keys()]); // fails (as expected)

sinon.assert.match(actualSet.keys(),      correctButArray.values());      // fails
sinon.assert.match([...actualSet.keys()], [...correctButArray.values()]); // succeeds (as expected)
```

**Expected behavior**
* The `wrongSet` assert should either fail since the values yielded don't match, _or_ at minimum should fail with an error message saying iterator comparison isn't supported.

* Ideally the `correctButArray` assert should pass since the values yielded do match. My thinking is this feels a bit like comparing two objects with getters that yield the same value: the getter _impls_ may be different, but Sinon matching is ok with that as long as the getters yield the same value. Iterators _could_ be considered similar.

**Context (please complete the following information):**

- Library version: 9.2.4
- Environment:
- Example URL: https://codepen.io/peterflynn/pen/wvedJoK?editors=0012
- Other libraries you are using:

**Additional context**
n/a

[...]

PR #2429, opened 2022-01-27
Bump samsam to 6.1.0

sdotson:
Ultimately addresses https://github.com/sinonjs/sinon/issues/2400.

Thought about adding tests here but then decided against it because it felt like I was retesting functionality already tested in samsam. 

Tested manually and it does appear that `Set`, `Map`, and `Array` iterators with different values fail and succeed with the same values. Comparing iterators of different type (e.g. Array Iterator to Set Iterator) that have the same values fails, which seems to be correct behavior, although open to alternative viewpoints on this. 

Let me know if this is isn't the way this is typically done.


fatso83:
Heh, nevermind, this was only setup in another repo. Fixed now: https://github.com/sinonjs/sinon/commit/2c68dbc84c82d6655340742bd9c6892075ecd970

PR #2437, opened 2022-01-28
Re-introduce version script

cincodenada:
#### Purpose (TL;DR)

Because of the merge mess with #2426, the `version` command in `package.json` was reverted, so that the new `version.sh` script isn't being used.

#### Solution

This PR also puts the `version.sh` script back, and persists a prettier fix in a generated file back to the generation of the file, so it doesn't keep popping up.

The only way to really, test this is to run `npm version`, but if #2436 is merged first, then we'll have a way to run that without publishing a release!

#### Checklist for author

- [x] `npm run lint` passes


cincodenada:
Ugh, sorry, wrong branch

PR #2438, opened 2022-01-28
Re-introduce new version script

cincodenada:
#### Purpose (TL;DR)

Because of the merge mess with #2426, the `version` command in `package.json` was reverted, so that the new `version.sh` script isn't being used.

#### Solution

This PR also puts the `version.sh` script back, and persists a prettier fix in a generated file back to the generation of the file, so it doesn't keep popping up.

The only way to really, test this is to run `npm version`, but if #2436 is merged first, then we'll have a way to run that without publishing a release!

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/scripts/update-changelog-page.sh
+++ b/scripts/update-changelog-page.sh
@@ -4,6 +4,7 @@ layout: default
 title: Changelog
 permalink: /releases/changelog
 ---
+
 # Changelog
 $(tail -n+2 CHANGES.md)
 EOL
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "prettier:check": "prettier --check '**/*.{js,css,md}'",
     "prettier:write": "prettier --write '**/*.{js,css,md}'",
     "preversion": "./scripts/preversion.sh",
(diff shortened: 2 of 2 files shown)


cincodenada:
@fatso83 Okay, I ran through and simulated a clean merge of #2426 to make sure the end-result of all of this merging was the same as if things had gone more smoothly, and this was the only difference I found.

I also snuck in the prettier change, and had a related question - do you have some sort of extra config for `changes` on your machine, or are you manually deleting the quoted bodies of the changelog, or is there something else going on? My `changes` command seems to generates a lot more detail in the changelogs than is showing up in your release commits, and I'm curious why!

fatso83:
@cincodenada Yeah, I manually deleted all I regarded as noise for people browsing the changelog.

scripts/update-changelog-page.sh
```diff
@@ -4,6 +4,7 @@ layout: default
 title: Changelog
 permalink: /releases/changelog
 ---
+
```
fatso83, line 7: Good call. Was not sure where that came from.

PR #2439, opened 2022-01-30
Fix fake tests

fatso83:

I just closed https://github.com/fatso83/mini-mocha/issues/23, which basically was about that the test runner should exit with a non-zero status if a test fails. As soon as that was updated, our example tests in Sinon failed 🤡 So this PR fixes the failing tests that the bugfix found.

I also flattened the tests for fakes, removing the `describe` block. Since it did not describe any specific issue, I felt it did not really do much in helping the understanding, so I thought it just as well could be removed. I also renamed the test files (for fakes only) to have them run in numerically sorted order.


mroderick (approved):
Very nice 👍

PR #2441, opened 2022-01-31
Add unimported to workflow

mroderick:
This PR adds a check for unimported files to the GitHub Actions workflow, using https://github.com/smeijer/unimported

As always, the PR is easier to understand when you read the commits individually.

#### How to verify

1. Add a file to `lib/` that is not used
2. `npm run unimported`
3. Observe that there's an error
4. Remove the file
5. `npm run unimported`
6. Observe that there is no error


fatso83:
Hmm ... not sure I follow. A fixup commit is supposed to be squashed with the previous when rebasing, right? And since all commits were squashed, the effect should be the same ... Or? 😬

https://github.com/sinonjs/sinon/commit/f329a01040bfa5a79e039419220b21eda56935d6

Zaram22:
Nice

Zaram22:
Project

fatso83:
@mroderick I blocked this account after the bottish behavior in #2441 #2444 and #2445.

PR #2456, opened 2022-05-12
fix: actaully reset 'injectedKeys'

mroderick:
Re-assigning the local variable `injectedKeys` would not change `sandbox.injectedKeys`, thus `restoreContext` doesn't fully restore the context.

See:

https://lgtm.com/projects/g/sinonjs/sinon/snapshot/9e09e7d79bac5808ca98fac4f7419a20be4fc43d/files/lib/sinon/sandbox.js?sort=name&dir=ASC&mode=heatmap#x9f770d565ef51b7d:1

### DIFF
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -197,17 +197,15 @@ function Sandbox() {
     };
 
     sandbox.restoreContext = function restoreContext() {
-        let injectedKeys = sandbox.injectedKeys;
-
-        if (!injectedKeys) {
+        if (!sandbox.injectedKeys) {
             return;
         }
 
-        forEach(injectedKeys, function (injectedKey) {
+        forEach(sandbox.injectedKeys, function (injectedKey) {
             delete sandbox.injectInto[injectedKey];
         });
 
-        injectedKeys = [];
+        sandbox.injectedKeys = [];
     };
 
     function getFakeRestorer(object, property) {
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -205,7 +205,7 @@ function Sandbox() {
             delete sandbox.injectInto[injectedKey];
         });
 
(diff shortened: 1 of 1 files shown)


fatso83:
Any way of adding a regression test to show the bug?

mroderick:
> Any way of adding a regression test to show the bug?

That's a fair question. We haven't detected it thus far with tests and users have also not reported it.
There might be something else that resets that property, so we've been hiding in the shadow of that for years.

I can look further

lib/sinon/sandbox.js
```diff
@@ -196,18 +196,15 @@ function Sandbox() {
     };
 
     sandbox.restoreContext = function restoreContext() {
-        var injectedKeys = sandbox.injectedKeys;
-        var injectInto = sandbox.injectInto;
-
-        if (!injectedKeys) {
+        if (!sandbox.injectedKeys) {
```
fearphage, line 199: Is this optimization needed? `forEach` can handle empty arrays.

lib/sinon/sandbox.js
```diff
@@ -196,18 +196,15 @@ function Sandbox() {
...
         }
 
-        forEach(injectedKeys, function (injectedKey) {
-            delete injectInto[injectedKey];
+        forEach(sandbox.injectedKeys, function (injectedKey) {
+            delete sandbox.injectInto[injectedKey];
         });
 
-        injectedKeys = [];
+        sandbox.injectedKeys = [];
```
fearphage, line 207: Do we actually need a new array?

```suggestion
        sandbox.injectedKeys.length = 0;
```

fearphage (changes requested):

lib/sinon/sandbox.js
mroderick: I don't know if it is, but it was there already.

Issue #2448, opened 2022-04-06
Custom formatter not respected by samsam/commons


ajaska:
Sinon supports setting a custom formatter, but it is not used by sub-libraries.

**To Reproduce**

```ts
const sinon = require("sinon");
const inspect = require("util").inspect;

const obj = { a: () => {} };
const aSpy = sinon.spy(obj, "a");
obj.a({ foo: { bar: "baz" } });

sinon.setFormatter(inspect);

sinon.assert.calledWith(aSpy, sinon.match({ foo2: { bar2: "baz2" } }));
```

Output:

```
Error [AssertError]: expected a to be called with arguments
{ foo: { bar: 'baz' } } match(foo2: [object Object])
```

**Expected behavior**

```
Error [AssertError]: expected a to be called with arguments
{ foo: { bar: 'baz' } } match(foo2: { bar2: 'baz2' })
```

**Context (please complete the following information):**

- Library version: sinon@latest -- 13.0.1
- Environment: node
- Example URL:
- Other libraries you are using:

**Additional context**
I believe this is because common's `valueToString` takes a simple approach and does not allow one to pass a custom formatter in. I believe a fix would involve updating samsam's `createMatcher` to accept a formatter, and using it (if specified) instead of common's `valueToString`.

mroderick:
Thank you for taking the time to create an easy to follow issue 👍 

Would you like to contribute a PR to `samsam` to improve the situation?

mroderick:
This is an interesting one.

The `setFomatter` method was originally added in #1503, in order to allow Cypress to override the formatter. 

We have a few options for making things consistent here.

1. Add a `setFormatter` option to `@sinonjs/samsam`
2. Remove `setFormatter` from `sinon` and make all `@sinonjs/` libraries  use node's `inspect` from the `util` package for formatting values
3. Use a different formatter?

#### Option 1 - Add a `setFormatter` option to `@sinonjs/samsam`

In order to solve the inconsistency and have some nicer output, we could expand `@sinonjs/samsam` with it's own `setFomatter` method, that could be called by `sinon.setFormatter`.

However, this would make `@sinonjs/samsam` stateful, which would have the potential to break people's tests in all sorts of interesting ways.

Ultimately, I don't think that libraries should be stateful.

#### Option 2 - Use `util.inspect`

Instead of making more libraries stateful, we remove `setFormatter` and use `util.inspect` for value formatting everywhere.

[...]

PR #2480, opened 2022-11-28
Fix 2448 remove custom formatter

mroderick:
Fix 2448: remove custom formatter
…
Remove option to pass a custom formatter.

The sub libraries of Sinon has long moved on to use `util.inspect` from Node. By using that in Sinon itself, we align all the libraries.

See #2448 for discussion


hexeberlin (approved):

mroderick:
This has been published to the npm registry as `v15.0.0`

Issue #2484, opened 2022-12-12
Enable and Set Assertion Logging Limits 


sgoossens:
when `sinon.assert` fails for a stubbed method that had potentially a large amount of data on it, it logs the entirety of the stub each time which can cause runoff on the terminal see [here ](https://github.com/sgoossens/SinonOverflowExample) making it very hard to debug.  It would be nice to be able to toggle on a log threshold limit such that we can control the amount that is logged out when an assertion fails.

fatso83:
Was closed in October 2023

PR #2485, opened 2022-12-12
fix: assertion log limit

sgoossens:

Allow for a log limit setting on the assertion object, that when true, will truncate the error.message that is returned via a failing assertion.  The limit for the log message will be set to a default if none was provided

https://github.com/sinonjs/sinon/issues/2484

When a stubbed method has a large property on it, it is logged in the failing assertion causing runoff in the terminal and making it hard to see what assertion actually failed.

By allowing for a log limitation setting and a default log limit, error logs can be truncated.

1. use [this app](https://github.com/sgoossens/SinonOverflowExample) and point the sinon instance to this version
2. see the logs are reduced when the setting is on

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -15,12 +15,40 @@ const forEach = arrayProto.forEach;
 const join = arrayProto.join;
 const splice = arrayProto.splice;
 
-function createAssertObject() {
+function applyDefaults(obj, defaults) {
+    for (const key of Object.keys(defaults)) {
+        const val = obj[key];
+        if (val === null || typeof val === "undefined") {
+            obj[key] = defaults[key];
+        }
+    }
+}
+
+/**
+ * @param {object}  [opts] options bag
+ * @param {boolean} [opts.shouldLimitAssertionLogs] default is false
+ * @param {number}  [opts.assertionLogLimit] default is 10K
+ * @returns {object} object with multiple assertion methods
+ */
+function createAssertObject(opts) {
+    const cleanedAssertOptions = opts || {};
+    applyDefaults(cleanedAssertOptions, {
+        shouldLimitAssertionLogs: false,
+        assertionLogLimit: 1e4,
+    });
+
     const assert = {
         failException: "AssertError",
 
         fail: function fail(message) {
-            const error = new Error(message);
+            let msg = message;
+            if (cleanedAssertOptions.shouldLimitAssertionLogs) {
+                msg = message.substring(
+                    0,
+                    cleanedAssertOptions.assertionLogLimit,
+                );
+            }
+            const error = new Error(msg);
             error.name = this.failException || assert.failException;
 
             throw error;
--- a/lib/sinon/assert.js
+++ b/lib/sinon/assert.js
@@ -333,20 +333,4 @@ function createAssertObject(opts) {
 }
 
 module.exports = createAssertObject();
-Object.defineProperty(module.exports, "createAssertObject", {
-    get() {
-        return createAssertObject;
-    },
-});
-
-// allow for easy mocking
-module.exports.mock = {};
-Object.defineProperty(module.exports.mock, "createAssertObject", {
-    get() {
-        return createAssertObject;
-    },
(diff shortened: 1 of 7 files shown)


fatso83 (changes requested):
Missing test coverage. ✌️

lib/sinon/assert.js
```diff
@@ -15,8 +15,11 @@ var forEach = arrayProto.forEach;
 var join = arrayProto.join;
 var splice = arrayProto.splice;
 
-function createAssertObject() {
+var DEFAULT_LOG_LIMIT = 10000;
+
+function createAssertObject(shouldLimitAssertionLogs) {
     var assert;
+    var assertionLogLimit;
```
fatso83, line 22:
```suggestion
```
This only makes sense to keep if you expose a way of setting it. So in the current context, it could be removed, but I assume it has been added to allow for testing this, which would 

1. Fix the only remaining issue
2. Make this line of code meaningful 😄 

P.S. We [allow newer EcmaScript syntax](https://github.com/sinonjs/sinon/blob/main/COMPATIBILITY.md#es2017), so you mist as well use `const`.

fatso83:
Would you be interested in help getting it over the finish line? Have some time these days.

sgoossens:
> Would you be interested in help getting it over the finish line? Have some time these days.

Sure that would be great. I lost track of this but think it would be a great add. I can also put some time in to addressing your feedback as well

fatso83 (approved):

fatso83:
OK, so I added test coverage, fixed up some logical errors, exposed the options on the sandbox creation, etc. 

This is ready for merge from my perspective, but it does not change the default, as that would be a breaking change, but we [can do that with version 18 ](https://github.com/sinonjs/sinon/milestone/10). It still allows creating sandboxes with log limits, which is probably what you want right now.

The way I stubbed out

fatso83:
@sgoossens You ok with this? Anything to add/change?

sgoossens (commented):
Looks great thank you for helping get this over the line

lib/sinon/assert.js
```diff
@@ -333,4 +333,20 @@ function createAssertObject(opts) {
...
-module.exports.createAssertObject = createAssertObject;
+Object.defineProperty(module.exports, "createAssertObject", {
+    get() {
+        return createAssertObject;
+    },
+});
+
+// allow for easy mocking
+module.exports.mock = {};
+Object.defineProperty(module.exports.mock, "createAssertObject", {
```
mroderick, line 344: I'm not a fan of this approach, as it adds test code to production code.

For a testing library, adding some bloat might not be that big a deal ... but for production code, especially code that ships over the internet, this adds a lot of bloat of unused code. (No, tree shaking is not a good argument to embrace this practice).

Additionally, if the code enabling tests is not added very carefully, it becomes entangled with the production code.



Since https://github.com/sinonjs/sinon/commit/11a8e3f24ccec981a280dbe63df316d1fbde4ddd, where proxyquire was removed, this seems to be one solution ... are there others?

I wish we had at least had a discussion about proxyquire and prettier conflict, so we could have evaluated solutions before they got merged into `main`.

fatso83: 👍 @mroderick . I got rid of the `mock` object. It was not necessary at all, as I got this a bit confused with ESM exports for a while, and seeing it work made me realize the whole getter thing was just a totally unneccessary

test/create-sandbox-test.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+const sinonAssert = require("../lib/sinon/assert");
+const createSandbox = require("../lib/sinon/create-sandbox");
+
+describe("create-sandbox", function () {
+    it("passes on options to the creation of the assert object", function () {
```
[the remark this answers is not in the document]
fatso83: Good call. I'll fix.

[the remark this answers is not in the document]
fatso83: Done! Much better to keep the implementation specifics out of it.

Issue #2496, opened 2023-03-11
throws() keeps raising exception with callsFake()


ziluvatar:
**Describe the bug**

When you create a stub like `sinon.stub().throws()`, if you later define another behavior for it with `.callsFake()`, while executing the call it will through the error instead of executing the fake call.

**To Reproduce**
```
const fn = sinon.stub().throws('not mocked');
fn.callsFake(() => 'mocked behavior')

console.log(fn())
```

**Expected behavior**
`mocked behavior`

**Actual behavior**
```
[...]/node_modules/sinon/lib/sinon/proxy-invoke.js:84
    throw exception;
    ^
not mocked
    at Object.fake.exceptionCreator ([...]/node_modules/sinon/lib/sinon/default-behaviors.js:18:32)
[...]
```

**Context (please complete the following information):**

- Library version: 15.0.1
- Environment: NodeJS v16.15.0

**Additional context**
N/A

fatso83:
Thanks for reminding me why we have tried moving people towards the Fake API 😄  Stub setup and testing is much harder to reason about than immutables. Like what _is_ the correct behavior of setting up two unconditional and incompatible behaviors? 

Also, the chaining API is fraught with peril. At first look, I thought this was a case of #2447, as you do not save the stub instance, but the result of the first behavior. That's worth knowing about at least, even though that was not the issue here.

Another issue is that I cannot see a use case for this that is not served better by more meaningful methods. If you want to have different behavior for different calls you could do this:
```
stub.onFirstCall().callsFake(() => "mocked behavior");
stub.onSecondCall().throws("not mocked");
```

Or if you want a default behavior, but changing it for some later invocation, you could use the `on<X>Call` family of methods or the `withArgs()` family, like:
```javascript
stub.callsFake(() => "mocked behavior"); // call 1, 3, 4
stub.onSecondCall().throws("not mocked"); // call 2
```

Anyway ... I _know_ that our current behavior is that subsequent (unconditional) stubbing should "overwrite"/shadow previous ones, so in that sense, aligning the `throws` behavior seems only reasonable to me. So definitely welcome the fix!

fatso83:
published as sinon 15.0.2

ziluvatar:
That was fast! Thanks!

[...]

PR #2497, opened 2023-03-11
fix throws().callsFake() precedence

ziluvatar:

Fixes https://github.com/sinonjs/sinon/issues/2496 by cleaning any previous `exception*` behavior when `callsFake` is used.

* Issue https://github.com/sinonjs/sinon/issues/2496

Precedence among behaviors is defined by the behavior itself (cleaning other previously set behaviors) and by the order [at invocation time](https://github.com/sinonjs/sinon/blob/ddd0f25/lib/sinon/behavior.js#L162-L210).

Since `throws` behavior (`exception` and `exceptionHandler`) is the first one to run at invocation time, other behaviors can only define precedence on its own definition. That's why I modified `callsFake` behavior to clean `throws`

 #### Solution
When `callsFake` is chained on a stub it cleans previous `exception` and `exceptionHandler` values which essentially invalidates any previous `throws` or `throwsException` behaviours.

**Note:** I think this fix should not break any current usage, because any attempt to run the stubbed function would have raised an exception already and the developer would have noticed the same I did notice: callsFake was ignored, so that, probably they applied some workaround like restoring + callsFake OR recreating the stub + callsFake.

Run:
```
const fn = sinon.stub().throws('not mocked');
fn.callsFake(() => 'mocked behavior')

console.log(fn())
```

You should see:
```
mocked behavior
```

#### Checklist for author

- [x] `npm run lint` passes
- n/a References to standard library functions are [cached](https://github.com/sinonjs/sinon/pull/1523).

### DIFF
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -31,6 +31,8 @@ function throwsException(fake, error, message) {
 var defaultBehaviors = {
     callsFake: function callsFake(fake, fn) {
         fake.fakeFn = fn;
+        fake.exception = undefined;
+        fake.exceptionCreator = undefined;
     },
 
     callsArg: function callsArg(fake, index) {
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -1407,6 +1407,28 @@ describe("stub", function () {
             refute(fakeFn.called);
             assert(returned === 3);
         });
+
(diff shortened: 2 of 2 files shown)


fatso83:
This is great and works just like the rest. Thanks!

fatso83:
published as sinon 15.0.2

PR #2505, opened 2023-03-27
Update sponsors and backers in README and homepage

fatso83:
- Use automated fetching to keep sponsors updated
- Update wording

The big difference is that this is now dynamically kept up to date on the homepage with just _current_ sponsors for the "sponsors" tier. If we create other tiers (like Gold, Silver, Bronze) like [Parcel is doing](https://opencollective.com/parcel#support), we could differentiate on support level, but we are not quite there. It would perhaps make sense to have a second level support option for orgs/businesses, though, as it seems backing businesses are currently not listed with < 100$ contributions, unless they are listed as "individuals".

In the README I chose to keep the top ten supporters so far. Without it, it would look a bit anemic when not showing the historic support we have gotten 😄 

1. Check out this branch
2. Run the Jekyll process (see /docs/CONTRIBUTING.md for how)

##### The new look
Skjermbilde 2023-03-27 kl  04 36 53

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,38 +64,10 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how you can contribute to 
 
 ## Backers
 
-Support us with a monthly donation and help us continue our activities. [[Become a backer](https://opencollective.com/sinon#backer)]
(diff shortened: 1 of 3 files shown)


mantoni (approved):
👍🙏

Issue #2491, opened 2023-02-20
sinon.restore cannot restore spies


runspired:
Version: 15.0.1

Assuming a class instance like so

```ts
class Thing {
  aMethod() {}
}
const thing = new Thing();
```

and a spy like so

```ts
sinon.spy(thing, 'aMethod');
```

a global restore will error

```ts
sinon.restore();
```

This is because sinon attempts to directly set `target[property]` back to the original method, which is caught by the spy proxy.

Example Error:

```
TypeError: Cannot assign to read only property 'setRequestReviewers' of object '[object Object]'
         at Function.restore 
```

runspired:
hrm, actually this runs deeper. I changed this to restoring the spy directly and it still errors in this way. So I dug a bit and I think sinon isn't accounting for decorated methods.

```
<img width="568" alt="image" src="https://user-images.githubusercontent.com/650309/220195088-30ff673e-680b-4291-bfaa-689734ca5863.png">
```

fatso83:
> sinon isn't accounting for decorated methods.

To be honest, I am not sure what you mean by this. Do you mean the not-yet-standardized (?) annotations employed by some frameworks and applied by Babel and such, or something else?

Would be great with a minimal reproduction if someone was to look into fixing this, such as [a repo with 2-3 files](https://github.com/fatso83/sinon-issues/tree/sinon-1877-esm-cjs-compat)

fatso83:
@runspired Would you be able to provide us with a reproduction case? As in with full setup and transpilation, etc. I assume that [TC39](https://github.com/tc39/proposal-decorators) will be accepted in the near future, and so it would be nice if we already had support in place for whatever that might throw at us.

runspired:
All you have to do is define with defineProperty with the config seen above in the screenshots. Doesn't even take a decorator, decorator is just the common case. Happy next week to give you a gist or some such.

fatso83:
Great. Looking forward to it, as even with those hints I am not totally sure what the bug actually is, so a repro is definitely warranted :)

stevenhair:
I also ran into this problem - it seems that it was introduced in 15.0.2 (I'm guessing in #2499). My code uses decorators, but I haven't been able to isolate the issue yet.

@runspired, did you get a chance to put a reproduction together? I could take a stab at it if you haven't.

fatso83:
[...]

Issue #2195, opened 2020-01-06
Conditional support for replacing props not defined by the object


fatso83:
**Is your feature request related to a problem? Please describe.**
In #1557 we removed support for stubbing undefined properties, aligning the behavior with how sandbox stubbing worked. Later on we have made the sinon methods use a default sandbox to further align how the default methods and sandboxes worked.

Still, as the thread in #1537 show, this remains a sorely missed feature for some, and while we had good reasons for removing it, maybe we could appease the ones wanting it by making it an opt-in? It remains a quite useful feature in some cases.

**Describe the solution you'd like**
I suggest adding a config field to the sandbox to allow defining fields (fakes) that are not originally defined by the objects we insert the fake into. Something like `sinon.createSandbox({allowUndefined: true})`? (Totally not sure about the actual name of the prop, though ... `allowNonOwn`?)

Given

```javascript
var obj = {a: 1};
```
it should allow
```javascript
sinon.stub(obj, 'b').value(2); // not defined anywhere on the prototype chain
sinon.replace(obj, 'c', sinon.fake()); 
```

Related to this, but really a different case, is props on the prototype chain. As explained in the "Solution" paragraph in #2193 we deal with this differently today, depending on whether we use spies or fakes:

```
var parent = { foo: function() {} };
var child = Object.create(parent):

// disallowed
sinon.stub(child, 'foo');
// allowed
sinon.replace(child, 'foo', sinon.fake());
```

If we simplified the "undefined" and "props on an ancestor" problems into a single issue: props not defined by the object whose fields are being stubbed, we could align this behavior for our spies, stubs and fakes under a default non-permissive line and have an opt-in solution for those who want it. By default, you would then only be allowed to replace fields owned/defined by the object.

Having the opt-in for stubs and spies would not be a breaking change, but as the `sinon.replace` method today allows stubbing defined props owned by an ancestor it would be a breaking change to disallow this by default.

**Describe alternatives you've considered**
Two flags instead of one: `allowUndefinedProps` (props are not present anywhere on the prototype chain) and `allowShadowingFieldsOnPrototype` (to stub fields defined by ancestors). I don't really see the benefit, though.

mantoni:
As I understand this, `sinon.replace` throws on undefined properties to highlight an issue in case the replaced property was removed or there is a typo while writing a test. I quite like this and I agree we should keep it this way.

[...]

PR #2508, opened 2023-04-20
Handling non-configurable object descriptors on the prototype

fatso83:

Fix for #2491 where it seems that we fail to restore spies set on instances if their prototypes have unconfigurable property descriptors.

A property descriptor must have `configurable: true` set for us to be able to restore it. If we get the property descriptor from a prototype we should be free to modify this to whatever we want.

_This is work in progress, as I hit issues when trying to achieve consistency for spies, stubs, mocks and fakes._

Fixing this in `wrap-method.js` only fixes the issue for _spies_ as we deal with the issues in four different ways:

**stubs**
Even though this module is also imported by the `stubs` module it fails for stubs _before_ getting to the `wrapMethod` call, as it has checks of their own on the object descriptors: `Descriptor for property aMethod is non-configurable and non-writable`

**mocks**
`wrap-method` is also used by the `mock` module, but it also fails before getting to it: `Attempted to wrap undefined property myMethod as function`

**fakes / sandbox.replace()**
Fakes do not implement anything to do with replacing properties themselves and leave that to `Sandbox#replace`, which fails with `Cannot assign to read only property 'aMethod' of object '#<BaseClass>'`. ~The "weird" thing here is that we do not use any of the old `wrap-method` logic and instead have resorted to build new functionality for restoring. I am not sure why.~ 
EDIT: by intent. See discussion in #2195

image

 #### Solution  - optional
Setting `configurable: true` in wrap-method was only helpful for spies. 

These are my thoughts for further work in this PR
- The `Descriptor for property aMethod is non-configurable and non-writable` check is good and could be re-used elsewhere, but it needs to _only apply to cases where the descriptor comes from the object,_ not its prototype. We can use the `isOwn` property from `get-property-descriptor.js` to do this.
- Could the also perhaps be moved to `wrap-method` so that it would also be used for spies?
- I am not sure why the mocks logic is not using any of the above (wrap-method) for spies and stubs. Probably should
- `Sandbox#replace` _probably_ should not use `wrap-method`, as it seems to have simplified the use-cases. Not sure though

1. Check out this branch
2. `npm i; npm test`

#### Checklist for author

- [x] `npm run lint` passes

**`isOwn` and `shadowsPropOnPrototype` is relevant**
image

### DIFF
--- a/lib/sinon/util/core/wrap-method.js
+++ b/lib/sinon/util/core/wrap-method.js
@@ -137,6 +137,7 @@ module.exports = function wrapMethod(object, property, method) {
         for (i = 0; i < types.length; i++) {
             mirrorProperties(methodDesc[types[i]], wrappedMethodDesc[types[i]]);
         }
+        methodDesc.configurable = true;
         Object.defineProperty(object, property, methodDesc);
 
         // catch failing assignment
--- a/lib/sinon/stub.js
+++ b/lib/sinon/stub.js
@@ -134,7 +134,7 @@ function assertValidPropertyDescriptor(descriptor, property) {
     if (!descriptor || !property) {
         return;
     }
-    if (!descriptor.configurable && !descriptor.writable) {
+    if (descriptor.isOwn && !descriptor.configurable && !descriptor.writable) {
         throw new TypeError(
             `Descriptor for property ${property} is non-configurable and non-writable`
         );
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -823,9 +823,11 @@ describe("issues", function () {
         });
     });
 
-    describe("#2491 - unable to restore stubs on an instance where the prototype has an unconfigurable property descriptor", function () {
-        it("should ensure object descriptors are always configurable", function () {
-            class BaseClass {}
+    describe("#2491 - unable to restore spies on an instance where the prototype has an unconfigurable property descriptor", function () {
+        function createInstanceFromClassWithReadOnlyPropertyDescriptor() {
+            class BaseClass {
+                instanceProperty = 1;
+            }
             Object.defineProperty(BaseClass.prototype, "aMethod", {
                 value: function () {
                     return 42;
@@ -834,13 +836,55 @@ describe("issues", function () {
 
             // anchor
             const instance = new BaseClass();
-            assert.isFunction(instance.aMethod);
             assert.equals(instance.aMethod(), 42);
-            this.sandbox.spy(instance, "aMethod")
+
+            return instance;
+        }
+
+        it("should ensure copied object descriptors are always configurable for spies", function () {
+            const instance =
+                createInstanceFromClassWithReadOnlyPropertyDescriptor();
+            this.sandbox.spy(instance, "aMethod");
 
             refute.exception(() => {
-                this.sandbox.restore(); // #2491: this throws "TypeError: Cannot assign to read only property 'myMethod' of object '#<BaseClass>'"
-           });
+                this.sandbox.restore(); // #2491: this throws TypeError: Cannot assign to read only property 'myMethod' of object '#<BaseClass>'
+            });
+        });
+
+        it("should not throw if the unconfigurable object descriptor to be used for a Stub is on the prototype", function () {
+            const instance =
+                createInstanceFromClassWithReadOnlyPropertyDescriptor();
+
+            // per #2491 this throws 'TypeError: Descriptor for property aMethod is non-configurable and non-writable'
+            // that makes sense for descriptors taken from the object, but not its prototype, as we are free to change
+            // the latter when setting it
+            refute.exception(() => {
+                this.sandbox.stub(instance, "aMethod").returns("a stub");
+            });
+        });
+
+        it("should not throw if the unconfigurable object descriptor to be used for a Mock is on the prototype", function () {
+            const instance =
+                createInstanceFromClassWithReadOnlyPropertyDescriptor();
+
+            const mock = this.sandbox.mock(instance);
+
+            // per #2491 this throws 'TypeError: Attempted to wrap undefined property myMethod as function
+            refute.exception(() => {
+                mock.expects("aMethod").once();
+            });
+        });
+
+        it("should not throw if the unconfigurable object descriptor to be used for a Fake is on the prototype", function () {
+            const instance =
+                createInstanceFromClassWithReadOnlyPropertyDescriptor();
+
+            // per #2491 this throws "TypeError: Cannot assign to read only property 'aMethod' of object '#<BaseClass>'"
+            // that makes sense for descriptors taken from the object, but not its prototype, as we are free to change
+            // the latter when setting it
+            refute.exception(() => {
+                this.sandbox.replace(instance, "aMethod", sinon.fake());
+            });
         });
     });
 });
--- a/test/issues/issues-test.js
+++ b/test/issues/issues-test.js
@@ -825,9 +825,7 @@ describe("issues", function () {
 
     describe("#2491 - unable to restore spies on an instance where the prototype has an unconfigurable property descriptor", function () {
         function createInstanceFromClassWithReadOnlyPropertyDescriptor() {
-            class BaseClass {
-                instanceProperty = 1;
-            }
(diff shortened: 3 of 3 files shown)


lib/sinon/util/core/wrap-method.js
```diff
@@ -137,6 +137,7 @@ module.exports = function wrapMethod(object, property, method) {
         for (i = 0; i < types.length; i++) {
             mirrorProperties(methodDesc[types[i]], wrappedMethodDesc[types[i]]);
         }
+        methodDesc.configurable = true
```
mroderick, line 140: Do we need to store the original value, so it can be restored when the method is restored?

mroderick:
> The Descriptor for property aMethod is non-configurable and non-writable check is good and could be re-used elsewhere, but it needs to only apply to cases where the descriptor comes from the object, not its prototype. We can use the isOwn property from get-property-descriptor.js to do this.

That's could help reduce the number of implementations 👍

mroderick:
> I am not sure why the mocks logic is not using any of the above (wrap-method) for spies and stubs. Probably should

It seems that `wrapMethod` predates `mock`, which was surprising to me.

`function wrapMethod` was introduced in https://github.com/sinonjs/sinon/commit/7fb6657f6290673ada8938aa454b312020fee427#diff-3cbef9e218193dbaed32c532449e176e0cf97d7c0427326cdd5c83ba18df20fbR2-R25 on Mar 30, 2010 

`function mock` was introduced in https://github.com/sinonjs/sinon/commit/84dbb22ae72a94380b846ee58a7d2b0238c5ea0d#diff-ab6968d5e523c9eac55c8cd92a72dcd87f6647f3587e8d1b8c691e3e7746cc2bR12-R18 on Apr 18, 2010

`mock` has seen very little love over the years, probably because none of the maintainers use it. It seems likely to me that very few of the users actually use `mock`, or there would have been more issues and pull requests for it over the years. With that in mind, I don't find it surprising that it has it's own way of doing things.

I agree, it should use `wrapMethod` for spies and stubs.

lib/sinon/util/core/wrap-method.js
fatso83: The `restore()` function in this module already restores the original value. It only does so in the case of _own_ descriptors, otherwise it just deletes the spy/stub (thus removing the shadowing of the prototype).

fatso83:
A side note

# A little detour into setting fakes

> Sandbox#replace only deals with existing properties, which made this much easier.

That means there is no easy way of cleaning up fakes set on instances of a class, right? Say we have this:
```javascript
class Foo{ method(){} }
const foo = new Foo();
```

We could easily clean up this with a sandbox.restore():
```javascript
sandbox.stub(foo, 'method')
sandbox.restore()
```
But setting and cleaning up using fakes looks like this:
```javascript
foo.method = sandbox.fake()
delete foo.method
```

Should we not try to aim for consistency here? I think we have touched on this before (without anyone doing anything), but I could have liked for us to have this:
```javascript
sandbox.replace(foo, 'method', sandbox.fake(), {allowNonExisting: true})
sandbox.restore(); // like for everywhere else
```

fatso83:
> I agree, it should use wrapMethod for spies and stubs.

`mock.js` _is_ actually using wrapMethod. That is exactly what is failing (for some reason I haven't looked into yet)

fatso83:
So this fixes the immediate issue in #2491 .

eugenet8k:
This change brought a decent amount of failures in our unit test coverage. We used sinon for stubbing various objects and ES imports. 

For example, a unit test where we stub some method in the Chaplin JS framework:

```
import Chaplin from 'chaplin'

describe(__filename, () => {

  beforeEach(() => {
    sandbox = sinon.createSandbox()
    ...
    sandbox.stub(Chaplin.utils, 'redirectTo')
    ...
})

```

fails with
```
[‣](http://localhost/test/)
TypeError: Cannot redefine property: redirectTo
    at Function.defineProperty (<anonymous>)
    at wrapMethod (webpack-internal:///./node_modules/sinon/pkg/sinon-esm.js:4651:16)
    at Function.stub (webpack-internal:///./node_modules/sinon/pkg/sinon-esm.js:3847:44)
    at Sandbox.stub (webpack-internal:///./node_modules/sinon/pkg/sinon-esm.js:3291:39)
    at Context.eval (webpack-internal:///./app/components/apps/app-row-layout.spec.js:54:13)
```

if I rollback this newly added one-liner
```
140:  methodDesc.configurable = true;
```

it works just fine.

The ChaplinJS source code looks something like
```
/*!
 * Chaplin 1.2.0
 *
 * Chaplin may be freely distributed under the MIT license.
 * For all details and documentation:
 * http://chaplinjs.org
 */

(function(root, factory) {
  if (typeof define === 'function' && define.amd) {
    define(['backbone', 'underscore'], factory);
  } else if (typeof module === 'object' && module && module.exports) {
    module.exports = factory(require('backbone'), require('underscore'));
  } else if (typeof require === 'function') {
    factory(window.Backbone, window._ || window.Backbone.utils);
  } else {
    throw new Error('Chaplin requires Common.js or AMD modules');
  }
}(this, function(Backbone, _) {
  function require(name) {
    return {backbone: Backbone, underscore: _}[name];
  }

  require =(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
'use strict';
module.exports = {
  Application: require('./chaplin/application'),
  Composer: require('./chaplin/composer'),
  Controller: require('./chaplin/controllers/controller'),
  Dispatcher: require('./chaplin/dispatcher'),
  Composition: require('./chaplin/lib/composition'),
  EventBroker: require('./chaplin/lib/event_broker'),
  History: require('./chaplin/lib/history'),
  Route: require('./chaplin/lib/route'),
  Router: require('./chaplin/lib/router'),
  support: require('./chaplin/lib/support'),
  SyncMachine: require('./chaplin/lib/sync_machine'),
  utils: require('./chaplin/lib/utils'),
  mediator: require('./chaplin/mediator'),
  Collection: require('./chaplin/models/collection'),
  Model: require('./chaplin/models/model'),
  CollectionView: require('./chaplin/views/collection_view'),
  Layout: require('./chaplin/views/layout'),
  View: require('./chaplin/views/view')
};
```

fatso83:
@eugenet8k I am sorry to hear that, but I have no idea how this could happen from the brief description. Would it be possible for you to file a bug report with a reproducible example? Guessing this happens in some proprietary product, so for comparison, this is what I usually do myself if I do not know how to recreate a minimal example from scratch:
1. Copy the proprietary source code to some other folder
2. Make/Find a failing test where the above issues happens
3. Delete all prod and test code that doesn't affect the test, backing up every time it crashes for other reasons
4. Modify/remove dependencies when something crashes, trimming away all dead code (like unused reducers in reduc stores)
5. Keep on pruning until you have an project consisting of only config files (babel, package.json, eslint, etc) and the minimal amount of code needed for reproddution (typically you should be able to just be left with 2-5 files)
6. Run some renaming script to rename variables/code
7. Make sure it still runs and push to github :)

Here's an example of what I run all production code through before it ends up on StackOverflow:
```
alias websafe='util.esed -e "s/(diffia|ptflow)/acme-org/gi" -e "s/$USER/my-user/gi" -e "s/(nimble|ptflow|pete)/foo-project/gi" -e "s/(clinic|people)/app-\1/gi" '
```
> diffia, pete, clinic, people and ptflow are project references I want to remove

So I can do 
```
pbpaste | websafe | pbcopy
```

to trim away all production references.

eugenet8k:
@fatso83 thank you for the extended comment, I created two code sandbox examples to demonstrate the issue, please review https://github.com/sinonjs/sinon/issues/2514

PR #2519, opened 2023-06-04
fix: bundling compatibility with webpack@5

AviVahl:
#### Purpose (TL;DR)
Ensure sinon bundles nicely for browsers without any errors or warnings.
https://github.com/sinonjs/sinon/issues/2411 re-fixed (after "exports" were added I think).

#### Background

When using webpack v5 to bundle cjs code that calls `require('sinon')`, it would have defaulted to "exports->require" and fail with multiple node API errors/warnings (util, timers, etc.).

This patch ensures that anyone who bundles sinon with a bundler that respects "exports" gets the (browser-compatible) esm version.

Tested on both webpack v5 and v4. should be noted that v4 worked even without this patch, as it automatically injected polyfills.  webpack@5 no longer does so.

#### Solution

The esm version bundles without any errors/warnings, so make sure bundlers pick that one up.

#### How to verify

Use the same steps to reproduce in https://github.com/sinonjs/sinon/issues/2411
Modify the node_modules with this patch and rerun ` npx webpack --mode development`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "LICENSE",
     "README.md"
   ],
-  "browser": "./pkg/sinon-esm.js",
+  "browser": "./lib/sinon.js",
   "main": "./lib/sinon.js",
   "module": "./pkg/sinon-esm.js",
   "exports": {
--- a/package.json
+++ b/package.json
@@ -116,11 +116,12 @@
     "LICENSE",
     "README.md"
   ],
-  "browser": "./lib/sinon.js",
+  "browser": "./pkg/sinon-esm.js",
   "main": "./lib/sinon.js",
   "module": "./pkg/sinon-esm.js",
   "exports": {
     ".": {
+      "browser": "./pkg/sinon-esm.js",
       "require": "./lib/sinon.js",
       "import": "./pkg/sinon-esm.js"
     },


fatso83:
Thanks, seems good! Needs a test run, of course :-)

AviVahl:
Looks like browserify doesn't like the esm in package.json->browser. I'll revert that one and leave the one in the exports. All should be well then.

AviVahl:
Reverted. Tests should (hopefully) now pass.

On a related note, please consider dropping browserify support... it's pretty outdated in 2023.
Might be better to test widely-used bundlers, like: webpack, rollup, esbuild, etc.

AviVahl:
@fatso83 heya, any chance we can rerun ci on the smaller change?

AviVahl:
Now passes on CI. :) Would appreciate if this could land main, as it would allow us to remove workarounds to the bundling issues in several repositories.

fatso83 (approved):

fatso83:
Ironically, this seems to have broken MUI's webpack setup: https://github.com/mui/material-ui/pull/37430 👀

AviVahl:
Yeah, webpack now resolves the esm version, so this config: https://github.com/mui/material-ui/blob/master/test/regressions/webpack.config.js#LL26C1-L29C8 probably needs to now say `process: 'process/browser.js'`

They might even be able to remove the ProvidePlugin completely if sinon was the only place needing this polyfill.

PR #2537, opened 2023-08-11
Enable use of assignment in the presence of accessors

fatso83:
#### Purpose

Allows specifying a fourth argument, an options bag, to `sandbox.replace()` to make it possible to override the default behavior of throwing when encountering accessor methods (getters/setters).

This means we will pass the supplied replacement to the setter and vice-versa get the original value from the getter when restoring.

Issue #2503 made the case for using assignment, instead of defining object descriptors, in injection methods. Enabling this in all existing methods would probably break some projects in unforeseen ways. It would also mean adding yet another argument to most of these, which would make them harder to use, as some of them (spy) already have optional third arguments today.

The discussion in #2503 revealed we already had a "special case" in our `Sandbox#replace` method, as this already uses basic assignment. The only thing preventing it from being used was that it tries to prevent false usage by throwing on encounter accessor methods for the given property. Thus this PR simply opens up for circumenting this.

This opens up for some exotic cases, such as manual setups for stubbing the exports of true ESM modules through the innards of the module, in the way described in that feature:

**my-module.mjs**
```javascript
export function doThing() {
    console.log("Wow, we did the thing!");
}

export function doThingTwice() {
    for (let i = 0; i < 2, i +=1) doThing();
}

export const myModule = {
    get doThing() {
        return doThing;
    },
    set doThing(value) {
        doThing = value;
    }
};
```

**my-module.test.mjs**
```javascript
import {doThingTwice, myModule} from "./my-module.mjs";
import sinon from "sinon";

const stub = sinon.stub(myModule, "doThing");

doThingTwice();

sinon.assert.calledTwice(stub);
```

Alternatively one could, of course, _not_ do this change, and instead say that this manual stubbing setup is already doable by exposing functions for setting the corresponding exports and manually invoking these. The downside is that this will cause lots of manual restore code, which would otherwise by handled by Sinon.

Run tests and alternatively try the copy-pasting the example above. You will typically need to say `import sinon from '../lib/sinon-esm'`.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -181,13 +181,13 @@ A convenience reference for [`sinon.assert`](./assertions)
 
 _Since `sinon@2.0.0`_
 
-#### `sandbox.replace(object, property, replacement);`
+#### `sandbox.replace(object, property, replacement, options);`
 
 Replaces `property` on `object` with `replacement` argument. Attempts to replace an already replaced value cause an exception. Returns the `replacement`.
 
 `replacement` can be any value, including `spies`, `stubs` and `fakes`.
 
-This method only works on non-accessor properties, for replacing accessors, use `sandbox.replaceGetter()` and `sandbox.replaceSetter()`.
+By default, this method only works on non-accessor properties, for replacing accessors, use `sandbox.replaceGetter()` and `sandbox.replaceSetter()`.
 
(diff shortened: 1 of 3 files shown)


fatso83:
Wrong name of branch. Closing and reopening

Issue #2403, opened 2021-09-17
Option to use assignment operator in injection methods.


sripberger:
**Is your feature request related to a problem? Please describe.**
Yes. Sinon's injection and cleanup features are difficult to use for truly unobtrusive unit testing of function modules-- i.e. modules that just export some functions for consumers to use. This is fundamentally a result of how CommonJS modules work, and something which *can* be solved in ES modules.

**Describe the solution you'd like**
I'd like to add an optional argument to the `sinon.stub` and `sinon.spy` methods, specifically the signatures which take an object and a method name, for example  `const stub = sinon.stub(object, "method");` This optional argument would cause the injection logic to use a simple assignment instead of `Object.defineProperty` both to put the double in place,  and to restore the original function during cleanup. Something like this:

```js
const stub = sinon.stub(object, "method", {useAssignment: true});
```

This feature could be used to implement a standards-compliant solution for getting test doubles in place that doesn't require major changes to how code is typically written in modules like these. Under additional context below I have a much more exhaustive explanation of why this is the case, if anybody cares to read a bunch.

**Describe alternatives you've considered**
I am considering writing a wrapper around sinon that duplicates some of this injection logic, but it would be easier and better if I didn't have to do that. Additional context explains several other alternatives I have used historically and why they don't really work with ES modules.

**Additional context**
Sinon is fantastic and one of my favorite things about working in Node.js. Not only is the API expressive and powerful, it typically makes injection of test doubles very easy and doesn't force developers to drastically restructure their code in order to accommodate it.

Want to stub some behavior in a class? Just make it into an instance method, construct an instance like normal, then call `sinon.stub(instance, "method").`  It doesn't really get much simpler than that. No need to say, implement this behavior in a separate class and then ask all consuming code to provide this implementation every time. You can still do this if you think it's appropriate, but you're not *forced* to do it all the time, which I love.

This simplicity breaks down in one case, though, and that is when dealing with function modules-- i.e. modules that just contain functions which are not meant to be used as class constructors. Historically, in CommonJS, you are forced to consider how variables defined in the top scope of a module are fully hidden and cannot be affected from outside the module. If you write your modules like this, for example (and these are all intentionally-trivial examples):

my-module.js
```js
function doThing() {
    console.log("Wow, we did the thing!");
}
exports.doThing = doThing;

function doThingTwice() {
    for (let i = 0; i < 2, i +=1) doThing();
}
exports.doThingTwice = doThingTwice;
```

[...]

PR #2538, opened 2023-08-11
Enable use of assignment in the presence of accessors

fatso83:
#### Purpose

Allows specifying a fourth argument, an options bag, to `sandbox.replace()` to make it possible to override the default behaviour of throwing when encountering accessor methods (getters/setters). 

This means we will pass the supplied replacement to the setter and vice-versa get the original value from the getter when restoring.

P.S. I am not sold on the options name, so could be something else like "allowUsingAccessor". Was just inspired by the original issue.

Issue #2403 made the case for using assignment, instead of defining object descriptors, in injection methods. Enabling this in all existing methods would probably break some projects in unforeseen ways. It would also mean adding yet another argument to most of these, which would make them harder to use, as some of them (spy) already have optional third arguments today.

The discussion in #2403 revealed we already had a "special case" in our `Sandbox#replace` method, as this already uses basic assignment. The only thing preventing it from being used was that it tries to prevent false usage by throwing on encounter accessor methods for the given property. Thus this PR simply opens up for circumenting this.

This opens up for some exotic cases, such as manual setups for stubbing the exports of true ESM modules through the innards of the module, in the way described in that feature:

**my-module.mjs**
```javascript
export function doThing() {
    console.log("Wow, we did the thing!");
}

export function doThingTwice() {
    doThing();
    doThing();
}

export const myModule = {
    get doThing() {
        return doThing;
    },
    set doThing(value) {
        doThing = value;
    },
};
```

**my-module.test.mjs**
```javascript
import { doThingTwice, myModule } from "./my-module.mjs";
import sinon from "./lib/sinon.js";

const stub = sinon.replace(myModule, "doThing", sinon.fake(), {
    forceAssignment: true,
});

doThingTwice();

sinon.assert.calledTwice(stub);
```

Alternatively one could, of course, _not_ do this change, and instead say that this manual stubbing setup is already doable by exposing functions for setting the corresponding exports and manually invoking these. The downside is that this will cause lots of manual restore code, which would otherwise by handled by Sinon.

Run tests and alternatively try the copy-pasting the example above into the root of the repo, then run `node *.test.mjs`.

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -236,7 +236,7 @@ console.log(myObject.myMethod());
 // strawberry
 ```
 
-#### `sandbox.replace.usingAccessor(obj, property, value);`
+#### `sandbox.replace.usingAccessor(object, property, value);`
 
 Usually one intends to _replace_ the value or getter of a field, but there are use cases where one actually wants to _assign_ a value to a property using an existing setter. `#replace.usingAccessor(object, property, value)` will do just that; pass the value into setter function and vice-versa use the getter to get the value used for restoring later on.
 
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -241,6 +241,7 @@ console.log(myObject.myMethod());
 Usually one intends to _replace_ the value or getter of a field, but there are use cases where one actually wants to _assign_ a value to a property using an existing setter. `#replace.usingAccessor(object, property, value)` will do just that; pass the value into setter function and vice-versa use the getter to get the value used for restoring later on.
 
 ##### Use case: no-frills dependency injection in ESM with cleanup
+
 One use case can be to conveniently allow ESM module stubbing using pure dependency injection, having Sinon help you with the cleanup, without resorting to external machinery such as module loaders or require hooks (see [#2403](https://github.com/sinonjs/sinon/issues/2403)). This would then work regardless of bundler, browser or server environment.
 
 #### `sandbox.replaceGetter(object, property, replacementFunction);`
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -181,7 +181,6 @@ A convenience reference for [`sinon.assert`](./assertions)
 
 _Since `sinon@2.0.0`_
 
-<<<<<<< HEAD
 #### `sandbox.define(object, property, value);`
 
 Defines the `property` on `object` with the value `value`. Attempts to define an already defined value cause an exception.
@@ -220,7 +219,7 @@ Replaces `property` on `object` with `replacement` argument. Attempts to replace
 
 `replacement` can be any value, including `spies`, `stubs` and `fakes`.
 
-By default, this method only works on non-accessor properties, for replacing accessors, use `sandbox.replaceGetter()` and `sandbox.replaceSetter()`.
+This method only works on non-accessor properties, for replacing accessors, use `sandbox.replaceGetter()` and `sandbox.replaceSetter()`.
 
 ```js
 var myObject = {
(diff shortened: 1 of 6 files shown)


mantoni:
Looks good. While the name is okay, I'm also wondering about alternatives.

Suggestion: `accessorInjection` or `useAccessorInjection`

Another question: should the whole thing fail with a specific error message if the flag is specified, but no accessor pair is found?

mantoni:
One more thought: Since this is a completely different approach to swapping out a function, we could also introduce `sandbox.inject(...)` to make it a bit less verbose. WDYT?

fatso83:
Maybe that's just as good? We would then have define, inject and replace, all with dedicated use cases.

fatso83:
I had just about finished the changes when I found that [we already _have_ Sandbox#inject](https://github.com/sinonjs/sinon/blob/5fc17c73681b095a0ebd69ac9be783bf6efe7406/lib/sinon/sandbox.js#L88). Pretty sure it is not the most well-known function and frankly I am not totally sure what its usecase is. We even do not have any documentation on it and it is only used once in lib/sinon/create-sandbox.js (line 53). All of that sandbox setup related code is basically shrouded in mystery by now, so I have no idea how all of this works.

So ... I need a new name. 
- `#accessorAssign` - too brief?
- `#injectUsingAccessor` - overloading the term "inject", confusing readers when comparing to the existing mentions of `injectTo` and `inject()`
- `#forceAssign` - I think the below gives more hints?
- `#forceAccessors` - will probably make one wonder what it does, but clearer on first reading of docs?

Currently aiming for something like `sinon.forceAccessors(obj, 'doFoo', fake)`. Whatcha think? 🤷 

`#naming_is_hard`

mantoni:
I agree that `sinon.assignUsingAccessors(obj, 'doFoo', fake)` is the most clear. Quite a handful to type though, but I'm also running out of ideas.

The very last thing I can thing of: I checked the work on `sinon.define(...)` and it uses assignments, if I understand correctly. So the use case here would work with it, however I don't think the `restorer`  would. Do you think it's worth trying to fold this functionality into `define`? Or do you prefer keeping them separate? I don't have a strong opinion here.

mantoni:
I agree with the _intent_ argument. It's a different thing and we should probably not mix them

I'm not really happy with the this, but can't think of anything better. Maybe @mroderick has more ideas?

fatso83:
I think we already agreed to forego the fourth argument. The discussion now is basically about which dedicated name to use.

This is not about replacing accessors, Morgan, so the existing functions for replacing accessors should stay as they are. This is just about assigning values using the accessors, without replacing them. See the expanded discussion.

fatso83:
I like it!

mroderick:
>  I can just update this and get this merged if there's agreement.

Go ahead!

fatso83:
When actually writing it out, I had second thoughts about `replace.accessor`, since it sounds like you want to replace the accessor, so instead I resorted to `usingAccessor`. I think it both reads better and makes more sense. Should be ready to merge, unless anyone wants more unit tests. I just copied the assertions from the normal `.replace`, but I haven't added additional checks for the method reusing them.

mantoni (approved):
Just one tiny comment. 👍

docs/release-source/release/sandbox.md
```diff
@@ -236,9 +236,17 @@ console.log(myObject.myMethod());
 // strawberry
 ```
 
-#### `sandbox.replaceGetter();`
+#### `sandbox.replace.usingAccessor(obj, property, value);`
```
fatso83, line 239:
```suggestion
#### `sandbox.replace.usingAccessor(object, property, value);`
```

PR #2539, opened 2023-08-11
.define method

### LINKED ISSUE #2195: Conditional support for replacing props not defined by the object

gukoff:

Add the `.define(obj, property, value)` method to the sandbox to allow temporarily define properties on the objects (i.e. until the sandbox is restored).

Fixes  #2195

Run the tests

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -4,7 +4,7 @@ title: Sandboxes - Sinon.JS
 breadcrumb: sandbox
 ---
 
-Sandboxes removes the need to keep track of every fake created, which greatly simplifies cleanup.
+Sandboxes remove the need to keep track of every fake created, which greatly simplifies cleanup.
 
 ```javascript
 var sandbox = require("sinon").createSandbox();
@@ -180,6 +180,40 @@ var sandbox = sinon.createSandbox({
 A convenience reference for [`sinon.assert`](./assertions)
 
 _Since `sinon@2.0.0`_
+
+#### `sandbox.define(object, property, value);`
+
+Defines the `property` on `object` with the value `value`. Attempts to define an already defined value cause an exception.
+
+`value` can be any value except `undefined`, including `spies`, `stubs` and `fakes`.
+
+```js
+var myObject = {};
+
+sandbox.define(myObject, "myValue", function () {
+    return "blackberry";
+});
+
+sandbox.define(myObject, "myMethod", function () {
+  return "strawberry";
+});
+
+console.log(myObject.myValue);
+// blackberry
+
+console.log(myObject.myMethod());
+// strawberry
+
+sandbox.restore();
+
+console.log(myObject.myValue);
+// undefined
+
+console.log(myObject.myMethod);
+// undefined
+```
+
+_Since `sinon@15.3.0`_
 
 #### `sandbox.replace(object, property, replacement);`
 
--- a/docs/release-source/release/sandbox.md
+++ b/docs/release-source/release/sandbox.md
@@ -191,7 +191,7 @@ Defines the `property` on `object` with the value `value`. Attempts to define an
 var myObject = {};
 
 sandbox.define(myObject, "myValue", function () {
-    return "blackberry";
+  return "blackberry";
 });
 
 sandbox.define(myObject, "myMethod", function () {
--- a/lib/sinon/sandbox.js
+++ b/lib/sinon/sandbox.js
@@ -296,7 +296,8 @@ function Sandbox() {
             throw new TypeError("Expected value argument to be defined");
         }
 
-        // likely a no-op, but just in case we allow to replace by undefined in the future
+        // check just in case we allow to replace properties by undefined in the future
+        // in the implementation at the moment of writing, this check always succeeds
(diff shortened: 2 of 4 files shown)


lib/sinon/sandbox.js
```diff
@@ -268,8 +271,6 @@ function Sandbox() {
             );
         }
 
-        verifyNotReplaced(object, property);
```
gukoff, line 271: Typo, this shouldn't be deleted.

Need a new test for this line.

lib/sinon/sandbox.js
```diff
@@ -278,6 +281,33 @@ function Sandbox() {
         return replacement;
     };
 
+    sandbox.define = function define(object, property, value) {
+        const descriptor = getPropertyDescriptor(object, property);
+
+        if (typeof descriptor !== "undefined") {
+            throw new TypeError(
+                `Cannot define the already existing property ${valueToString(
```
fatso83, line 289: Might leave a hint that one should use `replace`

lib/sinon/sandbox.js
```diff
@@ -278,6 +281,33 @@ function Sandbox() {
...
+                    property
+                )}`
+            );
+        }
+
+        if (typeof value === "undefined") {
+            throw new TypeError("Expected value argument to be defined");
+        }
```
fatso83, line 297: It is possible to have a descriptor with a `value` of `undefined`, which is different from not being defined, but this might be fine in practice. Edge cases :)

fatso83 (commented):
This looks like a great start! I'll leave this for next week, as I am off to a train to the mountains. Thanks!

docs/release-source/release/sandbox.md
```diff
@@ -181,6 +181,40 @@ A convenience reference for [`sinon.assert`](./assertions)
...
+sandbox.restore();
+
+console.log(myObject.myValue);
+// undefined
+
+console.log(myObject.myMethod);
+// undefined
+```
+
+_Since `sinon@15.3.0`_
```
gukoff, line 216: I just put the next minor version here. Let me know if the docs need to be updated separately.

lib/sinon/sandbox.js
fatso83: I think this is _fine_ 😄

fatso83 (changes requested):
Seems pretty ready for inclusion. Just added some comments. You are free to object.

lib/sinon/sandbox.js
```diff
@@ -278,6 +285,33 @@ function Sandbox() {
...
+                )}. Perhaps you meant sandbox.replace()?`
+            );
+        }
+
+        if (typeof value === "undefined") {
+            throw new TypeError("Expected value argument to be defined");
+        }
+
+        // Check just in case we allow to replace properties by undefined in the future.
+        // In the implementation at the moment of writing, this check always succeeds.
```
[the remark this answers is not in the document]
gukoff: I meant there's no way the value was replaced before, because we already checked that the descriptor is still undefined. I removed the comment, now I don't think it's really useful.

fatso83 (approved):

fatso83:
> [warn] Code style issues found in 2 files. Forgot to run Prettier?

Could you just address this and I'll merge 😄

Issue #2528, opened 2023-07-25
Bug mocking with SWC


assertnotnull:
**Describe the bug**
Sinon is unable to mock the function when compiled with SWC but works with ts-node

**To Reproduce**
clone [the sample repo with bug
](https://github.com/assertnotnull/sinon-swc-bug)
```
pnpm i
pnpm test
```
Outcome:
```
1) main
       should mock:
     TypeError: Descriptor for property toBeMocked is non-configurable and non-writable
      at assertValidPropertyDescriptor (node_modules/.pnpm/sinon@15.2.0/node_modules/sinon/lib/sinon/stub.js:138:15)
      at Function.stub (node_modules/.pnpm/sinon@15.2.0/node_modules/sinon/lib/sinon/stub.js:89:5)
      at Sandbox.stub (node_modules/.pnpm/sinon@15.2.0/node_modules/sinon/lib/sinon/sandbox.js:389:39)
      at Context.<anonymous> (src/main.spec.ts:12:22)
      at processImmediate (node:internal/timers:476:21)
```

To see it working with ts-node replace "@swc/register" with "ts-node/register"
Then run pnpm test again and it passes.

**Expected behavior**
It should mock

fatso83:
This is _clearly_ not an issue with Sinon, but with your transpiler tooling. It even says clearly what the issue is: 
> Descriptor for property toBeMocked is non-configurable and non-writable

While ts-node clearly exports a writeable [object descriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor), SWC does not. This is in-line with _how ES Modules are supposed to work_, so SWC seems to be the correct behavior here.

Just googling "swc writable object descriptor" comes up with [a few suggestions](https://stackoverflow.com/a/72793194/200987), some ok workarounds. I would still try another approach, if you just need to do this to have mutable namespaces for your modules, and that is to use the [Dalton's ESM loader](https://www.npmjs.com/package/esm). It has an option to make namespaces mutable and you can register it before your SWC module.

assertnotnull:
I've tried the suggestion in stackoverflow as I also found this before but it's not working either.
Other libraries complexifies the tests too much that it outweighs the benefits.
I guess running tests with mocks is not an option with SWC.

fatso83:
[...]

PR #2540, opened 2023-08-13
New article: making Sinon work with complex setups

fatso83:
#### Purpose 

Wrote an article with-step-by-step instructions in how one can figure out what is preventing you from getting Sinon to work in most environments, and show three different approaches that all enable you to use Sinon in mocking dependencies.

This is basically an elaboration of what I wrote in #2528 

Rendered as PDF: [Case study: real world dependency stubbing.pdf](https://github.com/sinonjs/sinon/files/12330557/Case.study.real.world.dependency.stubbing.pdf)

Testing in Jekyll:
- have Ruby 2.7.8 with Bundler installed
- `cd docs;  bundle install; npm run serve-docs`


docs/_howto/link-seams-commonjs.md
```diff
@@ -1,13 +1,19 @@
...
 
-This is the CommonJS version, so we will be using [proxyquire][proxyquire] to construct our seams.
+> If you want a better understand of the example and get a good description of what _seams_ are, we recommend that you read the [seams (all 3 web pages)][seams] excerpt from the classic [Working Effectively with Legacy Code][legacy], though it is not strictly necessary.
 
-To better understand the example and get a good description of what seams are, we recommend that you read the [seams (all 3 web pages)][seams] excerpt from [Working Effectively with Legacy Code][legacy] before proceeding.
+This guide targets the CommonJS module system, made popular by NodeJS. There are other module systems, but until recent years this was the de-facto module system and even when the actual EcmaScript Module standard arrived in 2015, transpilers and bundlers can still _output_ code as CJS modules. For instance, Typescript outputs CJS modules per default as of 2023, so it is still relevant, as your `import foo from './foo'` might still end up being transpiled into `const foo = require('./foo')` in the end.
+
+<!-- TODO: input link to the other article on stubbing ESM -->
```
fatso83, line 12: I should add that link now 😅

docs/_howto/typescript-swc.md
```diff
@@ -0,0 +1,257 @@
...
+    configurable: true
+  }
+}
+mocked
+    ✔ should mock
+```
+
+The important difference to note about the object `Other` is that the property `toBeMocked` is a simple writable _value_ in the case of `ts-node` and a non-configurable _getter_ in the case of SWC. It being a getter is not a problem for Sinon, as we have a multitude of options for replacing those, but if `configurable` is set to `false` Sinon cannot really do anything about it.
+
+If we take a look at
```
fatso83, line 150: Incomplete sentence 🤦‍♂️

docs/_howto/typescript-swc.md
```diff
@@ -0,0 +1,257 @@
...
+const stub = sandbox.fake.returns("mocked")
+sandbox.replaceGetter(Other, "toBeMocked", () => stub)
+```
+
+## Use pure dependency injection
+
+> [Working code][pure-di]
+
+This technique works regardless of language, module systems, bundlers and tool chains, but requires slight modifications of the SUT to allow modifying it. You also do not get help from Sinon in automatically resetting state.
+
```
fatso83, line 188: Could add a link to the ongoing PR I made to address this

docs/_howto/typescript-swc.md
```diff
@@ -0,0 +1,257 @@
...
+A getter _is_ different from a value, so you need to change your testcode slightly to replace the getter:
+
+```
+const stub = sandbox.fake.returns("mocked")
+sandbox.replaceGetter(Other, "toBeMocked", () => stub)
+```
+
+## Use pure dependency injection
+
+> [Working code][pure-di]
```
fatso83, line 185: Include link to section on testing in Single Page Apps In Depth

https://singlepageappbook.com/maintainability3.html#testing-interactions-between-modules

assertnotnull:
Nice article!

kasir-barati:
Thanks for the article @fatso83, just a comment. What if one does not want to use swc? and configure its jest in a way to export the configureable objects?

fatso83:
See the two highlighted sections. This is not meant as a guide on SWC or Typescript, which was just chosen as a popular and slightly complicated example that could be used to show different techniques. It's a general guide on how you can analyse the situation and figure out how to work around it once you know what's going on.

I don't know your particular case and what the issue is, but we're not really a support channel, use StackOverflow or ChatGPT for that. We are interested in improving docs to avoid confusion, of course, so any tips is worth mentioning.

Wrt Jest I used it with Typescript and Sinon in a previous project. I remember that as working fairly well. Jest has control over the entire runtime, so it can do magic a library cannot 🪄 AFAIK we used the built-in jest.mock(...) for replacing dependencies. You can then choose to either create the stubs using Sinon or Jest. It's mostly a matter of preference. I don't know how to configure Jest to do what you propose, but I think you would need a plugin at least. Maybe even write it yourself.

A newer approach of the old _manual_ DI, is the "no extra tooling needed" approach that was just exposed with the new sandbox.replace.usingAccessor. That allows for tests that will work the same in Jest or Mocha or any other runner. You should check it out. I should of course update the article to mention that.

Issue #2525, opened 2023-07-18
.throws with no message creates error with empty message


mroderick:
When calling `throwsException` with a `String` as the `error` argument and no `message` argument, it  will create errors with an empty message as seen in the code sample below.

```js
function throwsException(fake, error, message) {
    if (typeof error === "function") {
        fake.exceptionCreator = error;
    } else if (typeof error === "string") {
        fake.exceptionCreator = function () {
            const newException = new Error(message || "");
            newException.name = error;
            return newException;
        };
    } else if (!error) {
        fake.exceptionCreator = function () {
            return new Error("Error");
        };
    } else {
        fake.exception = error;
    }
}
```

While this is not a bug as such, I do think it is poor design to return an Error instance with an empty `message` property.

Instead, we could calculate a reasonable error message from the `error` argument.

**To Reproduce**
Steps to reproduce:

```js
const sinon = require("sinon");

const callback = sinon.stub();
callback.withArgs(1).throws("apple pie");

try {
    callback(1);
} catch (error) {
    console.log(`error.message: "${error.message}"`;
}
```

**Expected behavior**
Non-empty `error.message`

ahmed1hsn:
`function throwsException(fake, error, message) {
    if (typeof error === "function") {
        fake.exceptionCreator = error;
    } else if (typeof error === "string") {
        fake.exceptionCreator = function () {
            const newException = new Error(message || error || "");
            newException.name = error;
            return newException;
        };
    } else if (!error) {
        fake.exceptionCreator = function () {
            return new Error("Error");
        };
    } else {
        fake.exception = error;
    }
}`
@mroderick as `error` is already string in the block, and `message` is undefined (probably), we can conditionally use `error` in case there is no `message` inside this block. 
Would this solve the issue?

peanutenthusiast:
Hi, @mroderick, is this issue still open? I'd be happy to submit a PR with @ahmed1hsn's suggested changes if helpful.

peanutenthusiast:
Hi @fatso83, 

When I applied @ahmed1hsn's suggestion and ran `npm test`, I got one failing test

```
1472 passing (2s)
  11 pending
  1 failing

  1) sinonSpy.call
       call.toString
         includes exception:

      AssertionError: [assert.equals] 'doIt() !TypeError(TypeError)' expected to be equal to 'doIt() !TypeError'
      + expected - actual

      -doIt() !TypeError(TypeError)
      +doIt() !TypeError
      
[...]

PR #2544, opened 2023-09-12
fix(2525): ensure non empty message when error of type string is passed, but no message

peanutenthusiast:

When an error of type String and an undefined message are passed to `throwsException`, ensure a non-empty error.message

See issue: #2525 

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

Note: The branch currently causes [this test](https://github.com/sinonjs/sinon/blob/main/test/proxy-call-test.js#L1128) to fail, so I'd appreciate guidance on whether expectations change, given this will result in a major semver change.

### DIFF
--- a/test/stub-test.js
+++ b/test/stub-test.js
@@ -941,16 +941,19 @@ describe("stub", function () {
         });
 
         it("creates a non empty error message when error is a string and no message is passed", function () {
-            const stub = createStub()
+            const stub = createStub();
 
-            stub.withArgs(1).throws("TypeError")
+            stub.withArgs(1).throws("TypeError");
 
-            assert.exception(function () {
-                stub(1)
-            }, {
-                message: "Sinon-provided TypeError"
-            })
-        })
+            assert.exception(
+                function () {
+                    stub(1);
+                },
+                {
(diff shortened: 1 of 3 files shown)


test/stub-test.js
```diff
@@ -940,6 +940,18 @@ describe("stub", function () {
...
+        it("creates a non empty error message when error is a string and no message is passed", function () {
+            const stub = createStub()
+
+            stub.withArgs(1).throws("apple pie")
+
+            assert.exception(function () {
+                stub(1)
+            }, {
+                message: "Sinon-provided apple pie"
+            })
```
fatso83, line 952: Since the passed string is the actual error name, maybe use something that sounds like an error? Just in case someone casually looks at this for reference, it might confuse more than if you had passed something like "ValidationError".

fatso83:
Just fix the Prettier thingie and this is good to merge :)

mroderick (changes requested):

mroderick:
Thank you for your contribution! 🥳 

This has been published to the npm registry as part of `sinon@16.0.0`

PR #2553, opened 2023-10-10
Add job to run tests in bun

mroderick:
This PR adds a job to the default workflow to run tests in `bun`, now that [it has reached v1.0](https://bun.sh/blog/bun-v1.0).

1. Check out this branch
2. `bun install`
3. `bun run test-bun`

---

Some of the tests are still broken, and I haven't figured out why yet ... so why not a draft PR, to see if you, dear reader, might have some insights.


mroderick:
I'm going to close this for now, and may resume this in the future when Bun fixes some of it's incompatibility

PR #2565, opened 2023-10-28
breaking: Remove sinon.defaultConfig and related modules

### LINKED ISSUE #2561: Remove unused/meaningless exports and modules from the codebase
Both `sinon.defaultConfig` (`default-config.js`) and the `get-config` modules are unused in the sinon codebase. They originally existed to support sinonTest, which is a separate module, and should just be removed. They can be moved over into that project independently.

fatso83:
`default-config` and `get-config` are leftovers from when Sinon shipped with sinon.test (now the independent NPM module
 'sinon-test').

These serve no purpose internally, and really have no purpose but to help sinon-test create a base default. If needed, these [can be copied into the sinon-test](https://github.com/sinonjs/sinon-test/issues/204) project. No projects should depend on these (my assumption), but since it is a change of the API we mark it as a breaking change

When working with these I found out that some things had never been documented, so added docs for that.

fixes #2561


mroderick (approved):
Very nice PR! 🥇

PR #2569, opened 2023-11-05
Stop wrapping strings in Error

### LINKED ISSUE #1679: How can we override reject creating a new Error Object?
https://github.com/sinonjs/sinon/blob/4310343393b16b92b56526ab4614da4eb2886efa/lib/sinon/default-behaviors.js#L176

I'm maintaining some code that rejects strings and this is creating issues while testing. my bluebird catches are not behaving correctly.

```js
bluebird.rejects('Something')
    .catch(e => e === 'Something' /* true */, () => console.log(typeof e) /*string*/)
    .catch(() => console.log('should not be here'));
```
However using `sinon.stub().usingPromise('bluebird').rejects('Something');` ends up creating an error object with the name Something. Not expected at all.

Before someone says we are only supporting native Promises, let's investigate what they do!
```js
Promise.reject('Something')
     .catch(e => {
        console.log(e === 'Something'); // true
        console.log(typeof e); // string
    });
```

Hmmm, they don't create an Error class, they just toss the exception it was given like a throw.
```js
try { 
    throw 'Something'; 
} catch (e) { 
    console.log(e === 'Something');  // true
    console.log(typeof e); // string
}
```

fatso83:

Fix #1679 by changing the Stub and Fake API to not wrap strings in Error

In #1679 it was raised how stub.rejects("error") does not reject the promise with the passed in reason, but chooses to wrap it in an error. Some people found that behavior surprising that and had objections to that API design.

@mantoni chimed in to say that he thought stub.rejects should behave in line with `Promise.reject`.

I made the point that while `stub.rejects(string)` might be surprising to someone comparing it to the `Promise.reject` API _it is totally _consistent_ with the current behavior of `stub.throws(string)` and that we should keep the two aligned.

 #### Solution  

This is _one_ solution to the issue: it changes the `rejects` and `throws` API's of Stub and Fake to not treat the single-argument string in a special matter. This will potentially break a lot of existing code, but at least keeps consistency across the API's for throwing sync and async.

There are other takes, one being to simply keep things as they are and document in the docs how to these are _utility_ methods to achieve the most normal flow (which is throwing an error sync and async), and simply update the docs (both JSDoc and homepage docs) to point to other methods for achieving more specialized cases.

For instance, if you want to throw a string using the API that is today (before this change), you can just create a fake or stub that does that:
```javascript
sinon.stub(() => {throw "foo"; })
// throws a string "foo"
```
Same goes for Promises / throwing async
```javascript
sinon.stub(() => Promise.reject("foo"))
// alternatively
sinon.stub().returns(Promise.reject("foo"))
```

I think this is just as valid an approach: making people aware that 
the basic building blocks are there. That's what I like about the
Fake API: fewer methods in that API to rather create explicit 
behavior that creates less misconceptions about what something is doing.

1. Check out this branch
2. `npm install`
3. <your-steps-here>

#### Checklist for author

- [ ] `npm run lint` passes

### DIFF
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -213,7 +213,7 @@ const defaultBehaviors = {
 
     rejects: function rejects(fake, error, message) {
         let reason;
-        if (typeof error === "string") {
+        if (typeof error === "string" && typeof message !== "undefined") {
             reason = new Error(message || "");
             reason.name = error;
         } else if (!error) {
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -13,11 +13,9 @@ const useRightMostCallback = -2;
 function throwsException(fake, error, message) {
     if (typeof error === "function") {
         fake.exceptionCreator = error;
-    } else if (typeof error === "string") {
+    } else if (typeof error === "string" && typeof message !== "undefined") {
         fake.exceptionCreator = function () {
-            const newException = new Error(
-                message || `Sinon-provided ${error}`,
-            );
+            const newException = new Error(message);
             newException.name = error;
             return newException;
         };
--- a/lib/sinon/fake.js
+++ b/lib/sinon/fake.js
@@ -74,32 +74,26 @@ fake.returns = function returns(value) {
 
 /**
  * Creates a `fake` that throws an Error.
- * If the `value` argument does not have Error in its prototype chain, it will
- * be used for creating a new error.
  *
  * @example
  * var f1 = sinon.fake.throws("hello");
  *
- * f1();
- * // Uncaught Error: hello
- *
+ * try { f1(); } catch (err} { console.log(typeof err, err); }
+ * // string hello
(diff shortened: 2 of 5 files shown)


fatso83:
I am partial to closing this in favor of just improving the documentation with more examples. The API has been stable for ages and is well documented: https://sinonjs.org/releases/v17/stubs/#stubthrows, so I think this just comes down to some unfortunate tooling effects.

Right now, it seems that tooling (VSCode, etc) has a hard time dealing with typescript definitions from one source and JSDoc  strings from another, so people might lose out on the _actual docs_. Not sure how to deal with this ATM, unfortunately, but it's really another issue.

Any thoughts, @JemiloII and @dgreenwald-ccs (seeing that you commented in the original issue).

JemiloII:
When I made that comment years back, promises were fairly new. A lot has changed and now we have async/await to avoid some the issues we had before. So having the docs with more examples would be very helpful so if someone does get confused, they can see it in the docs

PR #2575, opened 2024-01-04
Make sandbox concurrency safe

### LINKED ISSUE #2472: calledImmediatelyBefore fails when executed in parallel mode using node-tap
**Describe the bug**

`spy1.calledImmediatelyBefore(spy2)` fails when used in a [node-tap](https://node-tap.org/docs/api/parallel-tests/) test suite that is being executed in parallel. In serial mode it works without issue. `calledBefore` continues to work as normal. 

I've even tried using sandboxes for each test but it did not fix the issue.

**To Reproduce**

Repo: https://github.com/Ethan-Arrowood/sinon-spy-parallel-bug-reproduction

```sh
git clone https://github.com/Ethan-Arrowood/sinon-spy-parallel-bug-reproduction
cd sinon-spy-parallel-bug-reproduction
pnpm install
pnpm run test
```

To see tests passing comment out line 27 in `test/index.ts`:

```diff
+ // tap.jobs = 2;
- tap.jobs = 2;
```

**Expected behavior**

I'd expect my spy assertions to pass in both serial and parallel mode

- Library version: 14.0.0
- Environment: Node.js 18
- Example URL:
- Other libraries you are using: node-tap, typescript

remorses:
#### Make sandbox concurrency safe

Fix #2472, sandbox not working when running tests concurrently because of global callId variable

 #### Solution

Store callId in a context passed with an optional argument, which is the sandbox object when using sandbox

Added a test with the #2472 reproduction

1. Check out this branch
2. `npm install`
3. npm test

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/proxy-invoke.js
+++ b/lib/sinon/proxy-invoke.js
@@ -12,6 +12,9 @@ const bind = Function.prototype.bind;
 
 module.exports = function invoke(func, thisValue, args, ctx = globalContext) {
     const matchings = this.matchingFakes(args);
+    if (ctx.callId == null) {
+        ctx.callId = 0;
+    }
     const currentCallId = ctx.callId++;
     let exception, returnValue;
 
--- a/lib/sinon/proxy-invoke.js
+++ b/lib/sinon/proxy-invoke.js
@@ -2,6 +2,7 @@
 
 const arrayProto = require("@sinonjs/commons").prototypes.array;
 const proxyCallUtil = require("./proxy-call-util");
+const globalContext = require("./util/core/global-context");
 
 const push = arrayProto.push;
 const forEach = arrayProto.forEach;
@@ -9,11 +10,9 @@ const concat = arrayProto.concat;
 const ErrorConstructor = Error.prototype.constructor;
 const bind = Function.prototype.bind;
 
-let callId = 0;
-
-module.exports = function invoke(func, thisValue, args) {
+module.exports = function invoke(func, thisValue, args, ctx = globalContext) {
     const matchings = this.matchingFakes(args);
-    const currentCallId = callId++;
+    const currentCallId = ctx.callId++;
     let exception, returnValue;
 
     proxyCallUtil.incrementCallCount(this);
--- a/lib/sinon/proxy.js
+++ b/lib/sinon/proxy.js
@@ -254,7 +254,7 @@ function createProxy(func, originalFunc, ctx = globalContext) {
     return proxy;
 }
 
-function wrapFunction(func, originalFunc, ctx) {
+function wrapFunction(func, originalFunc, ctx = globalContext) {
     const arity = originalFunc.length;
     let p;
     // Do not change this to use an eval. Projects that depend on sinon block the use of eval.
--- a/lib/sinon/proxy.js
+++ b/lib/sinon/proxy.js
@@ -2,6 +2,7 @@
 
 const arrayProto = require("@sinonjs/commons").prototypes.array;
 const extend = require("./util/core/extend");
+const globalContext = require("./util/core/global-context");
 const functionToString = require("./util/core/function-to-string");
 const proxyCall = require("./proxy-call");
 const proxyCallUtil = require("./proxy-call-util");
@@ -240,8 +241,8 @@ delegateToCalls(proxyApi, "alwaysReturned", false, "returned");
 delegateToCalls(proxyApi, "calledWithNew", true);
 delegateToCalls(proxyApi, "alwaysCalledWithNew", false, "calledWithNew");
 
-function createProxy(func, originalFunc) {
-    const proxy = wrapFunction(func, originalFunc);
+function createProxy(func, originalFunc, ctx = globalContext) {
+    const proxy = wrapFunction(func, originalFunc, ctx);
 
     // Inherit function properties:
     extend(proxy, func);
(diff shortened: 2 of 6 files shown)


Ethan-Arrowood (approved):
Great work. Test case looks good to me. Appreciate you making this happen 🚀

test/sandbox-test.js
```diff
@@ -2377,4 +2377,52 @@ describe("Sandbox", function () {
...
+                    await this.execute("second");
+                    await Promise.resolve();
+                    await this.execute("third");
+                }
+
+                async execute(s) {
+                    for (const f of this[s]) {
+                        await f();
+                    }
+                }
```
fatso83, line 2400: This test was a bit hard to wrap my head around, at least at midnight ... 😄 Could you clarify what you are really doing? I mean, I assume you are somehow testing quazi-concurrent runs, interlacing runs from different sandboxes, but I still find it a bit hard to understand and read.

Is there some way you could re-order or refactor this in some way that made the intent as expressed through the code clearer? 

To me, it is not at all clear how the sandbox runs are being interlaced and in what order functions are being called. Would it not be much clearer if doing something a la this pseudo-code

```js
const sandboxASpy1 = sandboxA.spy();
const sandboxASpy2 = sandboxA.spy();
const sandboxBSpy1 = sandboxB.spy();
const sandboxBSpy2 = sandboxB.spy();

sandboxASpy1();
sandboxBSpy1();
sandboxBSpy2();
sandboxASpy2();

assert(sandboxASpy1.calledImmediatelyBefore(sandboxASpy2));
assert(sandboxBSpy1.calledImmediatelyBefore(sandboxBSpy2));
```

There is no need for awaits AFAIK. JS is single threaded no matter what, and the asynchronicity just muddles up what is going on IMHO.

lib/sinon/sandbox.js
```diff
@@ -445,8 +446,13 @@ function Sandbox(opts = {}) {
...
     }
 
-    sandbox.spy = function spy() {
-        const createdSpy = sinonSpy.apply(sinonSpy, arguments);
+    sandbox.spy = function spy(object, property, types) {
+        const createdSpy = sinonSpy.apply(sinonSpy, [
+            object,
+            property,
+            types,
+            sandbox,
```
fatso83, line 454: If you are passing in the sandbox as the context, then what is the need for the global context object?

lib/sinon/proxy.js
```diff
@@ -253,7 +254,7 @@ function createProxy(func, originalFunc) {
     return proxy;
 }
 
-function wrapFunction(func, originalFunc) {
+function wrapFunction(func, originalFunc, ctx = globalContext) {
```
fatso83, line 257: Why are all of these context objects optional, with a default global? Even the default Sinon instance is a sandbox in itself.

lib/sinon/proxy-invoke.js
```diff
@@ -2,18 +2,20 @@
...
 
-let callId = 0;
-
-module.exports = function invoke(func, thisValue, args) {
+module.exports = function invoke(func, thisValue, args, ctx = globalContext) {
     const matchings = this.matchingFakes(args);
-    const currentCallId = callId++;
+    if (ctx.callId == null) {
+        ctx.callId = 0;
+    }
```
fatso83, line 17: When does this ever happen? I cannot see any tests exercising this.

I see the fallback global module, where it is already initialised, and I see the sandbox, which is also initialised.

lib/sinon/util/core/global-context.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
+
+module.exports = {
+    callId: 0,
+};
```
fatso83, line 5: Is this even needed?

fatso83 (changes requested):
Thanks for having a go at this, @remorses! I tried taking a proper look, and I think I got there in the end, even though it was late. 

There are two main issues I need covered:
1. this needs to work for fakes, stubs and mocks as well
2. Optional context seems like a bad idea

Since stubs are spies that should be a wrap. I _guess_ mocks are covered by being proxies ... ? But AFAIK fakes do not share much code with these, so that needs to be addressed (I noted that in https://github.com/sinonjs/sinon/issues/2472#issuecomment-1612901648).

I am pretty sure you could template the test case so that you can make the same test for both spies and fakes, btw, just parameterising it by passing in some factory method.

Wrt the context, the fact that it has been made optional concerns me. Why is it optional? Is it just to avoid having to change testcases? A good IDE should be able to fill in default values for you when adding a new parameter. If it's just for convenience, I would rather just fix it. All spies are created from a sandbox, even the ones on the default Sinon instance, so there should be no need for the global module, AFAIK.

fatso83:
This currently fails on your branch:
```javascript
✦ ❯ cat test/concurrency-test.js
const assert = require("node:assert");
const sinon = require("../lib/sinon");

const sbA = sinon.createSandbox();
const sbB = sinon.createSandbox();
const F = sbA.fake();

const f1 = sbB.fake();
const f2 = sbB.fake();

f1();
F();
f2();
assert(f1.calledImmediatelyBefore(f2));
```

fatso83:
@remorses Do you need some guidance in how to progress? AFAIK the old-skool API is covered (haven't checked the mocks API, but I think that should be covered by the proxy stuff).

fatso83:
Ping?

fatso83:
This does not seem like it is going places. Closing due to unresponsiveness.

PR #2577, opened 2024-01-05
Link to createStubInstance util.md docs in stubs.md

DanKaplanSES:

The Utilities documentation contains valuable information about `createStubInstance`, but I didn't see any way to know that unless you happen to have already viewed the Utilities page. 

Alternatively, could it make sense to move all of the `createStubInstance` docs into stubs.md? For me, that's where I would expect to find everything there is to know about `createStubInstance`, or at least the behavior it provides.

1. Check out this branch
2. `npm install`
3. <your-steps-here>

#### Checklist for author

- [ ] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/stubs.md
+++ b/docs/release-source/release/stubs.md
@@ -126,6 +126,8 @@ var stub = sinon.createStubInstance(MyConstructor);
 stub.foo.returns(3);
 ```
 
+See the [Utilities API](../utils) for more information.
+
 #### `stub.withArgs(arg1[, arg2, ...]);`
 
 Stubs the method only for the provided arguments.


fatso83 (approved):

fatso83:
> Alternatively, could it make sense to move all of the createStubInstance docs into stubs.md

I get it from the name, but it's not a method on the stubs object, and it's mostly similar by name only. If it should be moved somewhere, I guess it would be on the sandbox, if anything. But linking from different docs is highly effective and we should only do it more than we do. 

I know the docs are far from perfect, but it's what we have now. I know @mroderick is cooking on something, but I don't know what will become of it.

DanKaplanSES:
@fatso83 
> I guess it would be on the sandbox, if anything. But linking from different docs is highly effective and we should only do it more than we do.

Hm, that's interesting. Until [your comment on this issue](https://github.com/sinonjs/sinon/pull/2576#discussion_r1446125092), I didn't realize this was documented in three places. Should I submit a PR to add links to the other docs in these places?

fatso83:
It doesn't mean that the docs are correct, though ... I see that the docs say that there is a difference between the `sinon. createStubInstance` and `sandbox.createStubInstance`, but that might be leftovers from the Sinon 1.x days. As of Sinon 2.0, the default `sinon` instance _is a sandbox_. Cannot see how they are different (please test if you think otherwise)? So I'd rather just have the docs in one place and just link to that.

DanKaplanSES:
> So I'd rather just have the docs in one place and just link to that.

I think I understand but I'm not 100% sure. In other words, the sandbox documentation should be the one and only place to document `createStubInstance` and the stub/util documents should mostly link to it.

Issue #2629, opened 2024-11-27
Error when stubbing a property/function that does not exist


DanKaplanSES:
I'd like a more informative error message for this situation:

```
        const x = {
            getFoo: () => {
                return "bar"
            }
        }

        sinon.stub(x, 'getFoo').value("baz");

        console.log(x.getFoo()); // TypeError: x.getFoo is not a function
        console.log(typeof x.getFoo); // string
```

The intent here was to stub `x`'s `getFoo` method, but the user mistakenly used `value` to do so. In this situation, it would be more helpful for sinon to error on the `stub` line with an error message similar to: "`x.getFoo` is a function, not a getter. Use `returns` instead of `value`."

I'm aware this is a breaking change, so I think the user would need to configure some sort of "strict mode" to enable an error message like this.

senocular:
A property being a function doesn't mean it should always be a function. It may be desirable to replace the function entirely rather than what that function returns.

Consider something like the `HTMLElement.prototype.onclick` callback. This is a function property that can be a function or null.  As a function the function is called when the respective element is clicked. When null, no call is made. You may need to test for situations where that property is null meaning changing its value from a function to null, not changing what the function returns.  Granted this property normally starts out as null, but this is just an example (you might imagine someone creating a web component that defines a default `onclick` which could later be set to null).

DanKaplanSES:
@senocular That's a good point. Would that impact the suggested error message, or something else I'm not understanding?

PR #2631, opened 2024-12-26
feat: throw when trying to use .value() on a method

mroderick:

This is a solution for #2629

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -4,6 +4,7 @@ const arrayProto = require("@sinonjs/commons").prototypes.array;
 const isPropertyConfigurable = require("./util/core/is-property-configurable");
 const exportAsyncBehaviors = require("./util/core/export-async-behaviors");
 const extend = require("./util/core/extend");
+const getPropertyDescriptor = require("./util/core/get-property-descriptor");
 
 const slice = arrayProto.slice;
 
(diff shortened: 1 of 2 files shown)


mroderick:
@mantoni the [errors seen in `ubuntu-latest`](https://github.com/sinonjs/sinon/actions/runs/12502499977/job/34881594651) might also affect Mochify running on latest Chrome + Ubuntu.

image

If you're seeing similar errors in Mochify, you might want to stick to `ubuntu-22.04` for a bit, while the Chrome/Ubuntu conflict gets fixed in Ubuntu/Puppeteer/Chrome land.

- https://github.com/puppeteer/puppeteer/issues/13365

lib/sinon/default-behaviors.js
```diff
@@ -286,6 +287,20 @@ const defaultBehaviors = {
...
+            rootStub.propName,
+        );
+        const propertyIsGetter = propertyDescriptor.get !== undefined;
+        const propertyIsMethod =
+            !propertyIsGetter &&
+            typeof rootStub.rootObj[rootStub.propName] === "function";
+
+        if (propertyIsMethod) {
+            throw new Error(
+                `${rootStub.propName} is a function, not a getter or value. Use .returns() instead of .value()`,
```
fatso83, line 301: Are we supposed to allow setting `.value()` on a getter property? Seems erronous.

test/stub-test.js
```diff
@@ -3781,6 +3781,29 @@ describe("stub", function () {
...
+        it("allows stubbing getters", function () {
+            const y = {
+                get foo() {
+                    return "bar";
+                },
+            };
+            refute.exception(function () {
+                createStub(y, "foo").value("bar");
+            });
+        });
```
fatso83, line 3793: The original issue (#2629) does not deal with this issue, so this seems more like we are documenting existing (non-intentional) behavior? Dan does talk about getters, but that is the conventional Java-bean like getters, which are methods.

lib/sinon/default-behaviors.js
mroderick: I wasn't sure either ... to me it seemed more reasonable to use `.returns()` rather than `.value()` on a getter

test/stub-test.js
mroderick: If we change it like in https://github.com/sinonjs/sinon/pull/2631#discussion_r1900442363, then this test will obviously also need to be changed

lib/sinon/default-behaviors.js
fatso83: I think we should just do the simplest thing: prevent using `value()` on a method. So "`${rootStub.propName} is a function and should not be overridden by .value(). Use .returns() when creating a function that returns a value`.

There is little reason to add synthetic property getters and setters to the mix, as it just adds more noise/possible confusion, IMHO.

PR #2660, opened 2025-06-17
Also mirror the calledOnceWith assertion

bmeurer:

For Chrome DevTools front-end testing we generally prefer the `sinon.assert.foo(spy, ...)` form over `assert.isTrue(spy.foo(...))`, and `calledOnceWith` is missing from the `sinon.assert` object currently.

 #### Solution  - optional

It's pretty straight-forward to also expose the `calledOnceWith` on the `sinon.assert` object.

1. Check out this branch
2. `npm install`
3. `npm test` (specifically the `test/assert-test.js` file checks that the method is properly exposed)

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/docs/release-source/release/assertions.md
+++ b/docs/release-source/release/assertions.md
@@ -87,6 +87,12 @@ Passes if `spy` was called with the provided arguments.
 
 It's possible to assert on a dedicated spy call: `sinon.assert.calledWith(spy.firstCall, arg1, arg2, ...);`.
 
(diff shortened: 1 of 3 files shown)


bmeurer:
> Hi, thanks for doing this! Could you also just run Prettier on the changes? seems to fail on that step.

Done. Also added the method to the documentation.

bmeurer:
Should be ready now.

PR #2670, opened 2025-12-21
Fix issue 2668 - Perform additional cleanup on callThrough() to discard any previously defined behavior

### LINKED ISSUE #2668: stub.resolves(xxx) still applies after stub.callThrough()
**Describe the bug**

I have a test where a stub is initially configured to `stub.resolve(xxx)`.\
Later I `stub.callThrough()` to restore the the initial method.\
However, when calling the stubbed method, the original method is not called, `resolve(xxx)` still applies.

**To Reproduce**

```ts
it("#xxx - callThrough not clearing resolves()", async function () {
    const ns = {
        async foo() { return 'bar'; }
    };

    const stub = sinon.stub(ns, 'foo');
    stub.resolves('baz');

    assert.equals(await ns.foo(), 'baz');

    stub.callThrough();

    assert.equals(await ns.foo(), 'bar'); // fails, stub resolves with 'baz'.
});
```

Note that the same test without the async/await is pass.

**Expected behavior**

The original method should have been called.

- Sinon version : 17.0.2 to 21.0.0
- Runtime: Node 20

```
  Binaries:
    Node: 20.17.0 - /home/cfx/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - /home/cfx/.nvm/versions/node/v20.17.0/bin/npm
    pnpm: 10.26.0 - /home/cfx/.nvm/versions/node/v20.17.0/bin/pnpm
  Browsers:
    Chrome: 127.0.6533.72
```

- Other relevant environmental info:
- Example URL: 

**Additional context**
The issue appears to have been introduced by <https://github.com/sinonjs/sinon/pull/2593>, the issue does not occur with Sinon 17.0.1.

Inspired by various cleanup logic from `default-behaviors.js`, I've been able to quick fix the issue with:

```diff
diff --git a/lib/sinon/default-behaviors.js b/lib/sinon/default-behaviors.js
index ba85007e..76d64556 100644
--- a/lib/sinon/default-behaviors.js
+++ b/lib/sinon/default-behaviors.js
@@ -240,6 +240,14 @@ const defaultBehaviors = {
     },
 
     callThrough: function callThrough(fake) {
+        fake.returnValue = undefined;
+        fake.resolve = false;
+        fake.resolveThis = false;
+        fake.reject = false;
+        fake.returnValueDefined = false;
+        fake.exception = undefined;
+        fake.exceptionCreator = undefined;
+        fake.fakeFn = undefined;
         fake.callsThrough = true;
     },
 
```

However I'm not sure to understand why the cleanup logic varies so much from case to case in this file and what really needs to be cleaned-up, so this might not be correct/sufficient.

Seeing the implementation of `callThroughWithNew()` in _default-behaviors.js_, I'm wondering if it has the same issue.

Sigill:
#### Purpose

Fix #2668 by cleanup all other behaviors when calling callThrough()

1. Check out this branch
2. `npm install`
3. `npm run test`

#### Checklist for author

- [ ] `npm run lint` passes

### DIFF
--- a/src/sinon/default-behaviors.js
+++ b/src/sinon/default-behaviors.js
@@ -241,10 +241,40 @@ const defaultBehaviors = {
 
     callThrough: function callThrough(fake) {
         fake.callsThrough = true;
+
+        fake.callArgAt = undefined;
+        fake.callsThroughWithNew = false;
+        fake.exception = undefined;
+        fake.exceptionCreator = undefined;
+        fake.fakeFn = undefined;
+        fake.reject = false;
+        fake.resolve = false;
+        fake.resolveArgAt = undefined;
+        fake.resolveThis = false;
+        fake.returnArgAt = undefined;
+        fake.returnThis = false;
+        fake.returnValue = undefined;
+        fake.throwArgAt = undefined;
+
+        fake.callArgProp = undefined;
+        fake.callbackArguments = [];
+        fake.callbackContext = undefined;
+        fake.callbackAsync = false;
+        fake.returnValueDefined = false;
     },
 
     callThroughWithNew: function callThroughWithNew(fake) {
         fake.callsThroughWithNew = true;
+
+        fake.callArgAt = undefined;
+        fake.exception = undefined;
+        fake.exceptionCreator = undefined;
+        fake.throwArgAt = undefined;
+
+        fake.callArgProp = undefined;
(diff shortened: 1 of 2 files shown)


fatso83:
This seems fine to me! Could you rebase on the latest `main` and I'll have a new look?

Issue #2692, opened 2026-04-10
new version 21.1.0 causes ERR_REQUIRE_ESM


pascalverlinden:
Apparently the update to version 21.1.0 throws an error in sinon/lib/sinon/colorizer.js

Version 21.0.3 is working fine.

Sample output when trying to run tests:

```
/home/app/node_modules/sinon/lib/sinon/colorizer.js:3
var supportsColor = require('supports-color');
                    ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /home/app/node_modules/sinon/node_modules/supports-color/index.js from /home/app/node_modules/sinon/lib/sinon/colorizer.js not supported.
Instead change the require of index.js in /home/app/node_modules/sinon/lib/sinon/colorizer.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/colorizer.js:3:21)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/spy-formatters.js:4:19)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/proxy.js:10:21)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/stub.js:6:13)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/mock-expectation.js:10:12)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/mock.js:4:23)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/sandbox.js:9:12)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon/create-sandbox.js:4:15)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/create-sinon-api.js:4:21)
    at Object.<anonymous> (/home/app/node_modules/sinon/lib/sinon.js:3:22)
    at Object.<anonymous> (/home/app/node_modules/@tmining/connector-lib/src/test-utils.js:1:15)
    at Object.<anonymous> (/home/app/node_modules/@tmining/connector-lib/index.js:9:21)
    at Object.<anonymous> (/home/app/src/lib.js:10:39)
    at Object.<anonymous> (/home/app/src/app.js:8:13)
    at Object.<anonymous> (/home/app/src/index.js:1:13) {
  code: 'ERR_REQUIRE_ESM'
}
```

**Context (please complete the following information):**

[...]

PR #2693, opened 2026-04-10
fix(#2692): Remove ESM-only supports-color as it breaks CJS exports

fatso83:

Hopefully fixes 2692


package.json
```diff
@@ -132,7 +133,7 @@
     "@sinonjs/fake-timers": "^15.3.0",
     "@sinonjs/samsam": "^10.0.0",
     "diff": "^8.0.4",
-    "supports-color": "^10.2.2"
+    "npm-run-all": "^4.1.5"
```
AviVahl, line 136: @fatso83 was it intended to add this to dependencies? I believe it should be a devDep. noticed it because of a major jump in the number of dependencies in a downstream project.

AviVahl: created https://github.com/sinonjs/sinon/pull/2694 to fix

fatso83: nope, that was a post midnight slip up 🫣

Issue #2695, opened 2026-04-11
Change in 21.1: sinon.stub() throws with `--disable-proto=throw`


kevinoid:
**Describe the bug**

Calling `sinon.stub()` from Node.js running with [`--disable-proto=throw`](https://nodejs.org/api/cli.html#disable-protomode) causes `Error` to be thrown.

**To Reproduce**

1. Run `node --disable-proto=throw -e "require('sinon').stub()"` in a project with sinon 21.1 installed.
2. Observe the output similar to the following:

       /path/to/node_modules/sinon/lib/sinon/util/core/walk.js:29
               if (seen[k] !== true) {
                       ^
       
       Error: Accessing Object.prototype.__proto__ has been disallowed with --disable-proto=throw
           at /path/to/node_modules/sinon/lib/sinon/util/core/walk.js:29:17
           at Array.forEach (<anonymous>)
           at walkInternal (/path/to/node_modules/sinon/lib/sinon/util/core/walk.js:28:5)
           at walkInternal (/path/to/node_modules/sinon/lib/sinon/util/core/walk.js:41:9)
           at walkInternal (/path/to/node_modules/sinon/lib/sinon/util/core/walk.js:41:9)
           at walk (/path/to/node_modules/sinon/lib/sinon/util/core/walk.js:57:12)
           at collectOwnMethods (/path/to/node_modules/sinon/lib/sinon/collect-own-methods.js:39:5)
           at addReturnedMethodsToCollection (/path/to/node_modules/sinon/lib/sinon/sandbox.js:196:21)
           at Sandbox.stub (/path/to/node_modules/sinon/lib/sinon/sandbox.js:229:9)
           at [eval]:1:18 {
         code: 'ERR_PROTO_ACCESS'
       }

**Expected behavior**
The process would produce no output and exit with code 0, as it does with sinon 21.0.3 and earlier.

**Context (please complete the following information):**

- Sinon version : 21.1.1
- Runtime: Node 25

**Additional context**

I realize running with [`--disable-proto=throw`](https://nodejs.org/api/cli.html#disable-protomode) may not be a supported scenario, and I understand if you choose not to fix this issue.  However, I have found it useful for my own projects to flag use of `__proto__` and thought it may be worth considering, given that it worked previously.

Thanks for considering,
Kevin

fatso83:
[...]

PR #2699, opened 2026-04-14
fix: exclude `__proto__` from `walk()`

kevinoid:

Fix issue #2695 by excluding `__proto__` from `walk()`.

[`__proto__`] is a special property to access an object's prototype.  It has many pitfalls:

- Setting it to an object value changes an object's prototype, which is generally discouraged and may be unexpected by the `iterator` callback.
- Setting it to a non-object value does nothing (meaning `seen[k] = true` has no effect).
- When Node.js is run with the [`--disable-proto=throw`] option, getting or setting `__proto__` causes an exception with code `ERR_PROTO_ACCESS` to be thrown, which causes issues like #2695.

#### Solution  - optional

Since `__proto__` (and all properties of `Object.prototype`) are currently unused by consumers of `walk()` in this project, it is both safe and preferable to exclude it from enumeration by `walk()`.

[`__proto__`]: https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.prototype.__proto__
[`--disable-proto=throw`]: https://nodejs.org/api/cli.html#disable-protomode

1. Check out this branch
2. `npm install`
3. `NODE_OPTIONS=--disable-proto=throw npm test`

#### Checklist for author

- [x] `npm run lint` passes

#### Additional Work

It would be great if `--disable-proto=throw` could be added to the test suite, either by adding [`--node-option disable-proto=throw`](https://mochajs.org/running/cli/#--node-option-name--n-name) to the mocha invocation in `test-node`, or adding a separate run to test under these conditions.  However, this may introduce an additional maintenance burden, so requires additional discussion and is not currently part of this PR.

Thanks for considering,
Kevin

### DIFF
--- a/src/sinon/util/core/walk.js
+++ b/src/sinon/util/core/walk.js
@@ -28,7 +28,13 @@ function walkInternal(obj, iterator, context, originalObj, seen) {
     }
 
     forEach(Object.getOwnPropertyNames(obj), function (k) {
-        if (seen[k] !== true) {
+        // Skip __proto__ because:
+        // - It's not currently useful in this project.
+        // - It's automatic, problematic, and deprecated.
+        // - Assigning to it has special effects (or no effect for non-object)
+        //   which may not be expected/handled properly by iterator callback.
+        // - Getting it throws Error with node --disable-proto=throw.
+        if (k !== "__proto__" && seen[k] !== true) {
             seen[k] = true;
             const target =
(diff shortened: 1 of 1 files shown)


fatso83 (approved):

fatso83:
Thank you!

kevinoid:
My pleasure.  Thanks for reviewing, merging, and maintaining sinon!

PR #2715, opened 2026-05-16
fix: isolate callId counter per sandbox for parallel test support

### LINKED ISSUE #2472: calledImmediatelyBefore fails when executed in parallel mode using node-tap

singhvishalkr:
When tests run in parallel with separate sandboxes, the \calledImmediatelyBefore\ and \calledImmediatelyAfter\ assertions fail unexpectedly. This happens because proxy-invoke.js uses a global \callId\ counter shared across all sandboxes.

The fix introduces a per-sandbox context object that holds its own \callId\ counter. Each sandbox passes this context through the spy/stub/fake creation chain, and the proxy stores it for use during invocation.

Implementation:
- proxy-invoke.js: Use \	his.sinonContext.callId\ instead of a global variable, with a default context for backward compatibility when used outside sandboxes
- proxy.js: Accept optional context parameter and store it on the proxy
- spy.js/stub.js/fake.js: Thread context through creation functions
- sandbox.js: Create a sandbox-specific context and pass it to spy/stub/fake via new \withContext\ methods

Added tests that verify call order assertions work correctly across independent sandboxes.

The public API remains unchanged. Direct use of \sinon.spy()\ without a sandbox uses a shared default context (same behavior as before), while sandbox-created spies/stubs/fakes get isolated counters.

Fixes #2472

### DIFF
--- a/src/sinon/fake.js
+++ b/src/sinon/fake.js
@@ -18,11 +18,12 @@ const { slice } = prototypes.array;
  *
  * @param {SinonFunction|undefined} [f]
  * @param {object} [context] The sinon context for callId tracking
+ * @param {boolean} hasFunctionArgument
  * @returns {SinonFunction}
  * @namespace
  */
-function fakeImpl(f, context) {
-    if (typeof f !== "undefined" && typeof f !== "function") {
+function fakeImpl(f, context, hasFunctionArgument) {
+    if (hasFunctionArgument && typeof f !== "function") {
         throw new TypeError("Expected f argument to be a Function");
     }
 
@@ -39,7 +40,7 @@ function fakeImpl(f, context) {
  * @namespace
  */
 function fake(f) {
-    return fakeImpl(f, undefined);
+    return fakeImpl(f, undefined, arguments.length > 0);
 }
 
 /**
@@ -50,7 +51,7 @@ function fake(f) {
  * @returns {SinonFunction}
  */
 fake.withContext = function (context, f) {
-    return fakeImpl(f, context);
+    return fakeImpl(f, context, arguments.length > 1);
 };
 
 /**
--- a/src/sinon/fake.js
+++ b/src/sinon/fake.js
@@ -17,16 +17,41 @@ const { slice } = prototypes.array;
  * When an `f` argument is supplied, this implementation will be used.
  *
  * @param {SinonFunction|undefined} [f]
+ * @param {object} [context] The sinon context for callId tracking
+ * @returns {SinonFunction}
+ * @namespace
+ */
+function fakeImpl(f, context) {
+    if (typeof f !== "undefined" && typeof f !== "function") {
+        throw new TypeError("Expected f argument to be a Function");
+    }
+
+    return wrapFunc(f, context);
+}
+
+/**
+ * Returns a `fake` that records all calls, arguments and return values.
+ *
+ * When an `f` argument is supplied, this implementation will be used.
+ *
+ * @param {SinonFunction|undefined} [f]
  * @returns {SinonFunction}
  * @namespace
  */
 function fake(f) {
-    if (arguments.length > 0 && typeof f !== "function") {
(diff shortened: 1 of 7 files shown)


fatso83 (approved):

fatso83:
The implementation looks good, but the tests are failing 🙃

singhvishalkr:
Updated the branch.

The failing tests were from the sandbox wrappers using the array helper incorrectly when forwarding arguments into the context-aware spy/stub/fake constructors. That dropped the original stub target metadata, which broke accessor stubs and nested sandbox behavior. The wrappers now use the existing commons array helpers consistently, and `src/sinon/spy.js` has been formatted.

Verification:
- `npx prettier --check` on changed files
- `npx eslint --max-warnings 0` on changed files
- `npx mocha -R spec test/src/sandbox-test.js --grep "call order isolation"`
- `npx mocha -R dot test/src/sandbox-test.js --timeout 10000`

I could not use `npm run test-node` on Windows because Rollup fails locally before tests with: `Entry module "src/create-sinon-api.js" cannot be external`. The source-level sandbox tests above cover the failing behavior directly.

fatso83:
```
  1551 passing (2s)
  12 pending
  1 failing

  1) fake
       should reject non-Function argument:
     AssertionError: [assert.exception] Expected exception
      at Object.fail (node_modules/@sinonjs/referee/lib/create-fail.js:12:21)
      at Object.fail (node_modules/@sinonjs/referee/lib/define-assertion/index.js:57:17)
      at Object.assert (node_modules/@sinonjs/referee/lib/assertions/exception.js:44:21)
      at assertion (node_modules/@sinonjs/referee/lib/define-assertion/index.js:62:23)
      at referee.<computed>.<computed> [as exception] (node_modules/@sinonjs/referee/lib/define-assertion/index.js:111:22)
      at file:///home/runner/work/sinon/sinon/test/src/fake-test.js:94:20
      at Array.forEach (<anonymous>)
      at Context.<anonymous> (file:///home/runner/work/sinon/sinon/test/src/fake-test.js:93:18)
      at process.processImmediate (node:internal/timers:504:21)

```

singhvishalkr:
Fixed in `d216f3f9`. The sandbox context plumbing now distinguishes an omitted `fake()` argument from an explicit `fake(undefined)`, so the existing non-Function validation is preserved.

Ran:
- `npm.cmd exec -- mocha -R spec test/src/fake-test.js --grep "should reject non-Function argument"`
- `npm.cmd exec -- mocha -R dot test/src/fake-test.js`
- `npm.cmd exec -- mocha -R dot test/src/sandbox-test.js --grep "call order isolation"`
- `npx.cmd prettier --check src/sinon/fake.js`
- `npx.cmd eslint --max-warnings 0 src/sinon/fake.js`
- `git diff --check`

fatso83:
It seems that we missed createStubInstance() in this change. Any chance of patching the support to cover that? Ref #2760.

PR #2724, opened 2026-06-11
fix: let returns override returnsArg

### LINKED ISSUE #2656: returns() doesn't override returnsArg(0)
**Describe the bug**

Since #2593, calling `returns(...)` on a stub that was previously configured with `returnsArg(0)` doesn't update the behavior. Instead, the original `returnsArg(0)` behavior is preserved.

**To Reproduce**

```js
const stub = sandbox.stub();

stub.returnsArg(0);
console.log(stub(1));

stub.returns(42);
console.log(stub(2)); // should log 42, logs 2
```

- Sinon version : 20.0.0
- Runtime: Node 20

```
N/A
```

- Other relevant environmental info:
- Example URL: 

**Additional context**

pupuking723:
Fixes #2656.

This clears the previous `returnsArg()` behavior when `returns()` is configured afterwards. Without clearing `returnArgAt`, the behavior invocation path still prioritizes `returnsArg` over the explicit return value, so `stub.returnsArg(0).returns(42)` returns the first argument instead of `42`.

The added regression test covers that chaining order directly.

Verification:

- `npm run test-node -- test/src/stub-test.js --grep "is superseded by returns"`
- `npm run test-node -- test/src/stub-test.js --grep "returnsArg"`
- `npm run test-node -- test/src/stub-test.js`
- `npm run test-node`
- `npx eslint src/sinon/default-behaviors.js test/src/stub-test.js --max-warnings 0`
- `npx prettier --check src/sinon/default-behaviors.js test/src/stub-test.js`

### DIFF
--- a/src/sinon/default-behaviors.js
+++ b/src/sinon/default-behaviors.js
@@ -145,6 +145,7 @@ const defaultBehaviors = {
     returns: function returns(fake, value) {
         fake.callsThrough = false;
         fake.returnValue = value;
+        fake.returnArgAt = undefined;
         fake.resolve = false;
         fake.reject = false;
         fake.returnValueDefined = true;
--- a/test/src/stub-test.js
+++ b/test/src/stub-test.js
@@ -721,6 +721,15 @@ describe("stub", function () {
             assert.equals(obj.fn("myarg"), "myarg");
         });
 
+        it("is superseded by returns", function () {
+            const stub = createStub();
+
+            stub.returnsArg(0);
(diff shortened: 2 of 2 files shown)


mroderick:
Closing and re-opening to diagnose issues with GitHub Actions

mroderick:
That worked!

mroderick:
Thank you for your contribution to Sinon @pupuking723 ❤️

PR #2733, opened 2026-06-22
Fix #1741: record calls on stubbed getters/setters

### LINKED ISSUE #1741: Cannot spy on property getters, and stubbed getters do not record calls
* Sinon version : 4.4.6
* Environment   : Chrome 65.0.3325.162 on Mac
* Example URL   : N/A
* Other libraries you are using: N/A

I'm attempting to stub out - and monitor - a property getter,
using the stubbing approach recommended in the documentation at http://sinonjs.org/releases/v4.4.6/stubs/#stubgetgetterfn

I was hoping that the resulting stub would report when the getter was actually invoked, but no such reports are logged.

(In fact, I don't actually want to stub the getter: I merely want to spy on it, but Sinon provides no property getter spy API right now afaik.)

**What actually happens**

I can successfully stub out the getter, and control what it does, but Sinon's invocation recording reports no calls.

(This is also related to issue #1545.)

I do understand that, for a stubbed property, there are really two sets of function calls going on -  the getter and the setter - and so really there should be two sets of recordings going on.  But right now, there are none: when stubbing a property, as recommended in the docs above, no recording happens.

**How to reproduce**

There's a small example at https://codesandbox.io/s/l5m9loy21l; output is in the console.

youdie006:
## Purpose (TL;DR)

`stub.get(fn)` / `stub.set(fn)` installed the raw replacement function directly as the property accessor, so invoking a stubbed getter/setter ran untracked — `callCount` never incremented (#1741). This wraps the replacement in a spy and exposes it as `stub.getter` / `stub.setter`, so accessor invocations are recorded (mirroring the existing `spy.get` / `spy.set` surface). This is the scoped "tiny PR" @fatso83 green-lit in the issue thread.

## How to verify

```js
const obj = { get prop() { return 1; } };
const s = sinon.stub(obj, "prop").get(() => 42);
obj.prop; obj.prop;
s.getter.callCount; // 2  (previously: not recorded)
```

`npx mocha test/src/stub-test.js --grep "records calls to the stubbed"` → 2 passing.

## Checklist

- [x] Lint passes (`eslint --max-warnings 0`; prettier clean)
- [x] Scoped + non-breaking (+19 lines in `default-behaviors.js`, no API removed); full suite 1554 passing / 0 failing locally.

Fixes #1741

### DIFF
--- a/src/sinon/default-behaviors.js
+++ b/src/sinon/default-behaviors.js
@@ -4,6 +4,7 @@ const { prototypes } = commons;
 import isPropertyConfigurable from "./util/core/is-property-configurable.js";
 import exportAsyncBehaviors from "./util/core/export-async-behaviors.js";
 import extend from "./util/core/extend.js";
+import spy from "./spy.js";
 
 const { slice } = prototypes.array;
 
@@ -280,8 +281,15 @@ const defaultBehaviors = {
     get: function get(fake, getterFunction) {
         const rootStub = fake.stub || fake;
 
+        // Wrap the replacement getter in a spy so that reading the property
+        // records the access (callCount, calledOnce, ...). The spy is exposed
+        // on the stub as `stub.getter`, mirroring the `spy.get` accessor that
+        // `sinon.spy(obj, "prop", ["get"])` produces. See #1741.
+        const getterSpy = spy(getterFunction);
+        rootStub.getter = getterSpy;
+
         Object.defineProperty(rootStub.rootObj, rootStub.propName, {
-            get: getterFunction,
+            get: getterSpy,
             configurable: isPropertyConfigurable(
                 rootStub.rootObj,
                 rootStub.propName,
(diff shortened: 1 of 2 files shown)


mroderick:
This is a really clean fix. Thank you for picking it up, and for keeping it scoped — the diff is easy to follow and exactly addresses the issue.

A couple of nice things I noticed 👏

 - Wrapping in `spy()` is the right approach — routes through the existing call-recording machinery without reinventing anything.
 - this context is properly preserved through the spy proxy.
 - No circular dependency introduced (`spy.js` doesn't import `default-behaviors.js`).
 - The naming collision with `stub.get`/`stub.set` makes `.getter`/`.setter` the right call.

One small suggestion: `resetBehavior` in `stub.js` doesn't clean up `.getter`/`.setter`, so after a `resetBehavior()` the stub still holds stale spy references. It's a narrow window (calling `.get()` again overwrites it), but adding `delete this.getter; delete this.setter;` there would keep it clean. 

Nice work on this!

PR #2737, opened 2026-06-26
feat: make sinon.restoreObject idempotent

### LINKED ISSUE #2736: Make sinon.restoreObject idempotent

In most cases, I want to use `sinon.restoreObject(value)` in a test hook like `beforeEach`/`afterEach`. However, not every single test case necessarily stubs `value`.

``` typescript
afterEach(() => {
    sinon.restoreObject(value);
})

it('should do x', () => {
    sinon.stub(value, 'foo').returns(123);
    // ...
});


it('should do y', () => {
    // no value stub
})
```

This will throw after the second test because `value` does not contain a stub.

I want the `sinon.restoreObject` method to have symmetry with `sinon.restore()` and `sandbox.restore()`. Ie, passing an empty/un-mocked object should be a no-op.

Right now i have my own wrapper:

``` typescript
function restore(value) {
    if ('restore' in value && typeof value.restore === 'function') {
        value.restore()
    }
}
```

However, it doesn't handle nesting. Another alternative would be to leave `sinon.restoreObject` as-is and allow passing a value to `sinon.restore(value)`.

**Additional context**

The strict behaviour was originally introduced here: https://github.com/sinonjs/sinon/pull/2092
I think the stricter behaviour makes sense for the other methods, but not for `restoreObject`.

icholy:

Fix #2736 by making `sinon.restoreObject` a no-op on empty/stub-less objects.

1. Check out this branch
2. `npm install`
3. `npm test`

#### Checklist for author

- [x] `npm run lint` passes

### DIFF
--- a/src/sinon/restore-object.js
+++ b/src/sinon/restore-object.js
@@ -1,4 +1,4 @@
-import walkObject from "./util/core/walk-object.js";
+import { walkObject } from "./util/core/walk-object.js";
 
 function filter(object, property) {
     return object[property].restore && object[property].restore.sinon;
--- a/src/sinon/spy.js
+++ b/src/sinon/spy.js
@@ -5,7 +5,7 @@ import extend from "./util/core/extend.js";
 import getPropertyDescriptor from "./util/core/get-property-descriptor.js";
 import isEsModule from "./util/core/is-es-module.js";
 import * as proxyCallUtil from "./proxy-call-util.js";
-import walkObject from "./util/core/walk-object.js";
+import { walkObjectStrict } from "./util/core/walk-object.js";
 import wrapMethod from "./util/core/wrap-method.js";
 
 const { prototypes, functionName, valueToString } = commons;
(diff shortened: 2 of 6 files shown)


hexeberlin (commented):
@icholy thanks for your PR! 

We are not huge fans of using booleans as parameters (here is a [nice article](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) about the why).

What do you think? Is there a better way to make this improvement to sinon?

icholy:
@hexeberlin just tell me what you want the `walkObject` signature to look like. I was trying to following the conventions that already exited https://github.com/sinonjs/sinon/blob/544985b67032fdf5e7d563d8ba9f00c517db5a39/src/sinon/sandbox.js#L341

hexeberlin:
@icholy that is a fair point! We are definitely not consistent ... yet :)

Could you please create a new method for the strict version? E.g.

```js
export walkObject;
export walkObjectStrict;
```

... and then fix the usage sites to chose the desired version.

icholy:
@hexeberlin done

hexeberlin:
@icholy thanks a lot for your contribution and patience ❤️

PR #2759, opened 2026-08-08
fix: make return/throw behavior setters fully mutually exclusive

rajanpanth:
## Summary
Each of `returns`/`returnsArg`/`throwsArg`/`returnsThis`/`callsFake`/`resolves`/`resolvesArg`/`rejects`/`resolvesThis`/`callThrough`/`callThroughWithNew` only clears the specific fields its author happened to think of, rather than the full set `behavior.js#invoke` checks. So e.g. `returnsArg` followed by `throwsArg`, `callsFake`, `returnsThis`, or `resolves` leaves `returnArgAt` in place, which `invoke()` checks before any of those — silently keeping the earlier behavior instead of being overridden:

```js
const stub = sinon.stub();
stub.returnsArg(0);
stub.throwsArg(1);
stub("myarg", new Error("boom")); // returns "myarg" -- does NOT throw
```

#d36e921 fixed this for the `returns`-after-`returnsArg` pair specifically, but the same bug shape remained for every other combination.

## Changes
- `src/sinon/default-behaviors.js`: generalizes that fix into a shared `resetBehavior()`, called by every return/throw setter before it sets its own field(s). This is the same exhaustive-reset pattern `callThrough` already used — I just made every setter in this family use it consistently instead of each remembering to clear a handful of fields ad hoc.
  - Deliberately **excludes** the `callsArg`/`yields` callback-invocation fields from this reset: yielding a callback is a separate, additive side effect meant to combine with any of these return/throw behaviors regardless of call order (see the existing "returnsArg takes precedent over yielded return value" tests) — not another mutually-exclusive choice in this set. Left the `callsArg`/`yields` family of setters untouched.
  - `callThrough`/`callThroughWithNew` still additionally clear the callback config too, since "call the real method" is meant to replace all stub customization. This also fixes `callThroughWithNew` never having cleared `returnArgAt`/`returnThis`/`fakeFn`/`resolve*`/`reject`/`returnValue`, unlike `callThrough` (which already cleared everything).
- `test/src/stub-test.js`: 6 new "is superseded by X" tests alongside the existing ones, covering the previously-broken combinations.

## Testing
- Full suite: 1503/1504 passing. The 1 failure (`Sandbox > warns of potential leak when many fakes are created`, a 2000ms timeout) is pre-existing and unrelated — confirmed via `git stash`, it fails identically without this change.
- Confirmed all 6 new tests fail on the prior code and pass with the fix.
- `eslint --max-warnings 0` clean.


rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to update anything if needed.

Thanks!

PR #2762, opened 2026-08-19
docs: fix dead browser build download links

rajanpanth:
The Installation section links to `https://sinonjs.org/releases/` and `https://sinonjs.org/releases#npm-cdns`. Both return 404 — the VitePress rebuild of sinonjs.org dropped the releases section, and there is no releases page in `docs/` either, so there is nothing to point the old URLs at.

The `pkg/` browser bundle is still published (it's listed in `package.json` `files`), so this points readers at it directly on the two CDNs that serve it:

- https://cdn.jsdelivr.net/npm/sinon/pkg/sinon.js — 200
- https://unpkg.com/sinon/pkg/sinon.js — 200

I left the `nise` changelog link in `CHANGES.md` alone even though it also 404s, since that file is a historical record of past releases rather than live documentation.


lbesecker195 (approved):
Verified: `https://sinonjs.org/releases/` returns 404, and both the jsDelivr and unpkg `pkg/sinon.js` URLs return 200. LGTM.

PR #2764, opened 2026-08-31
fix: share call order with stub instances

### LINKED ISSUE #2760: createStubInstance() doesn't receive per-sandbox callId context, breaking calledAfter/calledBefore vs spy()

wangxpych:
Fixes #2760.

## Problem

The per-sandbox `callId` context added for parallel-test isolation is passed through `spy()`, `stub()`, and `fake()`, but `createStubInstance()` still creates its method stubs with the module-level default context. Comparisons against other fakes from the same sandbox can therefore report the wrong call order.

## Changes

- Thread the sandbox context through `createStubInstance()` when its methods are stubbed.
- Route the root `sinon.createStubInstance()` wrapper through its backing root sandbox so it uses the same context as root spies and stubs.
- Add regression coverage for both the root API and explicit sandboxes, placing a stub-instance call between two spy calls.

## Verification

- `npm test` — Node: 1,558 passing / 12 pending; headless browser: 1,551 passing / 13 pending; WebWorker: 1 passing
- `npm run lint`
- `npm run prettier:check`
- `npm run test-contract`
- `npm run check-dependencies`


wangxpych:
Withdrawing this contribution after re-evaluating the repository's recent human review and merge activity. No action is needed from maintainers. Thank you.

Issue #2714, opened 2026-05-15
How to stub ES module imports documentation is incorrect


AniaMakes:
Documentation was added in #2676 (~2 months ago)

I tried following it and discovered that it is, in fact, incorrect, at least when using the most recent version of `esm`.

The set up file, according to the instructions, should look like this:
```
require = require("esm")(module, {
  cjs: true,
  mutableNamespace: true,
});
```
however, if you look on `esm` documentation (https://www.npmjs.com/package/esm#options), you can see that `mutableNamespace` is NOT a top-level key. It should be set up like this:
```
require = require("esm")(module, {
  cjs: {mutableNamespace: true}
});
```

Given that the foundation of this documentation is incorrect, it makes me doubt if this has actually been verified by a human, and I think that incorrect docs are worse than no docs at all. 

I hope you are able to either delete the file, or mark it as "unverified" tutorial, after updating the esm config. 
Thank you in advance

fatso83:
I did not manually verify the PR, but I am sure @eduardbar can fix this ASAP?

PR #2767, opened 2026-09-05
docs: fix esm loader options in ESM stubbing guide

alorentiar:
The loader example in this guide passes `mutableNamespace` as a top-level
esm option:

```js
require = require("esm")(module, {
  cjs: true,
  mutableNamespace: true,
});
```

esm only reads it under `cjs` (see the options table in its README), and it
rejects unknown top-level keys, so this loader actually fails on startup with
`Unknown esm@3.2.25 option: mutableNamespace`. I reproduced that with mocha +
esm on an older Node (esm only runs there), then confirmed the corrected
options load cleanly.

Moved the flag under `cjs` in both places it appears. Keeping it scoped to
the reported config issue.

Refs #2714


alorentiar:
Gentle ping on this one. It is a small docs fix and has been sitting for a few days, so happy to rebase or adjust if anything needs changing.

lbesecker195 (approved):
Verified against the esm README: `cjs` is "a boolean or object", and `mutableNamespace` is listed as one of its sub-options, not a top-level option. The corrected form is the one esm actually reads. LGTM.

